Issue #18, opened 2022-04-14
Determine how to invalidate / retire message schemas


aramikm:
If a message schema is faulty or is to be deprecated, how do we do this?

Some thoughts -- 
Are there extra fields in schema registration storage that show if it's deprecated or invalidated?  Is it deleted from schema registration storage?   Either way there should almost certainly be an event deposited indicating what has been done to the schema.

If the schema is deleted from storage, then what happens to messages that were in the pool right beforehand which would then fail validation immediately afterward?  Similarly what does this mean for message batches that are retrieved and checked for validation historically?

If the schema is not deleted from storage, then almost certainly the schema should be marked as no longer usable for current messages.  This means there should also be a block number after which the schema is no longer valid, and this block number should be emitted with the event so that subscribers are alerted.

Perhaps there are multiple end states for a schema, such as:
Invalid -- it was never valid, so any messages using this schema can never be valid.  Needed only for registering a schema by mistake or with errors (e.g. it's missing a needed field).
Deprecated -- Still valid but slated to be expired after a certain block number
Unsupported/Expired -- used only for messages that were valid at one time, the message is no longer valid if it was posted after a certain block number

Maybe we just need Valid, Invalid, and Expired?  Do we need any other fields?

Related to https://github.com/Liberty30/mrc/issues/12

wilwade:
Archive

PR #21, opened 2022-04-19
Validity states and transitions

shannonwells:
Some ideas about how we might handle validity of MsaIds, Delegations and Schemas on chain.

One of the motivations for this model is that a message relating to a given ID number can be tied to one and only one block, because it exists in one and only one batch. Furthermore, the messages are either valid or invalid at the time of posting. Lastly, a message itself relates to one and only one each of Delegate, Delegator, and Schema.

I think this means we don't have to and should not concern ourselves with schema upgrades or backward compatibility. This model also doesn't prevent actually deleting MsaIds if we must; it's only for checking validity of messages we expect to still be around.

See #18 for more discussion and other ideas about retention.


shannonwells:
This was just a way to discuss some thoughts about schema and retirement but we've decided to go a different route.

PR #50, opened 2022-05-02
Helper library for avro

saraswatpuneet:
- [x] Convert raw schema to Avro Schema Type
- [x] Convert raw schema to validated and serialized schema bytes upon parsing 
- [x] Convert a list of raw schemas to validated and serialized avro schema
- [x] Convert a serialized schema back to Avro schema
- [x] Convert a list of serialized schemas to Avro schemas
- [x] Get a writer object for avro schema, writer holds data in a buffer of type Vec<u8>
- [x] Populate avro writer with records data for matching schema
- [x] Populate a avro schema with records (provided as hashmap) and return serialized data
- [x] Get a reader object populated with serialized data for a given Avro schema
- [x] Get a map of key value (hashmap) for serialized data mapped to an avro schema 

Tests:
 - [x] Write unit tests for all cases and end to end flow


common/helpers/src/avro.rs
```diff
@@ -0,0 +1,248 @@
...
+	#[error("I/O error")]
+	Io(#[from] std::io::Error),
+	#[error("Avro error")]
+	Avro(#[from] apache_avro::Error),
+	#[error("Invalid avro schema: {0}")]
+	InvalidSchema(String),
+	#[error("Invalid avro records")]
+	InvalidRecords(),
+}
+
```
saraswatpuneet, line 17: Typical errors returned by avro helper

common/helpers/src/avro.rs
```diff
@@ -0,0 +1,248 @@
...
+			},
+			Err(e) => Err(AvroError::Avro(e)),
+		}
+	})
+	.unwrap_or_else(|_| {
+		// if the unwind is caught, return an error
+		Err(AvroError::InvalidSchema(raw_schema.to_string()))
+	})
+}
+
```
saraswatpuneet, line 51: Given a valid raw avro schema translate it to avro which checks for validity on parsing or panic if schema is broken in either case return an error otherwise return a serialized schema to be stored on chain

common/helpers/src/avro.rs
```diff
@@ -0,0 +1,248 @@
...
+	let mut schema_fingerprints = Vec::new();
+	let mut schemas = Vec::new();
+	for s in raw_schema {
+		let (schema, fingerprint) = fingerprint_raw_schema(s)?;
+		schema_fingerprints.push(fingerprint);
+		schemas.push(schema);
+	}
+	Ok((schemas, schema_fingerprints))
+}
+
```
saraswatpuneet, line 78: Same as previous but do this for a list of raw schemas

common/helpers/src/avro.rs
```diff
@@ -0,0 +1,248 @@
...
+	let schema_str = str::from_utf8(&serialized_schema);
+	match schema_str {
+		Ok(schema_str) => {
+			let schema = Schema::parse_str(schema_str)?;
+			Ok(schema)
+		},
+		Err(error) => Err(AvroError::InvalidSchema(error.to_string())),
+	}
+}
+
```
saraswatpuneet, line 105: Given a serialized avro schema return its Avro Schema type to be used in further workflow

common/helpers/src/avro.rs
```diff
@@ -0,0 +1,248 @@
...
+/// assert!(translated_schema.is_ok());
+/// ```
+pub fn translate_schemas(serialized_schemas: Vec<Vec<u8>>) -> Result<Vec<Schema>, AvroError> {
+	let mut schemas = Vec::new();
+	for s in serialized_schemas {
+		schemas.push(translate_schema(s)?);
+	}
+	Ok(schemas)
+}
+
```
saraswatpuneet, line 129: same as previous this time translate a list of serialized avro schema to Avro Schema type

common/helpers/src/avro.rs
```diff
@@ -0,0 +1,248 @@
...
+/// Function to get the schema writer with default container as Vec<u8>
+/// # Examples
+/// ```
+/// use avro::*;
+/// let writer = get_schema_data_writer();
+/// ```
+pub fn get_schema_data_writer<'a>(schema: &'a Schema) -> Writer<'a, Vec<u8>> {
+	Writer::with_codec(schema, Vec::new(), Codec::Snappy)
+}
+
```
saraswatpuneet, line 139: returns a Avro writer object will be passed to populate for writing data to avro schema

common/helpers/src/avro.rs
```diff
@@ -0,0 +1,248 @@
...
+	records: &HashMap<String, SchemaValue>,
+) -> Result<(), AvroError> {
+	let mut record_list = Record::new(writer.schema()).unwrap();
+	for (field_name, field_value) in records.iter() {
+		record_list.put(field_name, field_value.clone());
+	}
+	writer.append(record_list)?;
+	Ok(())
+}
+
```
saraswatpuneet, line 162: Given an Avro writer populate a hashmap as records for a given schema

common/helpers/src/avro.rs
```diff
@@ -0,0 +1,248 @@
...
+		record_list.put(field_name, field_value.clone());
+	}
+	writer.append(record_list)?;
+	let serialized_bytes = writer.into_inner();
+	match serialized_bytes {
+		Ok(serialized_bytes) => Ok(serialized_bytes),
+		Err(error) => Err(AvroError::Avro(error)),
+	}
+}
+
```
saraswatpuneet, line 191: populate a schema with data and returned serialized data to be stored by message store for example

common/helpers/src/avro.rs
```diff
@@ -0,0 +1,248 @@
...
+/// let serialized_record = populate_schema_and_serialize(schema_fingerprint.unwrap().1, record);
+/// let reader = get_schema_data_reader(schema_fingerprint.unwrap().1, serialized_record.unwrap());
+/// ```
+pub fn get_schema_data_reader<'a>(
+	serialized_data: &'a Vec<u8>,
+	schema: &'a Schema,
+) -> Result<Reader<'a, &'a [u8]>, AvroError> {
+	Ok(Reader::with_schema(&schema, &serialized_data[..])?)
+}
+
```
saraswatpuneet, line 212: return a Avro Reader populated with serialized data obtained from previous call

common/helpers/src/avro.rs
```diff
@@ -0,0 +1,248 @@
...
+                    result_record.insert(field_name.clone(), field_value.clone());
+                }
+            }
+            _ => {
+                return Err(AvroError::InvalidRecords());
+            }
+        }
+    }
+    Ok(result_record)
+}
```
saraswatpuneet, line 248: return a hashmap of key value pair for a serialized data adhering to specific schema

README.md
```diff
@@ -51,33 +51,39 @@ source .env
 
 ### Start local Relay chain(alice and bob) and MRC(alice)
```
saraswatpuneet, line 53: Ignore README.md, init.sh, spec files my branch was pulled from older ref before bootnodes branch was merged so will fix conflicts

common/helpers/Cargo.toml
```diff
@@ -0,0 +1,26 @@
...
+thiserror = "1"
+sha2 = "0.10.2"
+
+[dev-dependencies]
+proptest = "1.0"
+
+[features]
+default = ['std']
+std = [
+]
```
saraswatpuneet, line 26: Having this common/helper crate for any helper mods/functions etc.

common/helpers/src/avro.rs
saraswatpuneet: ? Handles the panic and return error no need to panic_unwind

common/helpers/src/avro.rs
```diff
@@ -0,0 +1,235 @@
...
+/// * `record` - record to be written as HashMap<String, SchemaValue> where SchemaValue is common types
+/// # Examples
+/// ```
+/// use avro::*;
+/// let schema = r#"{"type": "record", "name": "User", "fields": [{"name": "name", "type": "string"}, {"name": "favorite_number", "type": "int"}]}"#;
+/// let record = vec![("name".to_string(), SchemaValue::String("John".to_string())), ("favorite_number".to_string(), SchemaValue::Int(42))];
+/// let schema_fingerprint = fingerprint_raw_schema(schema);
+/// assert!(schema_fingerprint.is_ok());
+/// let serialized_record = populate_schema_and_serialize(schema_fingerprint.unwrap().1, record);
+pub fn populate_schema_and_serialize(
```
aramikm, line 163: It seems like you can reuse already existing functions `get_schema_data_writer` and `populate_records`

common/helpers/src/avro.rs
```diff
@@ -0,0 +1,235 @@
...
+	schema: &Schema,
+	record: &HashMap<String, SchemaValue>,
+) -> Result<Vec<u8>, AvroError> {
+	let mut writer = Writer::with_codec(schema, Vec::new(), Codec::Snappy);
+	let mut record_list = Record::new(writer.schema()).unwrap();
+	for (field_name, field_value) in record.iter() {
+		record_list.put(field_name, field_value.clone());
+	}
+	writer.append(record_list)?;
```
aramikm, line 172: So one of the main reasons to use a binary serialization scheme like `avro` was to compress the stored message.
This library as far as I've checked always writes header (which includes schema) when serializing. This actually increases the size of each individual message instead of decreasing it. We need a way to avoid serializing the schema on each write.

https://github.com/apache/avro/blob/master/lang/rust/avro/src/writer.rs#L94
https://github.com/apache/avro/blob/master/lang/rust/avro/src/writer.rs#L325

common/helpers/src/avro.rs
```diff
@@ -0,0 +1,235 @@
...
+/// let serialized_record = populate_schema_and_serialize(schema_fingerprint.unwrap().1, record);
+/// let record_map = get_schema_data_map(schema_fingerprint.unwrap().1, serialized_record.unwrap());
+pub fn get_schema_data_map<'a>(
+	serialized_data: &'a Vec<u8>,
+	schema: &'a Schema,
+) -> Result<HashMap<String, SchemaValue>, AvroError> {
+	let reader = get_schema_data_reader(serialized_data, schema)?;
+	let mut result_record = HashMap::<String, SchemaValue>::new();
+	for record in reader {
+		let record_value = record.unwrap();
```
aramikm, line 223: I think this will panic instead of returning an error result

aramikm (commented):
Added some questions and comments

common/helpers/Cargo.toml
```diff
@@ -0,0 +1,26 @@
...
+name = 'common-helpers'
+publish = false
+repository = "https://github.com/liberty30/mrc/"
+version = "0.1.0"
+
+[package.metadata.docs.rs]
+targets = ['x86_64-unknown-linux-gnu']
+
+[dependencies]
+apache-avro = { git = "https://github.com/apache/avro", features = ["snappy"] }
```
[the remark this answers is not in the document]
saraswatpuneet: Updated dependency as suggested in docs here https://github.com/apache/avro/tree/master/lang/rust

common/helpers/src/avro.rs
saraswatpuneet: Good point, refactored this

common/helpers/src/avro.rs
saraswatpuneet: Good point, serialized data can be bad or wrong schema is provided i added ? and wrote a negative test

common/helpers/src/avro.rs
saraswatpuneet: @aramikm 
https://github.com/apache/avro/blob/75e74ee45c6a6f250a8f1321537e0fbb6c97517d/lang/rust/avro/src/writer.rs#L94

by default has_header is false i suppose and hence it will skip writing header by default

saraswatpuneet: I think you are right, by default it sets the header, and looks like setter on has_header is set to skip

saraswatpuneet: @aramikm  this looks like does our job , what you think

https://github.com/apache/avro/blob/75e74ee45c6a6f250a8f1321537e0fbb6c97517d/lang/rust/avro/src/writer.rs#L382

saraswatpuneet: @aramikm  i think this is the answer https://github.com/apache/avro/blob/75e74ee45c6a6f250a8f1321537e0fbb6c97517d/lang/rust/avro/src/writer.rs#L382

We can safely use it for internal use, what you think?

saraswatpuneet: @aramikm  I found a way, there is a lesser known function to_avro_datum and from_avro_datum that will not write to header and since we are using for internal purposes we can enforce getter of data via this process, check the latest update to function ```get_schema_data_map```
Which is only data getter from a serialized_data and its schema

common/helpers/src/avro.rs
saraswatpuneet: This is the only getter which accepts a serialized data of which was converted to avro datum and load it back in schema and return hashmap

saraswatpuneet:
@wilwade  added a commit to fix doc tests and imports

common/helpers/src/avro.rs
```diff
@@ -0,0 +1,242 @@
...
+			let schema = fingerprint_raw_schema(r);
+			match schema {
+				Ok(schema) => schema,
+				Err(_error) => (Schema::Null, vec![]),
+			}
+		})
+		.collect();
+
+	Ok(schemas)
+}
```
saraswatpuneet, line 73: @wilwade  here using rayon

saraswatpuneet: Just because it parallel we might not throw error for list of schemas but set it to Null if schema is invalid, also good so we dont reject schemas just because one was broken

common/helpers/Cargo.toml
```diff
@@ -0,0 +1,27 @@
...
+thiserror = "1"
+sha2 = "0.10.2"
+rayon = "1.5"
+
```
saraswatpuneet, line 20: using rayon for parallel iterator

common/helpers/src/avro.rs
saraswatpuneet: using rayon parallel iterators

common/helpers/Cargo.toml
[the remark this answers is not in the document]
aramikm: I'm a bit confused about `apache-avro` and `avro-rs`. I think apache-avro existing in crates.io is discontinued and avro-rs is what people are using via crates.

common/helpers/src/avro.rs
aramikm: This would be great. Now the only thing that I'm curious about it how smaller is a data serialized via headerless avro vs it's json but we can do a check later.

common/helpers/Cargo.toml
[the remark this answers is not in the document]
saraswatpuneet: Avro-rs got archived by developer and moved under apache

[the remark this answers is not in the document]
aramikm: I see.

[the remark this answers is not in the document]
saraswatpuneet: But good reminder let me update it with crate version rather git

aramikm (approved):
Looks good to me!

common/helpers/Cargo.toml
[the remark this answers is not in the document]
saraswatpuneet: We figured they have not released latest crate so some functionality was missing that we needed, we will keep an eye on this library to update when new release happens

enddynayn:
Please squash all your commits into one easy to read commit.

common/helpers/src/avro.rs
```diff
@@ -0,0 +1,242 @@
...
+/// # Examples
+/// ```
+/// use common_helpers::avro;
+/// use common_helpers::types::*;
+/// let raw_schema = r#"{"type": "record", "name": "User", "fields": [{"name": "name", "type": "string"}, {"name": "favorite_number", "type": "int"}]}"#;
+/// let schema_result = avro::fingerprint_raw_schema(raw_schema);
+/// assert!(schema_result.is_ok());
+/// let serialized_schema = schema_result.unwrap().1;
+pub fn fingerprint_raw_schema(raw_schema: &str) -> Result<(Schema, Vec<u8>), AvroError> {
+	// parse_str will fail if schema is not valid
```
enddynayn, line 34: rm

common/helpers/src/avro.rs
```diff
@@ -0,0 +1,242 @@
...
+
+	let schema_result = Schema::parse_str(raw_schema)?;
+	let schema_canonical_form = schema_result.canonical_form();
+	// return the schema and the fingerprint
```
enddynayn, line 38: rm

common/helpers/src/avro.rs
```diff
@@ -0,0 +1,242 @@
...
+/// use common_helpers::types::*;
+/// let raw_schema = r#"{"type": "record", "name": "User", "fields": [{"name": "name", "type": "string"}, {"name": "favorite_number", "type": "int"}]}"#;
+/// let vec_raw_schema: [&str; 1] = [raw_schema];
+/// let schema_result = avro::fingerprint_raw_schema_list(&vec_raw_schema);
+/// assert!(schema_result.is_ok());
+/// let serialized_schemas = schema_result.unwrap().1;
+/// ```
+pub fn fingerprint_raw_schema_list(
+	raw_schema: &[&str],
+) -> Result<(Vec<Schema>, Vec<Vec<u8>>), AvroError> {
```
enddynayn, line 60: when does this error?

saraswatpuneet: it errors when the function call on fingerprint_raw_schema errors

PR #55, opened 2022-05-03
PR CI will cancel previous runs

wilwade:
Goal: Get tests to run faster

Solution in this PR: Stop the runners for prior commits in PRs.

### DIFF
--- a/.github/workflows/tests-pr.yml
+++ b/.github/workflows/tests-pr.yml
@@ -1,8 +1,9 @@
+name: Tests Pull Requests
+concurrency:
+  group: tests-${{ github.ref }}
+  cancel-in-progress: true
 on:
-  push:
-    branches: [main]
   pull_request:
-name: Tests
 jobs:
   tests:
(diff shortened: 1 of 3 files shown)


.github/workflows/tests-pr.yml
```diff
@@ -1,8 +1,9 @@
+name: Tests Pull Requests
+concurrency:
+  group: tests-${{ github.ref }}
+  cancel-in-progress: true
```
wilwade, line 4: This is what does the work for us. It cancels previous running tests and lints so pushing several times in a row doesn't flood the queue.

.github/workflows/tests-pr.yml
```diff
@@ -1,8 +1,9 @@
...
 on:
-  push:
-    branches: [main]
```
wilwade, line 3: There was no (nice) way to have the concurrency only apply to non-main branches, so moved main branch lint and test over to the main branch workflow.

Issue #24, opened 2022-04-22
[MRC Message Store] Network consumer should be able to fetch messages on chain


shannonwells:
As a consumer of the network, I want to be able to 
* Retrieve all messages on chain
* for a given schema
* within a block range
* and be able to page the results

Blocked by #27

# Description
See ["Read"](https://github.com/Liberty30/mrc/blob/main/designdocs/MESSAGE_STORAGE.md#read) section of Design Doc, implement endpoint with checks specified in the document

# Acceptance Criteria
- [x] Expose custom RPC that lets user retrieve the range of messages, given:
    * starting block number
    * ending block number
    * schema id
    * page
    * page size
- [x] tests for same

PR #57, opened 2022-05-05
add rpc to get messages issue #24

aramikm:
# Problem
Described in https://github.com/LibertyDSNP/mrc/issues/24

# What
- [x] Expose custom RPC that lets user retrieve the range of messages
- [x] minor optimisation `on_initialize`
- [x] tests for same

# Verification Evidence
- after creating messages was able to successfully get messages

Screen Shot 2022-05-05 at 1 39 12 PM

Screen Shot 2022-05-05 at 1 40 20 PM

Screen Shot 2022-05-05 at 1 38 49 PM


pallets/messages/rpc/src/lib.rs
```diff
@@ -0,0 +1,77 @@
...
+		let runtime_api_result = api.get_messages_by_schema(&at, schema_id, pagination);
+		map_result(runtime_api_result)
+	}
+}
+
+fn map_result<T>(response: CoreResult<CoreResult<T, DispatchError>, ApiError>) -> Result<T> {
+	match response {
+		Ok(Ok(res)) => Ok(res),
+		Ok(Err(DispatchError::Module(e))) => Err(RpcError {
+			code: ErrorCode::ServerError(100), // No real reason for this value
```
aramikm, line 62: We have to decide about what error codes to use here for different type of failures

pallets/messages/src/lib.rs
```diff
@@ -174,6 +254,10 @@ impl<T: Config> Pallet<T> {
 		let message_count = block_messages.len() as u32;
 		let mut schema_count = 0u32;
 
+		if message_count == 0 {
```
aramikm, line 257: optimization suggested by @shannonwells

pallets/messages/rpc/src/lib.rs
saraswatpuneet: Good point especially separating out a errors.rs or something in common primitives

node/src/rpc.rs
```diff
@@ -51,7 +52,9 @@ where
 	let FullDeps { client, pool, deny_unsafe } = deps;
 
 	io.extend_with(SystemApi::to_delegate(FullSystem::new(client.clone(), pool, deny_unsafe)));
-	io.extend_with(TransactionPaymentApi::to_delegate(TransactionPayment::new(client)));
-
+	io.extend_with(TransactionPaymentApi::to_delegate(TransactionPayment::new(client.clone())));
+	io.extend_with(pallet_messages_rpc::MessagesApi::to_delegate(
+		pallet_messages_rpc::MessagesHandler::new(client.clone()),
+	));
```
saraswatpuneet, line 58: Should we group out mrc related handlers here, since schema and other pallets will expose rpc

pallets/messages/runtime-api/src/lib.rs
```diff
@@ -0,0 +1,23 @@
...
+		AccountId: Codec,
+		BlockNumber: Codec,
+	{
+		fn get_messages_by_schema(schema_id: SchemaId,
+			pagination: BlockPaginationRequest<BlockNumber>) ->
+		Result<
+			BlockPaginationResponse<BlockNumber, MessageResponse<AccountId, BlockNumber>>,
+			DispatchError,
+		>;
+	}
```
saraswatpuneet, line 22: How are we ensuring it is signed and paid for? I been thinking how we can execute an extrinsic via rpc (to do small offchain stuff)
Like schemas will submitted to an rpc that will quickly validated and submit to register_schema extrinsic instead of public function

saraswatpuneet (approved):
Pretty standard and neat! just few comments to think about rpc submitting to extrinsic which is believe is done by payments pallet to calculate fees of given rpc

node/src/rpc.rs
aramikm: Sounds good. Going to apply for additional RPC

pallets/messages/rpc/src/lib.rs
aramikm: Sounds good. Going to apply for additional RPCs

pallets/messages/runtime-api/src/lib.rs
aramikm: That is a good question. So read-only RPC's like this one don't need to be signed or paid for IMO since it is not writing any data on-chain. We only need some kind of `quota limits` implementation using offchain db to ensure this is not going to get abused.

All extrinsics are queued using the following. I think looking into how this is implemented is going to be useful
```
	#[rpc(name = "author_submitExtrinsic")]
	fn submit_extrinsic(&self, extrinsic: Bytes) -> FutureResult<Hash>;
```

Issue #54, opened 2022-05-02
[MRC Message Store] Replace Vec with BoundedVec


aramikm:
As a best practice in Substrate we need to replace all `Vec` used in storage with `Boundedvec`

A new problem not seen before caused to split this into a new issue.

Acceptance Critera:
--------------------

- [x] Replace all `Vec` used in `Messages` pallet with `BoundedVec`

Issue #53, opened 2022-05-02
[MRC Message Store] Use Msa Pallet to fetch Msa account


aramikm:
We need to use Msa pallet to get Msa account from origin

Acceptance criteria
------------------

- [x] Use Msa pallet via loosely coupling
- [x] getting Msa account from origin and use it in message creation (`add` extrinsic)
- [x] update tests
- [x] update benchmarks
- [x] generate spec

aramikm:
Implemented inside https://github.com/LibertyDSNP/mrc/pull/51

PR #65, opened 2022-05-06
replace vec with boundedvec issue #54

aramikm:
# Problem
Replace Vec with BoundedVec issue #53

# What
- replaced vec with BoundedVec
- fixed the tests

# Notes
- The issue was mostly due to using `u16` instead of `u32`. `Decode` Trait is only implemented for `Get<T : u32>`

### DIFF
--- a/pallets/messages/src/lib.rs
+++ b/pallets/messages/src/lib.rs
@@ -22,7 +22,7 @@ pub mod weights;
 
 mod types;
 
-use frame_support::{dispatch::DispatchResult, ensure, pallet_prelude::Weight, traits::Get};
+use frame_support::{dispatch::DispatchResult, ensure, pallet_prelude::Weight, traits::Get, BoundedVec};
 use sp_runtime::{traits::One, DispatchError};
 use sp_std::{collections::btree_map::BTreeMap, convert::TryInto, prelude::*};
 
@@ -51,7 +51,7 @@ pub mod pallet {
 
 		/// The maximum number of messages in a block
 		#[pallet::constant]
-		type MaxMessagesPerBlock: Get<u16>;
+		type MaxMessagesPerBlock: Get<u32>;
 
 		/// The maximum size of a message [Byte]
 		#[pallet::constant]
@@ -65,8 +65,14 @@ pub mod pallet {
 
 	#[pallet::storage]
 	#[pallet::getter(fn get_block_messages)]
-	pub(super) type BlockMessages<T: Config> //TODO: convert Vec to BoundedVec
-		= StorageValue<_, Vec<(Message<T::AccountId>, SchemaId)>, ValueQuery>;
+	pub(super) type BlockMessages<T: Config> = StorageValue<
+		_,
+		BoundedVec<
+			(Message<T::AccountId, T::MaxMessageSizeInBytes>, SchemaId),
+			T::MaxMessagesPerBlock,
+		>,
+		ValueQuery,
+	>;
 
 	#[pallet::storage]
 	#[pallet::getter(fn get_messages)]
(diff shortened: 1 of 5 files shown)


pallets/messages/src/mock.rs
```diff
@@ -54,10 +55,37 @@ impl system::Config for Test {
 }
 
 parameter_types! {
-	pub const MaxMessagesPerBlock: u16 = 500;
+	pub const MaxMessagesPerBlock: u32 = 500;
 	pub const MaxMessageSizeInBytes: u32 = 100;
 }
 
+impl std::fmt::Debug for MaxMessagesPerBlock {
```
aramikm, line 62: These traits was needed in tests. So I had to implement them

pallets/messages/src/lib.rs
```diff
@@ -138,24 +146,23 @@ pub mod pallet {
...
 				.try_into()
 				.map_err(|_| Error::<T>::TypeConversionOverflow)?;
-			ensure!(
-				current_size < T::MaxMessagesPerBlock::get(),
-				Error::<T>::TooManyMessagesInBlock
-			);
 
 			let m = Message {
-				data: message,
+				data: message.try_into().map_err(|_| Error::<T>::TooLargeMessage)?,
```
enddynayn, line 156: I think you already have this validation above? Why do you need it again in here?

pallets/messages/src/lib.rs
```diff
@@ -138,24 +146,23 @@ pub mod pallet {
 
 			// TODO: validate schema existence and validity from schema pallet
 
-			let mut block_messages = <BlockMessages<T>>::get();
+			let mut block_messages = <BlockMessages<T>>::get().into_inner();
```
enddynayn, line 149:
```







```
BlockMessage::try_mutate(|messages| -> DispatchResult {
   messages.try_push(message).map_err(|_| Error::<T>::TooManyMessagesInBlock)?
  Ok(())
})?
```
Also, why do we need to store the current size in each element? I don't think we are using that index?

File: pallets/messages/src/lib.rs [hunk 4]
```diff
@@ -51,7 +53,7 @@ pub mod pallet {
 
 		/// The maximum number of messages in a block
 		#[pallet::constant]
-		type MaxMessagesPerBlock: Get<u16>;
+		type MaxMessagesPerBlock: Get<u32>;
```
shannonwells, line 56: What motivated this decision?

shannonwells (approved):
Just the one question, nothing blocking.

pallets/messages/src/mock.rs
enddynayn: I dont know why we need this. When we actually need to use it we do T:: MaxMessagesPerBlock::get().
so I needed to print something out
```
let a  = T:: MaxMessagesPerBlock::get()
println!("hello {:?}", a);
```

pallets/messages/src/lib.rs
aramikm: So the reason is that `Decode` Trait is only implemented for Get<T : u32> when used in `BoundedVec`. So if you are using `Get<16>` to define the max bound for any storage it complains that `Decode` is not implemented for the storage.

https://github.com/paritytech/substrate/blob/polkadot-v0.9.18/frame/support/src/storage/bounded_vec.rs#L89

pallets/messages/src/mock.rs
aramikm: It does not work in tests because it needs the `Debug` trait to be implemented in case you are checking 2 objects inside a `assert_eq!` And I'm checking these objects quality inside `add_message_should_store_message_on_temp_storage`

pallets/messages/src/lib.rs
aramikm: Redundant. Removed!

pallets/messages/src/lib.rs
aramikm: Size is for storing total order. I will modify to use `try_mutate`

pallets/messages/src/mock.rs
aramikm: Actually we just need those only for `MaxMessageSizeInBytes` so I'm removing the `MaxMessagesPerBlock` implementations

PR #68, opened 2022-05-09
Feat/register schema 39

shannonwells:
## Problem
Closes #68, schema registration
Includes PR #58 RPC basic layout

* Schema pallet with register_schema extrinsic, no validation except for size
* Includes basic benchmark
* Tests pulled from polkadot-experiments

with @saraswatpuneet


pallets/schemas/src/lib.rs
```diff
@@ -0,0 +1,143 @@
...
+		StorageMap<_, Twox64Concat, SchemaId, BoundedVec<u8, T::MaxSchemaSize>, ValueQuery>;
+
+	#[pallet::call]
+	impl<T: Config> Pallet<T> {
+		#[pallet::weight((10_000, Pays::Yes))]
+		pub fn register_schema(origin: OriginFor<T>, schema: Vec<u8>) -> DispatchResult {
+			let sender = ensure_signed(origin.clone())?;
+
+			let cur_count = Self::schema_count();
+			ensure!(cur_count < T::MaxSchemaRegistrations::get(), <Error<T>>::TooManySchemas);
```
enddynayn, line 101: how do we determine this? I though we did not have a limit on schemas?

saraswatpuneet: U16 type on schemaId will have max around 65k before it starts overwriting so thats our hard limit by default though I believe we will need governance and policies to retire schemas or remove stale ones to add new ones I believe

pallets/schemas/src/lib.rs
```diff
@@ -0,0 +1,143 @@
...
+			Ok(cur_count)
+		}
+
+		pub fn require_valid_schema_size(
+			schema: Vec<u8>,
+		) -> Result<BoundedVec<u8, T::MaxSchemaSize>, Error<T>> {
+			let bounded_fields: BoundedVec<u8, T::MaxSchemaSize> =
+				schema.try_into().map_err(|()| Error::<T>::TooLongSchema)?;
+			ensure!(
+			bounded_fields.len() >= T::MinSchemaSize::get() as usize,
```
enddynayn, line 137: why do we check the min schema size?

pallets/schemas/src/lib.rs
```diff
@@ -0,0 +1,143 @@
...
+	impl<T: Config> Pallet<T> {
+		#[pallet::weight((10_000, Pays::Yes))]
+		pub fn register_schema(origin: OriginFor<T>, schema: Vec<u8>) -> DispatchResult {
+			let sender = ensure_signed(origin.clone())?;
+
+			let cur_count = Self::schema_count();
+			ensure!(cur_count < T::MaxSchemaRegistrations::get(), <Error<T>>::TooManySchemas);
+			let schema_id = cur_count.checked_add(1).ok_or(<Error<T>>::TooManySchemas)?;
+
+			Self::require_valid_schema_size(schema.clone())?;
```
enddynayn, line 104: maybe add the require_valid_schema_size inside the add_schema function.

enddynayn: I think you can remove this line.

pallets/schemas/src/lib.rs
```diff
@@ -0,0 +1,143 @@
...
+
+	#[pallet::storage]
+	#[pallet::getter(fn schema_count)]
+	pub(super) type SchemaCount<T: Config> = StorageValue<_, SchemaId, ValueQuery>;
+
+	// storage for message schemas hashes
+	#[pallet::storage]
+	#[pallet::getter(fn get_schema)]
+	pub(super) type Schemas<T: Config> =
+		StorageMap<_, Twox64Concat, SchemaId, BoundedVec<u8, T::MaxSchemaSize>, ValueQuery>;
```
saraswatpuneet, line 92: can we declare a "Schema" type Vec<u8> in common primitives? so it can be referenced by others

pallets/schemas/src/lib.rs
```diff
@@ -0,0 +1,143 @@
...
+
+			let cur_count = Self::schema_count();
+			ensure!(cur_count < T::MaxSchemaRegistrations::get(), <Error<T>>::TooManySchemas);
+			let schema_id = cur_count.checked_add(1).ok_or(<Error<T>>::TooManySchemas)?;
+
+			Self::require_valid_schema_size(schema.clone())?;
+			Self::add_schema(schema.clone(), schema_id)?;
+
+			let current_block = <frame_system::Pallet<T>>::block_number();
+			Self::deposit_event(Event::SchemaRegistered(sender, schema_id, current_block));
```
enddynayn, line 108: why do we need to log out the current_block?

enddynayn (approved):

pallets/schemas/src/lib.rs
enddynayn: Should MaxSchemaRegistrations go in the bounded vector?

pallets/schemas/src/lib.rs
```diff
@@ -0,0 +1,143 @@
...
+
+	#[pallet::call]
+	impl<T: Config> Pallet<T> {
+		#[pallet::weight((10_000, Pays::Yes))]
+		pub fn register_schema(origin: OriginFor<T>, schema: Vec<u8>) -> DispatchResult {
+			let sender = ensure_signed(origin.clone())?;
+
+			let cur_count = Self::schema_count();
+			ensure!(cur_count < T::MaxSchemaRegistrations::get(), <Error<T>>::TooManySchemas);
+			let schema_id = cur_count.checked_add(1).ok_or(<Error<T>>::TooManySchemas)?;
```
enddynayn, line 102: This is an overflow error than not a TooManySchemas?

enddynayn (approved):

pallets/schemas/src/lib.rs
shannonwells: I guess we don't, I was just thinking it might be helpful for consumers to know when a schema became active before sending messages with it.

pallets/schemas/src/lib.rs
shannonwells: @enddynayn No that's the max size for the schema data.  Almost all of this code is straight from polkadot-experiments.

shannonwells: @enddynayn :  Puneet is right, this was part of the design doc, the limit is the Uint16 bounds.  We discussed the possibility to change it later.

pallets/schemas/src/lib.rs
shannonwells: I don't know why we'd draw the distinction though, either way it's too many.

pallets/schemas/src/lib.rs
shannonwells: That was a requirement that zero-length should fail but I did not set it in the config to zero.  Anyway, there will be a minimum size under which it's not even possible to have a valid schema, and I think we should use that as a quick validation failure once we determine what that is.

shannonwells:
I'm going to put this back in a WIP state as I realized I forgot a couple of things.

pallets/schemas/src/lib.rs
shannonwells: good thought.

pallets/schemas/src/lib.rs
enddynayn: I think the bounded vector should determine how many schemas are allowed to be put in.

pallets/schemas/src/lib.rs
enddynayn: because these are two different errors:
```
ensure!(cur_count < T::MaxSchemaRegistrations::get(), <Error<T>>::TooManySchemas);

let schema_id = cur_count.checked_add(1).ok_or(<Error<T>>::TooManySchemas)?;
```

common/primitives/src/schema.rs
```diff
@@ -0,0 +1,5 @@
+use frame_support::BoundedVec;
+
+pub type SchemaId = u16;
```
aramikm, line 3: There is an existing `SchemaId` type defined in messages.rs. Please remove that and replace all of references with this one.

pallets/schemas/src/benchmarking.rs
```diff
@@ -0,0 +1,35 @@
...
+fn random_schema() -> Vec<u8> {
+	Vec::from("aasdf,sfdfb,eeec,dfsdfs,egsdkld,sldfklzz,sffs,sdjkl,lslsls,lfkjsf".as_bytes())
+}
+
+fn register_random_schema<T: Config>(sender: T::AccountId) -> DispatchResult {
+	SchemasPallet::<T>::register_schema(RawOrigin::Signed(sender).into(), random_schema())
+}
+
+benchmarks! {
+	register_schema {
```
aramikm, line 19: I think schema registration benchmark should be dependent on the schema size as a param. You can take a look at benchmark for `add` in messages pallet as an example.

pallets/schemas/src/lib.rs
shannonwells: After chatting we decided to keep both of these and change the second error to indicate an overflow even though (hopefully) that error is highly unlikely, it's a failsafe.

pallets/schemas/src/lib.rs
```diff
@@ -0,0 +1,140 @@
...
+	use common_primitives::schema::{Schema, SchemaId};
+	use frame_support::pallet_prelude::*;
+	use frame_system::pallet_prelude::*;
+
+	#[pallet::config]
+	pub trait Config: frame_system::Config {
+		type Event: From<Event<Self>> + IsType<<Self as frame_system::Config>::Event>;
+		type WeightInfo: WeightInfo;
+
+		// TODO: may need to be moved into MSA and import traits into schemas pallet.
```
aramikm, line 35: I'm not sure what is the use case for this. But in case of consuming MSA pallet we have a `AccountProvider` trait that can be used to get msa accounts from keys and it can easily extended for new functionalities.

I think we will need a new issue for any of existing todos to be able to track

pallets/schemas/src/lib.rs
```diff
@@ -0,0 +1,140 @@
...
+
+	// storage for message schemas hashes
+	#[pallet::storage]
+	#[pallet::getter(fn get_schema)]
+	pub(super) type Schemas<T: Config> =
+		StorageMap<_, Twox64Concat, SchemaId, Schema<T::MaxSchemaSize>, ValueQuery>;
+
+	#[pallet::call]
+	impl<T: Config> Pallet<T> {
+		#[pallet::weight((10_000, Pays::Yes))]
```
aramikm, line 98: should be replaced with calculated weights

pallets/schemas/src/rpc/src/lib.rs
```diff
@@ -0,0 +1,63 @@
...
+use jsonrpc_derive::rpc;
+use schemas_runtime_api::SchemasRuntimeApi;
+use sp_api::{ApiError, ProvideRuntimeApi};
+use sp_blockchain::HeaderBackend;
+use sp_runtime::{generic::BlockId, traits::Block as BlockT, DispatchError};
+use std::sync::Arc;
+
+#[rpc]
+pub trait SchemasApi<BlockHash, AccountId> {
+	#[rpc(name = "schema_get_latest_schema_id")]
```
aramikm, line 13: 🛑 Blocker:  this should be `schemas_getLatestSchemaId` based on how js library expects it

the convention is
`<pallet name>_<rpc name camel case>`

runtime/mrc/src/lib.rs
```diff
@@ -300,14 +301,33 @@ impl frame_system::Config for Runtime {
...
+impl pallet_schemas::Config for Runtime {
+	type Event = Event;
+	type WeightInfo = pallet_schemas::weights::SubstrateWeight<Runtime>;
+
+	// TODO: may need to be moved into MSA and import traits into schemas pallet.
+	type Signature = MultiSignature;
+	type Public = <MultiSignature as Verify>::Signer;
+
+	type MinSchemaSize = ConstU32<5>;
+	type MaxSchemaSize = ConstU32<300>;
```
aramikm, line 322: 300 bytes seems low? isn't it?

aramikm (changes requested):
Massive effort!
Added 1 blocker and a number of improvements.

runtime/mrc/src/lib.rs
shannonwells: Yes, it was 255 in experiments; I'm not sure what to put there and I feel like it should be discussed since we were talking about what limits.  It's just a place holder, so I'll put a TODO there and file an issue.

pallets/schemas/src/lib.rs
shannonwells: ah thanks, I meant to do this and got distracted with other changes.

pallets/schemas/src/lib.rs
shannonwells: This was meant to be removed anyway.

aramikm (approved):

PR #72, opened 2022-05-10
feat(Keys): feat(Keys): add ability to add multiple keys 

enddynayn:
Add ability to add additional keys to an MSA.
This provides users with ability to authorize
multiple devices and others.

Screen Shot 2022-05-10 at 4 32 08 PM


pallets/msa/Cargo.toml
```diff
@@ -20,14 +20,14 @@ frame-benchmarking = {default-features = false, git = "https://github.com/parity
 frame-support = {default-features = false, git = "https://github.com/paritytech/substrate.git", branch = "polkadot-v0.9.18"}
 frame-system = {default-features = false, git = "https://github.com/paritytech/substrate.git", branch = "polkadot-v0.9.18"}
 scale-info = {version = "2.0.0", default-features = false, features = ["derive"]}
+sp-core = {default-features = false, git = "https://github.com/paritytech/substrate.git", branch = "polkadot-v0.9.18"}
```
aramikm, line 23: is this used? If so it should be added under `features`

`"sp-core/std"`

pallets/msa/src/lib.rs
```diff
@@ -68,11 +82,35 @@ pub mod pallet {
 
 			Ok(())
 		}
+
+		#[pallet::weight(10_000)]
```
aramikm, line 86: If the benchmark is separated then we should have a separate benchmark issue for this to make sure it is tracked

pallets/msa/src/lib.rs
```diff
@@ -101,6 +139,26 @@ impl<T: Config> Pallet<T> {
...
+		Ok(())
+	}
+
+	pub fn verify_signature(
+		signature: MultiSignature,
+		signer: T::AccountId,
+		payload: Vec<u8>,
+	) -> DispatchResult {
+		let key = T::ConvertIntoAccountId32::convert(signer.clone());
+		let wrapped_payload = wrap_binary_data(payload);
```
aramikm, line 154: so we assume all the payloads should be wrapped for now

pallets/msa/src/tests.rs
```diff
@@ -19,29 +24,137 @@ fn it_throws_msa_identifier_overflow() {
...
+		let (new_msa_id, _) = Msa::create_account(account.into()).unwrap();
+
+		let new_key = key_pair_2.public();
+
+		let add_new_key_data = AddKeyData { nonce: 1, msa_id: new_msa_id };
+		let encode_data_new_key_data = wrap_binary_data(add_new_key_data.encode());
+
+		let signature: MultiSignature = key_pair_2.sign(&encode_data_new_key_data).into();
+
+		let _ = Msa::add_key_to_msa(
```
aramikm, line 151: using `assert_ok!` might make sense here

aramikm (approved):
Looks good, Added some comments but none was blocker.

pallets/msa/src/lib.rs
enddynayn: Yes.

pallets/msa/src/tests.rs
enddynayn: good idea.

Issue #34, opened 2022-04-22
Schema Cost Calculator


wilwade:
DEPENDENT: #28 

As a User, I want to know how much my schema is going to cost.

**Acceptance Criteria**
- RPC: `schema_estimator`
  - params: same as `register_schema`
  - Response: Tokens it would cost to register the schema (not counting tips)

saraswatpuneet:
Following two issues created out of this
1. #77  - Currency, its adapters, converters etc,
2. #79 - MRC weight to fee conversion etc.

PR #75, opened 2022-05-11
Add schema cost calculator and add runtime api 

saraswatpuneet:
Tasks
- [x] schemas runtime api integration with node got lost in merges, adding it back.
- [x] Add another rpc endpoint to return feedetails of extrinsic (register schema)
- [x] Add TODO statements in code to track future changes 

Tests:
 - [x] Tested two rpc endpoints in schemas pallet and they work as expected
 - [x] Unit tests for cost_calc_function 
 
 Closes: #34 
 Create Issue(s): #77


pallets/schemas/src/lib.rs
```diff
@@ -19,12 +24,16 @@ pub use pallet::*;
 pub mod weights;
 pub use weights::*;
 
+type BalanceOf<T> =
+	<<T as Config>::Currency as Currency<<T as frame_system::Config>::AccountId>>::Balance;
+
 #[frame_support::pallet]
```
saraswatpuneet, line 30: This is generic BalanceType, at some point we will have to define our token system for enabling payments for such things

saraswatpuneet: We will have to implement currency and adaptors for custom tokens

pallets/schemas/src/lib.rs
```diff
@@ -42,6 +51,10 @@ pub mod pallet {
 		// Maximum number of schemas that can be registered
 		#[pallet::constant]
 		type MaxSchemaRegistrations: Get<SchemaId>;
+
+		// TODO: Definition of Currency/Currency Adaptors for MRC etc.
+		// Issue: https://github.com/LibertyDSNP/mrc/issues/77
+		type Currency: Currency<Self::AccountId>;
 	}
```
saraswatpuneet, line 58: For now using current currency default will need changes according to mrc requirements

pallets/schemas/src/lib.rs
```diff
@@ -131,5 +147,14 @@ pub mod pallet {
...
 			Ok(bounded_fields)
 		}
+
+		pub fn calculate_schema_cost(schema: Vec<u8>) -> FeeDetails<BalanceOf<T>> {
+			// TODO weight to currency/token conversion needed
+			// reference: https://github.com/paritytech/substrate/blob/0ba251c9388452c879bfcca425ada66f1f9bc802/frame/transaction-payment/src/lib.rs#L445
+			// Issue: https://github.com/LibertyDSNP/mrc/issues/77
+			let schema_weight = T::WeightInfo::register_schema(schema.len() as u32);
+			let tip = 0u32.into();
+			FeeDetails { inclusion_fee: None, tip }
```
saraswatpuneet, line 157: Actual function calculating weights and there will be a converter needed to translate weights in mrc currency 
Have added issue #77

pallets/schemas/src/rpc/src/lib.rs
```diff
@@ -3,15 +3,45 @@ use core::result::Result as CoreResult;
...
+pub trait SchemasApi<BlockHash, AccountId, ResponseType> {
 	#[rpc(name = "schemas_getLatestSchemaId")]
 	fn get_latest_schema_id(&self, at: Option<BlockHash>) -> Result<u16>;
+	#[rpc(name = "schemas_calculateSchemaFee")]
+	fn get_schema_registration_fee(
+		&self,
+		at: Option<BlockHash>,
+		schema: Vec<u8>,
+	) -> Result<FeeDetails<NumberOrHex>>;
 }
```
saraswatpuneet, line 46: RPC call for getting schema registration fee as FreeDetails type from balances pallet or we can have custom returns .

pallets/schemas/src/rpc/src/lib.rs
```diff
@@ -25,20 +55,57 @@ impl<C, M> SchemasHandler<C, M> {
...
+					base_fee: try_into_rpc_balance(inclusion_fee.base_fee)?,
+					len_fee: try_into_rpc_balance(inclusion_fee.len_fee)?,
+					adjusted_weight_fee: try_into_rpc_balance(inclusion_fee.adjusted_weight_fee)?,
+				})
+			} else {
+				None
+			},
+			tip: Default::default(),
+		})
+	}
```
saraswatpuneet, line 108: Gets the schema cost from calculate_schema_cost implemented in Schemas pallet and translate into  return type

pallets/schemas/src/runtime-api/src/lib.rs
```diff
@@ -2,12 +2,18 @@
...
 sp_api::decl_runtime_apis! {
-	pub trait SchemasRuntimeApi<AccountId> where
-	AccountId: codec::Codec
+	pub trait SchemasRuntimeApi<AccountId, Balance> where
+	AccountId: codec::Codec,
+	Balance: Codec + MaybeDisplay,
 	{
 		fn get_latest_schema_id() -> Result<u16, DispatchError>;
+		fn calculate_schema_cost(schema: Vec<u8>) -> FeeDetails<Balance>;
 	}
```
saraswatpuneet, line 18: runtime api call

runtime/mrc/src/lib.rs
```diff
@@ -324,6 +324,9 @@ impl pallet_schemas::Config for Runtime {
 	type MinSchemaSizeBytes = ConstU32<5>;
 	type MaxSchemaSizeBytes = ConstU32<4096>;
 	type MaxSchemaRegistrations = MaxSchemaRegistrations;
+	// TODO: Currency for MRC and adapters needs to be implemented.
+	// See Issue: https://github.com/LibertyDSNP/mrc/issues/77
+	type Currency = Balances;
```
enddynayn, line 329: where is this used?

pallets/schemas/src/lib.rs
enddynayn: can this go in the rpc side of things? Can we flip the dependency? I think all we need from here is the weights for the rpc?

runtime/mrc/src/lib.rs
```diff
@@ -666,11 +668,14 @@ impl_runtime_apis! {
...
 	}
 
-	// TODO should this be here or under a collection of mrc runtime apis?
-	impl pallet_schemas_runtime_api::SchemasRuntimeApi<Block, AccountId> for Runtime {
+	impl pallet_schemas_runtime_api::SchemasRuntimeApi<Block, AccountId, Balance> for Runtime {
 		fn get_latest_schema_id() -> Result<u16, DispatchError> {
 			Schemas::get_latest_schema_id()
 		}
+
+		fn calculate_schema_cost( schema:Vec<u8>) -> pallet_transaction_payment::FeeDetails<Balance> {
```
enddynayn, line 676: spacing

enddynayn (approved):

pallets/schemas/src/lib.rs
enddynayn: I dont see this being used

saraswatpuneet: It is used to define type BalanceOf right above

pallets/schemas/Cargo.toml
```diff
@@ -26,13 +26,15 @@ sp-io = {default-features = false, git = "https://github.com/paritytech/substrat
 sp-runtime = {default-features = false, git = "https://github.com/paritytech/substrate.git", branch = "polkadot-v0.9.18"}
 sp-std = {default-features = false, git = "https://github.com/paritytech/substrate.git", branch = "polkadot-v0.9.18"}
 sp-core = {default-features = false, git = "https://github.com/paritytech/substrate.git", branch = "polkadot-v0.9.18"}
+pallet-transaction-payment = {git = "https://github.com/paritytech/substrate", default-features = false, branch = "polkadot-v0.9.18"}
```
aramikm, line 29: should this be added under `features` like
`pallet-transaction-payment/std` ?

pallets/schemas/src/lib.rs
aramikm: One more generic question. If we are just going to use the `len` for calculation does it make sense to just get the data size from user instead of the whole schema? It will reduce used resources on the node but it will also might cause some discrepancy in case the user does not send the correct data len

saraswatpuneet: Yes it is done in pallet and at rpc side we want to ensure correct format is being returned because pallet and rpc (off chain) arent connected, rpc will just try to translate before sending it out

saraswatpuneet: @aramikm  good question, we might also run validation along with cost calculation when avro no_std is available (if it is available), also its safer to not give away raw numbers for a given input  len only, if someone does not have a schema should not be able to playground with this and we might reject in rpc if schema is not valid

pallets/schemas/Cargo.toml
saraswatpuneet: Added

runtime/mrc/src/lib.rs
saraswatpuneet: Done

pallets/schemas/src/lib.rs
```diff
@@ -131,5 +146,29 @@ pub mod pallet {
...
+			let base_fee = 0u32.into();
+			FeeDetails {
+				inclusion_fee: Some(InclusionFee {
+					base_fee: base_fee,
+					len_fee: fixed_len_fee,
+					adjusted_weight_fee: unadjusted_weight_fee,
+				}),
+				tip,
+			}
+		}
```
saraswatpuneet, line 172: This calculation needs to be refined, this is separated in issue #77 , also will be useful as a reference for other pallets needing  such computation

pallets/schemas/src/lib.rs
enddynayn: I think we should remove these dependencies from here.

common/primitives/src/weight_to_fees.rs
```diff
@@ -0,0 +1,29 @@
...
+		let p = MILLIUNIT / 10;
+		let q = 100 * Balance::from(ExtrinsicBaseWeight::get());
+		smallvec![WeightToFeeCoefficient {
+			degree: 1,
+			negative: false,
+			coeff_frac: Perbill::from_rational(p % q, q),
+			coeff_integer: p / q,
+		}]
+	}
+}
```
saraswatpuneet, line 29: @enddynayn  what you think now? Pulled it out in separate implementation, easy to tweak and shared

pallets/schemas/src/lib.rs
saraswatpuneet: Done

Issue #31, opened 2022-04-22
Get Schema Details


wilwade:
As a user, I want to be able to get the details about a given schema id.

**Acceptance Criteria**
- [x] RPC: get_schema
  - Might be custom or built in. tbd in implementation 
  - param: schema_id
  - reutrns: schema_id, serialized_schema, created_at (block number)
- [x] Tests

aramikm:
currently it seems like we are not storing the creation block number. I'm not sure if there is any efficient way to get it if we don't store it.

saraswatpuneet:
@wilwade  do we need block number? if so any reasons to know? like retiring or deprecating based on block number or based on if a schema is being used or not, like a stale schema

saraswatpuneet:
Or should we have a blocknumber in schema policy that gets updated by message store when the last message for that schema was stored, That way schema seeing no actions for N number of blocks can be marked as point of interest for possible retirement

wilwade:
@saraswatpuneet I guess we don't need the block number. We might need the expire block for #33, but we can worry about that then.

I cannot see a reason for a created at block number at this time.

PR #76, opened 2022-05-11
rpc: get schema details issue #31

aramikm:
# Problem
Described in https://github.com/LibertyDSNP/mrc/issues/31

# What
- [x] Expose custom RPC that lets user retrieve schema details
- [x] tests for same
- [x] generated new spec
- [x] pulled out common rpc mappings into common 
- [x] changed `Schema` to struct

# Verification Evidence

- getting an existing schema
Screen Shot 2022-05-11 at 4 17 53 PM

- getting non existing schema
Screen Shot 2022-05-11 at 4 18 21 PM


common/primitives/src/schema.rs
```diff
@@ -1,5 +1,19 @@
...
+use scale_info::TypeInfo;
+#[cfg(feature = "std")]
+use serde::{Deserialize, Serialize};
+use sp_std::prelude::*;
+#[cfg(feature = "std")]
+use utils::*;
 
 pub type SchemaId = u16;
 
-pub type Schema<T> = BoundedVec<u8, T>;
```
aramikm, line 5: we might need the `Schema` in here later but we can move it when we needed it in the worst case. I think `SchemaResponse` will reduce the requirement to access `Schema` directly, but let's see.

common/primitives/src/schema.rs
```diff
@@ -1,5 +1,19 @@
...
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+#[derive(Default, Clone, Encode, Decode, PartialEq, Debug, TypeInfo, Eq)]
+pub struct SchemaResponse {
+	pub schema_id: SchemaId,
+	#[cfg_attr(feature = "std", serde(with = "as_string"))]
```
aramikm, line 17: does this make sense or we want to encode `as_hex`?

pallets/schemas/src/types.rs
```diff
@@ -0,0 +1,23 @@
+use codec::{Decode, Encode};
+use common_primitives::schema::{SchemaId, SchemaResponse};
+use frame_support::{traits::Get, BoundedVec};
+use scale_info::TypeInfo;
+use sp_std::prelude::*;
+
+#[derive(Default, Clone, Encode, Decode, PartialEq, Debug, TypeInfo, Eq)]
+#[scale_info(skip_type_params(MaxDataSize))]
+pub struct Schema<MaxDataSize>
```
aramikm, line 9: since we are going to store other metadata like `expired_block` related to `Schema` I think it makes sense to use a struct.

pallets/schemas/src/rpc/src/lib.rs
```diff
@@ -1,17 +1,19 @@
...
-use sp_api::{ApiError, ProvideRuntimeApi};
+use sp_api::ProvideRuntimeApi;
 use sp_blockchain::HeaderBackend;
-use sp_runtime::{generic::BlockId, traits::Block as BlockT, DispatchError};
+use sp_runtime::{generic::BlockId, traits::Block as BlockT};
 use std::sync::Arc;
 
 #[rpc]
-pub trait SchemasApi<BlockHash, AccountId> {
+pub trait SchemasApi<BlockHash> {
```
enddynayn, line 11: Do we need BlockHash? Maybe I did not catch how we are using it.

pallets/schemas/src/mock.rs
```diff
@@ -30,16 +30,23 @@ frame_support::construct_runtime!(
...
 impl pallet_schemas::Config for Test {
 	type Event = Event;
 	type WeightInfo = ();
 	type MinSchemaSizeBytes = ConstU32<5>;
-	type MaxSchemaSizeBytes = ConstU32<100>;
+	type MaxSchemaSizeBytes = MaxSchemaSizeBytes;
 	type MaxSchemaRegistrations = MaxSchemaRegistrations;
 }
 
+impl Clone for MaxSchemaSizeBytes {
```
enddynayn, line 44: why do you need this?

common/primitives/src/schema.rs
enddynayn: I think hex is a bit more fitting?

pallets/schemas/src/lib.rs
```diff
@@ -86,7 +90,7 @@ pub mod pallet {
 	#[pallet::storage]
 	#[pallet::getter(fn get_schema)]
 	pub(super) type Schemas<T: Config> =
-		StorageMap<_, Twox64Concat, SchemaId, Schema<T::MaxSchemaSizeBytes>, ValueQuery>;
+		StorageMap<_, Twox64Concat, SchemaId, Schema<T::MaxSchemaSizeBytes>, OptionQuery>;
```
enddynayn, line 93: OptionQuery makes more sense here

pallets/schemas/src/lib.rs
```diff
@@ -120,6 +124,15 @@ pub mod pallet {
 			Ok(cur_count)
 		}
 
+		pub fn get_schema_by_id(
+			schema_id: SchemaId,
+		) -> Result<Option<SchemaResponse>, DispatchError> {
+			if let Some(schema) = Self::get_schema(schema_id) {
+				return Ok(Some(schema.map_to_response(schema_id)))
```
enddynayn, line 131: It not very clear to me on how we could return DispatchError?
maybe we just need `<Option<SchemaResponse>`

pallets/schemas/src/tests.rs
```diff
@@ -77,3 +77,35 @@ fn register_schema_id_deposits_events_and_increments_schema_id() {
...
+	new_test_ext().execute_with(|| {
+		// arrange
+		let sender: AccountId = 1;
+		let serialized_fields = Vec::from("foo,bar,bazz".as_bytes());
+		assert_ok!(SchemasPallet::register_schema(
+			Origin::signed(sender),
+			serialized_fields.clone()
+		));
+
+		// act
```
enddynayn, line 92: left over comments, there are a few in the file

enddynayn (approved):

pallets/schemas/src/lib.rs
```diff
@@ -86,7 +90,7 @@ pub mod pallet {
 	#[pallet::storage]
 	#[pallet::getter(fn get_schema)]
 	pub(super) type Schemas<T: Config> =
-		StorageMap<_, Twox64Concat, SchemaId, Schema<T::MaxSchemaSizeBytes>, ValueQuery>;
+		StorageMap<_, Twox64Concat, SchemaId, Schema<T::MaxSchemaSizeBytes>, OptionQuery>;
 
 	#[pallet::call]
 	impl<T: Config> Pallet<T> {
```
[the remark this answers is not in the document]
aramikm: I think based on conversation from Spec we will need to test it and see if it will be effective or not

pallets/schemas/src/lib.rs
aramikm: So I tried returning `<Option<SchemaResponse>` but my rpc was giving an unrelated error. I tried for a while but only changing the return type from `<Option<SchemaResponse>` to `Result<Option<SchemaResponse>, DispatchError>` solved it.

I think we might be able to come back and change this later if we have some extra time.

pallets/schemas/src/mock.rs
aramikm: For the Bounded vector to be cloned() the size type should also be clone able.

pallets/schemas/src/rpc/src/lib.rs
aramikm: it is used in previously written rpc `get_latest_schema_id`

pallets/schemas/src/tests.rs
aramikm: If you mean `Act` it is related to Arrange/Act/Assert pattern
https://automationpanda.com/2020/07/07/arrange-act-assert-a-pattern-for-writing-good-tests/

saraswatpuneet (approved):
Looks Good to me

pallets/schemas/src/lib.rs
enddynayn: does map_to_response return any error?

pallets/schemas/src/tests.rs
enddynayn: oh cool did not know. Thanks for sharing.

pallets/schemas/src/types.rs
enddynayn: I mean this would probably simplify the code that you added. But my real concern is that we do not do this. After todays discussion it looks like we are not clear on the direction we are going to take with this. Lets keep it simple for now.

pallets/schemas/src/rpc/src/lib.rs
enddynayn: We want to avoid to be able to query by blockhash right?

pallets/schemas/src/mock.rs
enddynayn: I have not seen other pallets do this when passing in the bound from the config.

common/primitives/src/schema.rs
shannonwells: lol @saraswatpuneet asked me to move it to there

pallets/schemas/src/lib.rs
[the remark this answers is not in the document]
shannonwells: @enddynayn that is the plan already to use it in the params

runtime/mrc/src/lib.rs
```diff
@@ -660,17 +674,14 @@ impl_runtime_apis! {
...
-	}
-
-	// TODO should this be here or under a collection of mrc runtime apis?
-	impl pallet_schemas_runtime_api::SchemasRuntimeApi<Block, AccountId> for Runtime {
+	impl pallet_schemas_runtime_api::SchemasRuntimeApi<Block> for Runtime {
 		fn get_latest_schema_id() -> Result<u16, DispatchError> {
 			Schemas::get_latest_schema_id()
 		}
+
+		fn get_by_schema_id(schema_id: SchemaId) -> Result<Option<SchemaResponse>, DispatchError> {
```
enddynayn, line 682: I think we can just return Option<SchemaResponse>

common/primitives/src/schema.rs
aramikm: we are not sure on this yet. Keeping it as string until it changes

pallets/schemas/src/types.rs
aramikm: Converting back to `BoundedVec` for the sake of simplicity for now

pallets/schemas/src/mock.rs
aramikm: removed

pallets/schemas/src/lib.rs
aramikm: returning `Option`

runtime/mrc/src/lib.rs
aramikm: Fixed

pallets/schemas/src/lib.rs
aramikm: Outdated comment

PR #78, opened 2022-05-12
feat(Key): add ability to revoke

enddynayn:
Add ability to revoke a key. Revoking a key does not
remove it from storage. Instead it expires the key and
does not allow for submitting transactions with it.

[issue-37](https://github.com/LibertyDSNP/mrc/issues/37)


saraswatpuneet (approved):
Neat! Dont have any questions on the code

pallets/msa/src/lib.rs
```diff
@@ -131,17 +154,20 @@ impl<T: Config> Pallet<T> {
 	}
 
 	pub fn add_key(msa_id: MessageSenderId, key: &T::AccountId) -> DispatchResult {
-		<KeyOwnerOf<T>>::try_mutate(key, |maybe_msa| {
+		<KeyInfoOf<T>>::try_mutate(key, |maybe_msa| {
 			ensure!(maybe_msa.is_none(), Error::<T>::DuplicatedKey);
 
-			*maybe_msa = Some(msa_id);
+			*maybe_msa =
+				Some(KeyInfo { msa_id, expired: T::BlockNumber::zero(), nonce: Zero::zero() });
```
saraswatpuneet, line 161: @enddynayn  any chance of possible panic here? if maybe_msa is nill pointer?

pallets/msa/src/lib.rs
```diff
@@ -68,6 +69,10 @@ pub mod pallet {
 		KeyVerificationFailed,
 		NotMsaOwner,
 		InvalidSignature,
+		NotKeyOwner,
+		NoKeyExist,
```
shannonwells, line 73:
```suggestion
		NoKeyExists,
```

pallets/msa/src/lib.rs
```diff
@@ -159,6 +185,33 @@ impl<T: Config> Pallet<T> {
...
 		Ok(())
 	}
+
+	pub fn revoke_key(key: &T::AccountId) -> DispatchResult {
+		KeyInfoOf::<T>::try_mutate(key, |maybe_info| -> DispatchResult {
+			let mut info = maybe_info.take().ok_or(Error::<T>::NoKeyExist)?;
+
+			ensure!(info.expired == T::BlockNumber::zero(), Error::<T>::KeyRevoked);
+
+			let current_block = frame_system::Pallet::<T>::block_number();
```
shannonwells, line 195: We don't want to allow a block number to be specified for the revocation?

pallets/msa/src/lib.rs
```diff
@@ -159,6 +185,33 @@ impl<T: Config> Pallet<T> {
...
+
+		Ok(())
+	}
+
+	pub fn try_get_key_info(key: &T::AccountId) -> Result<KeyInfo<T::BlockNumber>, DispatchError> {
+		let info = Self::get_key_info(key).ok_or(Error::<T>::NoKeyExist)?;
+		Ok(info)
+	}
+
+	pub fn get_owner_of(key: &T::AccountId) -> Option<MessageSenderId> {
```
shannonwells, line 212: This function isn't being used anywhere in this file right now.

pallets/msa/src/lib.rs
```diff
@@ -131,17 +154,20 @@ impl<T: Config> Pallet<T> {
...
+				Some(KeyInfo { msa_id, expired: T::BlockNumber::zero(), nonce: Zero::zero() });
 
 			Ok(())
 		})
 	}
 
-	pub fn is_msa_owner(who: &T::AccountId, msa_id: MessageSenderId) -> DispatchResult {
-		ensure!(Self::get_owner_of(&who) == Some(msa_id), Error::<T>::NotMsaOwner);
+	pub fn ensure_msa_owner(who: &T::AccountId, msa_id: MessageSenderId) -> DispatchResult {
+		let info = Self::get_key_info(who).ok_or(Error::<T>::NoKeyExist)?;
```
shannonwells, line 168: did you mean to use `get_owner_of` here?

pallets/msa/src/lib.rs
shannonwells, line 161: how is the nonce to be used?  It's not being incremented anywhere.

shannonwells (commented):
I'm not sure about the need for a nonce or the get_owner_of functions since neither is used.

pallets/msa/src/lib.rs
enddynayn: Its used below. The window needs to be expand.

pallets/msa/src/lib.rs
enddynayn: You are right. It was a suggestion made during planning.

pallets/msa/src/lib.rs
enddynayn: I did not, but now that you point it out I can use it.

pallets/msa/src/lib.rs
enddynayn: I think so.

pallets/msa/src/lib.rs
```diff
@@ -96,14 +101,32 @@ pub mod pallet {
...
+
+		#[pallet::weight(10_000)]
+		pub fn revoke_msa_key(origin: OriginFor<T>, key: T::AccountId) -> DispatchResult {
+			let who = ensure_signed(origin)?;
+
+			ensure!(who != key, Error::<T>::InvalidSelfRevoke);
+
+			let who = Self::try_get_key_info(&who)?;
+			let key_info = Self::try_get_key_info(&key)?;
+			ensure!(who.expired == T::BlockNumber::zero(), Error::<T>::KeyRevoked);
```
aramikm, line 121: nit: can it be BlockNumber::default() ?

shannonwells (approved):
All questions answered, thanks looks good to go, to me.

pallets/msa/src/lib.rs
aramikm: Nonce is to make the data which is going to be signed unique, so nobody can use that signature for something similar. Nonce should not get stored in here as far as I know.

pallets/msa/src/types.rs
```diff
@@ -8,3 +9,10 @@ pub struct AddKeyData {
 	pub msa_id: MessageSenderId,
 	pub nonce: u32,
 }
+
+#[derive(TypeInfo, Debug, Clone, Decode, Encode, PartialEq, Default, MaxEncodedLen)]
+pub struct KeyInfo<BlockNumber> {
+	pub msa_id: MessageSenderId,
+	pub nonce: u32,
```
aramikm, line 16: I don't think we need to store every nonce.

pallets/msa/src/lib.rs
enddynayn: I think there are great reasons to added but it was added a bit too soon.

pallets/msa/src/types.rs
enddynayn: I was not too crazy about the idea of adding nonce to start with. I thought it was a good idea but should have been left for another story. I do not see any issues with it being here since every key needs to have a nonce. Can you explain why this should go else where?

enddynayn: I don't think we even need nonce.

aramikm: I might be missing something but would you expand more on why every key needs to store a nonce? like how that nonce is going to be used later?

enddynayn: Keys are used for signing. Any of these keys can sign a transaction.

aramikm: Me and @enddynayn had a conversation about these and the TLDR was that since in capacity the keys are not used in the standard way that the substrate account keys are used, the nonce related logic is not taken care of for any of these keys to prevent **replay attacks** and we need to take care of these ourselves

aramikm (approved):
looks good!

PR #81, opened 2022-05-13
[Draft] Design Doc: Data Pruning

### LINKED ISSUE #74: [Design Doc] Storage
Design doc around pruning historical state

Basics
- While people could run archive nodes, we want to make it such that there is limited reason to do so.
- Why? Cost of running nodes. For comparison, Ethereum Archive node is at >10TB while a full node is < 1TB.
- Not forcing, but not required to read state (and any historical state could still be validated)
- Outcomes:
  - We store more information in state than otherwise need to, but total storage is lower.
  - Lower gossip data requests. Effectively full nodes have all the data you could need.
  - Schema Messages don't have to be permanent
- Trust model changes
  - Ethereum you can trust (like most people) the consensus, but you could also do your own validation and run an archive node.
  - MRC still offers that choice, but only from (about) the point in time that you started archiving. If you didn't start early, then you might have to trust that the consensus had the correct state at the point you start archiving.

wilwade:
Design doc around pruning and data storage
Closes #74

### DIFF
--- a/designdocs/pruning_data.md
+++ b/designdocs/pruning_data.md
@@ -1,0 +1,64 @@
+# Pruning Data
+
+MRC is focused on being able to send a high volume of messages.
+To this end, we must worry about the cost of storing those messages over time.
+As of right now, Ethereum has an [archive node](https://etherscan.io/chartsync/chainarchive) storage requirement of over 10TB of data.
+This makes running an archive node expensive to do.
+How is MRC going to solve this problem?
+
+## Where is Data?
+
+We have two data storage locations that we can adjust.
+First, we have the historical state in blocks.
+For Ethereum this would be log and transaction information.
+For MRC, it is extrinsic information and prior state information.
+
+Second, is the active state.
+For Ethereum, this would be balances, smart contracts, and smart contract state.
+For MRC, this is balances, schemas, messages, and to a lesser degree staking information and governance voting.
+
+## Historical Block Pruning
+
+Many blockchains hold that anyone should be able to do a complete validation of the chain from the genesis block.
+If you want to pull all 10TB+ of data down (after several restarts), you can test to make sure that each state transition of each transaction and block were applied correctly and that it matches the rest of the world.
+Or you could just be a "full node" pull the block headers, the current state, and the past few hundred blocks for about 1TB of data.
+The full node version trusts that the information prior to the few hundred blocks is good.
+
+MRC is structured such that we can survive without any historical block data beyond the last few hundred.
+Effectively, we don't require archive nodes.
+Someone could run one, or at least run one that starts from the time they started it, but the network functions without it.
+Also as we still maintain block headers, a historical block is still able to be validated.
+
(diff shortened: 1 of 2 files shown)


aramikm (approved):
I think this sums up the problem nicely

saraswatpuneet (approved):

designdocs/pruning_data.md
```diff
@@ -0,0 +1,64 @@
...
+If you want to pull all 10TB+ of data down (after several restarts), you can test to make sure that each state transition of each transaction and block were applied correctly and that it matches the rest of the world.
+Or you could just be a "full node" pull the block headers, the current state, and the past few hundred blocks for about 1TB of data.
+The full node version trusts that the information prior to the few hundred blocks is good.
+
+MRC is structured such that we can survive without any historical block data beyond the last few hundred.
+Effectively, we don't require archive nodes.
+Someone could run one, or at least run one that starts from the time they started it, but the network functions without it.
+Also as we still maintain block headers, a historical block is still able to be validated.
+
+Effectively, after a given point in time, we trust that a given block was validated correctly by consensus and by the [relay chain](https://polkadot.network/technology/).
```
shannonwells, line 32:
```suggestion
This means that after a given point in time, we trust that a given block was validated correctly by consensus and by the [relay chain](https://polkadot.network/technology/).
```

designdocs/pruning_data.md
```diff
@@ -0,0 +1,64 @@
...
+The relay chain's secure validation provides a greater degree of security for MRC without having to rely on the MRC's collator nodes.
+
+The outcome of this choice does mean that slightly more information is held in active state.
+Messages store additional information that would normally be stored in the transaction alone, because that transaction might not be retrievable after some time.
+The total storage require is still reduced.
```
shannonwells, line 37:
```suggestion
The total storage required is still reduced.
```

designdocs/pruning_data.md
```diff
@@ -0,0 +1,64 @@
...
+
+## Active State Pruning
+
+Since we have historical block pruning, we have to be extra careful what we prune from active state.
+Any information we prune, must be expected to be lost forever.
```
shannonwells, line 42:
```suggestion
Any information we prune, must be expected to be lost forever, however,
```

designdocs/pruning_data.md
```diff
@@ -0,0 +1,64 @@
...
+However that is also a benefit.
```
shannonwells, line 43:
```suggestion
that is also a benefit.
```

designdocs/pruning_data.md
```diff
@@ -0,0 +1,64 @@
...
+MRC can be structured such that selected information is forgotten.
+
+We leverage this with allowing schemas to be configured with a time to live or [TTL](https://en.wikipedia.org/wiki/Time_to_live) for messages.
```
shannonwells, line 46: Unclear what "this" means.

shannonwells (changes requested):
This isn't really a design document; it's too general and high level and is more of a starting point for discussions.  You could not generate implementation stories from this.  I'm not sure where it belongs instead, it just should not be in designdocs.

Issue #77, opened 2022-05-11
Spike: Extrinsic fee computation for MRC (a generic extensible pallet)


saraswatpuneet:
- [x] decouple fee computation in its own pallet to provide room for future updates and adjustments
- [x] Extend transaction payment with base implementation as well as rpc endpoint to compute mrc extrinsic cost 
- [x] Provide an endpoint to run schema validation checks  #67

saraswatpuneet:
https://github.com/paritytech/substrate/blob/e9d79daaa782f41680ada04719151371a737c6fb/frame/transaction-payment/src/lib.rs#L517

Above is an example for weighttofee and aspects of weight computation

saraswatpuneet:
https://docs.substrate.io/rustdocs/latest/frame_system/limits/struct.BlockWeights.html

PR #82, opened 2022-05-13
Validate incoming schema on cost calculation

saraswatpuneet:
- [x] Within rpc call to calculate schema cost, do a schema validation prior to on chain call
- [x] Validate schema to be a valid avro schema, using avro helper functions

Unit Tests:
- [x] pass a invalid schema and rpc should fail
- [x] pass a valid schema and rpc should pass

Part of #77


pallets/schemas/src/rpc/src/lib.rs
```diff
@@ -93,6 +94,20 @@ where
...
 		schema: Vec<u8>,
 	) -> Result<FeeDetails<NumberOrHex>> {
+		// Validate serialized avro schema
+		let schema_string = String::from_utf8(schema.clone()).map_err(|_| RpcError {
+			code: ErrorCode::InvalidParams,
+			message: format!("invalid avro schema"),
+			data: Some(format!("{:?}", schema).into()),
+		})?;
+		let finger_print = avro::fingerprint_raw_schema(&schema_string);
+		if finger_print.is_err() {
```
saraswatpuneet, line 104: Run schema validator on incoming serialized schema before computing cost, this enable user to know if their schema is valid before registration and cost associated if valid

aramikm (approved):
:white_check_mark:

PR #83, opened 2022-05-13
[Draft] messages: lightweight retention implementation

aramikm:
# Problem
We need to be able to prune messages that outside of their retention policy to maintain a manageable on-chain state size.

# Main Idea
To use circular arrays style override in `StorageMap`

### DIFF
--- a/designdocs/MESSAGE_STORAGE.md
+++ b/designdocs/MESSAGE_STORAGE.md
@@ -25,23 +25,12 @@ using [StorageDoubleMap](https://docs.substrate.io/rustdocs/latest/frame_support
 
 ### Main Storage types
 - **Messages**
-    - _Type_: `DoubleStorageMap<BlockNumber, SchemaId, BoundedVec<Message>>`
+    - _Type_: `DoubleStorageMap<BlockNumber, SchemaId, (BlockNumber, BoundedVec<Message>)>`
     - _Purpose_: Main structure To store all messages for a certain block number and schema id
 - **BlockMessages**
     - _Storage Type_: `StorageValue<BoundedVec<(Message, SchemaId)>>`
     - _Purpose_: A temporary storage to put each messages with it's schemaId for current block to
   increase write throughput
-- **RetentionPeriods**
-    - _Type_: `StorageMap<SchemaId, BlockNumber>`
-    - _Purpose_: To store the retention period for each SchemaId (allows future adjustments)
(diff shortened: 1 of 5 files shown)


pallets/messages/src/lib.rs
```diff
@@ -185,8 +191,14 @@ impl<T: Config> Pallet<T> {
...
+			let retention_block = Self::get_target_retention_block(block_number);
+			let tuple = <Messages<T>>::get(retention_block, schema_id);
+			if tuple.0 != block_number {
+				// this is a left-over block from retention and should be ignored!
+				continue
+			}
 
+			let list = tuple.1.into_inner();
 			let list_size: u32 =
 				list.len().try_into().map_err(|_| Error::<T>::TypeConversionOverflow)?;
```
saraswatpuneet, line 203: Naming variables with meaningful names instead of tuple, list would help readability

designdocs/MESSAGE_STORAGE.md
```diff
@@ -110,13 +99,22 @@ from `StartingBlockNumber` until is reaches one of values from `EndBlockNumber`
...
 
+#### Cleanup (Retention policy - light)
+The idea is to implement inserts using the same concept as a circular array. We will calculate a
+`target_block` which is going to be `current_block % retention_policy_block_size` and insert
+messages into this calculated target block. This will allow us to override messages which are
+already outside retention policy with new ones.
+
+Note: There are going to be some left-over blocks and messages that still needs another process to
+clean up.
+
```
saraswatpuneet, line 117: Some clarity around what are left over blocks and their significance, say along with this note

pallets/messages/src/lib.rs
```diff
@@ -58,6 +58,10 @@ pub mod pallet {
 		/// The maximum size of a message [Byte]
 		#[pallet::constant]
 		type MaxMessageSizeInBytes: Get<u32> + Clone;
+
+		/// Retention period for messages [Block]
+		#[pallet::constant]
+		type RetentionPeriodBlocks: Get<u32>;
 	}
```
saraswatpuneet, line 65: is this temporary? will it be coming from Schema pallet and variable at per schema basis if so we can just define a primitives with some constant value and mark todo for schema pallet with a new issue

saraswatpuneet (approved):
I like the approach, just simple comments would like to on see any pros to be listed in design doc taking this approach and potential impact of schemas implementation

PR #90, opened 2022-05-13
Designdoc/batching

### LINKED ISSUE #10: [Design Doc] Batched Messages
Blocked by Liberty30/mrc#9, https://github.com/Liberty30/polkadot-experiments/issues/27

**Notes:**
- Batches messages might need to be handled differently than on chain messages in Liberty30/mrc#9
- Or maybe not?
- The biggest difference is that these reference a batch file
- Do we need to have requirements that it be a parquet file?
- We should have requirements around where it can be stored (IPFS only for now? perhaps this is a different design doc?)
- Need to give the byte-length of the file for capacity transaction reasons as well as limiting super massive files or attacks

shannonwells:
Design document for creating and fetching batches. Closes #10 

Also fixed a couple of issues in the delegation design doc.


designdocs/batched_messages.md
```diff
@@ -0,0 +1,81 @@
...
+### Constants
+* `BatchPageSizeMax` - the maximum number of batch announcements that will be returned by the `get_batches` query
+* `BatchSizeMinBytes` - the minimum possible size of a valid batch file with 1 row
+
+### Enums
+* `MessageType` - all values specified in [DSNP Spec Announcement Types](https://spec.dsnp.org/DSNP/Announcements.html)
+
+### Types
+* `BatchAnnouncementParams<T:Config>`: generic
+    * `batch_url`:`str` the URL of the batch file.  Must be IPFS or HTTPS URL
```
shannonwells, line 39: I don't know if we want to use both IPFS and HTTPS.

designdocs/batched_messages.md
```diff
@@ -0,0 +1,81 @@
...
+### Constants
+* `BatchPageSizeMax` - the maximum number of batch announcements that will be returned by the `get_batches` query
+* `BatchSizeMinBytes` - the minimum possible size of a valid batch file with 1 row
+
+### Enums
+* `MessageType` - all values specified in [DSNP Spec Announcement Types](https://spec.dsnp.org/DSNP/Announcements.html)
+
+### Types
+* `BatchAnnouncementParams<T:Config>`: generic
+    * `batch_url`:`Vec` the URL of the batch file.  Must be IPFS or HTTPS URL. The URL must be correctly formatted.
```
shannonwells, line 39: This was a `&str`, @enddynayn suggested it be a `Vec` and now I wonder if it should be a `BoundedVec`, and if so what the bounds would be.

designdocs/batched_messages.md
```diff
@@ -0,0 +1,81 @@
...
+
+### Constants
+* `BatchPageSizeMax` - the maximum number of batch announcements that will be returned by the `get_batches` query
+* `BatchSizeMinBytes` - the minimum possible size of a valid batch file with 1 row
+
+### Enums
+* `MessageType` - all values specified in [DSNP Spec Announcement Types](https://spec.dsnp.org/DSNP/Announcements.html)
+
+### Types
+* `BatchAnnouncementParams<T:Config>`: generic
```
aramikm, line 38: I'm a bit confused about the relationship between `BatchAnnouncements`  and Regular `Messages`. Are these supposed to be encoded to Vec<u8> then stored via Messages pallet or do they get store separately?

designdocs/batched_messages.md
```diff
@@ -0,0 +1,81 @@
...
+    * `batch_url`:`Vec<u8>` the URL of the batch file.  Must be IPFS or HTTPS URL. The URL must be correctly formatted.
+    * `message_type`: `MessageType`  the type of messages in this batch
+    * `file_size`: `usize`, the size of the batch file, used to determine message fee.  Must be &gt;= the minimum possible DSNP batch file size.
+    * `file_hash`: `<T::Hash>`, the hash of the batch file. Must not be 0 hash
+    * `msa_id`: `MsaId`, the id to use for the announcer.  Must exist and be active
+
+* `BatchAnnouncement`: implements `BatchAnnouncementParams`, returned by `get_batches`
+
+* `BatchAnnouncementOptions<T:Config>`: struct
+    * `msa_id`:  `MsaId`, the announcer's MSA id.  If 0, get all announcements
```
aramikm, line 48: seems like this is better to be `Option`

designdocs/batched_messages.md
```diff
@@ -0,0 +1,81 @@
...
+    * `block_number`: `<T::BlockNumber>`, retrieve messages starting at the given block number (inclusive). If 0, get all messages
```
aramikm, line 49: I think using default of any param if not set (here `0`) to get all is dangerous. I don't think there should be a get all option since depend on the size stored it can be huge.

For pagination via block number I've defined a request and response. Please take a look if makes sense to use something like that.
https://github.com/LibertyDSNP/mrc/blob/main/common/primitives/src/messages.rs#L26-L58

aramikm (commented):
A number of questions and suggestions!

enddynayn (approved):

designdocs/batched_messages.md
shannonwells: I assumed this would take the parameters and make a Message out of them and would store via Messages pallet.

designdocs/batched_messages.md
shannonwells: I have some questions about your implementation and also about paging data from the db but it'll be quicker to talk about it in Zoom.

designdocs/batched_messages.md
aramikm: So in that case you'll also need a schema to be able to store on `Messages`. It can come from a configuration.

designdocs/batched_messages.md
aramikm: Sure. Let me know

saraswatpuneet (approved):
:100:

designdocs/batched_messages.md
```diff
@@ -0,0 +1,81 @@
...
+* `BatchAnnouncementParams<T:Config>`: generic
+    * `batch_url`:`Vec<u8>` the URL of the batch file.  Must be IPFS or HTTPS URL. The URL must be correctly formatted.
+    * `message_type`: `MessageType`  the type of messages in this batch
+    * `file_size`: `usize`, the size of the batch file, used to determine message fee.  Must be &gt;= the minimum possible DSNP batch file size.
+    * `file_hash`: `<T::Hash>`, the hash of the batch file. Must not be 0 hash
+    * `msa_id`: `MsaId`, the id to use for the announcer.  Must exist and be active
+
+* `BatchAnnouncement`: implements `BatchAnnouncementParams`, returned by `get_batches`
+
+* `BatchAnnouncementOptions<T:Config>`: struct
```
wilwade, line 47: Should have schema_id as well.

designdocs/batched_messages.md
shannonwells: Yes actually, I had thought to configure a single batch schema, but @wilwade points out that we are planning to use schemas for any type of message, and we will also want to configure whether a schema is 'batchable', so schema id is now in there.

designdocs/batched_messages.md
shannonwells: done

designdocs/batched_messages.md
shannonwells: Yes actually, I had thought to configure a single batch schema, but @wilwade points out that we are planning to use schemas for any type of message, and we will also want to configure whether a schema is 'batchable', so schema id is now in there.

designdocs/batched_messages.md
```diff
@@ -0,0 +1,81 @@
...
+## Proposal
+* All names are placeholders and may be changed.
+* Errors in the extrinsic(s) must have different, reasonably-named error enums for each type of error for ease of debugging.
+
+### Constants
+* `BatchPageSizeMax` - the maximum number of batch announcements that will be returned by the `get_batches` query
+* `BatchSizeMinBytes` - the minimum possible size of a valid batch file with 1 row
+
+### Enums
+* `MessageType` - all values specified in [DSNP Spec Announcement Types](https://spec.dsnp.org/DSNP/Announcements.html)
```
wilwade, line 35:
```suggestion
```

designdocs/batched_messages.md
```diff
@@ -0,0 +1,81 @@
...
+
+### Types
+* `BatchAnnouncementParams<T:Config>`: generic
+    * `batch_url`:`Vec<u8>` the URL of the batch file. Must be IPFS and correctly formatted.
+    * `schema_id`: `SchemaId`  the schema id for messages in this batch.  The Schema Id must be for a "batchable" schema.
+    * `file_size`: `usize`, the size of the batch file, used to determine message fee.  Must be &gt;= the minimum possible DSNP batch file size.
+    * `file_hash`: `<T::Hash>`, the hash of the batch file. Must not be 0 hash
+    * `msa_id`: `MsaId`, the id to use for the announcer.  Must exist and be active
```
wilwade, line 43: I don't think this is the id of the announcer, it is the id of the claimed origin (although for batches usually I would expect that to be the announcer, but I could imagine a schema that requires a single user to announce a batch of stuff). DSNP uses batches that are multi-user (and derived by the DSNP Id). Other schemas might not.

jvasile: `batch_url` as a Vec<u8> makes sense, but "Must be IPFS" seems under-specified.

Do we want to accept `https://ipfs.io/ipfs/<cid>`?  If so, do we really mean only get that from this specific gateway, which would be... weird.

Or do we mean `ipfs://<cid>` and if so, do we mean to require inclusion of `ipfs://` or do we mean to require *only* the CID?  Or do we mean either is ok?  I prefer just specifying the CID.

Which version of CID are we specifying?  0, 1, both?  IPFS docs currently say " CIDv0 is still used by default for many IPFS operations, so you should generally support v0" but also [switch to v1 for future-proofing](https://docs.ipfs.io/concepts/content-addressing/#identifier-formats).  I prefer supporting only CIDv1.

At any rate, if CIDv2 arrives some years from now, do we want to support it automatically and immediately (which is what the current text implies) or do we want to consider the matter and then amend the spec?  I prefer the latter.

Maybe the text could look more like:  "the IPFS CIDv1 of the batch file".  I'd be happy to justify my preferences if they're not shared by others.

And do we want to support all of CIDv1 or do we want to force something a little more narrow?  I lean toward getting more specific.

Also: I've been slowly coming around to the idea that we should limit URL length to something like 4096. Given that filesystems limit path length, it's hard to imagine a need for longer URLs, even if the URL is <cid>/path.  The big upside here is just having limits on places where people can stuff arbitrary data.

shannonwells: If we're accepting _only_ IPFS, there's no reason to take up extra storage with a redundant URL scheme. 

One way to deal with lengths is to pick only one allowed hash algorithm, version, codec, and base encoding and strictly enforce it. Then I think you can either enforce one length or pick a minimum and maximum length.

designdocs/batched_messages.md
```diff
@@ -0,0 +1,96 @@
...
+* `BatchPageSizeMax` - the maximum number of batch announcements that will be returned by the `get_batches` query
+* `BatchSizeMinBytes` - the minimum possible size of a valid batch file with 1 row
+
+### Enums
+* `BatchFormat` - a list of supported formats for batch files. Currently only Parquet file format is supported, however,
+other formats are being considered.
+
+### Types
+* `BatchAnnouncementParams<T:Config>`: generic
+    * `batch_url`:`Vec<u8>` the URL of the batch file. Must be an IPFS [CIDv1](https://github.com/multiformats/cid/) URL. Accepted codec, algorithm, and base are to be determined.
```
jvasile, line 40: Thinking about this one a little more, I'm trying to figure out what is an IPFS CIDv1 URL, and there isn't an official way to do this unless you are talking about using a web gateway, which we aren't.

First, there's no location, so I guess we mean a URI.

Second, the IPFS folks think of this in terms of a path, not a URI.  It's the interplanetary filesystem, after all.  [They don't like `ipfs://` as a scheme specifier.](https://github.com/ipfs/go-ipfs/issues/1678)  They seem to require `foo://ip[f|n]s/<cid>[/<path>]`, where you indicate ipfs or ipns by making it the root of the path, not the scheme.  We don't have to conform to their views.  Some don't.  See, e.g., [Opensea](https://docs.opensea.io/docs/metadata-standards), which adopted `ipfs://`.  I tend toward using `ipfs://`.

I'm assuming there's no reason to support directories or ipns.  And IPNS would open the door to mutable content, which we don't want.  So maybe we can pare this down to just the CID (no path, no root prefix) and maybe we can rename this variable `batch_uri`.  Something like this:

    `batch_uri`:`Vec<u8>`  a URI with an `ipfs` scheme and a path consisting solely of the IPFS CIDv1 of the batch file in multibase format.  Accepted codec, algorithm, and base are to be determined.
    
 This would allow us to expand support to other content-addressed schemas, or even ipfs-based schemas like `nix:///ipfs/<cid>` in the future.

wilwade:
Options:
1. No Delegation on chain, you are on your own for inside the batch `BatchMessage(signer_has_msa_id)`
2. No delegation on chain, but the schema MUST have msa_id `BatchMessage(signer_has_msa_id)`
3. Allow batch message on chain to have a signer and msa_id, but **DSNP** requires signer == chain's msa_id `BatchMessage(signer_has_msa_id ?? msa_id), but DSNP wants ==`
4. Option on schema to force signer == chain's msa_id `BatchMessage(signer_has_msa_id == msa_id)`


We like 1.
- Simple
- Cannot enforce it anyway because batches are off-chain
- Flexible on the schemas

designdocs/batched_messages.md
shannonwells: We'll be using a pre-defined, single schema for batch messages.

designdocs/batched_messages.md
shannonwells: I've renamed the field to uri (as you correctly point out), but other details will have to be deferred to when the work is being done.  The revised plan was to just use the CID for now.  I thought I'd put this into another comment but apparently I didn't.

wilwade (approved):
🚢

PR #91, opened 2022-05-16
Feat/set governance schema max

### LINKED ISSUE #66: SUDO Extrensic: Set the maximum size of a schema
DEPENDENT: #28

As governance, I should be able to adjust the maximum size of a schema.

**Acceptance Criteria**
- [ ] An extrinsic that only can be called via sudo
- [ ] Alters the MAXIMUM_SCHEMA_SIZE
- [ ] Only effects future schema, not prior schema
- [ ] Schema longer than the set size should be rejected from registration

shannonwells:
## Problem 
As a Governance,
to prevent excessive schema sizes and 
therefore excessive cost to the network, 
and also adjust it if we need to,
We should be able to set a size limit for schemas.

Closes #66 SUDO extrinsic: set maximum size for a schema

## Solution
Two parts of this, the first is to create a value store for the governance-set schema max, which can be updated through an extrinsic.

Next is to use a BoundedVec for the schema parameter whose bound is (for now) just shy of a u16 size. Since the bound is a Config constant it can't be changed through an extrinsic.

## Notes
It was suggested to create a SignedExtension so that it could be applied to the call before it even got put into the message pool, however, it was uncertain whether this was necessary, and that further tests would need to be conducted.

In the interests of time, I decided to implement the two things that we could do and leave the rest to another story which can be prioritized.


pallets/schemas/src/lib.rs
```diff
@@ -78,40 +79,89 @@ pub mod pallet {
 	#[pallet::generate_store(pub(super) trait Store)]
 	pub struct Pallet<T>(_);
 
+	#[pallet::storage]
+	#[pallet::getter(fn get_schema_max_bytes)]
+	pub(super) type GovernanceSchemaMaxBytes<T: Config> = StorageValue<_, u32, ValueQuery>;
```
aramikm, line 84: I think eventually we might have a separate pallet for governance but for now this seems fine

pallets/schemas/src/lib.rs
```diff
@@ -78,40 +79,89 @@ pub mod pallet {
...
 			Ok(())
 		}
+
+		/// Set a new value for the Schema maximum number of bytes.  Must be <= the limit of the
+		/// Schema BoundedVec used for registration.
+		#[pallet::weight(30_000)]
+		pub fn set_max_schema_bytes(origin: OriginFor<T>, max_size: u32) -> DispatchResult {
+			ensure_root(origin.clone())?;
+			ensure!(max_size <= T::SchemaMaxBytesBoundedVecLimit, Error::<T>::InvalidSchemaMaxValue);
+			GovernanceSchemaMaxBytes::<T>::set(max_size);
```
aramikm, line 153: I think we do need events for any governance related changes

aramikm (changes requested):
Everything looks good except the genesis config setup. 
I think you might need to modify genesis Spec file and verify by running the chain and verifying that the value stored correctly.

enddynayn:
supercalifragilisticexpialidocious! 💯

pallets/schemas/src/lib.rs
shannonwells: Argh, you're right!

aramikm (approved):
Since the genesis value part is separated in another issue, this is good to go.

Issue #38, opened 2022-04-22
Users should be able to fetch all public keys associated with an MSA


dopatraman:
# Description
There needs to be a record of what public keys are associate with an MSA.

## Note
- This will involve a 1:M mapping between MSA ids and public keys.
- Currently there is no storage for directly going from an MSA id to keys.

# Acceptance Criteria
- [x] Users should be able to fetch a list of keys given an MSA identifier
  - There should be logic to keep both the map created by #22 and the map above up to date on every insert and delete.
- [x] RPC: `get_msa_keys`
  - Param: `msa_id`
  - Returns: list of public keys and when (if) they were deactivated

PR #92, opened 2022-05-17
msa: fetch keys by msa id issue #38

aramikm:
# Problem
Defined in https://github.com/LibertyDSNP/mrc/issues/38

User should be able to fetch keys related to a MSA id

# What
- Added new storage for reverse mapping
- Added new rpc endpoint for MSA pallet to fetch keys using msa_id

# Notes/Questions
- Should we supply msa_id on all requests in a way that we don't need to get it via accountId? In this case we might just need to have a DoubleMap of (MsaId, Accountd) -> keyInfo. The assumption for this is that the user should always know their MsaId

# Remaining Tasks

- [x] show key detail (is revoked in rpc)

# RPC verification

Screen Shot 2022-05-16 at 6 19 08 PM


aramikm:
After having a conversation with the team it turned out that due to requirement that an account-id can be connected to only one MSa account at a time, we will need to keep the mapping between `account-id` -> `Msa` on-chain.

enddynayn (approved):

enddynayn:
Look Fabulous! ⛹️‍♂️

pallets/msa/src/lib.rs
```diff
@@ -154,13 +168,22 @@ impl<T: Config> Pallet<T> {
...
 	pub fn add_key(msa_id: MessageSenderId, key: &T::AccountId) -> DispatchResult {
-		<KeyInfoOf<T>>::try_mutate(key, |maybe_msa| {
+		KeyInfoOf::<T>::try_mutate(key, |maybe_msa| {
 			ensure!(maybe_msa.is_none(), Error::<T>::DuplicatedKey);
 
 			*maybe_msa =
 				Some(KeyInfo { msa_id, expired: T::BlockNumber::default(), nonce: Zero::zero() });
 
-			Ok(())
+			// adding reverse lookup
```
enddynayn, line 177: leftover comment

pallets/msa/src/lib.rs
```diff
@@ -213,6 +236,20 @@ impl<T: Config> Pallet<T> {
 	pub fn get_owner_of(key: &T::AccountId) -> Option<MessageSenderId> {
 		Self::get_key_info(&key).map(|info| info.msa_id)
 	}
+
+	/// Fetches all the keys associated with a message Source Account
+	/// NOTE: This should only be called from RPC due to heavy database reads
```
enddynayn, line 241: This Note message makes me think that we are not doing a good job at compartmentalizing logic.

runtime/mrc/src/lib.rs
```diff
@@ -679,6 +680,12 @@ impl_runtime_apis! {
 		}
 	}
 
+	impl pallet_msa_runtime_api::MsaApi<Block, AccountId, BlockNumber> for Runtime {
+		fn get_msa_keys(msa_id: MessageSenderId) ->	Result<Vec<KeyInfoResponse<AccountId, BlockNumber>>, DispatchError> {
```
enddynayn, line 684: spacing

pallets/msa/src/types.rs
```diff
@@ -16,3 +16,23 @@ pub struct KeyInfo<BlockNumber> {
 	pub nonce: u32,
 	pub expired: BlockNumber,
 }
+
+impl<BlockNumber> KeyInfo<BlockNumber>
```
enddynayn, line 20: for future thoughts maybe we have a trait and implement that trait on this type.

pallets/msa/src/rpc/src/lib.rs
```diff
@@ -0,0 +1,56 @@
...
+use jsonrpc_core::Result;
+use jsonrpc_derive::rpc;
+use pallet_msa_runtime_api::MsaApi as MsaRuntimeApi;
+use sp_api::ProvideRuntimeApi;
+use sp_blockchain::HeaderBackend;
+use sp_runtime::{generic::BlockId, traits::Block as BlockT};
+use std::sync::Arc;
+
+#[rpc]
+pub trait MsaApi<BlockHash, AccountId, BlockNumber> {
```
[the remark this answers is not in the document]
aramikm: It's not but It should be part of RPC, I've tried removing it before and it didn't work.

It is used in `MessagesHandler` -> `std::marker::PhantomData<M>`

my rust is not advances enough yet to know why but It seems necessary.

pallets/msa/src/tests.rs
```diff
@@ -157,13 +196,18 @@ fn it_logs_key_added_event() {
...
 
+		// act
 		assert_ok!(Msa::add_key_to_msa(
 			Origin::signed(account.into()),
 			new_key.into(),
 			signature,
 			add_new_key_data.clone(),
 		));
 
+		// assert
```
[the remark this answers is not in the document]
aramikm: so the action is creation, what you assert on is the effects of creation which is log + data in DB. That's why I put in the same test. I think maybe the test name is confusing. I'll update

pallets/msa/src/lib.rs
```diff
@@ -154,13 +168,22 @@ impl<T: Config> Pallet<T> {
...
-		<KeyInfoOf<T>>::try_mutate(key, |maybe_msa| {
+		KeyInfoOf::<T>::try_mutate(key, |maybe_msa| {
 			ensure!(maybe_msa.is_none(), Error::<T>::DuplicatedKey);
 
 			*maybe_msa =
 				Some(KeyInfo { msa_id, expired: T::BlockNumber::default(), nonce: Zero::zero() });
 
-			Ok(())
+			// adding reverse lookup
+			<MsaKeys<T>>::try_mutate(msa_id, |ref mut key_list| {
```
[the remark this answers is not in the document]
aramikm: thanks, not necessary

pallets/msa/src/lib.rs
aramikm: that's just a small description comment for anyone to know why we have this second mutate inside

pallets/msa/src/lib.rs
aramikm: what is the alternative?

I can pull out the logic from here inside RPC but having all the logic inside one place is nice but don't have any strong opinions on it.

pallets/msa/src/types.rs
aramikm: yeah, that would be nice. Will refactor in future tickets.

Issue #64, opened 2022-05-06
Setup scripts to onboard local mrc to relay/validators


saraswatpuneet:
**Details**:

- [ ] Infra will need to expose bootnodes related ports
- [ ] Define script command to onboard mrc to relay running in infra

**Tests**
- [ ] run tests to register a para_id on infra relay
- [ ] update mrc spec to point to para_id from above
- [ ] run onboard mrc to relay and ensure successful onboarding

PR #94, opened 2022-05-17
Feat/devnet script offboard #64

saraswatpuneet:
Rationale
Devenet instead of reserving a new para id with each new deployment (without runtime upgrades), job will off-board mrc and onboard latest update again. This will work for short term till we have runtime upgrades in place

Tasks

- [x] Devnet will need a way to offboard mrc and onboard with new  updates without bricking the chain
- [x] Add a script to offboard parachain for the same

Tests:
1. Start mrc
2. Onboard mrc
3. Offboard mrc

Note: this will only be used initially in devnet until we figure out proper procedure to run runtime upgrades instead


README.md
```diff
@@ -87,6 +87,14 @@ source .env
 
 1. Link to parachain [dashboard](https://polkadot.js.org/apps/?rpc=ws%3A%2F%2F127.0.0.1%3A11946)
 
+1. Off-boarding MRC from relay chain
+
+    ```bash
+    ./scripts/init.sh offboard-mrc
```
wilwade, line 93: @saraswatpuneet is the offboard script expected to open up the paraId for reuse?

I tried it and while it offboarded the id 2000, when running register, it chose 2001 instead.

saraswatpuneet: @wilwade  no, para_id once registered is taken, this ensures we safely offboard mrc and onboarding it again for same para_id but with updated genesis/wasm with new changes we push in. This in part is only safe to do in devenet as in testnet and betanet we would be doing runtime upgrades etc.

saraswatpuneet: This will prevent us from always having to register new para_ids but that option is also on the table

wilwade: Ah I see. So it doesn't require re-running the register step, only the onboard step.

saraswatpuneet: yes, we will only run register step when we re-deploy relay for any reason so register mrc step is tied to relay deployment

aramikm (approved):
Looks good!

wilwade (approved):
All worked for me!

PR #99, opened 2022-05-19
feat(Delegate): add ability to revoke

enddynayn:
Add ability to revoke a delegation. Revoking
a delegation means that a user has expired
the relationship and messages submitted by
delegates are no longer valid. Note that
delegates are not removed from storage and
instead flagged expired at a block.

[issue-89](https://github.com/LibertyDSNP/mrc/issues/89)


pallets/msa/src/lib.rs
```diff
@@ -304,6 +328,33 @@ impl<T: Config> Pallet<T> {
 		Ok(())
 	}
 
+	// TODO: add test
```
enddynayn, line 331: done

pallets/msa/src/lib.rs
```diff
@@ -304,6 +328,33 @@ impl<T: Config> Pallet<T> {
...
+		delegate_msa_id: Delegate,
+		delegator_msa_id: Delegator,
+	) -> DispatchResult {
+		DelegateInfoOf::<T>::try_mutate_exists(
+			delegate_msa_id,
+			delegator_msa_id,
+			|maybe_info| -> DispatchResult {
+				let mut info = maybe_info.take().ok_or(Error::<T>::DelegateNotFound)?;
+
+				// maybe remove optionvalue
```
enddynayn, line 342: this is good.

pallets/msa/src/lib.rs
```diff
@@ -104,6 +105,9 @@ pub mod pallet {
 		DuplicateDelegate,
 		AddDelegateVerificationFailed,
 		UnauthorizedDelegator,
+		DelegateRevoked,
+		DelegateError,
+		DelegateNotFound,
```
enddynayn, line 110: help with naming. Maybe DelegateRelationshipNotFound

pallets/msa/src/lib.rs
```diff
@@ -149,6 +153,26 @@ pub mod pallet {
 			Ok(())
 		}
 
+		#[pallet::weight(10_000)]
+		pub fn revoke_msa_delegate(
+			origin: OriginFor<T>,
+			delegate_msa_id: MessageSenderId,
+		) -> DispatchResult {
+			let key = ensure_signed(origin)?;
```
enddynayn, line 161: delegator_key

pallets/msa/src/lib.rs
```diff
@@ -149,6 +153,27 @@ pub mod pallet {
 			Ok(())
 		}
 
+		#[pallet::weight(10_000)]
+		pub fn revoke_msa_delegate(
```
aramikm, line 157: since the delegate can be revoked by any of 2 parties. I think it might makes sense to have 3 params

```
origin: OriginFor<T>,
delegate_msa_id: MessageSenderId,
delegator_msa_id: MessageSenderId,
```
and the origin can be a key associated to delegate or delegator

aramikm (approved):
Looks good. Just added one comment

pallets/msa/src/lib.rs
enddynayn: I think maybe we have a different function to make things a little easier. Here I do not know the submitter is the delegate or the delegator.

Maybe we have a `revoke_delegator`. I'll add an issue for this.

Issue #85, opened 2022-05-13
[MRC Delegate] Creating MSA via and with delegation


aramikm:
#### create_account_with_delegate
Creates a new `MsaId` on behalf of an MSA and adds the origin as the MSA's delegate.

### Acceptance Criteria
- [x] a user with already existing MSA account creates a new MSA account and add itself as a delegate  https://github.com/LibertyDSNP/mrc/pull/100
- [x] unit tests https://github.com/LibertyDSNP/mrc/pull/100
- [x] benchmark
- [x] spec generation 

## Notes
- Some details mentioned in design doc
  -  * Parameters:
        1. `payload`:
           1. `data` - this is what the MSA owner must sign and provide to the delegate beforehand.
               * `msa_id` - the delegate, of type `MsaId`
               * `permission` a value indicating the permission to be given to the delegate
           2. `public_key` - The authorizing key used to create `signature`
           3. `signature` - The signature of the hash of `data`
      * Event:  `IdentityCreated`, with the `public_key`, the newly created `MsaId`, and `msa_id`
      * Restrictions:  origin must own `msa_id` in the payload.
 - Signature related data might need some kind of nonce or random input just like one of already implemented ones in MSA
 - The permission details need to be defined in more details. It can be deferred.

PR #100, opened 2022-05-19
creating msa via and with delegate

aramikm:
# Problem
Chain should support creation of a new Msa account from another Msa account. This creation also means that the creator is delegated to do actions on behalf of new account. Issue #85 

# What
- [x] added new extrinsic to support this action
- [x] tests

# Notes
- Benchmark
- Spec generation

will be done in a separate PR

### DIFF
--- a/pallets/msa/src/lib.rs
+++ b/pallets/msa/src/lib.rs
@@ -105,6 +105,7 @@ pub mod pallet {
 		DuplicateDelegate,
 		AddDelegateVerificationFailed,
 		UnauthorizedDelegator,
+		UnauthorizedDelegate,
 		DelegateRevoked,
 		DelegateError,
 		DelegateNotFound,
@@ -116,44 +117,46 @@ pub mod pallet {
 		pub fn create(origin: OriginFor<T>) -> DispatchResult {
 			let who = ensure_signed(origin)?;
 
-			let (identifier, _) = Self::create_account(who.clone(), EMPTY_FUNCTION)?;
-
-			Self::deposit_event(Event::MsaCreated { msa_id: identifier, key: who });
+			let (_, _) = Self::create_account(who.clone(), |new_msa_id| -> DispatchResult {
+				Self::deposit_event(Event::MsaCreated { msa_id: new_msa_id, key: who });
+				Ok(())
+			})?;
 
 			Ok(())
 		}
 
 		#[pallet::weight(T::WeightInfo::create(10_000))]
-		pub fn create_account_with_delegate(
+		pub fn create_sponsored_account_with_delegation(
 			origin: OriginFor<T>,
+			delegator_key: T::AccountId,
 			proof: MultiSignature,
-			creation_delegate_payload: AccountCreationDelegate<T::AccountId>,
+			add_delegate_payload: AddDelegate,
 		) -> DispatchResult {
-			let who = ensure_signed(origin)?;
+			let delegate_key = ensure_signed(origin)?;
 
-			Self::verify_signature(
-				proof,
-				creation_delegate_payload.key.clone(),
-				creation_delegate_payload.encode(),
-			)?;
+			Self::verify_signature(proof, delegator_key.clone(), add_delegate_payload.encode())?;
 
-			let delegate_msa_id =
-				Self::ensure_valid_creation_delegate(&creation_delegate_payload.key, &who)?;
+			let delegate_msa_id = Self::ensure_valid_msa_key(&delegate_key)?.msa_id;
+			ensure!(
+				add_delegate_payload.authorized_msa_id == delegate_msa_id,
+				Error::<T>::UnauthorizedDelegate
+			);
 
-			let (_, _) = Self::create_account(
-				creation_delegate_payload.key.clone(),
-				|new_msa_id: MessageSenderId| -> DispatchResult {
-					let _ = Self::add_delegate(delegate_msa_id, new_msa_id.into())?;
+			let (_, _) =
+				Self::create_account(delegator_key.clone(), |new_msa_id| -> DispatchResult {
+					let _ = Self::add_delegate(delegate_msa_id.into(), new_msa_id.into())?;
 
-					Self::deposit_event(Event::MsaCreatedAndDelegated {
-						new_msa_id: new_msa_id.into(),
-						delegator_key: creation_delegate_payload.key.clone(),
-						delegate: delegate_msa_id,
+					Self::deposit_event(Event::MsaCreated {
+						msa_id: new_msa_id.clone(),
+						key: delegator_key.clone(),
 					});
 
+					Self::deposit_event(Event::DelegateAdded {
+						delegator: new_msa_id.into(),
+						delegate: delegate_msa_id.into(),
+					});
 					Ok(())
-				},
-			)?;
+				})?;
 
 			Ok(())
 		}
@@ -224,9 +227,10 @@ pub mod pallet {
 			let msa_id = add_key_payload.msa_id;
 			Self::ensure_msa_owner(&who, msa_id)?;
 
-			Self::add_key(msa_id, &key, EMPTY_FUNCTION)?;
-
-			Self::deposit_event(Event::KeyAdded { msa_id, key });
+			Self::add_key(msa_id, &key.clone(), |new_msa_id| -> DispatchResult {
+				Self::deposit_event(Event::KeyAdded { msa_id: new_msa_id, key });
+				Ok(())
+			})?;
 
 			Ok(())
 		}
@@ -431,17 +435,6 @@ impl<T: Config> Pallet<T> {
 		ensure!(info.expired == T::BlockNumber::zero(), Error::<T>::KeyRevoked);
 
 		Ok(info)
-	}
-
-	fn ensure_valid_creation_delegate(
-		delegator_key: &T::AccountId,
-		delegate_key: &T::AccountId,
-	) -> Result<Delegate, DispatchError> {
-		let delegate_msa_id = Self::ensure_valid_msa_key(&delegate_key)?.msa_id;
-
-		ensure!(Self::try_get_key_info(delegator_key).is_err(), Error::<T>::AlreadyAssignedKey);
-
-		Ok(delegate_msa_id.into())
 	}
 }
 
--- a/pallets/msa/src/lib.rs
+++ b/pallets/msa/src/lib.rs
@@ -116,9 +116,44 @@ pub mod pallet {
 		pub fn create(origin: OriginFor<T>) -> DispatchResult {
 			let who = ensure_signed(origin)?;
 
-			let (identifier, _) = Self::create_account(who.clone())?;
+			let (identifier, _) = Self::create_account(who.clone(), EMPTY_FUNCTION)?;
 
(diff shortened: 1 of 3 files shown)


pallets/msa/src/lib.rs
```diff
@@ -81,10 +82,26 @@ pub mod pallet {
...
+			key: T::AccountId,
+		},
+		KeyRevoked {
+			key: T::AccountId,
+		},
+		DelegateAdded {
+			delegate: Delegate,
+			delegator: Delegator,
+		},
+		MsaCreatedAndDelegated {
```
aramikm, line 100: Let me know if this makes sense or maybe we might want to have something else as an event for this action

pallets/msa/src/lib.rs
```diff
@@ -112,13 +130,48 @@ pub mod pallet {
 		pub fn create(origin: OriginFor<T>) -> DispatchResult {
 			let who = ensure_signed(origin)?;
 
-			let (identifier, _) = Self::create_account(who.clone())?;
+			let (identifier, _) = Self::create_account(who.clone(), EMPTY_FUNCTION)?;
```
aramikm, line 133: I wish I could find a better alternative way instead of passing in an empty function, any alternative that would achieve the same effect is appreciated. (without any code duplication)

pallets/msa/src/lib.rs
```diff
@@ -112,13 +130,48 @@ pub mod pallet {
...
 
 			Self::deposit_event(Event::MsaCreated { msa_id: identifier, key: who });
 
 			Ok(())
 		}
 
+		#[pallet::weight(T::WeightInfo::create(10_000))]
```
aramikm, line 140: this is going to be replaced with fresh calculated weights in the next PR which will be dedicated to benchmarks!

pallets/msa/src/types.rs
```diff
@@ -4,6 +4,8 @@ use scale_info::TypeInfo;
 
 use codec::{Decode, Encode};
 
+pub const EMPTY_FUNCTION: fn(MessageSenderId) -> DispatchResult = |_| Ok(());
```
aramikm, line 7: a no-op empty function to pass as a param when there is nothing needs to be done inside creation transaction

pallets/msa/src/lib.rs
```diff
@@ -229,7 +292,7 @@ impl<T: Config> Pallet<T> {
 					.try_insert(index, key.clone())
 					.map_err(|_| Error::<T>::KeyLimitExceeded)?;
 
-				Ok(())
+				on_success(msa_id)
```
aramikm, line 295: why do we need to this?

Substrate does not take care of transactions correctly. When we have multiple mutates(non-nested) after each other if the first one is successful and the second one fails the first mutate changes will not be reverted. That's why we need to nest these mutates so if one of them fail then all the rest also reverted.

pallets/msa/src/lib.rs
shannonwells: I wonder if we should separate these into two events since we're also allowing adding delegates, and also account creation via token payment.

pallets/msa/src/lib.rs
shannonwells: Why not make the function deposit the event on success?

shannonwells (approved):
It looks good, I had a suggestion but since I don't know if it would work, it's not a blocker.

pallets/msa/src/types.rs
```diff
@@ -48,6 +50,12 @@ pub struct AddDelegate {
 	pub permission: u8,
 }
 
+#[derive(TypeInfo, Debug, Clone, Decode, Encode, PartialEq)]
+pub struct AccountCreationDelegate<AccountId> {
+	pub key: AccountId,
```
enddynayn, line 55: i think this the msa_id?

enddynayn: should this be the msa_id that you are giving permission to sponsor your account creation?

pallets/msa/src/lib.rs
```diff
@@ -112,13 +130,48 @@ pub mod pallet {
...
+				creation_delegate_payload.key.clone(),
+				creation_delegate_payload.encode(),
+			)?;
+
+			let delegate_msa_id =
+				Self::ensure_valid_creation_delegate(&creation_delegate_payload.key, &who)?;
+
+			let (_, _) = Self::create_account(
+				creation_delegate_payload.key.clone(),
+				|new_msa_id: MessageSenderId| -> DispatchResult {
```
enddynayn, line 159: nice!

pallets/msa/src/lib.rs
```diff
@@ -112,13 +130,48 @@ pub mod pallet {
...
 			Ok(())
 		}
 
+		#[pallet::weight(T::WeightInfo::create(10_000))]
+		pub fn create_account_with_delegate(
+			origin: OriginFor<T>,
+			proof: MultiSignature,
+			creation_delegate_payload: AccountCreationDelegate<T::AccountId>,
+		) -> DispatchResult {
+			let who = ensure_signed(origin)?;
```
enddynayn, line 146: who ==> key

pallets/msa/src/lib.rs
```diff
@@ -337,6 +400,17 @@ impl<T: Config> Pallet<T> {
 
 		Ok(info)
 	}
+
+	fn ensure_valid_creation_delegate(
+		delegator_key: &T::AccountId,
+		delegate_key: &T::AccountId,
+	) -> Result<Delegate, DispatchError> {
+		let delegate_msa_id = Self::ensure_valid_msa_key(&delegate_key)?.msa_id;
```
enddynayn, line 408: we should probably ensure_valid_msa_key for delegator

pallets/msa/src/lib.rs
```diff
@@ -112,13 +130,48 @@ pub mod pallet {
...
 
 			Self::deposit_event(Event::MsaCreated { msa_id: identifier, key: who });
 
 			Ok(())
 		}
 
+		#[pallet::weight(T::WeightInfo::create(10_000))]
+		pub fn create_account_with_delegate(
+			origin: OriginFor<T>,
+			proof: MultiSignature,
```
enddynayn, line 143: your missing the key of the signer?

pallets/msa/src/lib.rs
```diff
@@ -112,13 +130,48 @@ pub mod pallet {
...
+			creation_delegate_payload: AccountCreationDelegate<T::AccountId>,
+		) -> DispatchResult {
+			let who = ensure_signed(origin)?;
+
+			Self::verify_signature(
+				proof,
+				creation_delegate_payload.key.clone(),
```
enddynayn, line 150: creation_delegate_payload.key ==> delegator_key

pallets/msa/src/lib.rs
```diff
@@ -112,13 +130,48 @@ pub mod pallet {
...
+				creation_delegate_payload.encode(),
+			)?;
+
```
enddynayn, line 153: we need add in here that they key is valid
```
Self::ensure_valid_msa_key(&delegate_key (submitter))?
I think we want to check that the signing key of the transaction is valid.
```

pallets/msa/src/lib.rs
enddynayn, line 408: we cant really use this here because what we want to know is if this key already exist if so than we cannot assign this key. I think we can actually remove this function.

pallets/msa/src/lib.rs
```diff
@@ -112,13 +130,48 @@ pub mod pallet {
...
+			let delegate_msa_id =
+				Self::ensure_valid_creation_delegate(&creation_delegate_payload.key, &who)?;
+
+			let (_, _) = Self::create_account(
+				creation_delegate_payload.key.clone(),
+				|new_msa_id: MessageSenderId| -> DispatchResult {
+					let _ = Self::add_delegate(delegate_msa_id, new_msa_id.into())?;
```
enddynayn, line 160: nice use of the call back

pallets/msa/src/types.rs
enddynayn: you want to sign who you are giving permission to, what msa.

pallets/msa/src/lib.rs
enddynayn: I like that idea.

aramikm: OK, I can modify it to add two events but then it would be harder to know if this was created using delegation or not

pallets/msa/src/lib.rs
aramikm: sounds good

pallets/msa/src/lib.rs
aramikm: Was not sure if the old structure was making sense or not. I can modify it to be compatible with the rest

pallets/msa/src/lib.rs
aramikm: we have this inside `ensure_valid_creation_delegate`

pallets/msa/src/lib.rs
enddynayn: Yeah the reason that I pulled out was to make it more practical to use. It also actually simplified the code.

pallets/msa/src/lib.rs
aramikm: there are two things we are checking here

1- if the delegate key has a valid msa account (to ensure it can delegate)
2- if the delegator key has not any registered msa account

pallets/msa/src/types.rs
aramikm: you can sign whatever the payload is as long as the payload and signature are valid, but for the sake of consistency I'll remove this and use the existing one

pallets/msa/src/lib.rs
enddynayn: ensure_valid_creation_delegate does not really do anything? Lets jump on a call.

pallets/msa/src/lib.rs
```diff
@@ -112,13 +130,48 @@ pub mod pallet {
...
-			let (identifier, _) = Self::create_account(who.clone())?;
+			let (identifier, _) = Self::create_account(who.clone(), EMPTY_FUNCTION)?;
 
 			Self::deposit_event(Event::MsaCreated { msa_id: identifier, key: who });
 
 			Ok(())
 		}
 
+		#[pallet::weight(T::WeightInfo::create(10_000))]
+		pub fn create_account_with_delegate(
```
enddynayn, line 141: create_account_with_sponsor

pallets/msa/src/lib.rs
```diff
@@ -112,13 +130,48 @@ pub mod pallet {
...
+		) -> DispatchResult {
+			let who = ensure_signed(origin)?;
+
+			Self::verify_signature(
+				proof,
+				creation_delegate_payload.key.clone(),
+				creation_delegate_payload.encode(),
+			)?;
+
+			let delegate_msa_id =
```
enddynayn, line 154: we are usally missing one important check.

```
ensure!(delegate_msa_id == authorized_msa_id, UnauthorizedDelegate)

enddynayn (approved):

saraswatpuneet (approved):
Looks good 👍

PR #101, opened 2022-05-19
Transaction fee computation/ common pallet

### LINKED ISSUE #77: Spike: Extrinsic fee computation for MRC (a generic extensible pallet)

saraswatpuneet:
## Description
As a mrc participation, I would like to know the cost of interactions with MRC via on chain MRC specific  methods.

Tasks
While pallet transaction payment provide an rpc endpoint to get basic fee details for a given extrinsic. MRC would require an extension of the said pallet to add custom fee, adjustment etc. Which requires following tasks to be completed as a part of this PR.

- [x] A common pallet for all transaction cost calcultion
- [x]  Define pallet config derive from pallet_transaction_payment, where config will be extensible for future adjustment to costs.
- [x]  Define a generic function to compute basic FeeDetails for a given extrinsic for now 

- [x] add rpc endpoint to compute extrinsic cost for signed transactions.

## Testing

- [x] Unit tests covering the generic function for signed and unsigned transactions 
- [x] Run MRC and make rpc query for MRC related extrinsic and verify fee results 

Here sending a unsigned extrinsic returns null inclusion fee which is expected 
image

Closes #77


README.md
```diff
@@ -95,6 +95,20 @@ source .env
...
+    ```bash
+    ./scripts/init.sh stop-relay-chain
+    ```
+
+1. Stop MRC running in the terminal.
+
+1. Run ```docker volume prune``` to remove unused volumes.
+
+1. Remove mrc chain data via ```rm -rf /tmp/mrc```.
+
```
saraswatpuneet, line 111: Update readme to list clean up steps

common/primitives/src/lib.rs
```diff
@@ -6,4 +6,3 @@ pub mod msa;
 pub mod rpc;
 pub mod schema;
 pub mod utils;
-pub mod weight_to_fees;
```
saraswatpuneet, line 9: Not needed replaced with tx fee pallet

pallets/schemas/src/lib.rs
```diff
@@ -183,11 +182,5 @@ pub mod pallet {
 			}
 			None
 		}
-
-		pub fn calculate_schema_cost(schema: Vec<u8>) -> Weight {
-			let schema_len = schema.len() as u32;
-			let schema_weight = T::WeightInfo::register_schema(schema_len);
-			schema_weight
-		}
```
saraswatpuneet, line 191: Move it away from schema to tx pallet where we will have all custom configs for MRC related costs

pallets/schemas/src/rpc/src/lib.rs
```diff
@@ -104,40 +91,7 @@ where
...
-		let fee_return = FeeDetails {
-			inclusion_fee: Some(InclusionFee {
-				base_fee: try_into_rpc_balance(base_fee)?,
-				len_fee: try_into_rpc_balance(len_fee.into())?,
-				adjusted_weight_fee: try_into_rpc_balance(unadjusted_schema_fee.into())?,
-			}),
-			tip,
-		};
-		Ok(fee_return)
+		Ok(true)
```
saraswatpuneet, line 94: Convert this rpc at schema to simply verify if a schema is valid or not
Raises #107

pallets/tx_fee/src/lib.rs
```diff
@@ -0,0 +1,57 @@
...
+// Balance type alias.
+pub(crate) type BalanceOf<T> = <OnChargeTransactionOf<T> as OnChargeTransaction<T>>::Balance;
+
+#[frame_support::pallet]
+pub mod pallet {
+	use super::*;
+
+	#[pallet::config]
+	pub trait Config: frame_system::Config + pallet_transaction_payment::Config {}
+
```
saraswatpuneet, line 25: Extend transaction payment config for this pallet here we can add customization specific to MRC

pallets/tx_fee/src/lib.rs
```diff
@@ -0,0 +1,57 @@
...
+			if unchecked_extrinsic.is_signed().unwrap_or(false) {
+				<pallet_transaction_payment::Pallet<T>>::compute_fee_details(
+					len,
+					&dispatch_info,
+					tip,
+				)
+			} else {
+				// Unsigned extrinsics have no inclusion fee.
+				FeeDetails { inclusion_fee: None, tip }
+			}
```
saraswatpuneet, line 54: Compute feedetails for a given signed extrinsic via rpc call

pallets/tx_fee/src/lib.rs
```diff
@@ -0,0 +1,57 @@
...
+
+	#[pallet::pallet]
+	#[pallet::generate_store(pub(super) trait Store)]
+	pub struct Pallet<T>(_);
+
+	#[pallet::error]
+	pub enum Error<T> {}
+
+	impl<T: Config> Pallet<T> {
+		pub fn compute_extrinsic_cost<Extrinsic: sp_runtime::traits::Extrinsic + GetDispatchInfo>(
```
enddynayn, line 34: These are RPC endpoints which should exist in their corresponding pallet. Why are we taking these out?

enddynayn: Can you please point me to where you got this pattern from?

saraswatpuneet: Not really! then you would introduce dependencies of transaction pallets in each one of those, ideally I would handle all such translations at one spot, especially knowing this will evolve into complexities of doing some math around fee adjustments for specific cases etc.

Otherwise, I will have all pallets coupled with transaction pallet and do its own thing.

saraswatpuneet: Further, we can have traits or say enums defined where logic can be updated around say Schema, or Message Cost etc.

enddynayn: I guess I dont understand what special transactions we are doing?

saraswatpuneet: A schema could have different fee based on its retention policy for example or a message storage cost could differ based on its storage period, and so on

enddynayn, line 34: is there a method that computes cost already?

pallets/tx_fee/src/tests.rs
```diff
@@ -0,0 +1,245 @@
...
+		self
+	}
+	pub fn weight_fee(mut self, weight_to_fee: u64) -> Self {
+		self.weight_to_fee = weight_to_fee;
+		self
+	}
+	fn set_constants(&self) {
+		EXTRINSIC_BASE_WEIGHT.with(|v| *v.borrow_mut() = self.base_weight);
+		TRANSACTION_BYTE_FEE.with(|v| *v.borrow_mut() = self.byte_fee);
+		WEIGHT_TO_FEE.with(|v| *v.borrow_mut() = self.weight_to_fee);
```
shannonwells, line 186: I find this method surprising, given that these are "all caps" constants one would expect these to be set explicitly at the file scope and not programmatically.  Is this how it's typically done?  Why wouldn't these be pallet configs?

pallets/tx_fee/src/lib.rs
saraswatpuneet: Because we will have to add adjustments to basic cost but that is future work
And here I am using pallet transaction to calculate base cost , check the pallet function

pallets/schemas/src/rpc/src/lib.rs
enddynayn: this should return an error?

pallets/tx_fee/Cargo.toml
```diff
@@ -0,0 +1,56 @@
...
+targets = ["x86_64-unknown-linux-gnu"]
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+codec = {package = "parity-scale-codec", version = "3.0.0", default-features = false, features = [
+	"derive",
+]}
+smallvec = "1.7.0"
+frame-benchmarking = {default-features = false, git = "https://github.com/paritytech/substrate.git", branch = "polkadot-v0.9.18", optional = true}
```
enddynayn, line 22: are you using all these dependecies?

pallets/tx_fee/src/tests.rs
```diff
@@ -0,0 +1,245 @@
...
+	let xt = TestXt::new(call.clone(), Some((origin, extra)));
+	let info = xt.get_dispatch_info();
+	let ext = xt.encode();
+	let len = ext.len() as u32;
+
+	let unsigned_xt = TestXt::<_, ()>::new(call, None);
+
+	ExtBuilder::default().base_weight(5).weight_fee(2).build().execute_with(|| {
+		// all fees should be x1.5
+		<NextFeeMultiplier<Runtime>>::put(Multiplier::saturating_from_rational(3, 2));
```
shannonwells, line 225: follow up to above question - I see here this is where those are set, wouldn't this be run at compile time?  Does this mean you could change the base weight and weight fees at runtime?

shannonwells (approved):
It seems fine?  I just had the one question but it's mostly out of not knowing this part of substrate code at all.

pallets/tx_fee/src/lib.rs
enddynayn: If this is future work lets merge it when we have that problem. We don't have that problem. I looked through other substrate base code and I could not find one that does this.

Like I mention I think we should solve for problems with have. At the moment we dont have this problem. Lets not create one.

enddynayn: We don't have this flushed out yet. We are already well aware that we do not know the details of our retention policy. Lets wait for when we have this problem.

saraswatpuneet: I am just going off of discussion but can always hold off though you confused me with "we may not have that need" which is again may or may not question because this is not trying to solve it but put a placeholder because I am sure we have that requirement.

enddynayn:
I see you did some really great work and it is impressive. I think we can leverage some of this work soon. At this point, I would not recommend merging this as we do not know for certain if we need this. When you responded "a schema could..." it raised a flag for me. We should try to merge things that we know for certain will add immediate value.

pallets/tx_fee/src/tests.rs
aramikm: Yeah, I have the same question. It seems like it's thread local since it's only specific to each test then why it's shared with other parts? Definitely a rust pattern that I don't know about.

aramikm (approved):
I really like the test builder pattern but still don't have an overall grasp of how everything wires together.

pallets/tx_fee/src/tests.rs
saraswatpuneet: TestExternalitties provide some off the shelf builders to create these runtime constructs 
https://docs.substrate.io/v3/runtime/testing/

pallets/tx_fee/src/tests.rs
saraswatpuneet: I learnt this pattern from TestExternalitties provide some off the shelf builders to create these runtime constructs
https://docs.substrate.io/v3/runtime/testing/ and some other parity code

pallets/tx_fee/src/tests.rs
```diff
@@ -0,0 +1,245 @@
...
+			})
+			.build_or_panic()
+	}
+}
+
+parameter_types! {
+	pub static WeightToFee: u64 = 1;
+	pub static TransactionByteFee: u64 = 1;
+	pub static OperationalFeeMultiplier: u8 = 5;
+}
```
saraswatpuneet, line 66: @shannonwells  @aramikm  with respect to question below, it was a random find for me that this WeightToFree maps to
WEIGHT_TO_FEE, and same for  TRANSACTION_BYTE_FEE

Due to them being static types. I haven't figured why it happens but if i define them again with EXTRINSIC_BASE_WEIGHT compiler complains

pallets/tx_fee/src/tests.rs
```diff
@@ -0,0 +1,245 @@
...
+		System: system::{Pallet, Call, Config, Storage, Event<T>},
+		Balances: pallet_balances::{Pallet, Call, Storage, Config<T>, Event<T>},
+		TransactionPayment: pallet_transaction_payment::{Pallet, Storage},
+		MrcTxFeePallet: pallet_tx_fee::{Pallet},
+	}
+);
+
+thread_local! {
+	static EXTRINSIC_BASE_WEIGHT: RefCell<u64> = RefCell::new(0);
+}
```
saraswatpuneet, line 45: @shannonwells  this is kept out of config for tests so, this can be modified for various test cases, one cannot change it at runtime

pallets/schemas/src/rpc/src/lib.rs
```diff
@@ -104,40 +91,7 @@ where
 			}
```
saraswatpuneet, line 91: @aramikm  @enddynayn  here is the error returned if schema is not valid

pallets/schemas/src/rpc/src/lib.rs
saraswatpuneet: Yes error is return when schema is not valid and it is valid simple true is sent back

PR #106, opened 2022-05-20
add schema max bytes initial value to genesis config and make sudo work

shannonwells:
* Sets the initial value for GovernanceSchemaMaxBytes to 1024
* Configures the Sudo pallet with Alice as the sudoer, so Sudo extrinsics can be used from the Polkadot dashboard in development/local testnet chains.

with @aramikm 

Here it is working:

Sudo menu item appears:
Screen Shot 2022-05-20 at 4 29 20 PM

Initial GovernanceSchemaMaxBytes:
Screen Shot 2022-05-20 at 4 29 48 PM

After running the command to set the new max value, via Sudo:
Screen Shot 2022-05-20 at 4 31 01 PM


node/src/chain_spec.rs
```diff
@@ -178,6 +180,7 @@ pub fn local_testnet_config() -> ChainSpec {
 
 fn testnet_genesis(
 	invulnerables: Vec<(AccountId, AuraId)>,
+	root_key: AccountId,
```
shannonwells, line 183: The root_key param to was added to this function, so "get_account_id_from_seed" call above is how to set the root key, i.e. the sudoer.

aramikm: I think we are going to need to have a pre-prod checklist and I think removing `Alice` or replacing it with else would be in that list.

runtime/mrc/src/lib.rs
```diff
@@ -529,7 +534,9 @@ construct_runtime!(
 		// MRC related pallets
 		Msa: pallet_msa::{Pallet, Call, Storage, Event<T>} = 34,
 		Messages: pallet_messages::{Pallet, Call, Storage, Event<T>} = 35,
-		Schemas: pallet_schemas::{Pallet, Call, Storage, Event<T>} = 36,
+		Schemas: pallet_schemas::{Pallet, Call, Storage, Event<T>, Config} = 36,
+		Sudo: pallet_sudo::{Pallet, Call, Config<T>, Storage, Event<T> }= 37,
```
aramikm, line 538: nit: sudo is not a MRC pallet

aramikm (approved):
Looks good!

runtime/mrc/src/lib.rs
enddynayn: I thought we already had this pallet.

enddynayn (approved):

pallets/schemas/src/rpc/Cargo.toml
```diff
@@ -30,8 +30,6 @@ sp-blockchain = { git = "https://github.com/paritytech/substrate", default-featu
 sp-std = {default-features = false, git = "https://github.com/paritytech/substrate.git", branch = "polkadot-v0.9.18"}
 pallet-transaction-payment = {git = "https://github.com/paritytech/substrate", default-features = false, branch = "polkadot-v0.9.18"}
 pallet-transaction-payment-rpc-runtime-api = {git = "https://github.com/paritytech/substrate", default-features = false, branch = "polkadot-v0.9.18"}
-
-[dev-dependencies]
```
enddynayn, line 34: why does this need to be included in schemas?

saraswatpuneet: @enddynayn  my next PR removes it, this was just a first

saraswatpuneet (approved):
👍

pallets/schemas/src/rpc/Cargo.toml
shannonwells: You need it for setting the Governance Schema Max as that's a sudo command, and for running it in the UI I assumed you would need it in regular dependences here too. I can try it with it back in dev-dependencies (because we needed it for testing)

enddynayn: how do you use it in testing?

runtime/mrc/src/lib.rs
shannonwells: I updated this so you can disable a sudoer by passing rootkey param as None and it will work.

pallets/schemas/src/rpc/Cargo.toml
shannonwells: taking it out, not needed.

node/src/chain_spec.rs
shannonwells: We should have such a checklist anyway. Someone could copy the chain spec to a production one and forget about the sudo.  I set testnet root key to 'None' and regenerated the specs.

PR #121, opened 2022-05-26
Killing a few snakes (renaming some errors and events in MSA pallet)

shannonwells:
Based on something I said in the Ethics & Governance meeting, I realized there was no way to tell who revoked a delegation.  So I started out by adding a new event to indicate who revoked a delegation. That got me to commenting what the errors and events were for, and while I was doing that, realized that we had one that's not used, and a couple that could use some better names.

A.k.a. "If you see a snake, just kill it, don't appoint a special committee on snakes"

A.k.a. shallow rabbit hole


pallets/msa/src/lib.rs
```diff
@@ -80,32 +80,54 @@ pub mod pallet {
...
 		NoKeyExists,
+		// An operation was attempted with a revoked Key
 		KeyRevoked,
+		// The number of key values has reached its maximum
 		KeyLimitExceeded,
+		// A transaction's Origin AccountId may not revoke itself
 		InvalidSelfRevoke,
+		// An MSA may not be its own delegate
 		InvalidSelfProvider,
-		InvalidMsaId,
```
shannonwells, line 103: This is the one that was unused.

saraswatpuneet (approved):

pallets/msa/src/lib.rs
saraswatpuneet: 🐍

PR #122, opened 2022-05-26
Added check_delegations method to MSA api

dopatraman:
# Goal
The goal of this PR is to add an RPC that checks the delegation status of a set of MSA ids.

# Implementation
This PR adds a `check_delegations` method to the MSA pallet's runtime API.


saraswatpuneet (approved):
Will clone and check but otherwise looks good

You can check the lint section and run a lint so the checks pass

👍

runtime/mrc/src/lib.rs
```diff
@@ -693,6 +694,16 @@ impl_runtime_apis! {
 		fn get_msa_id(key: AccountId) -> Result<Option<MessageSenderId>, DispatchError> {
 			Ok(Msa::get_owner_of(&key))
 		}
+
+		fn check_delegations(delegator_msa_ids: Vec<MessageSenderId>, provider_msa_id: MessageSenderId) -> Result<BTreeMap<MessageSenderId, bool>, DispatchError> {
+			let mut map = BTreeMap::new();
+			let provider = pallet_msa::types::Provider(provider_msa_id);
+			for id in delegator_msa_ids {
+				let delegator = pallet_msa::types::Delegator(id);
```
saraswatpuneet, line 702: @dopatraman  you can use rayon to make it a parallel iterator

saraswatpuneet: And move this to rpc call itself if doable .
An example https://github.com/LibertyDSNP/mrc/blob/5b2a52db5ee681e9f36c630b7ffa2fce4c7aec28/common/helpers/src/avro.rs#L58

But if it's not doable due to rayon it's fine as it is 

Just a suggestion

enddynayn (approved):

enddynayn:
spec file

pallets/msa/src/rpc/src/lib.rs
```diff
@@ -21,6 +22,9 @@ pub trait MsaApi<BlockHash, AccountId, BlockNumber> {
 
 	#[rpc(name = "msa_getMsaId")]
 	fn get_msa_id(&self, key: AccountId) -> Result<Option<MessageSenderId>>;
+
+	#[rpc(name = "msa_checkDelegations")]
+	fn check_delegations(&self, delegator_msa_ids: Vec<MessageSenderId>, provider_msa_id: MessageSenderId) -> Result<BTreeMap<MessageSenderId, bool>>;
```
aramikm, line 27: I think using a BTreeMap here is a bit overkill. It can just be `Vec` of some custom type or a pair. The reason is that this is a tree and populating it will be `O(NlogN)` but it can just be `O(N).` Maybe this is over optimization but since our resources are limited it might add up.

aramikm (approved):
Added an optimization comment but looks good and behaves as expected!

Screen Shot 2022-05-27 at 1 18 09 PM

Issue #20, opened 2022-04-19
[Design Doc] Delegate Permissions


wilwade:
We talked about various delegation permissions and now it is time to actually have some.

Output: A [design doc](https://github.com/LibertyDSNP/meta/blob/main/DESIGN_DOCS.md) that covers what Delegation Permissions is, why, and a version of how.

Ideas:
- Account recovery (Likely a different design doc)
- Limited Schema Ids (you can only send schemas 2 and 7 on my behalf)
- Time limited (after n blocks you must request re-authorization)
- Terms of Service? Permissions includes a hash of the ToS. This locks the ToS unless there is a re-delegation.
  - Chain/schema options for ToS instead of app specific

Terms:
- Account/User: The MSA of a user
- Provider: The MSA of the provider/app
- Permission: The generic option for any user
- Grant: The user level action/result. "A user grants a permission to a provider"

## Things We are Sure About
- User MUST opt-in to each permission
- Grant modification is a dual opt-in (User and Provider)
- Version 1:
  - Per schema grants by users for Creation
  - Revocation of a delegate ends all grants
  - Revocation of a grant should be "pre-paid" (Not staking it, just paying more for the create so the revoke is "covered")
  - All user account and delegation modification are token OR capacity

## Big Open Questions
- Public "Read" permission (DSNP or MRC?)
  - Future
- Do we track "use" of an app separate from write permissions
  - No

## Discussion Notes from 2022-05-25

Recording: https://drive.google.com/file/d/1wgsqx5KWTj2_K_NpCk00Ykk53P_GN8gX/view?usp=sharing

- Allowing a site to read my private data
  - Not posting on my behalf, reading my (private) data instead of writing data
  - Key management for reading private data
- Delegation for reading vs writing
- Do we need to have a way for the user to express how their data is used on "3rd party" sites (aka sites you don't have an explicit delegation with)
  - Maybe not MRC, but perhaps at the DSNP layer?
  - DSNP data use policy for each announcement (We already assume friends/followers can see it even on 3rd party sites)
- Do we need extendable/modular permissions or "roles"?
  - Grouping Schemas together somehow?
  - Roles are useful for users, but hide complexity (see Android/iOS permission mess)
  - What if there are 100's of schemas?
  - Could the wallet solve that problem?
- Delegation is a wallet level action
  - Users have a wallet interaction for delegation
- Do schemas have custom permissions?
  - Sub-permissions CRUD? We really only have Create as a permission
[...]

PR #136, opened 2022-06-02
Provider should be able to submit a message for a delegator#103

### LINKED ISSUE #103: Provider should be able to submit a message for a delegator
As a provider I want to be able to submit a message on behalf of another user (MSA) who has delegated to me

### Acceptance Criteria
- [x] Submitting a message on behalf should a user is successful
- [x] Submitting a message for a non existing delegate relationship should fail

## Architecture
- update message creation extrinsic with validation
- Watch out for pallet coupling issues

saraswatpuneet:
# Purpose
**From a provider's perspective:** I should be able to add messages to mrc on behalf of user as an option
**From a delegator's perspective:** An authorized Provider should be able to submit messages on my behalf
**Security perspective:** an unauthorized provider or not permissioned (#20) provider should not be able to submit messages on_behalf

Closes #103 

## Solution
- Extend add message extrinsic to allow for an optional ```on_behalf``` parameter which is the MSA id of actual delegator.
- Run checks to ensure a valid and authorized relationship between provider and delegator.
- Add a TODO comment as placeholder for future checks on permissions delegated to provider.
 
## Change summary
* Updated add extrinsic on messages
* Move some msa related primitives to common primitives for sharing 
* Extend AccountProvider trait to implement get_provider_info ensuring decoupling from MSA pallet
* Unit tests

## Outcomes:
 Design doc (#20) should lead to permission level checks on provider for acting on messages
 Creates issue: #140
## Steps to Verify
1. Two unit tests are added to mock two scenarios where a delegator has authorized provider and also a case where provider is not authorized 
2. Run MRC and verify using extrinsic calls from dashboard


pallets/messages/src/lib.rs
```diff
@@ -146,6 +149,21 @@ pub mod pallet {
 			let msa_id = T::AccountProvider::get_msa_id(&who);
 			ensure!(msa_id.is_some(), Error::<T>::InvalidMessageSourceAccount);
 
+			let message_sender_msa = msa_id.unwrap();
+			match on_behalf_of {
+				Some(producer) => {
+					let current_provider = Provider(message_sender_msa);
+					let current_delegator = Delegator(producer);
+					let provider_info = T::AccountProvider::get_provider_info_of(
```
enddynayn, line 157: maybe move this into a function inside of msa pallet that way you dont have to use Provider and Delegator here. I think this something we will be using often.

```
T::AccountProvider::ensure_valid_provider(provider_msa_id.into(), producer_msa_id.into())
```

enddynayn (approved):

pallets/messages/src/lib.rs
```diff
@@ -146,6 +149,21 @@ pub mod pallet {
...
+						current_provider,
+						current_delegator,
+					);
+					// TODO: check for specific permissions, such as if message schema
```
enddynayn, line 161: remove this comment and create an issue.

saraswatpuneet: #140

pallets/messages/src/lib.rs
saraswatpuneet: Since at runtime we are passing MSA singleton as AccountProvider it needs to be added there, however i believe this is still not clean because it depends on #20, So will require some refactoring at that point,  added a task in #140

pallets/schemas/src/rpc/src/lib.rs
```diff
@@ -79,7 +79,7 @@ where
 
 	fn check_schema_validity(
 		&self,
-		at: Option<<Block as BlockT>::Hash>,
+		_at: Option<<Block as BlockT>::Hash>,
```
shannonwells, line 82: thank you for fixing this warning lol

saraswatpuneet: 🐍

Issue #130, opened 2022-05-31
Investigate using Rayon for delegation storage


dopatraman:
Currently we iterate sequentially over delegation ids before checking their delegation statuses. Let's investigate using Rayon's parallel iterators for this.

aramikm:
We can probably do this by pulling out the iteration logic into RPC which can be use std libraries

saraswatpuneet:
Agreed with Aramik we should keep logic away from runtime

PR #138, opened 2022-06-02
Use parallel iteration for checking delegations

dopatraman:
# Goal
The purpose of this PR is to replace the sequential iteration we were using to check delegations with a parallel iteration, courtesy of `Rayon`.

# Implementation
We iterate over the `delegator_msa_ids` like before, but instead of inserting the delegation boolean into a map, we return a Vec of `(MessageSenderId, bool)` tuples. This way, we don't have to worry about Substrate/Rust's map thread safety.

Closes [#130](https://github.com/LibertyDSNP/mrc/issues/130)

### DIFF
--- a/pallets/msa/src/rpc/src/lib.rs
+++ b/pallets/msa/src/rpc/src/lib.rs
@@ -81,14 +81,13 @@ where
 		let at = BlockId::hash(self.client.info().best_hash);
 
 		let provider = Provider(provider_msa_id);
-		let tups = delegator_msa_ids
+
+		Ok(delegator_msa_ids
 			.par_iter()
 			.map(|&id| {
 				let delegator = Delegator(id);
 				(id, map_rpc_result(api.has_delegation(&at, delegator, provider)))
 			})
-			.collect();
-
-		Ok(tups)
+			.collect())
 	}
 }
--- a/pallets/msa/src/rpc/src/lib.rs
+++ b/pallets/msa/src/rpc/src/lib.rs
@@ -28,7 +28,7 @@ pub trait MsaApi<BlockHash, AccountId, BlockNumber> {
 		&self,
 		delegator_msa_ids: Vec<MessageSenderId>,
 		provider_msa_id: MessageSenderId,
-	) -> Result<BTreeMap<MessageSenderId, bool>>;
+	) -> Result<Vec<(MessageSenderId, Result<bool>)>>;
 }
 
 /// A struct that implements the `MessagesApi`.
--- a/pallets/msa/src/rpc/src/lib.rs
+++ b/pallets/msa/src/rpc/src/lib.rs
@@ -28,7 +28,7 @@ pub trait MsaApi<BlockHash, AccountId, BlockNumber> {
 		&self,
 		delegator_msa_ids: Vec<MessageSenderId>,
 		provider_msa_id: MessageSenderId,
-	) -> Result<Vec<(MessageSenderId, Result<bool>)>>;
+	) -> Result<Vec<(MessageSenderId, bool)>>;
 }
 
 /// A struct that implements the `MessagesApi`.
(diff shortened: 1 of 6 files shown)


enddynayn (approved):

pallets/msa/src/rpc/src/lib.rs
```diff
@@ -76,9 +76,19 @@ where
...
+		let provider = Provider(provider_msa_id);
+		let tups = delegator_msa_ids
+			.par_iter()
+			.map(|&id| {
+				let delegator = Delegator(id);
+				(id, map_rpc_result(api.has_delegation(&at, delegator, provider)))
+			})
+			.collect();
+
+		Ok(tups)
```
saraswatpuneet, line 92: "tups" made me giggle

runtime/mrc/src/lib.rs
```diff
@@ -695,14 +694,8 @@ impl_runtime_apis! {
...
-		fn check_delegations(delegator_msa_ids: Vec<MessageSenderId>, provider_msa_id: MessageSenderId) -> Result<BTreeMap<MessageSenderId, bool>, DispatchError> {
-			let mut map = BTreeMap::new();
-			let provider = pallet_msa::types::Provider(provider_msa_id);
-			for id in delegator_msa_ids {
-				let delegator = pallet_msa::types::Delegator(id);
-				map.insert(id, Msa::get_provider_info_of(provider, delegator).is_some());
-			}
-			Ok(map)
+		fn has_delegation(delegator: Delegator, provider: Provider) -> Result<bool, DispatchError> {
+			Ok(Msa::get_provider_info_of(provider, delegator).is_some())
```
saraswatpuneet, line 698: Question:
So will this always return true or false since is_some will evaluate only those two
Not sure you need a Result here then as Option will do, give you are Oking and no erring but maybe worth a try running a local test to throw an error from get_provider_info_of and see what happens

saraswatpuneet: If not you should do a match statement here and return error too

dopatraman: I believe that this method needs to return a `Result` to conform to substrate runtime constraints. I believe we paired on this, no?

saraswatpuneet: Rational behind return an error is that false could happen due to delegator not authorizing provider or false could also happen that they don't even exists

Hence my point about match check and return a error if not found

dopatraman: I'm not sure I follow. What are you trying to find out?

A couple of things we should keep in mind here:

1. All methods declared as runtime api methods must return `Result<_, DispatchError>`, so I'm not sure we can return a boolean here. I agree that since we are returning an `Ok` regardless, the Result seems superfluous. However, there is a chance that substrate reserves the error case here for issues dispatching calls to the runtime (that may be why the `Result<_, DispatchError>` is required.
2. The `get_provider_info_of` method is defined via macro so I'm not sure I can manipulate it to throw an error. There may be a way to mock it, but I'm not sure what insight we're trying to get here.

saraswatpuneet: I don't think you need to mock it 

I think I can now boil down to simple question


But I think we are good I checked the actual rpc and there is an error returned there given that was checked in before 


Approving it

saraswatpuneet (approved):

runtime/mrc/src/lib.rs
saraswatpuneet: https://github.com/LibertyDSNP/mrc/blob/81e290c474cd12438ff37d0cebf563eb980cfd0e/pallets/msa/src/rpc/src/lib.rs#L82

This is what i was searching so we are good

shannonwells (approved):
Much better with the unwrapped bool

runtime/mrc/src/lib.rs
aramikm, line 698: What is the delegation was revoked ? we are not removing any delegations, so a revoked delegation would still be some

aramikm (commented):
There seems to a bug with this implementation!

runtime/mrc/src/lib.rs
saraswatpuneet: That was my exact point that is_some just tells you it exists and still may have permissions

Glad you analyzed the right place 👏

saraswatpuneet: It should be matched and rather a permission and expired notice be returned to show the caller the actual config than booleans they can make their decision based on providerinfo

saraswatpuneet: Now I that I think about it

PR #145, opened 2022-06-06
Check expiration as a part of evaluating a valid delegation

dopatraman:
# Goal
The purpose of this PR is to introduce a check for a delegation's expiration as a part of evaluating delegation validity.

# Implementation
A provider's expiration is set at the time of revocation. When a delegation is queried for validity, we check to make sure that the expiration is greater than the current block. If it is, then the delegation has not expired (yet).

### DIFF
--- a/pallets/msa/src/lib.rs
+++ b/pallets/msa/src/lib.rs
@@ -130,6 +130,8 @@ pub mod pallet {
 		DelegationRevoked,
 		/// The operation was attempted with an unknown delegation
 		DelegationNotFound,
+		/// The operation was attempted with an expired delegation
+		DelegationExpired,
 	}
 
 	#[pallet::call]
@@ -401,6 +403,16 @@ impl<T: Config> Pallet<T> {
 			Ok(())
 		})?;
 
+		Ok(())
+	}
+
+	pub fn ensure_valid_delegation(provider: Provider, delegator: Delegator) -> DispatchResult {
+		let current_block = frame_system::Pallet::<T>::block_number();
+		let info = Self::get_provider_info_of(provider, delegator).ok_or(Error::<T>::DelegationNotFound)?;
+		if info.expired == T::BlockNumber::zero() {
+			return Ok(())
+		}
+		ensure!(info.expired >= current_block, Error::<T>::DelegationExpired);
 		Ok(())
 	}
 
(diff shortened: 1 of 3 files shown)


pallets/msa/src/lib.rs
```diff
@@ -404,6 +406,17 @@ impl<T: Config> Pallet<T> {
...
 
+	pub fn ensure_valid_delegation(provider: Provider, delegator: Delegator) -> DispatchResult {
+		let current_block = frame_system::Pallet::<T>::block_number();
+		let info = Self::get_provider_info_of(provider, delegator)
+			.ok_or(Error::<T>::DelegationNotFound)?;
+		if info.expired == T::BlockNumber::zero() {
+			return Ok(())
+		}
+		ensure!(info.expired >= current_block, Error::<T>::DelegationExpired);
+		Ok(())
```
enddynayn, line 417: ok((info))

pallets/msa/src/tests.rs
```diff
@@ -911,3 +911,53 @@ pub fn remove_delegation_by_provider_errors_when_no_delegator_msa_id() {
...
+	new_test_ext().execute_with(|| {
+		let provider = Provider(1);
+		let delegator = Delegator(2);
+
+		assert_ok!(Msa::add_provider(provider, delegator));
+
+		System::set_block_number(System::block_number() + 1);
+		assert_ok!(Msa::ensure_valid_delegation(provider, delegator));
+
+		// System::set_block_number(System::block_number() + 1);
```
enddynayn, line 953: rm

enddynayn (approved):

enddynayn:
great! ⛹️‍♂️

pallets/msa/src/lib.rs
```diff
@@ -404,6 +406,17 @@ impl<T: Config> Pallet<T> {
...
 	}
 
+	pub fn ensure_valid_delegation(provider: Provider, delegator: Delegator) -> DispatchResult {
+		let current_block = frame_system::Pallet::<T>::block_number();
+		let info = Self::get_provider_info_of(provider, delegator)
+			.ok_or(Error::<T>::DelegationNotFound)?;
+		if info.expired == T::BlockNumber::zero() {
+			return Ok(())
+		}
+		ensure!(info.expired >= current_block, Error::<T>::DelegationExpired);
```
aramikm, line 416: I think it probably should be `info.expired > current_block`. Maybe shorten it like the following

`ensure!(info.expired == T::BlockNumber::default() || info.expired > current_block, Error::<T>::DelegationExpired);`

PR #150, opened 2022-06-07
Design Doc Permission/ Grants 

saraswatpuneet:
Design doc intended to discuss and plan about permissions/grants implementation as a version 1 of delegator<->provider trust relationship

Closes: #20


designdocs/provider_permissions.md
```diff
@@ -0,0 +1,207 @@
...
+    Block,
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub struct Grant {
+    pub grant_type: GranType,
+    pub tos_hash: Vec<u8>,
+    pub expiry_time: u64,
+}
+```
```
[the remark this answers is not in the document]
saraswatpuneet: In doing show i think we can just keep tos and expiry at permission levels and not do it at grant level for first few versions, if a need arises

designdocs/provider_permissions.md
```diff
@@ -0,0 +1,207 @@
...
+### add_schema_grants()
+
+An extrinsic to allow a provider to request publish write to list of schemas. Rendering them **Restricted** status.
+
+- Parameters:
+    1. **provider_msa**: The MSA of the provider/app.
+    2. **delegator_msa**: The MSA of a user.
+    3. **schemas**: The list of schemas for which the provider wants to grant publish write typically ```Vec<SchemaId>```.
+    4. **tos_hash**: The hash of terms of service between a delegator and provider.
+
```
saraswatpuneet, line 111: provider msa can be extract from origin but these are just placeholders of what api will require from design perspective

designdocs/provider_permissions.md
```diff
@@ -0,0 +1,205 @@
...
+The proposed solution is to provide delegate level permissions and schema level grants to delegators and providers alike. This will enable a provider to publish data on behalf of a delegator, while also allowing a delegator to restrict a provider from publishing data on their behalf for specific schema.
+
+Note: The terminology and implementation are subject to change at issue resolution.
+
+### Permissions
+
+Permission is a generic option for any user. For version 1 of this implementation, the following options are available:
+
+- ***PUBLISHER***: Where a user grants full publication rights to a provider for any schema available to provider via MRC. This can be modified to be called a dsnp publisher where all dsnp related schemas are granted to provider. In other sense this could be super admin permission that can be granted via a governance mechanism.
```
wilwade, line 46: I suggest we remove the Publisher option for version 1.

Why?
- Simplicity. The Restricted option can do everything
- User safety: Explicit user permission while annoying for upgrading to a different version, it protects against misuse. A grant could be given to someone who then creates a new ("bad" for the user) protocol and immediately start using it.

designdocs/provider_permissions.md
```diff
@@ -0,0 +1,205 @@
...
+    2. **delegator_msa**: The MSA of a user.
+    3. **schema_ids**: The list of schemas for which the provider/delegator wants to block publishing rights, typically ```Vec<SchemaId>```.
+
+- Restrictions: Origin must own provider ```msa_id``` delegated by delegator ```msa_id```.
+
+- Outcomes: Provider permissions are set to **Restricted**. While schema level grants is set to blocked for a given ```schema_id```.
+
+- Notes:
+  - Revocation of a grant should be "pre-paid" (Not staking it, just paying more for the create so the revoke is "covered")
+  - If this action is already paid off while [adding schema grants](#add_schema_permissions), this can be done via rpc.
```
wilwade, line 152: I'm not sure I understand this line. Revocation is still an extrinsic, it is just free for the caller (and requires that the caller have that delegation)

wilwade (changes requested):
Two notes. I think the only big change is removing the wildcard publisher option.

designdocs/provider_permissions.md
```diff
@@ -0,0 +1,205 @@
+# Delegator <-> Provider Permissions/Grants
```
wilwade, line 1: p.s. Don't forget to update the readme file

designdocs/provider_permissions.md
saraswatpuneet: Makes sense for version 1 we can have restricted and can extend things if needed

designdocs/provider_permissions.md
saraswatpuneet: reworded it, in summary we want to fee to call this extrinsic be accounted for during

saraswatpuneet:
@wilwade  while its in review is it wise to create couple of stories given its pretty standard design we are going for

designdocs/provider_permissions.md
```diff
@@ -0,0 +1,203 @@
...
+
+The proposed solution is to provide delegate level permissions and schema level grants to delegators and providers alike. This will enable a provider to publish data on behalf of a delegator, while also allowing a delegator to restrict a provider from publishing data on their behalf for specific schema.
+
+Note: The terminology and implementation are subject to change at issue resolution.
+
+### Permissions
+
+Permission is a generic option for any user. For version 1 of this implementation, the following options are available:
+
+- ***PUBLISHER***: Where a user grants full publication rights to a provider for any schema available to provider via MRC. This can be modified to be called a dsnp publisher where all dsnp related schemas are granted to provider. In other sense this could be super admin permission that can be granted via a governance mechanism. For version 1 we will not implement this role.
```
wilwade, line 46:
```suggestion
```

designdocs/provider_permissions.md
```diff
@@ -0,0 +1,203 @@
...
+}
+```
+
+### Grants
+
+Grants enable delegators as well as providers to restrict one another from publishing data on specific schema(s). For version 1 of this implementation, the following options are available:
+
+- ***PUBLISH***: Where a delegator grants a provider to publish data on their behalf for specific schema(s) only. This is the default state of a provider on MRC, where a provider has to explicitly provide a list of schema(s) for which they are allowed to publish data on behalf of the delegator. This also enables a delegator to opt in to publish their data.
+
+- ***BLOCK***: When a delegator or provider want to restrict publication of data on specific schema(s). This is default state of any schemas, not authorized by delegator or provider as part of schema grants request.
```
wilwade, line 72:
```suggestion
```

saraswatpuneet:
Edited and reduced to minimal viable version for permissions/grants

PR #155, opened 2022-06-08
fix(messages): ensure delegation is valid

enddynayn:
Ensure that when submitting a message on behalf of
a delegator that the provider has not been revoked.

issue-153

### DIFF
--- a/pallets/messages/src/lib.rs
+++ b/pallets/messages/src/lib.rs
@@ -152,17 +152,14 @@ pub mod pallet {
 			let message_sender_msa = info.msa_id;
 
 			match on_behalf_of {
-				Some(producer) => {
-					let current_provider = Provider(message_sender_msa);
-					let current_delegator = Delegator(producer);
-					let provider_info = T::AccountProvider::get_provider_info_of(
-						current_provider,
-						current_delegator,
-					);
-					ensure!(provider_info.is_some(), Error::<T>::UnAuthorizedDelegate);
-				},
+				Some(producer) => T::AccountProvider::ensure_valid_delegation(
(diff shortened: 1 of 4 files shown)


pallets/messages/src/lib.rs
```diff
@@ -152,17 +152,14 @@ pub mod pallet {
...
-						current_provider,
-						current_delegator,
-					);
-					ensure!(provider_info.is_some(), Error::<T>::UnAuthorizedDelegate);
-				},
+				Some(producer) => T::AccountProvider::ensure_valid_delegation(
+					Provider(message_sender_msa),
+					Delegator(producer),
+				)
+				.map_err(|_| Error::<T>::UnAuthorizedDelegate)?,
```
enddynayn, line 159: maybe we should not be remapping these errors...could be an improvement on a different pr.

pallets/messages/src/mock.rs
```diff
@@ -118,6 +119,14 @@ impl AccountProvider for AccountHandler {
 		let info = KeyInfo { msa_id: get_msa_from_account(*key), nonce: 0, expired: 100 };
 		Ok(info)
 	}
+
+	fn ensure_valid_delegation(provider: Provider, _delegator: Delegator) -> DispatchResult {
+		if provider == Provider(2000) {
+			return Err(DispatchError::Other("some delegation error"))
+		};
+
```
saraswatpuneet, line 127: I think it is ok but message pallet will need permissions data which was returned previously to make decision around a message is allowed to be published or not given the permissions/grants
So this will require refactoring again

Issue #156, opened 2022-06-08
Message original sender


enddynayn:
When message is being sent by a provider, the original message sender msa should get stored.

iamianM:
Hey @enddynayn I was just messing around with this and wondering the same thing. If alice is a provider of like 4 msa_id's it just seems to use the most recently delegated id. How would it decide though? Alice is signing the Messages.add transactions in order to pay the fees, so I don't see where the delegator is involved. Should the delegator by signing something?

enddynayn:
Let's say that Alice is a provider to Bob. In other words, Bob is a delegator and Alice is the delegate.
If Alice wants to sends a message on behalf of Bob the second parameter of Messages.add, on_behalf, is populated with Bobs MsaId. We can verify that the message is valid, offchain, by checking that Alice was indeed a provider to Bob.

iamianM:
@enddynayn I only see schemaId and message as parameters for Messages.add
I'm confused as to what you mean by the second parameter `on_behalf`

saraswatpuneet:
@enddynayn  you mean if on_behalf is provided, do we need to store if message was sent on behalf of "this msa _id"? to be clear

PR #159, opened 2022-06-08
add delegator msa_id to message for delegated message publish

saraswatpuneet:
A bug was reported #156 
- MRC message should include if message was stored on behalf of a delegator
- Add a field to message type to hold that information as optional (i.e. if actually message was sent on behalf or not)
- Update unit tests and logic to refer to sender_msa_id as the id of actualy sender of message (could be a provider or delegator) and on_behalf_of as msa_id if the message is sent to publish on behalf of a user

pair: @enddynayn  verify if the intention behind the bug is what being addressed here

### DIFF
--- a/common/primitives/src/messages.rs
+++ b/common/primitives/src/messages.rs
@@ -15,9 +15,10 @@ use utils::*;
 pub struct MessageResponse<AccountId, BlockNumber> {
 	#[cfg_attr(feature = "std", serde(with = "as_hex"))]
 	pub data: Vec<u8>, //  Serialized data in a user-defined schema format
-	pub signer: AccountId,       //  Signature of the signer
-	pub msa_id: MessageSenderId, //  Message source account id (the original sender)
-	pub index: u16,              // index in block to get total order
+	pub signer: AccountId,                     //  Signature of the signer
+	pub sender_msa_id: MessageSenderId,        //  Message source account id (the original sender)
+	pub on_behalf_of: Option<MessageSenderId>, //  Message source account id (the original message producer)
+	pub index: u16,                            // index in block to get total order
 	pub block_number: BlockNumber,
 }
 
--- a/pallets/messages/src/lib.rs
+++ b/pallets/messages/src/lib.rs
@@ -174,7 +174,8 @@ pub mod pallet {
 					data: message.try_into().unwrap(), // size is checked on top of extrinsic
 					signer: key,
 					index: current_size,
-					msa_id: message_sender_msa,
+					sender_msa_id: message_sender_msa,
+					on_behalf_of,
 				};
 				existing_messages
 					.try_push((m, schema_id))
--- a/pallets/messages/src/tests.rs
+++ b/pallets/messages/src/tests.rs
@@ -14,7 +14,8 @@ fn populate_messages(schema_id: SchemaId, message_per_block: Vec<u32>) {
 		let mut list = BoundedVec::default();
 		for _ in 0..*count {
 			list.try_push(Message {
-				msa_id: 10,
+				sender_msa_id: 10,
+				on_behalf_of: None,
 				data: payload.clone().try_into().unwrap(),
 				index: counter,
 				signer: 1,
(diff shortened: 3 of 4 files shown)


common/primitives/src/messages.rs
```diff
@@ -15,9 +15,10 @@ use utils::*;
...
 	pub data: Vec<u8>, //  Serialized data in a user-defined schema format
-	pub signer: AccountId,       //  Signature of the signer
-	pub msa_id: MessageSenderId, //  Message source account id (the original sender)
-	pub index: u16,              // index in block to get total order
+	pub signer: AccountId,                     //  Signature of the signer
+	pub sender_msa_id: MessageSenderId,        //  Message source account id (the original sender)
+	pub on_behalf_of: Option<MessageSenderId>, //  Message source account id (the original message producer)
+	pub index: u16,                            // index in block to get total order
 	pub block_number: BlockNumber,
 }
```
saraswatpuneet, line 24: @aramikm  what are your thoughts, given messages can be added by sender msa_id but sender could be a provider submitting on behalf of a delegator

So is it a goo update to message being stored?

common/primitives/src/messages.rs
```diff
@@ -15,9 +15,10 @@ use utils::*;
 pub struct MessageResponse<AccountId, BlockNumber> {
 	#[cfg_attr(feature = "std", serde(with = "as_hex"))]
 	pub data: Vec<u8>, //  Serialized data in a user-defined schema format
-	pub signer: AccountId,       //  Signature of the signer
-	pub msa_id: MessageSenderId, //  Message source account id (the original sender)
-	pub index: u16,              // index in block to get total order
+	pub signer: AccountId,                     //  Signature of the signer
+	pub sender_msa_id: MessageSenderId,        //  Message source account id (the original sender)
```
enddynayn, line 19: I dont know if we need this

common/primitives/src/messages.rs
aramikm: I don't think `on_behalf_of` is necessary since we can get msa_id of the signer via stored `signer` key.

common/primitives/src/messages.rs
saraswatpuneet: We should do a parking lot on it tomorrow, I think we might need it, if we want to keep a history of who delegated what for a delegator, like same schema can be published by one provider as well as others, having a placeholder of who sent it will help differentiate anyone look at it for example

aramikm (commented):
I like the renaming of fields to make them more descriptive but I don't think storing that additional field is necessary. If we didn't have any storage restrictions I would definitely agree with storing it but considering on-chain limitation I think we can operate without it.

common/primitives/src/messages.rs
aramikm: I think since when we revoke a delegation we still keep the history and when it got revoked, then it would be easy to track this back to the actual provider.

saraswatpuneet: So we agree @enddynayn  and @aramikm  we can remove that field and just rename it 


I will push an update

common/primitives/src/messages.rs
```diff
@@ -15,9 +15,9 @@ use utils::*;
 pub struct MessageResponse<AccountId, BlockNumber> {
 	#[cfg_attr(feature = "std", serde(with = "as_hex"))]
 	pub data: Vec<u8>, //  Serialized data in a user-defined schema format
-	pub signer: AccountId,       //  Signature of the signer
-	pub msa_id: MessageSenderId, //  Message source account id (the original sender)
-	pub index: u16,              // index in block to get total order
+	pub signer: AccountId,              //  Signature of the signer
+	pub sender_msa_id: MessageSenderId, //  Message source account id (the original sender)
```
enddynayn, line 19: I think sender might still cause confusion, how about `origin_msa_id`?

enddynayn (approved):

common/primitives/src/messages.rs
aramikm: lol, I think origin might also be confusing since in extrinsic we have `origin` caller which is our case can be provider

aramikm (approved):

PR #160, opened 2022-06-08
fix(messages): ensure origin sender is stored

enddynayn:
Ensure that the origin sender of the message is store
when it exist.

issue-156


enddynayn:
@saraswatpuneet added this in his pr https://github.com/LibertyDSNP/mrc/pull/159

Issue #129, opened 2022-05-31
Design Doc for Registering Providers


wilwade:
We want to limit which MSAs can do certain things and require additional information so that we can provide better safety and guarantees to users.

`Provider`: An MSA that is registered for being able to be delegated to and being the target of capacity rewards when a person stakes to the network for token rewards.

## Things to Include
- How to identify a provider (msa registration, separate?)
- Getting a list of providers
- How to discontinue being a provider
- Governance approval is NOT required
- What happens if the provider MSA id is removed?
- How can I become a provider?
  - Expected to be a token transaction
- Who can become a provider?
  - Anyone who pays the token transaction
- Run Discussion around gatekeeping and making sure that providers are possible for all (community run providers etc...)

## Limits
- Delegating. Users cannot just delegate to anyone and is limited to just providers
- Callback safety. Wallets or other applications interacting with a registered provider should use the chain published domain for callbacks
  - @shannonwells has some flows around callbacks that might be helpful

## Additional Information
- MSA Id attaches to this information
- Callback (domain)
  - prove ownership of domain?
    - Registrar pallet perhaps?
- Name (What if this is close to another? Should these be unique? utf-8 or ascii?)
- Logo URL? (optional? or forced to be on the domain?)
- Description? (optional?)
- Other?

## Notes
- Interactions are token transactions
- Future need: Capacity rewards when network staking (#40)

mehtaishita:
To continue off of Enddy's comment about allowing delegation to a 'good friend' who isn't registered as a provider: 

What if a user can ask to delegate to another non-provider msa id, and in that workflow also allow the user to register as a temporary provider? Effectively, allowing various stages of being a provider such as 'Samaritan', 'professional' etc?

PR #186, opened 2022-06-16
Documentation Day Renaming: Message Source Id and Message.provider_key

wilwade:
Blocked pending merge of `doc/documentation_day_msa` into `doc/documentation_day` first.

- Rename type `MessageSenderId` -> `MessageSourceId`
- Rename Message field `sender` to `provider_key` for clarity

### DIFF
--- a/designdocs/MESSAGE_STORAGE.md
+++ b/designdocs/MESSAGE_STORAGE.md
@@ -47,8 +47,8 @@ using [StorageDoubleMap](https://docs.substrate.io/rustdocs/latest/frame_support
 Following is a proposed data structure for storing a Message on chain.
 ```rust
 pub struct Message<AccountId> {
-    pub data: Vec<u8>,		    //  Serialized data in a user-defined schemas format
-    pub signer: AccountId,	    //  Signature of the signer
+    pub payload: Vec<u8>,		    //  Serialized data in a user-defined schemas format
+    pub provider_key: AccountId,	    //  Signature of the signer
     pub msa_id: u64,                //  Message source account id (the original source of the message)
     pub index: u16,		    //  Stores index of message in block to keep total order
 }
--- a/designdocs/MESSAGE_STORAGE.md
+++ b/designdocs/MESSAGE_STORAGE.md
@@ -49,7 +49,7 @@ Following is a proposed data structure for storing a Message on chain.
 pub struct Message<AccountId> {
     pub data: Vec<u8>,		    //  Serialized data in a user-defined schemas format
     pub signer: AccountId,	    //  Signature of the signer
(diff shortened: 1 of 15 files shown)


enddynayn (approved):

designdocs/MESSAGE_STORAGE.md
```diff
@@ -47,9 +47,9 @@ using [StorageDoubleMap](https://docs.substrate.io/rustdocs/latest/frame_support
 Following is a proposed data structure for storing a Message on chain.
 ```rust
 pub struct Message<AccountId> {
-    pub data: Vec<u8>,		    //  Serialized data in a user-defined schemas format
-    pub signer: AccountId,	    //  Signature of the signer
-    pub msa_id: u64,                //  Message source account id (the original sender)
+    pub payload: Vec<u8>,		    //  Serialized data in a user-defined schemas format
+    pub provider_key: AccountId,	    //  Signature of the signer
```
enddynayn, line 51: I think you mentioned there are two types of provider:
1. registered provider
2. Provider 

A provider is some one who has capacity but cannot delegate?
A registered provider is someone who can delegate and is verified?

wilwade: True. This can be either. #129 talks about registering providers

enddynayn: In this case this would be a registered provider since we have a delegation validation.

wilwade: It *might* be a registered provider. Since `on_behalf_of` is Optional. If `on_behalf_of` is `Some` then it must be a registered provider. Otherwise `msa_id` is the id from the `provider_key`.

It is a bit confusing as we have that msa_id as optional. I'm open for ideas on how to reduce that confusion.

Here's how it works in my head:
- Every `Message` must have a `provider` (stored via the `provider_key`). The provider of the capacity that sent the message.
- Every message with an on-chain payload will have a source (the `msa_id`) which might be the msa_id of the provider, but could be a validated point in time delegation
- Every message with off-chain payload will ONLY have a provider. (The source, if any, must be in the off-chain payload and validated at read time)

Another way to put it

| -- | Provider | Sender |
| --- | --- | --- |
| Referenced via | key | msa_id |
| On-chain Msg Self Provider | Required | Same as Provider |
| On-chain Msg Delegated | Required | Required, Validated Delegation |
| Off-chain Msg | Required | Optional & Off-chain |

saraswatpuneet: If on_behalf_of is some we know its a delegation, now about msa_id being optional, at the code level per se we can define bunch of macros 
However we can also enforce to always include on behalf of in cases its same as origin we know its provider else registered provider

PR #203, opened 2022-06-24
[Draft] Support Multiple Formats for Payloads: Format Avro

johnmela:
# Goal
Add the ability to support multiple formats(Avro, Parquet, etc.) for message schemas

### DIFF
--- a/pallets/schemas/src/lib.rs
+++ b/pallets/schemas/src/lib.rs
@@ -59,8 +59,11 @@ mod mock;
 #[cfg(feature = "runtime-benchmarks")]
 mod benchmarking;
 
+mod types;
+
 pub use pallet::*;
 pub mod weights;
+pub use types::*;
 pub use weights::*;
 
 #[frame_support::pallet]
@@ -145,7 +148,7 @@ pub mod pallet {
 		_,
 		Twox64Concat,
 		SchemaId,
-		BoundedVec<u8, T::SchemaModelMaxBytesBoundedVecLimit>,
+		Schema<T::SchemaModelMaxBytesBoundedVecLimit>,
 		OptionQuery,
 	>;
 
@@ -230,11 +233,11 @@ pub mod pallet {
 
 	impl<T: Config> Pallet<T> {
 		fn add_schema(
-			model: BoundedVec<u8, T::SchemaModelMaxBytesBoundedVecLimit>,
+			schema: Schema<T::SchemaModelMaxBytesBoundedVecLimit>,
 			schema_id: SchemaId,
 		) -> DispatchResult {
 			<SchemaCount<T>>::set(schema_id);
-			<Schemas<T>>::insert(schema_id, model);
+			<Schemas<T>>::insert(schema_id, schema);
 			Ok(())
 		}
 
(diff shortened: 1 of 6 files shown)


pallets/schemas/src/lib.rs
```diff
@@ -243,10 +251,9 @@ pub mod pallet {
 		}
 
 		pub fn get_schema_by_id(schema_id: SchemaId) -> Option<SchemaResponse> {
-			// TODO: This will eventually be a struct with other properties. Currently it is just the model
-			if let Some(model) = Self::get_schema(schema_id) {
+			if let Some(schema) = Self::get_schema(schema_id) {
 				// this should get a BoundedVec out
-				let model_vec = model.into_inner();
+				let model_vec = schema.model.into_inner();
 				let response = SchemaResponse { schema_id, model: model_vec };
```
aramikm, line 257: I think `SchemaResponse` should also return the type

aramikm (commented):
Overall looks good. Added a comment and I think you need to run the following to check linting.

`cargo fmt --check`

and following to fix

`cargo fmt`

pallets/schemas/src/lib.rs
wilwade: It is `Option<SchemaResponse>` because there might not be a schema with that id.

johnmela:
> Overall looks good. Added a comment and I think you need to run the following to check linting.
> 
> `cargo fmt --check`
> 
> and following to fix
> 
> `cargo fmt`

cargo fmt isn't producing any code changes but I have VSCode autoformat on save enabled. I'm not sure if its doing this through cargo's built in linting or not.

So something I noticed with GH actions, it appears to run through two jobs in parallel (tests, lint) when executing a workflow on a pull request. If I am understanding correctly, both jobs are doing the same thing but the job named lint is the one that failed first. It doesn't seem to like `Default` trait in the derive macro on `ModelType`. 

My local `cargo check` doesn't throw this error. It could be because we are using an old nightly build for the rust toolchain in CI `nightly-2021-11-07`. We probably want to update that to reflect a newer build or maybe even target 
a stable release.

common/primitives/src/schema.rs
```diff
@@ -1,11 +1,12 @@
-#[cfg(feature = "std")]
 use crate::utils;
 use codec::{Decode, Encode};
 use scale_info::TypeInfo;
 #[cfg(feature = "std")]
 use serde::{Deserialize, Serialize};
 use sp_std::prelude::*;
 #[cfg(feature = "std")]
+use types::ModelType;
```
aramikm, line 8: I don't think this is correct. Since `#[cfg(feature = "std")]` is on top of `use types::ModelType;` this include will only happen in `std` builds and it will cause issues in run time build.

I think lines [8-10] should probably be like the following
```
#[cfg(feature = "std")]
use utils::*;
use types::ModelType;
```

pallets/schemas/src/lib.rs
aramikm: You still need to assign mode_type from schema to SchemaResponse

common/primitives/src/schema.rs
```diff
@@ -1,11 +1,12 @@
-#[cfg(feature = "std")]
```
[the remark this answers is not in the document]
johnmela: No, this was a mistake thanks for the catch

common/primitives/src/schema.rs
johnmela: Yes you are correct, I'll pay closer attention to the cfg macros above use statements

enddynayn (approved):

common/primitives/src/schema.rs
```diff
@@ -11,6 +11,20 @@ use utils::*;
...
+	/// Message payload modeled with Apache Avro
+	AvroBinary,
+}
+
+impl Default for ModelType {
+	fn default() -> Self {
+		Self::AvroBinary
+	}
+}
+
```
dopatraman, line 27: just curious -- why do we need this trait implementation?

enddynayn:
fabulous!  💯

dopatraman (commented):
Looks good to me.

aramikm (approved):
icons8-ship-48 it!

PR #209, opened 2022-06-29
MRC -> Frequency: Readmes

wilwade:
We are renaming the chain from MRC to Frequency. This kicks it off by updating the README files switching to the new name and adding a Glossary entry for MRC as the old name.


enddynayn (approved):
Awesome 🙌

PR #215, opened 2022-06-30
Update Design Docs to use Frequency

### LINKED ISSUE #211: Update Documentation MRC -> Frequency
Our Design Docs and Code documentation use the old MRC a lot. We need to update them to all use Frequency

- Should also work with the authors of any design docs prs to update those without conflicts. Specifically #90
- Should NOT update any code paths or code variable names. That is #212
- Careful with find and replace!

dopatraman:
# Goal
The goal of this PR is to replace occurrences of "MRC" with "Frequency" so our documentation stays up to date with current naming.

Closes #211 

# Checklist
- [x] Design doc(s) updated

### DIFF
--- a/designdocs/ACCOUNTS.md
+++ b/designdocs/ACCOUNTS.md
@@ -1,8 +1,8 @@
 ## Context and Scope
 
-The ability to create an on-chain account without a token to increase user accessibility for MRC.
+The ability to create an on-chain account without a token to increase user accessibility for Frequency.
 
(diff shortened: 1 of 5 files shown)


aramikm (approved):

saraswatpuneet (changes requested):
Just leaving this blocker for @wilwade  to confirm if design docs should mention "Frequency chain" instead of only frequency for a better readability

wilwade (approved):

wilwade:
> Just leaving this blocker for @wilwade to confirm if design docs should mention "Frequency chain" instead of only frequency for a better readability

Good question. We went back and forth on this when talking about the chain name. We landed on just "Frequency".

PR #216, opened 2022-07-01
Feat/on chain json validation rebase/111

### LINKED ISSUE #111: On-chain Schema JSON Validation Prior to Registration

## Acceptance Criteria

## Notes

mehtaishita:
# Goal
The goal of this PR is to add Json model validation on chain

Closes #111 

# Discussion
Null or empty Json schemas will be invalidated and not registered (even though serde-json does not consider them invalid)

# Checklist
- [ ] Chain spec updated
- [ ] Updated the js/rpc code if a custom RPC added/changed 
- [ ] Design doc(s) updated
- [x] Tests added
- [ ] Benchmarks added
- [ ] Weights updated


mehtaishita:
Comments from previous PR:

[x] Add negative json validation test
[x] Add utf 8 encoding test
[x] Clean up files changed
[x] Move serde to schema pallet since it needs to be no-std

Discussion:
Do I need to provide feedback to the user trying to register their schema as to why their schema failed to register? 
The only error code that exists on the schema pallet currently is 'invalid schema'

pallets/schemas/src/mock.rs
```diff
@@ -53,7 +53,7 @@ impl WeightToFeePolynomial for WeightToFee {
 impl pallet_schemas::Config for Test {
 	type Event = Event;
 	type WeightInfo = ();
-	type MinSchemaModelSizeBytes = ConstU32<5>;
+	type MinSchemaModelSizeBytes = ConstU32<15>;
```
dopatraman, line 56: Just curious, what does `15` represent here?

dopatraman (approved):

pallets/schemas/src/mock.rs
mehtaishita: the number of bytes of string representation of the schema model to be registered

enddynayn (approved):
Nice! 👍

pallets/schemas/src/lib.rs
```diff
@@ -66,6 +66,7 @@ pub use pallet::*;
 pub mod weights;
 pub use types::*;
 pub use weights::*;
+mod serde;
```
enddynayn, line 69: do you need to do:
```
pub use serde:: validate_json_model
```

PR #229, opened 2022-07-07
Delegation Revocation should be free 

### LINKED ISSUE #205: Delegation Revocation should be free
As a user, I want to be able to revoke delegation, even if I have _zero_ tokens.

## Acceptance Criteria
- [ ] Given that an MSA has delegated to a provider, I should be able to revoke that delegation without any tokens.
- [ ] There should be a validate check to ensure the delegation exists for ddos prevention by stopping propagation
- [ ] Create a story [chore] to update the base weight citing the new signed extension.

## Notes
- Method: `revoke_msa_delegation_by_delegator`
- See: `Pays::No`
- Two options
    - `Pays::No + Signed Extension` https://docs.substrate.io/reference/transaction-format/#signed-extensions
    - `Unsigned Transaction + signature as a param` like heartbeat https://docs.substrate.io/main-docs/fundamentals/transaction-types/#unsigned-transactions
- We should use `Pays::No + Signed Extension`
    - easier for wallets and clients
- Helpful links
    - https://substrate.stackexchange.com/questions/2923/signed-extensions-and-weights
    - https://github.com/paritytech/substrate/issues/6657

johnmela:
# Goal
The goal of this PR is to make Delegation Revocation have no cost, even for accounts with no token balance.

Closes #205

# Acceptance Criteria
- [x] Given that an MSA has delegated to a provider, I should be able to revoke that delegation without any tokens.
- [x] There should be a validate check to ensure the delegation exists for ddos prevention by stopping propagation

# Checklist
- [x] Tests added
- [x] Weights updated


pallets/msa/src/lib.rs
```diff
@@ -738,3 +740,75 @@ impl<T: Config> AccountProvider for Pallet<T> {
...
+
+	fn validate(
+		&self,
+		who: &Self::AccountId,
+		call: &Self::Call,
+		_info: &DispatchInfoOf<Self::Call>,
+		_len: usize,
+	) -> TransactionValidity {
+		match call.is_sub_type() {
+			Some(Call::revoke_msa_delegation_by_delegator { provider_msa_id, .. }) => {
```
aramikm, line 800: Looks like this will work, the only concern that I have is that someone flooding the queue with these requests and since the request has to do DB lookup(s) it would have effect of performance. One thing I was hoping we might be able to achieve was to group-by these events in queue in a way that you can not have more than one request per account at the same time, this would eliminate spamming using **one** account but does not help with using multiple different accounts to spam.

aramikm (approved):
Implemented approach looks like a good initial check to have, I added a potential attack vector which we might need to take a look later.

pallets/msa/src/lib.rs
```diff
@@ -738,3 +740,75 @@ impl<T: Config> AccountProvider for Pallet<T> {
...
+		self,
+		who: &Self::AccountId,
+		call: &Self::Call,
+		info: &DispatchInfoOf<Self::Call>,
+		len: usize,
+	) -> Result<Self::Pre, TransactionValidityError> {
+		self.validate(who, call, info, len).map(|_| ())
+	}
+
+	fn validate(
```
shannonwells, line 792: This needs a comment about what is going on and why

pallets/msa/src/lib.rs
```diff
@@ -738,3 +740,75 @@ impl<T: Config> AccountProvider for Pallet<T> {
 		Ok(result.unwrap())
 	}
 }
+
+/// Check to ensure that a provider has not yet been revoked if the
```
shannonwells, line 744: This deserves much more explanation, would be nice to call it out as a SignedExtension here at least.

pallets/msa/src/tests.rs
```diff
@@ -955,3 +982,101 @@ pub fn delegation_expired() {
 		);
 	})
 }
+
+#[test]
+fn signed_extension_revoke_msa_delegation_by_delegator() {
```
shannonwells, line 987: These tests are long enough that Reviewer and Future Developer could use some comments to understand what's going on and what is being tested.

pallets/msa/src/tests.rs
```diff
@@ -955,3 +982,101 @@ pub fn delegation_expired() {
...
+			len,
+		);
+		assert_ok!(result);
+	});
+}
+
+#[test]
+fn signed_extension_validation_failure_on_revoked() {
+	new_test_ext().execute_with(|| {
+		let (key_pair, _) = sr25519::Pair::generate();
```
shannonwells, line 1024: Non-blocking: A function that does this setup (create keypairs and providers and then add one as a delegate) instead of duplicating it would be really useful, especially for future tests of this pallet.

shannonwells (changes requested):
Since we're getting more eyes on this project and people will be potentially using this code, this needs some more comments and also there are a few places in the tests with copy-pasted code that make it harder to read through the tests.

pallets/msa/src/lib.rs
shannonwells: we've talked about adding pre-checks to some of these calls before anything ever hits the message pool. I forgot what we decided about that, is this possible?

wilwade: I think I understand. You are worried that duplicate transactions flood the mempool right?

So while an invalid request would be stopped at the first honest node, a "valid" request could be duplicated and still pass this test?

aramikm: @wilwade yes

enddynayn: I think we were talking about making this an unsigned transaction. I do not think this PR is ready.

wilwade: @enddynayn I thought we had rejected the idea of an unsigned transaction due to the tradeoff in complexity in getting the signature as a parameter setup.

I am curious as to the de-duplication more generally. I thought that was just testing nonces.

pallets/msa/src/lib.rs
```diff
@@ -738,3 +740,75 @@ impl<T: Config> AccountProvider for Pallet<T> {
...
+		&self,
+		who: &Self::AccountId,
+		call: &Self::Call,
+		_info: &DispatchInfoOf<Self::Call>,
+		_len: usize,
+	) -> TransactionValidity {
+		match call.is_sub_type() {
+			Some(Call::revoke_msa_delegation_by_delegator { provider_msa_id, .. }) => {
+				let delegator_msa_id: Delegator = Pallet::<T>::ensure_valid_msa_key(&who)
+					.map_err(|_| InvalidTransaction::Payment)?
```
johnmela, line 802: Would this make more sense using the `Call` exception since payment is not relevant to this call?

pallets/msa/src/lib.rs
```diff
@@ -738,3 +740,75 @@ impl<T: Config> AccountProvider for Pallet<T> {
...
+					.msa_id
+					.into();
+				let provider_msa_id = Provider(*provider_msa_id);
+
+				Pallet::<T>::ensure_valid_delegation(provider_msa_id, delegator_msa_id)
+					.map_err(|_| InvalidTransaction::Payment)?;
```
johnmela, line 808: Same as above:
Would this make more sense using the `Call` exception since payment is not relevant to this call?

enddynayn: Do a Custom one.

saraswatpuneet: Yes we should have custom ones for such unique implementations

pallets/msa/src/lib.rs
saraswatpuneet: This looks eerily similar 
https://hackernoon.com/zero-cost-blockchain-transactions-are-a-reality-on-polkadot

But looks like SignedExtension is way to go 

Are we sure we are not setting us up for vulnerabilities

pallets/msa/src/lib.rs
johnmela: This comment was added

pallets/msa/src/tests.rs
johnmela: Comments added to this test

pallets/msa/src/lib.rs
johnmela: I added a prefix + tag unique to the origin of the revoke delegation call in the signed extension return value `ValidTransaction` of the validate function. This seems to affect the priority such that when a transaction with a tag is already queued, subsequent calls error out with a "Priority is too low" error. This might serve as a mechanism to prevent duplicate calls in the transaction queue.

aramikm (approved):
Looks good to me.

You've probably done this, but other thing to test to make sure this is behaving as expected is to test with 2 different valid accounts in the same block and both of them should go through.

pallets/msa/src/tests.rs
```diff
@@ -785,6 +785,33 @@ fn revoke_provider_throws_errors() {
 	});
 }
 
+#[test]
+pub fn revoke_provider_call_has_correct_costs() {
```
shannonwells, line 789: missed a comment

shannonwells (approved):
a comment was missed but I don't think it should hold up merging; it can be added later.

pallets/msa/src/tests.rs
johnmela: Thanks, got it

johnmela:
> Looks good to me.
> 
> You've probably done this, but other thing to test to make sure this is behaving as expected is to test with 2 different valid accounts in the same block and both of them should go through.

I missed that case in QA, will test it before it is merged

saraswatpuneet (approved):
:+1:

johnmela:
I have QA'ed the case where two separate accounts submit the revokeMsaDelegationByDelegator extrinsic in the same block. Each separate account is able to revoke within the same block.

Here is what Polkadot js UI's error banner looks like when submitting a revokeMsaDelegationByDelegator call when the same account has already submitted for the next block.
Screen Shot 2022-07-29 at 10 15 54 AM

johnmela:
Something else worth documenting is that Custom `InvalidTransaction` types don't come back on the WS connection with the name of the `enum` element or the comment describing it. Either the client will have to be cognizant of what the mapped custom transaction errors are or we have to do something else to attach custom error information on the rpc response. That is of course, assuming the client needs to know why the extrinsic submission failed
Screen Shot 2022-07-29 at 10 36 31 AM
.

PR #230, opened 2022-07-08
Declaring the Shape of Schema Models

### LINKED ISSUE #188: Support Multiple Formats for Payloads: Format Parquet
As a user of MRC, I want to be able to create a format type for a Schema of Parquet, so that I can use Parquet payloads on MRC.

## Acceptance Criteria
- [ ] When creating a new Schema, I should be required to specify an allowed Format Type of Parquet
- [ ] I should be able to locate documentation on how to properly structure the Format for a schema with Format Type of Parquet (Pending Spike: https://github.com/LibertyDSNP/frequency/issues/214)
- [ ] When getting a Schema of type Parquet from RPC, I should be able to know that it is. (ex. Type Parquet)
.
## Discussion Questions
- What should the Parquet Format be? Spike: https://github.com/LibertyDSNP/frequency/issues/214

dopatraman:
# Goal
The goal of this PR is to sketch how to model different file types (Parquet).

Closes #188 

# Discussion
The idea behind this changeset is to introduce a trait that unifies different file types. This trait, `Model` will allow the return of serialized data (the serialized file shape, which, in this case, is a `Vec<u8>`). The actual file types themselves can be implemented separately.

# Checklist
- [ ] Chain spec updated
- [ ] Design doc(s) updated
- [ ] Tests added
- [ ] Benchmarks added
- [ ] Weights updated


common/primitives/src/parquet_model.rs
```diff
@@ -0,0 +1,57 @@
...
+  _type: ParquetType,
+  compression: CompressionCodec,
+  statistics: bool,
+  bloom_filter: bool,
+  optional: bool
+}
+
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+#[derive(Clone, Encode, Decode, PartialEq, Debug, TypeInfo, Eq, MaxEncodedLen)]
+pub enum ParquetType {
```
wilwade, line 20: @dopatraman I think I like this. Let's keep going down this path. Two complications to work through on ParquetTypes around [Logical Types](https://github.com/apache/parquet-format/blob/master/LogicalTypes.md):

1. Some are easy. DSNP needs string types for example, but those fit into the enum pattern
    - We also use numerical types and those take parameters so an enum might not work. Not sure how we would want to handle those, but they are fairly simple still.
2. We don't use the Nested Types in DSNP, but I am very unsure about how we could model those
    - Likely we should not support those v1, but might be worth considering.

dopatraman: re: types that take parameters, I believe Rust enums can be modeled with params: https://doc.rust-lang.org/rust-by-example/custom_types/enum.html

re: nested types, good point. I will tinker a bit and see what we can do.

dopatraman:
@aramikm I want to make sure I understand correctly. Are you saying the payload should be hardcoded as a String (and so we should not pursue casting it as a type)?

aramikm:
> @aramikm I want to make sure I understand correctly. Are you saying the payload should be hardcoded as a String (and so we should not pursue casting it as a type)?

To enforce specific structures for avro or any other type we don't need to define it explicitly on chain. We just need to check that the schema payload syntax is compatible with our desired structure. To achieve this we can just parse input (string) to ensure it has correct syntax. I might be missing something but I don't see what are other use-cases of storing and casting it as a type if we are not going to do any type specific logic. In parquet case since we are using our custom type maybe it would be easier to validate syntax by defining the type itself but other than that I prefer to keep everything as generic as possible to ensure we are going to have flexibility to add new stuff without maybe chain upgrade requirements.

common/primitives/src/schema.rs
```diff
@@ -51,8 +51,8 @@ pub struct SchemaResponse {
 	/// The unique identifier for this Schema
 	pub schema_id: SchemaId,
 	/// The data that represents how this schema is structured
-	#[cfg_attr(feature = "std", serde(with = "as_string"))]
-	pub model: Vec<u8>,
+	#[cfg_attr(feature = "std", serde(with = "as_model"))]
+	pub model: ParquetModel,
```
aramikm, line 55: what if it's not parquet?

aramikm (commented):
Added some comments

common/primitives/src/schema.rs
dopatraman: Good point, this part isn't ready yet. Ideally we'd use a trait here, but bc of the dynamic memory allocation associated with traits, we'd have to either define an `impl` for a dynamic trait or use an enum. I'm investigating both approaches.

common/primitives/src/schema.rs
```diff
@@ -51,8 +51,8 @@ pub struct SchemaResponse {
 	/// The unique identifier for this Schema
 	pub schema_id: SchemaId,
 	/// The data that represents how this schema is structured
-	#[cfg_attr(feature = "std", serde(with = "as_string"))]
-	pub model: Vec<u8>,
+	#[cfg_attr(feature = "std", serde(with = "as_model"))]
```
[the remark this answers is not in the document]
dopatraman: I was building on the prior approach here (one would assume that `Vec<u8>`s come with serialization out of the box as well). But let me see if I can remove this.

common/primitives/src/utils.rs
```diff
@@ -38,6 +40,23 @@ pub mod as_string {
 	}
 }
 
+/// Handle serializing and deserializing from `Vec<u8>` to a UTF-8 string
+#[cfg(feature = "std")]
+pub mod as_model {
+	use super::*;
+	use serde::{ser::Error, Deserialize, Deserializer, Serialize, Serializer};
+
+	/// Serializes a `Parquet Model` into a UTF-8 string
```
wilwade, line 49: This will likely just complicate things, but is it worthwhile to serialize into SCALE for storage?

I would guess that would just be more work on chain as the upload is in JSON, but I could imagine that we deserialize the JSON and serialize into SCALE for storage. Then RPC deserializes out of SCALE.

Call -> Storage -> RPC
JSON -> SCALE -> Object in JSON response

common/primitives/Cargo.toml
```diff
@@ -14,9 +14,9 @@ targets = ['x86_64-unknown-linux-gnu']
 
 [dependencies]
 codec = {package = "parity-scale-codec", version = "3.0.0", default-features = false, features = ["derive"]}
-impl-serde = {version = "0.3.1", optional = true}
+impl-serde = {version = "0.3.1", default-features = false}
```
mehtaishita, line 17: What does this provide?

common/primitives/src/utils.rs
dopatraman: hmm, i see the upside, but im not immediately sure what this would entail. i can investigate this.

common/primitives/src/parquet.rs
```diff
@@ -0,0 +1,42 @@
...
+  /// Compression for bloom filter
+  compression: CompressionCodec,
+  /// Whether or not to use a bloom filter
+  bloom_filter: bool,
+}
+
+impl ParquetModel {
+  /// Creates instance of struct
+  pub fn new(_type: ParquetType, compression: CompressionCodec, bloom_filter: bool) -> ParquetModel {
+    ParquetModel {
```
mehtaishita, line 36: You do not need to repeat parameter name and struct field, since they are the same
https://doc.rust-lang.org/book/ch05-01-defining-structs.html#using-the-field-init-shorthand

common/primitives/src/utils.rs
wilwade: This would be an even larger switch up, but just saw this today: https://github.com/paritytech/scale-value

common/primitives/src/parquet.rs
```diff
@@ -0,0 +1,42 @@
...
+pub mod types;
+/// Importing all compression codec types
+pub mod compression_codec;
+
+use crate::parquet::types::ParquetType;
+use crate::parquet::compression_codec::CompressionCodec;
+
+/// The model for Parquet data
+#[derive(Default, Clone, Encode, Decode, PartialEq, Debug, TypeInfo, Eq, MaxEncodedLen, Serialize, Deserialize)]
+pub struct ParquetModel {
```
wilwade, line 24: Example for DSNP Broadcast
```javascript
const broadcastParquetModel =[
  {
    field: "announcementType",
    type: {
      name: "TODO",
      bit_width: 32,
      sign: true
    },
    compression: "GZIP",
    bloom_filter: false,
  },
  {
    field: "contentHash",
    type: "BYTE_ARRAY",
    compression: "TODO",
    bloom_filter: true,
  },
  {
    field: "fromId",
    type: {
      name: "TODO",
      bit_width: 64,
      sign: false
    },
    compression: "GZIP", // TODO: Pick something
    bloom_filter: true,
  },
  {
    field: "url",
    type: "String",
    compression: "GZIP", // TODO: Pick something
    bloom_filter: false,
  },
];
```

common/primitives/src/utils.rs
dopatraman: As discussed, we'll table this for now.

shannonwells (commented):
Have you had a look at [parquet-rs](https://github.com/sunchao/parquet-rs), particularly the [data types](https://github.com/sunchao/parquet-rs/blob/master/src/data_type.rs) ?  It's not an active project but maybe we could use it, or use part of it.

dopatraman:
@shannonwells I don't know if using the whole library would be practical... to my knowledge we don't need a full implementation of Parquet. That being said, some of those data types may be helpful when we tackle Nested Types and Enums. Maybe we can look into using the data types this library provides at that point?

shannonwells (approved):

common/primitives/src/lib.rs
```diff
@@ -6,7 +6,6 @@
 #![cfg_attr(not(feature = "std"), no_std)]
 // Strong Documentation Lints
 #![deny(
-	missing_docs,
```
mehtaishita, line 9: Did we all agree to remove this?

dopatraman: it was a duplicate -- the rule is still at the bottom of the list.

common/primitives/src/parquet/column_compression_codec.rs
```diff
@@ -0,0 +1,32 @@
...
+use serde::{Deserialize, Serialize};
+use sp_std::prelude::*;
+
+/// Compression codecs: https://github.com/apache/parquet-format/blob/master/LogicalTypes.md
+#[derive(
+	Clone, Encode, Decode, PartialEq, Debug, TypeInfo, Eq, MaxEncodedLen, Serialize, Deserialize,
+)]
+pub enum ColumnCompressionCodec {
+	/// The compression used in the Bloom filter
+	/// NOTE: more research required here
```
wilwade, line 12: Old docs

common/primitives/src/parquet/string.rs
```diff
@@ -0,0 +1,23 @@
...
+	Clone, Encode, Decode, PartialEq, Debug, TypeInfo, Eq, MaxEncodedLen, Serialize, Deserialize,
+)]
+pub enum ParquetStringType {
+	/// Parquet strings
+	String,
+	/// Parquet UUIDs
+	UUID,
+}
+
+impl Default for ParquetStringType {
```
wilwade, line 19: Is serde forcing us to have a default?

pallets/schemas/src/lib.rs
```diff
@@ -285,13 +286,24 @@ pub mod pallet {
...
 			Ok(())
 		}
+
+		/// Ensures that a given u8 Vector conforms to a recognized Parquet shape
+		pub fn ensure_valid_model(model_type: &ModelType, model_vec: &Vec<u8>) -> DispatchResult {
+			if model_type == &ModelType::Parquet {
+				let _p: ParquetModel =
+					serde_json::from_slice(model_vec).map_err(|_| Error::<T>::InvalidSchema)?;
+			} else {
+				unimplemented!("Avro models not implemented yet.");
```
wilwade, line 303: Avro is just `ensure_valid_json`

pallets/schemas/src/lib.rs
```diff
@@ -269,8 +270,8 @@ pub mod pallet {
 		/// Retrieve a schema by id
 		pub fn get_schema_by_id(schema_id: SchemaId) -> Option<SchemaResponse> {
 			if let Some(schema) = Self::get_schema(schema_id) {
-				// this should get a BoundedVec out
-				let model_vec = schema.model.into_inner();
+				let model_vec: Vec<u8> = schema.model.into_inner();
+				Self::ensure_valid_model(&schema.model_type, &model_vec).ok()?;
```
wilwade, line 274: Not sure why we are doing `ensure_valid_model` here. Only valid models can be stored.

pallets/schemas/src/tests.rs
```diff
@@ -156,6 +163,7 @@ fn register_schema_id_deposits_events_and_increments_schema_id() {
 }
 
 #[test]
+#[ignore]
```
wilwade, line 166: Old Ignore

pallets/schemas/src/tests.rs
```diff
@@ -194,7 +202,71 @@ fn get_non_existing_schema_by_id_should_return_none() {
...
+			serde_json::to_string(&p).unwrap(),
+			r#"{"name":"Foo","type":"Boolean","compression":"Uncompressed","bloom_filter":true}"#
+		);
+	})
+}
+
+#[test]
+fn validate_parquet_model() {
+	new_test_ext().execute_with(|| {
+		let test_str_raw = r#"[{"name": "Foo", "type": "Boolean", "compression": "Uncompressed", "bloom_filter": true}]"#;
```
wilwade, line 239: Is it worth doing a test for the DSNP Broadcast model?

wilwade (dismissed):

common/primitives/src/parquet/column_compression_codec.rs
```diff
@@ -0,0 +1,32 @@
+use codec::{Decode, Encode, MaxEncodedLen};
+use scale_info::TypeInfo;
+use serde::{Deserialize, Serialize};
+use sp_std::prelude::*;
+
+/// Compression codecs: https://github.com/apache/parquet-format/blob/master/LogicalTypes.md
+#[derive(
+	Clone, Encode, Decode, PartialEq, Debug, TypeInfo, Eq, MaxEncodedLen, Serialize, Deserialize,
+)]
+pub enum ColumnCompressionCodec {
```
mehtaishita, line 10: These enum is case sensitive. Add an attribute like: 
[#[serde(rename_all = "lowercase")]](https://serde.rs/container-attrs.html#rename_all) (or any other option)

pallets/schemas/src/lib.rs
dopatraman: In theory, yes, but we don't _know_ that... especially since the type of the schema is a vector.

common/primitives/src/parquet/temporal.rs
```diff
@@ -0,0 +1,82 @@
...
+
+/// Parquet temporal types: https://github.com/apache/parquet-format/blob/master/LogicalTypes.md
+#[derive(
+	Clone, Encode, Decode, PartialEq, Debug, TypeInfo, Eq, MaxEncodedLen, Serialize, Deserialize,
+)]
+pub enum ParquetTemporalType {
+	/// Parquet dates
+	Date,
+	/// Parquet intervals
+	Interval,
```
mehtaishita, line 14: for documentation here maybe include a bit more information from their readMe
annotate a fixed_len_byte_array of length 12.

pallets/schemas/src/lib.rs
wilwade: True, but this is an on-chain call right?. Everytime we get a schema, I don't think we want to be testing that it is a valid schema.

common/primitives/src/parquet/temporal.rs
```diff
@@ -0,0 +1,82 @@
...
+	Clone, Encode, Decode, PartialEq, Debug, TypeInfo, Eq, MaxEncodedLen, Serialize, Deserialize,
+)]
+enum ParquetTimeUnit {
+	/// milliseconds
+	MILLIS,
+	/// microseconds
+	MICROS,
+	/// nanoseconds
+	NANOS,
+}
```
mehtaishita, line 76: could also include quick details like millis is int32 and the other two are int64
I know this comes from parquet type documentation, but could be worthwhile to include some details?

mehtaishita (changes requested):
I added a few comments, and also there's a change in the register schema method that was merged incorrectly

common/primitives/src/parquet/column.rs
```diff
@@ -0,0 +1,32 @@
...
+	_type: ParquetType,
+	/// Compression for column
+	compression: ColumnCompressionCodec,
+	/// Whether or not to use a bloom filter
+	bloom_filter: bool,
+}
+
+impl ParquetColumn {
+	/// Creates instance of struct
+	pub fn new(
```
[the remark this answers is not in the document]
dopatraman: If we need to instantiate this struct outside of this crate (in our case, our pallet tests) we need to either have a constructor or declare each field public (the latter is messy).

pallets/schemas/src/lib.rs
```diff
@@ -285,13 +286,24 @@ pub mod pallet {
...
 		}
+
+		/// Ensures that a given u8 Vector conforms to a recognized Parquet shape
+		pub fn ensure_valid_model(model_type: &ModelType, model_vec: &Vec<u8>) -> DispatchResult {
+			if model_type == &ModelType::Parquet {
+				let _p: ParquetModel =
+					serde_json::from_slice(model_vec).map_err(|_| Error::<T>::InvalidSchema)?;
+			} else {
+				unimplemented!("Avro models not implemented yet.");
+			}
```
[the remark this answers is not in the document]
dopatraman: to my knowledge, we can't coerce `_p` to a type within a match clause (matches have to be expressions).

common/primitives/src/parquet/temporal.rs
dopatraman: I don't know if that particular type info is super useful for this enum (the person using it does not necessarily need to know the integer type for their schema)... but you're right maybe there is more information I can add.

common/primitives/src/parquet/column_compression_codec.rs
dopatraman: i think if anything we should either keep the casing we have or make everything uppercase.

aramikm (approved):
Overall looks good, I added some nits and comments.

saraswatpuneet (approved):

common/primitives/src/parquet/column.rs
```diff
@@ -0,0 +1,32 @@
...
+		name: String,
+		_type: ParquetType,
```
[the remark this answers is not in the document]
saraswatpuneet: Yes this should be commented if and why underscore is put in place
Generally to suppress some warning?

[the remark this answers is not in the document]
dopatraman: > Generally to suppress some warning?

Yes, `type` is a reserved word.

I like `column_type`. I will change it to that.

PR #232, opened 2022-07-12
Feat/on chain json validation rebase/111

### LINKED ISSUE #111: On-chain Schema JSON Validation Prior to Registration

mehtaishita:
# Goal
The goal of this PR is to add a rudimentary Json validator and replace Serde

Closes #111 

# Discussion
Serde-Json-Core uses a type-specific Json deserializer. Since we do not want to force consumers to adhere to a specific Json structure, we would need to implement a custom deserializer that validates left-hand and right-hand sides independently (essentially deconstructs the Json of type "any"). Such a deserializer would be too inefficient to validate any Json on-chain. 

Instead, we are adding a rudimentary validator that enforces that the Json is an object (surrounded by "{ }"). Register schema RPC will still accept schemas with an invalid Json structure, as long it is a Json object. 

# Checklist
- [x] Chain spec updated
- [x] Design doc(s) updated
- [y] Tests added
- [ ] Benchmarks added
- [ ] Weights updated


mehtaishita:
Look at https://github.com/LibertyDSNP/frequency/pull/235 for an updated PR

PR #235, opened 2022-07-13
Feat/on chain json validation review/111

### LINKED ISSUE #111: On-chain Schema JSON Validation Prior to Registration

mehtaishita:
# Goal
The goal of this PR is to fix the use of serde-json library with no-std support for on-chain JSON validation

Closes #111 

# Discussion
Serde-Json uses 'Value' enum to deserialize Json to any type. Serde-Json provides no-std support if
- default features are turned off
- only imports that do not rely on std libraries are used. (For example, the compiler will complain if we tried to use serde-Json::Error, which relies on std::Error)

# Checklist
- [x] Tests added

### DIFF
--- a/pallets/schemas/src/lib.rs
+++ b/pallets/schemas/src/lib.rs
@@ -213,7 +213,6 @@ pub mod pallet {
 		) -> DispatchResult {
 			let sender = ensure_signed(origin)?;
 
-			Self::ensure_valid_schema(&model)?;
 			ensure!(
 				model.len() > T::MinSchemaModelSizeBytes::get() as usize,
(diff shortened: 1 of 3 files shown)


saraswatpuneet (approved):
Awesome! Can we add one test for Deserialization error

wilwade (approved):
Correctly invalidated and validated when I tested running it locally from Polkadot apps!

mehtaishita:
@saraswatpuneet 

I added a few more deserializer error test cases. Serde-Json has a much longer list of error codes though: https://docs.serde.rs/src/serde_json/error.rs.html#51-76

PR #247, opened 2022-07-22
Validate payload location

### LINKED ISSUE #189: Support Different Payload Locations: On-Chain
As a schema creator, I want to be able to set the location type for where messages for a schema are stored, so that in the future I can create a schema with off-chain support.

## Acceptance Criteria
- [x] When creating a new Schema, I should be required to specify an allowed Payload Location of "on-chain"
- [x] When getting a Schema from RPC, I should be able to know what the Payload Location is
- [x] When creating a new message, I should be required to use the proper method given the schema's payload location

dopatraman:
# Goal
The goal of this PR is to create an extrinsic dedicated to offchain storage so users do not store large payloads onchain.

Closes #189

# Discussion
This PR introduces a new extrinsic, `add_offchain` that includes an IPFS content address and payload length as parameters. The existence of this extrinsic makes it less likely for a user to mistaken store large payloads onchain.

# Checklist
- [x] Chain spec updated
- [ ] Tests added
- [x] Benchmarks added
- [x] Weights updated


pallets/messages/src/lib.rs
```diff
@@ -165,6 +167,50 @@ pub mod pallet {
...
+		/// * `origin` - A signed transaction origin from the provider
+		/// * `on_behalf_of` - Optional. The msa id of delegate.
+		/// * `schema_id` - Registered schema id for current message.
+		/// * `cid` - The content address for an IPFS payload
+		/// * `payload_length` - The size of the payload
+		/// * Returns
+		/// * [DispatchResultWithPostInfo](https://paritytech.github.io/substrate/master/frame_support/dispatch/type.DispatchResultWithPostInfo.html)
+		#[pallet::weight(T::WeightInfo::add_offchain(cid.get().len() as u32, 1_000))]
+		pub fn add_offchain(
+			origin: OriginFor<T>,
```
saraswatpuneet, line 181: Can we give it more descriptive name given it will be exposed to outside consumers

Like add_messages_offchain or add_messages_ipfs

Offchain could also mean offchain db so might create confusion

pallets/messages/src/lib.rs
```diff
@@ -165,6 +167,50 @@ pub mod pallet {
...
+			on_behalf_of: Option<MessageSourceId>,
+			schema_id: SchemaId,
+			cid: CIDv2,
+			payload_length: u32,
+		) -> DispatchResultWithPostInfo {
+			let provider_key = ensure_signed(origin)?;
+
+			// TODO: Is there a size limit on offchain payloads?
```
wilwade, line 189: Perhaps 100 MiB? @shannonwells thoughts?

pallets/messages/src/lib.rs
```diff
@@ -236,6 +291,17 @@ impl<T: Config> Pallet<T> {
 		})
 	}
 
+	/// Converts a Payload to a message string
+	pub fn payload_to_message(payload: Payload) -> BoundedVec<u8, T::MaxMessagePayloadSizeBytes> {
+		match payload {
+			Payload::Onchain(payload_vec) => payload_vec.try_into().unwrap(),
+			// TODO: What is the right way to format an offchain payload? May need to
+			// table until #190 is picked up.
```
wilwade, line 299: This is just internal to the chain right? So I don't think it matters much outside of performance. Perhaps a SCALE encoded struct?

pallets/messages/src/lib.rs
dopatraman: Good point.

pallets/messages/src/types.rs
```diff
@@ -40,3 +40,43 @@ where
...
+
+impl OffchainPayload {
+	/// Constructor
+	pub fn new(cid: CIDv2, payload_length: u32) -> OffchainPayload {
+		OffchainPayload { cid, payload_length }
+	}
+}
+
+/// Sum type for payloads
+pub enum Payload {
```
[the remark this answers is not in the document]
dopatraman: I think that _might_ end up making things confusing. The `PayloadLocation` is used as a an attribute of the schema, whereas the payload is the data described by the schema.

In theory we could parametrize the `PayloadLocation` enum, but we would still need to store the `CID` and payload length in another struct, since they could differ within a single schema definition.

pallets/messages/src/lib.rs
```diff
@@ -190,10 +236,19 @@ pub mod pallet {
...
+				Error::<T>::InvalidPayloadLocation
+			);
 
 			let message_source_id = Self::find_msa_id(&provider_key, on_behalf_of)?;
 
-			let message = Self::add_message(provider_key, message_source_id, payload, schema_id)?;
+			let message = Self::add_message(
+				provider_key,
+				message_source_id,
+				Payload::Onchain(payload),
```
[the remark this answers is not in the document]
wilwade: I was thinking the extrinsic name (currently `add`). The external calls should have a pattern.

pallets/messages/src/benchmarking.rs
```diff
@@ -35,6 +35,20 @@ benchmarks! {
...
+		let caller: T::AccountId = whitelisted_caller();
+		let input = vec![1; n as usize];
+		let cid = CIDv2::new(input);
+		let payload_length = 1_000;
+
+		// for j in 0 .. m {
+		// 	let sid = j % SCHEMAS;
+		// 	assert_ok!(add_message::<T>(sid.try_into().unwrap()));
+		// }
+	}: _ (RawOrigin::Signed(caller), None, 65535, cid, payload_length)
```
aramikm, line 50: So when you are using `..` in benchmarks that means it should be used in some shape or form as an input. Instead of payload_length you probably should pass n inside here.

random numbers like `65535` are confusing

pallets/messages/src/benchmarking.rs
```diff
@@ -35,6 +35,20 @@ benchmarks! {
...
 
+	add_ipfs_message {
+		let n in 0 .. T::MaxMessagePayloadSizeBytes::get() - 1;
+		let m in 1 .. MESSAGES;
+		let caller: T::AccountId = whitelisted_caller();
+		let input = vec![1; n as usize];
+		let cid = CIDv2::new(input);
+		let payload_length = 1_000;
+
+		// for j in 0 .. m {
```
aramikm, line 46: I think you should uncomment and use m to register messages to have a more accurate benchmark.

pallets/messages/src/lib.rs
```diff
@@ -165,6 +167,50 @@ pub mod pallet {
...
+		/// Adds a message for an offchain resource
+		/// # Arguments
+		/// * `origin` - A signed transaction origin from the provider
+		/// * `on_behalf_of` - Optional. The msa id of delegate.
+		/// * `schema_id` - Registered schema id for current message.
+		/// * `cid` - The content address for an IPFS payload
+		/// * `payload_length` - The size of the payload
+		/// * Returns
+		/// * [DispatchResultWithPostInfo](https://paritytech.github.io/substrate/master/frame_support/dispatch/type.DispatchResultWithPostInfo.html)
+		#[pallet::weight(T::WeightInfo::add_ipfs_message(cid.get().len() as u32, 1_000))]
```
aramikm, line 179: is cid size dynamic? I don't think it is. If not it should not be used as in input here. We usually use it like this if the input is an array with dynamic size.

pallets/messages/src/lib.rs
```diff
@@ -165,6 +167,50 @@ pub mod pallet {
...
+			let payload = IPFSPayload::new(cid, payload_length);
+			let message = Self::add_message(
+				provider_key,
+				message_source_id,
+				Payload::IPFS(payload),
+				schema_id,
+			)?;
+
+			Ok(Some(T::WeightInfo::add_ipfs_message(
+				message.payload.len() as u32,
```
aramikm, line 209: shouldn't this be `payload_length` ?

pallets/messages/src/types.rs
```diff
@@ -40,3 +40,43 @@ where
 		}
 	}
 }
+
+/// Newtype for CIDv2 addresses. We can change the inner type as we understand
+/// more about how to structure IPFS addressable content.
+#[derive(Clone, Encode, Decode, PartialEq, Debug, TypeInfo, Eq)]
+pub struct CIDv2(Vec<u8>);
```
aramikm, line 47: I'm not totally sure this should be Vec<u8> ?
I thought this is just sha256 hash with I think cos a constant size. If so it might make sense to use an array with constant size.

aramikm (changes requested):

pallets/messages/src/types.rs
aramikm: It seems like CID size is depend on the Hash function being used, so if we don't specify the hash type then it might makes sense to have vec<u8> to have more flexibility for different hash functions but we loose the checks to make sure if a provided CID is a valid one by checking the size.

pallets/messages/src/lib.rs
```diff
@@ -212,7 +271,7 @@ impl<T: Config> Pallet<T> {
 	pub fn add_message(
 		provider_key: T::AccountId,
 		message_source_id: MessageSourceId,
-		payload: Vec<u8>,
+		payload: Payload,
```
enddynayn, line 274: Looking good.

Question: do we need to pass in an enum here? I don't think we are doing anything with knowing if the payload is offchain or onchain.

dopatraman: We serialize the payload into the message differently based on what it is. Either way, I think the type of the `payload` param would have to be something other than a vector. Thoughts?

pallets/messages/src/lib.rs
```diff
@@ -236,6 +295,14 @@ impl<T: Config> Pallet<T> {
 		})
 	}
 
+	/// Converts a Payload to a message string
+	pub fn payload_to_message(payload: Payload) -> BoundedVec<u8, T::MaxMessagePayloadSizeBytes> {
```
enddynayn, line 299: if you decide to keep this, maybe define this method on the enum Payload.
```
impl Payload {
}
```

pallets/messages/src/lib.rs
```diff
@@ -165,6 +167,50 @@ pub mod pallet {
...
+		/// * `cid` - The content address for an IPFS payload
+		/// * `payload_length` - The size of the payload
+		/// * Returns
+		/// * [DispatchResultWithPostInfo](https://paritytech.github.io/substrate/master/frame_support/dispatch/type.DispatchResultWithPostInfo.html)
+		#[pallet::weight(T::WeightInfo::add_ipfs_message(cid.get().len() as u32, 1_000))]
+		pub fn add_ipfs_message(
+			origin: OriginFor<T>,
+			on_behalf_of: Option<MessageSourceId>,
+			schema_id: SchemaId,
+			cid: CIDv2,
```
enddynayn, line 184: you can make this:
```
pub fn add_ipfs_message(
 ..,
  payload: IPFSPayload
)
```

pallets/messages/src/lib.rs
enddynayn: This is already passed in the Config. Are we trying to reduce the size?

pallets/messages/src/types.rs
shannonwells: It looks like [RFC2392](https://datatracker.ietf.org/doc/html/rfc2392) is the CIDv2 specification. It's from 1998 and I don't think it ever was accepted.  Let's consider it YAGNI and kill this code unless there is a CIDv2 AND we decide to support it.  Someone correct me if I'm wrong (in which case my search skills are terrible).  But definitely IPFS has nothing about it [in their docs](https://docs.ipfs.io/concepts/content-addressing/).

pallets/messages/src/lib.rs
```diff
@@ -165,6 +167,50 @@ pub mod pallet {
...
+			payload_length: u32,
+		) -> DispatchResultWithPostInfo {
+			let provider_key = ensure_signed(origin)?;
+
+			// TODO: Is there a size limit on offchain payloads?
+
```
enddynayn, line 190: Don't forget to check that the signing keys have not been revoked. I think we will later move that logic out elsewhere but for now lets add it.

scripts/generate_weights.sh
```diff
@@ -0,0 +1,10 @@
+#! /bin/sh
```
shannonwells, line 1: This generates weights only for the messages pallet; I don't think you meant to check this in.  Either way it should be removed.

pallets/messages/src/benchmarking.rs
shannonwells: Also why does this use CIDv2? That's not real.

shannonwells (changes requested):
Need to removed the generate_weights.sh script, remove the CIDv2 code, and the benchmark needs to preload some messages in `add_ipfs_message`, using `m`

pallets/messages/src/lib.rs
aramikm: I think the important thing in weight generation is that how much dynamic is the size of this cid? If its a difference between 32 bytes and 64 bytes then it's still close enough that I don't think will contribute significantly in weights. So the question might be what is the absolute min and max size for a CID.

shannonwells (approved):
All my feedback was addressed, good to go from me.

pallets/messages/src/lib.rs
```diff
@@ -174,8 +216,8 @@ pub mod pallet {
 		/// # Returns
 		/// * [DispatchResultWithPostInfo](https://paritytech.github.io/substrate/master/frame_support/dispatch/type.DispatchResultWithPostInfo.html) The return type of a Dispatchable in frame.
 		/// When returned explicitly from a dispatchable function it allows overriding the default PostDispatchInfo returned from a dispatch.
-		#[pallet::weight(T::WeightInfo::add(payload.len() as u32, 1_000))]
-		pub fn add(
+		#[pallet::weight(T::WeightInfo::add_onchain_message(payload.len() as u32, 1_000))]
+		pub fn add_onchain_message(
```
aramikm, line 220: Just wanted to remind that this rename might break some client side apps (not sure if we have for frequency yet)

pallets/messages/src/types.rs
```diff
@@ -40,3 +40,36 @@ where
...
+
+	/// Exposes the inner vector
+	pub fn get(&self) -> &Vec<u8> {
+		&self.0
+	}
+}
+
+/// IPFS payloads are defined by two members, a CID address and a payload length
+#[derive(Clone, Encode, Decode, PartialEq, Debug, TypeInfo, Eq)]
+pub struct IPFSPayload {
```
aramikm, line 63: you might need to define typescript definitions for these types also unless we are generating them from chain metadata.

pallets/schemas/src/lib.rs
```diff
@@ -316,11 +316,16 @@ impl<T: Config> SchemaProvider<SchemaId> for Pallet<T> {
 	fn get_schema_by_id(schema_id: SchemaId) -> Option<SchemaResponse> {
 		// To account for db read
 		Self::get_schema_by_id(schema_id);
+
+		// This is janky, but it allows us to benchmark both message extrinsics from
+		// this pallet.
+		let location: PayloadLocation =
```
aramikm, line 322: is it possible to define a method on `PayloadLocation` with `#[cfg(feature = "runtime-benchmarks")]` that gives you location by schema_id and you move this logic inside that.

pallets/messages/src/benchmarking.rs
```diff
@@ -22,26 +22,52 @@ fn add_message<T: Config>(schema_id: SchemaId) -> DispatchResultWithPostInfo {
...
+		let m in 1 .. MESSAGES;
+		let caller: T::AccountId = whitelisted_caller();
+		let input = vec![1; n as usize];
+		let cid = CID::new(input);
+		let payload_length = 1_000;
+		let payload: IPFSPayload = IPFSPayload::new(cid, payload_length);
+
+		for j in 0 .. m {
+			let sid = (j % SCHEMAS) + 51;
+			assert_ok!(offchain_message::<T>(sid.try_into().unwrap()));
```
aramikm, line 60: nit: we ideally want to fill temporary list with a  mix of both kinds of messages to have more accurate weights

aramikm (commented):
Added some suggestions and comments

pallets/messages/src/lib.rs
dopatraman: good point. should we leave the old method defined and just have it point to this one?

pallets/schemas/src/lib.rs
dopatraman: we'd have to define this method on the `SchemaProvider` class. If we're ok with that (as in, defining a method purely for the purpose of benchmarking) then we can do it.

pallets/messages/src/lib.rs
```diff
@@ -190,12 +235,25 @@ pub mod pallet {
...
 
 			let message_source_id = Self::find_msa_id(&provider_key, on_behalf_of)?;
 
-			let message = Self::add_message(provider_key, message_source_id, payload, schema_id)?;
-
-			Ok(Some(T::WeightInfo::add(message.payload.len() as u32, message.index as u32)).into())
+			let message = Self::add_message(
+				provider_key,
+				message_source_id,
+				payload.try_into().unwrap(),
```
enddynayn, line 248: ⬆️

pallets/messages/src/lib.rs
```diff
@@ -190,12 +235,25 @@ pub mod pallet {
 
 			let schema = T::SchemaProvider::get_schema_by_id(schema_id);
```
enddynayn, line 236: I think we should push the bounded vector up here 
```
let bounded_payload: BoundedVec<u8, T::MaxMessagePayloadSizeBytes> =  payload.try_into()..map_err(|_| Error::<T>:: ExceedsMaxMessagePayloadSizeBytes)?

```

pallets/messages/src/lib.rs
```diff
@@ -212,7 +270,7 @@ impl<T: Config> Pallet<T> {
 	pub fn add_message(
 		provider_key: T::AccountId,
 		message_source_id: MessageSourceId,
-		payload: Vec<u8>,
+		payload: BoundedVec<u8, T::MaxMessagePayloadSizeBytes>,
```
enddynayn, line 273: 👍

pallets/messages/src/lib.rs
```diff
@@ -222,7 +280,7 @@ impl<T: Config> Pallet<T> {
 				.map_err(|_| Error::<T>::TypeConversionOverflow)?;
 
 			let msg = Message {
-				payload: payload.try_into().unwrap(), // size is checked on top of extrinsic
```
enddynayn, line 225: 👍

pallets/messages/src/lib.rs
```diff
@@ -165,6 +167,49 @@ pub mod pallet {
...
+			ensure!(
+				schema.unwrap().payload_location == PayloadLocation::IPFS,
+				Error::<T>::InvalidPayloadLocation
+			);
+
+			let message_source_id = Self::find_msa_id(&provider_key, on_behalf_of)?;
+			let message = Self::add_message(
+				provider_key,
+				message_source_id,
+				payload.cid.get().clone().try_into().unwrap(),
```
enddynayn, line 203: I think we want to handle the error here. Perhaps doing something similar from `add_on_chain_message`

enddynayn:
looking just a couple comments.

pallets/messages/src/lib.rs
dopatraman: what error? payload size or possible null pointer from the unwrap?

enddynayn: For the cid exceeding the size of the bounded vector.

dopatraman: Done.

enddynayn (approved):

aramikm (approved):

PR #252, opened 2022-07-27
Retire Msa through revocation

### LINKED ISSUE #44: As a user, I should be able to retire my MSA

# Description

## Notes

# Acceptance Criteria
- [ ] Benchmarks
- [x] Unit tests
  - [x] FAIL: Add new key
  - [x] FAIL: Add Message

## State Maps for Deletion

- Pays::No is in #340 

## Arch Notes

## Other Notes

mehtaishita:
# Goal
The goal of this PR is to retire an Msa for free by revoking all keys and delegation relationships associated with it.

Closes #44 

# Discussion

# Checklist
- [ ] Chain spec updated
- [ ] Design doc(s) updated
- [ ] Tests added
- [ ] Benchmarks added
- [ ] Weights updated


pallets/msa/src/lib.rs
```diff
@@ -645,6 +672,37 @@ impl<T: Config> Pallet<T> {
...
+	/// # Errors
+	/// * [`Error::<T>::_`] - MSA is already retired
+	pub fn retire_msa(
+		msa_id: MessageSourceId
+	) -> DispatchResult {
+
+		ProviderInfoOf::<T>::iter_key_prefix(msa_id.into())
+			.for_each(|delegator: Delegator| -> Result<(), Error<T>> {
+				Self::revoke_provider(msa_id.into(), delegator.into()).map_err(|_| Error::<T>::MsaRevoked)
+			});
```
mehtaishita, line 692: There needs to be a different error handling here, because right now the operation fails at the first error, instead of continuing the loop. I need to handle the closure of 'for each' to skip forward.
- Do we need to throw an error here in case there are any errors with revoking delegations at all? or can those errors be swallowed?

pallets/msa/src/tests.rs
```diff
@@ -233,6 +235,40 @@ fn it_revokes_msa_key_successfully() {
...
+		assert_ok!(Msa::retire_my_msa(test_origin_signed(1)));
+
+		for i in 1..4 {
+			assert_eq!(
+				Msa::get_provider_info_of(Provider(i), Delegator(1)).unwrap(),
+				ProviderInfo { expired: 1, permission: 0 },
+			);
+		}
+
+	})
```
mehtaishita, line 269: I added the happy path tests, but I would like to figure out the error handling portion before pushing the rest of the tests

pallets/msa/src/lib.rs
```diff
@@ -459,6 +466,26 @@ pub mod pallet {
...
+		/// ### Errors
+		/// - Returns ['NotMsaOwner'](Error::NotMsaOwner)
+		#[pallet::weight(T::WeightInfo::retire_my_msa())]
+		pub fn retire_my_msa(origin: OriginFor<T>) -> DispatchResult {
+
+			// fetch msa id from origin
+			let account_id = ensure_signed(origin)?;
+			let msa_id = Self::ensure_valid_msa_key(&account_id)?.msa_id.into();
+
+			// call self::retire_msa()
```
aramikm, line 482: I don't think you need these comments since the functionality and names are self explanatory

pallets/msa/src/weights.rs
```diff
@@ -179,4 +186,10 @@ impl WeightInfo for () {
 			.saturating_add(RocksDbWeight::get().reads(2 as Weight))
 			.saturating_add(RocksDbWeight::get().writes(1 as Weight))
 	}
+	//
+	fn retire_my_msa() -> Weight {
```
aramikm, line 190: these weights should be generated

pallets/msa/src/lib.rs
aramikm: Why you are not using a regular loop instead of for_each?

aramikm (changes requested):
This problem is more complicated than it looks. You can not iterate over all keys and delegation without having a limit on the number of them. What if the number was more enough to fit in the block?

Weight calculations are also more tricky. Here is an example of a such an extrinsic https://github.com/paritytech/polkadot/blob/master/runtime/common/src/crowdloan/mod.rs#L506-L547

One other way to make this simpler is to have a vec and add msa inside the vec in extrinsic call and do the actual removal of keys `on_initialize` and keep that msa inside that vec as long as there are existing keys or delegates to revoke

mehtaishita:
Pausing work on this to clarify what should happen to revoked/retired keys, especially since it can be a unbounded transaction.

Pushed all current work. There is a failing test 'it_retires_msa_despite_key_revoked_error". Again, the validity of this test will be determined after discussion with @wilwade next week

PR #298, opened 2022-08-03
Offchain payload length validation

### LINKED ISSUE #190: Support Different Payload Locations: IPFS
#189 is finished and no longer blocking.

Related to: #163 , Also blocked by "Research and Test IPFS Latency"

As a user of Frequency, I want to be able to set a schema to support payloads on IPFS, so that I can store payloads there.

## Acceptance Criteria
- [x] When creating a new Schema, I should be allowed to specify a Payload Location of "IPFS"
- [x] When creating a new message for an IPFS payload location schema, I should be required to use the IPFS supported add message method
- [ ] When creating a new message for an IPFS payload location schema, I should be required to properly format the CIDv2
- [x] When creating a new message for an IPFS payload location schema, I should be required to submit a payload length in bytes so that in the future I can be required to pay capacity for off-chain bytes

dopatraman:
# Goal
The goal of this PR is to add onto #189 for IPFS payload validation.

Closes #190.

# Discussion
This PR validates payload length according to limits set in the DSNP [spec](https://spec.dsnp.org/DSNP/BatchPublications.html#file-requirements) for batch file size limits. 

# Checklist
- [x] Tests added
- [x] Benchmarks added
- [x] Weights updated


pallets/messages/src/lib.rs
```diff
@@ -180,20 +184,25 @@ pub mod pallet {
...
 		/// * [DispatchResultWithPostInfo](https://paritytech.github.io/substrate/master/frame_support/dispatch/type.DispatchResultWithPostInfo.html)
-		#[pallet::weight(T::WeightInfo::add_ipfs_message(payload.cid.get().len() as u32, 1_000))]
+		#[pallet::weight(T::WeightInfo::add_ipfs_message(cid.len() as u32, 1_000))]
 		pub fn add_ipfs_message(
 			origin: OriginFor<T>,
 			on_behalf_of: Option<MessageSourceId>,
 			schema_id: SchemaId,
-			payload: IPFSPayload,
+			cid: Vec<u8>,
+			payload_length: u32,
```
dopatraman, line 193: As per @enddynayn 's suggestion, I don't think we need a dedicated struct for the payload (at least not yet).

pallets/messages/src/types.rs
```diff
@@ -43,33 +43,12 @@ where
...
-}
-
-impl IPFSPayload {
-	/// Constructor
-	pub fn new(cid: CID, payload_length: u32) -> IPFSPayload {
-		IPFSPayload { cid, payload_length }
+impl FrequencyCid {
+	/// Checks if a given string is a valid CID
+	pub fn is_valid(&self) -> bool {
+		unimplemented!("No way to check CID validity yet")
```
dopatraman, line 52: The `rust-cid` crate requires `std`, which we can't work with in the pallet code. Working with the crate has been pretty time intensive -- if we need CID validation it might be worth creating a dedicated story for it.

runtime/frequency/src/lib.rs
```diff
@@ -521,6 +521,7 @@ impl pallet_collator_selection::Config for Runtime {
 parameter_types! {
 	pub const MaxMessagesPerBlock: u32 = 7000;
 	pub const MaxMessagePayloadSizeBytes: u32 = 1024 * 50; // 50K
+	pub const MaxDsnpOffchainPayloadSizeBytes: u32 = 1024 * 128;
```
dopatraman, line 524: As per [DSNP spec](https://spec.dsnp.org/DSNP/BatchPublications.html#file-requirements).

pallets/messages/src/lib.rs
```diff
@@ -89,6 +89,10 @@ pub mod pallet {
 		/// The maximum size of a message payload bytes.
 		#[pallet::constant]
 		type MaxMessagePayloadSizeBytes: Get<u32> + Clone;
+
+		/// The maximum size for an offchain payload.
+		#[pallet::constant]
+		type MaxDsnpOffchainPayloadSizeBytes: Get<u32>;
```
enddynayn, line 95: you can drop Dsnp from here.

pallets/messages/src/lib.rs
```diff
@@ -180,20 +184,25 @@ pub mod pallet {
...
 		/// * Returns
 		/// * [DispatchResultWithPostInfo](https://paritytech.github.io/substrate/master/frame_support/dispatch/type.DispatchResultWithPostInfo.html)
-		#[pallet::weight(T::WeightInfo::add_ipfs_message(payload.cid.get().len() as u32, 1_000))]
+		#[pallet::weight(T::WeightInfo::add_ipfs_message(cid.len() as u32, 1_000))]
 		pub fn add_ipfs_message(
 			origin: OriginFor<T>,
 			on_behalf_of: Option<MessageSourceId>,
 			schema_id: SchemaId,
-			payload: IPFSPayload,
+			cid: Vec<u8>,
```
enddynayn, line 192: Nice!

pallets/messages/src/types.rs
enddynayn: I like this Idea. You can remove this code and add a ticket for it.

runtime/frequency/src/lib.rs
enddynayn: Drop Dsnp from it.

enddynayn (approved):
Looks great! Minor comments.

saraswatpuneet (approved):

runtime/frequency/src/lib.rs
```diff
@@ -521,6 +521,7 @@ impl pallet_collator_selection::Config for Runtime {
 parameter_types! {
 	pub const MaxMessagesPerBlock: u32 = 7000;
 	pub const MaxMessagePayloadSizeBytes: u32 = 1024 * 50; // 50K
+	pub const MaxOffchainPayloadSizeBytes: u32 = 1024 * 128;
```
aramikm, line 524: I'm not totally sure if this is correct. DSNP spec has "Batch files MUST have NO MORE THAN 128*1024 rows." it doesn't say anything about row size. So We are talking about max payload size I'm not sure this is correct

dopatraman: Good point. Maybe we dont need to check a max offchain payload size right now then?

PR #299, opened 2022-08-03
Designdoc: Provider Registration Updates

### LINKED ISSUE #290: Update Provider Registration Design Doc
Various Updates are needed to the Provider Design Doc to align with current thinking.

https://github.com/LibertyDSNP/frequency/blob/main/designdocs/provider_registration.md

## Known Updates
- [x] Registration Fee is Burned instead of staked
- [x] Registration Fee amount should be set by governance
- [x] Only Providers can be delegated to
- [x] Only Providers can have Capacity (This might be more part of the capacity design doc, but should be mentioned)
- [x] Providers v1 does **not** allow the Provider MSA to be retired. Move references to being able to retire a Provider MSA to the alternatives. Remove `deregister_provider`
- [x] User **does not** "stakes balance to become a provider" Instead, the user merely pays the registration fee.
- [x] It should be mentioned in the alternatives that instead of or in addition to a fee payment, that the provider could be required to post a bond.

**Note:** Also check for other updates that are needed.

mehtaishita:
# Goal
The goal of this PR is update Provider Registration design doc to align with current thinking.

Closes #290

# Discussion
* Will "Registration Fee" be credited to the Chain Treasury?
* If only Providers can have Capacity, how does a User register to be a Provider - since they need enough Capacity to create said transaction?
* How detailed does the Glossary section have to be? Will we create backlinks for all basic terms such as 'Provider' or 'Capacity'?

# Checklist
- [ ] ~~Chain spec updated~~
- [ ] ~~Updated the js/api-augment code if a custom RPC added/changed~~
- [x] Design doc(s) updated
- [ ] ~~Tests added~~
- [ ] ~~Benchmarks added~~
- [ ] ~~Weights updated~~


mehtaishita:
Once there is a shared understanding of the Design Doc, I can work together to create related stories

designdocs/provider_registration.md
```diff
@@ -139,18 +116,23 @@ providers ("mom and pop" local businesses and/or "freelancers").
 So to allow inclusion for all actors, it may be that the best way of verifying
 legitimateness is through a fee that is high enough to discourage malicious actors.
 
+### Provider Commitment
+Users can only delegate to Providers. Only Providers can have Capacity and can therefore generate capacity transactions.
```
mehtaishita, line 120: Do we need more descriptive information here?

designdocs/provider_registration.md
```diff
@@ -24,8 +24,6 @@ Among those questions are:
 1. How does a user become a provider?
 1. What are the attributes of a provider?
 1. How will providers be identified to other network actors?
-1. How does one stop being a provider?
-1. Can provider status be revoked from a provider?
```
shannonwells, line 28: Are we not going to answer these questions?

mehtaishita: We will, but can't at the moment. Should I keep the questions in, and leave an asterisk - that will be answered in future versions?

shannonwells: Yes, especially since we're still talking about it.  I'd rather have it in there and "TBD" or similar than to not have the question there.

mehtaishita: I added the questions back, and explained in 'Provider De-registration' as a future feature

wilwade (approved):

saraswatpuneet (approved):

Issue #286, opened 2022-07-28
Key Removal Part 2: Deleting the Key at Revocation Time


wilwade:
As the chain, we want to be able to keep a limit on the number of keys associated with an MSA. Currently the key to MSA relationship has the requirement that the key can only be associated with one MSA and that association is immutable.

## Open Questions
- Should we rename `revoke_msa_key` to `delete_msa_key`? YES.

## Acceptance Criteria
- [x] When msa.revoke_msa_key is called, the key should be removed from the database.
- [x] Update any documentation to the new details that key to MSA is still 1:1, but not immutable. There is nothing to stop a previously revoked key from being associated with a new MSA.
- [x] Update the js RPC data if needed

### Test cases that should be true given ^
- [x] *Before* calling msa.revoke_msa_key, I *cannot* add the key to a different MSA
- [x] *After* calling msa.revoke_msa_key, I *can* add the key to a different MSA
- [x] *After* calling msa.revoke_msa_key, if I use the RPC `get_msa_id` with the deleted key, it should respond with null
- [x] *After* calling msa.revoke_msa_key, if I use the RPC `get_msa_keys` with the msa of the deleted key, it should not respond with the key
- [x] If I have added `MaxKeys` keys...
    - [x] *Before* calling msa.revoke_msa_key, I *cannot* add a new key to the same MSA
    - [x] *After* calling msa.revoke_msa_key, I *can* add a new key to the same MSA

PR #300, opened 2022-08-04
Key Removal Part 1: Switch Message provider_key to use MSA Id instead of Signing Key

### LINKED ISSUE #285: Key Removal Part 1: Switch Message provider_key to use MSA Id instead of Signing Key
As the chain, we want to be able to keep a limit on the number of keys associated with an MSA. Currently the key to MSA relationship has the requirement that the key can only be associated with one MSA and that association is immutable.

## Part 1: Replace provider_key with provider MSA

We only store the provider key in one place, and before we can do part 2 of deleting the key (Issue: #286) the data use needs to be fixed.

## Acceptance Criteria
- [x] Update the storage of on-chain Messages to use the provider MSA instead of the provider key
- [ ] Verify that no other storage of MSA signing keys outside of the MSA pallet exist
- [ ] If I use the RPC to get an on-chain message that used a provider, it should return the provider msa
- [ ] Update the js RPC data

johnmela:
# Goal
The goal of this PR is to removal public key storage from the messages pallet. Instead, an Msa Id associated with the signer should be stored with the message. In order to proceed with https://github.com/LibertyDSNP/frequency/issues/286, public keys should only be stored in storage in the msa pallet.

Closes #285

## Linked Issue
https://github.com/LibertyDSNP/frequency/issues/285

# Discussion
I propose a change in naming of fields on the Message struct differing from instructions in #285. 
Instead using the term provider to reference the MSA that signed the transaction to create the message, I propose the new field should be named `signer_msa_id`. The reason for this is that there may not always be a provider / delegator relationship when sending a message. The term provider felt confusing to me since the provider could be self signed in the case that the MSA is sending a message on its own behalf.

Another option, if retaining the term provider is absolutely necessary:
Name the field `provider_msa_id` and add logic to set the field to `0` when there is no provider / delegator relationship to signify the self signed case.

# Checklist
- [ ] Chain spec updated
- [x] Design doc(s) updated
- [x] Tests updated
- [ ] Benchmarks added
- [x] Weights updated


common/primitives/src/messages.rs
```diff
@@ -13,15 +13,18 @@ use utils::*;
...
-	/// Message source account id (the original source).
-	pub msa_id: MessageSourceId,
-	/// Index in block to get total order
+	/// Message source account id of the MSA that signed the transaction. This may be the same id
+	/// as contained in `original_msa_id`, indicating that the transaction to add the message was
+	/// signed by the originator. An id differing from that of `original_msa_id` indicates that a
+	/// provider was delegated by `original_msa_id` to send this message on its behalf.
+	pub signer_msa_id: MessageSourceId,
+	///  Message source account id (the original source).
+	pub original_msa_id: MessageSourceId,
```
johnmela, line 26: This could also be left alone as `msa_id`. So long as documentation indicates that this field represents the MSA that the message is from, regardless if a provider signed the transaction

enddynayn: Looks good to me.

enddynayn (approved):

common/primitives/src/messages.rs
wilwade, line 26: Hmm.... I'm not sure about this naming over:
- `source_msa_id` or just `msa_id`
- `provider_msa_id`

Mostly to match the official data model:
image

https://github.com/LibertyDSNP/frequency/tree/main/designdocs

wilwade (changes requested):
Not sure about the naming, but either way we need to also update: https://github.com/LibertyDSNP/frequency/tree/main/designdocs which can be updated via updates to https://github.com/LibertyDSNP/DesignDocs/blob/main/img/BasicDataModel.drawio.png which can be opened and edited by draw.io

common/primitives/src/messages.rs
wilwade: Hmm... also what about `origin` instead of `original` or `source`?

johnmela: `provider_msa_id` didn't make sense to me to use unless there will always be a delegator / provider relationship. I suggested `signer_msa_id` because it signifies the holder of the MSA that signed the transaction to add the message. If however, we document very well that an MSA can be its own provider in the case that there is no delegation to a different MSA, then `provider_msa_id` makes sense to me (I am thinking of confusion for new developers).

`msa_id` instead of `original_msa_id` works for me also, `original` doesn't seem to add much value to the field name.

wilwade: I'm not sure it is any more or less documentation than what we already have to have around signer_msa_id. That introduces more questions. Who can be a signer? What if the signer is someone that isn't delegated to by the msa_id?

provider_msa_id deals with a lot of those issues. And the documentation around being one's own provider I think would be minimal (especially as it should be uncommon).

johnmela: Good points; I will refactor the variable names as:
`signer_msa_id`  -> `provider_msa_id`
`original_msa_id` -> `msa_id`

Also will update design docs to signify that instead of a provider key, we will store the provider msa on the `Message`

shannonwells (approved):
Nothing blocking, looks ready to go.

pallets/messages/src/lib.rs
```diff
@@ -201,9 +198,10 @@ pub mod pallet {
...
-			let message_source_id = Self::find_msa_id(&provider_key, on_behalf_of)?;
-			let message =
-				Self::add_message(provider_key, message_source_id, bounded_payload, schema_id)?;
+			let provider_msa_id = Self::find_msa_id(&provider_key, None)?;
+			let msa_id = Self::find_msa_id(&provider_key, on_behalf_of)?;
+
+			let message = Self::add_message(provider_msa_id, msa_id, bounded_payload, schema_id)?;
 
 			Ok(Some(T::WeightInfo::add_ipfs_message(cid.len() as u32, message.index as u32)).into())
 		}
```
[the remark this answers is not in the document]
johnmela: May as well change it now while I'm in here

wilwade (approved):
Nice. Thank you for remembering to update the js as well.

PR #308, opened 2022-08-09
Key remove part2 #286

### LINKED ISSUE #286: Key Removal Part 2: Deleting the Key at Revocation Time

saraswatpuneet:
# Goal
The goal of this PR is Deleting the Key at Revocation Time

Closes #286 #309 

## Acceptance Criteria
- [x] When msa.revoke_msa_key is called, the key should be removed from the database.
- [x] Update any documentation to the new details that key to MSA is still 1:1, but not immutable. There is nothing to stop a previously revoked key from being associated with a new MSA.
- [x] Remove KeyInfo.expired
- [x] Remove msa pallet `Error::<T>::KeyRevoked`
- [x] Update js/api-augment
- [x] update weights

### Test cases that should be true given ^
- [x] *Before* calling msa.revoke_msa_key, I *cannot* add the key to a different MSA
- [x] *After* calling msa.revoke_msa_key, I *can* add the key to a different MSA
- [x] *After* calling msa.revoke_msa_key, if I use the RPC `get_msa_id` with the deleted key, it should respond with null
- [x] *After* calling msa.revoke_msa_key, if I use the RPC `get_msa_keys` with the msa of the deleted key, it should not respond with the key
- [x] If I have added `MaxKeys` keys...
    - [x] *Before* calling msa.revoke_msa_key, I *cannot* add a new key to the same MSA
    - [x] *After* calling msa.revoke_msa_key, I *can* add a new key to the same MSA

# Checklist
- [x] Tests added
- [x] Weights updated


common/primitives/src/msa.rs
```diff
@@ -27,31 +27,21 @@ impl From<Delegator> for MessageSourceId {
 
 /// KeyInfo holds the information on the relationship between a key and an MSA
```
saraswatpuneet, line 28: Updates in this file is simply removing Blocknumber which was used in expiry which is now deleted

pallets/msa/src/lib.rs
```diff
@@ -393,8 +391,11 @@ pub mod pallet {
 				.map_err(|_| Error::<T>::AddKeySignatureVerificationFailed)?;
 
 			let msa_id = add_key_payload.msa_id;
+
 			Self::ensure_msa_owner(&who, msa_id)?;
 
+			Self::is_key_taken(msa_id, &key)?;
+
```
saraswatpuneet, line 398: Check if a key is taken else continue

pallets/msa/src/lib.rs
```diff
@@ -586,30 +596,31 @@ impl<T: Config> Pallet<T> {
...
-			ensure!(info.expired == T::BlockNumber::default(), Error::<T>::KeyRevoked);
-
-			let current_block = frame_system::Pallet::<T>::block_number();
-
-			info.expired = current_block;
-
-			*maybe_info = Some(info);
+	pub fn delete_key_info(key: &T::AccountId) -> DispatchResult {
+		KeyInfoOf::<T>::remove(key);
+		Ok(())
```
saraswatpuneet, line 606: deleting key should also delete its keyinfo

pallets/msa/src/lib.rs
```diff
@@ -586,30 +596,31 @@ impl<T: Config> Pallet<T> {
...
+	/// * [`DispatchResult`]
+	/// # Errors
+	/// * [`Error::<T>::NoKeyExists`] - If the key does not exist in the MSA
+	pub fn delete_key_for_msa(msa_id: MessageSourceId, key: &T::AccountId) -> DispatchResult {
+		<MsaKeysOf<T>>::try_mutate(msa_id, |key_list| {
+			let index = key_list.binary_search(key);
+			ensure!(index.is_ok(), Error::<T>::NoKeyExists);
+			key_list.remove(index.unwrap());
 			Ok(())
-		})?;
```
saraswatpuneet, line 610: added this function to remove the key from MSAKeysOf when delete is invoked
two things happen
1. Remove the key from msa_id
2. Remove the keyinfo associated with it

pallets/msa/src/lib.rs
```diff
@@ -682,13 +691,9 @@ impl<T: Config> Pallet<T> {
...
 
 	/// Checks that a key is associated to an MSA and has not been revoked.
-	pub fn ensure_valid_msa_key(
-		key: &T::AccountId,
-	) -> Result<KeyInfo<T::BlockNumber>, DispatchError> {
+	pub fn ensure_valid_msa_key(key: &T::AccountId) -> Result<KeyInfo, DispatchError> {
 		let info = Self::try_get_key_info(key)?;
 
-		ensure!(info.expired == T::BlockNumber::zero(), Error::<T>::KeyRevoked);
-
```
saraswatpuneet, line 691: Not needed as we are removing expiration semantics

pallets/msa/src/tests.rs
```diff
@@ -1053,3 +1046,54 @@ fn signed_extension_validation_valid_for_others() {
...
+				Origin::signed(new_account2.into()),
+				new_account1.into(),
+				signature,
+				add_new_key_data
+			),
+			Error::<Test>::KeyAlreadyRegistered
+		);
+	})
+}
+
```
saraswatpuneet, line 1074: above unit test check for errors when a key which is already associated with a msa is being added to another msa

pallets/msa/src/tests.rs
```diff
@@ -1053,3 +1046,54 @@ fn signed_extension_validation_valid_for_others() {
...
+		let add_new_key_data = AddKeyData { nonce: 1, msa_id: msa_id2 };
+		let encode_data_new_key_data = wrap_binary_data(add_new_key_data.encode());
+		let signature: MultiSignature = key_pair1.sign(&encode_data_new_key_data).into();
+		assert_ok!(Msa::add_key_to_msa(
+			Origin::signed(new_account2.into()),
+			new_account1.into(),
+			signature,
+			add_new_key_data
+		));
+	})
```
saraswatpuneet, line 1098: Test check  to ensure that if a key is removed from msa it can be added to another msa or the same one, just that key should be available

js/api-augment/interfaces/augment-api-query.ts
```diff
@@ -315,6 +315,26 @@ declare module '@polkadot/api-base/types/storage' {
        **/
       validationData: AugmentedQuery<ApiType, () => Observable<Option<PolkadotPrimitivesV2PersistedValidationData>>, []>;
     };
+    preimage: {
```
enddynayn, line 318: Thanks for doing this! I was going to open up a different PR.

pallets/msa/src/lib.rs
```diff
@@ -207,10 +207,10 @@ pub mod pallet {
 		NotKeyOwner,
 		/// An operation was attempted with an unknown Key
 		NoKeyExists,
-		/// An operation was attempted with a revoked Key
-		KeyRevoked,
 		/// The number of key values has reached its maximum
 		KeyLimitExceeded,
+		/// The key is already registered to the MSA
+		KeyAlreadyRegistered,
 		/// A transaction's Origin (AccountId) may not revoke itself
```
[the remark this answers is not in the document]
enddynayn: Is this a same as duplicatedKey?

pallets/msa/src/lib.rs
```diff
@@ -393,8 +391,11 @@ pub mod pallet {
 				.map_err(|_| Error::<T>::AddKeySignatureVerificationFailed)?;
 
 			let msa_id = add_key_payload.msa_id;
+
 			Self::ensure_msa_owner(&who, msa_id)?;
 
+			Self::is_key_taken(msa_id, &key)?;
```
enddynayn, line 397: https://github.com/LibertyDSNP/frequency/pull/308/files#diff-a15f544dbfef959186da2f9b4e7d8bd27aae8901824537de4f260e7a9b4f6c3bR506

I think this is checked already here.

pallets/msa/src/lib.rs
```diff
@@ -408,22 +409,23 @@ pub mod pallet {
...
 
 			ensure!(who != key, Error::<T>::InvalidSelfRevoke);
 
 			let who = Self::try_get_key_info(&who)?;
 			let key_info = Self::try_get_key_info(&key)?;
-			ensure!(who.expired == T::BlockNumber::zero(), Error::<T>::KeyRevoked);
 			ensure!(who.msa_id == key_info.msa_id, Error::<T>::NotKeyOwner);
 
-			Self::revoke_key(&key)?;
+			Self::delete_key_for_msa(who.msa_id, &key)?;
```
enddynayn, line 427: I think we need a transaction here

pallets/msa/src/lib.rs
```diff
@@ -408,22 +409,23 @@ pub mod pallet {
...
+			Self::delete_key_info(&key)?;
```
enddynayn, line 428: we are not deleting the key_info we are deleting the key.  I suggest not making two functions for this and instead combining them into one here:
```
Self::delete_key(&key)?;
```

Either we make this transactional or we combine them. Lets pair.



delete_key_info is not only deleting the key but the

pallets/msa/src/lib.rs
```diff
@@ -545,6 +545,16 @@ impl<T: Config> Pallet<T> {
 		Ok(())
 	}
 
+	/// Checks that the key is not already associated with another MSA
+	pub fn is_key_taken(msa_id: MessageSourceId, key: &T::AccountId) -> DispatchResult {
```
enddynayn, line 549: I dont think we need this.

pallets/msa/src/lib.rs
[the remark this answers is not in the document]
saraswatpuneet: Not really
1. Duplicate  Key, if an msa who owns the key try to add it again
2. KeyAlreadyRegistered: if an account with different msa try to add a key which is already registered to another msa

Keeping a separation will ensure correct errors tell users whats wrong

pallets/msa/src/lib.rs
saraswatpuneet: We also need to check if keyinfo msa_id == msa_id of who is trying to add that key, if not this a cross registration if yes we continue and next step i.e. try to add_key and returns duplicate key error

pallets/msa/src/lib.rs
saraswatpuneet: Or we can account for weights for delete_key_for_msa and delete_key_info within this extrinsic.
What you think?

pallets/msa/src/lib.rs
saraswatpuneet: Currently yes delete_key_for_msa deletes the association of key with msa from MSAKeysOf storage while delete_key_info deletes the keyinfo of that key from KeyInfo storage map

We can pair for sure

pallets/msa/src/lib.rs
saraswatpuneet: I am ok with combining it somewhere, just want to make sue we differentiate between DuplicateKey and key which is already registered to another MSA

pallets/msa/src/lib.rs
saraswatpuneet: @enddynayn  I combined the two in one where key is deleted from MSAKeys as well as KeyInfo

lets talk about its transactional nature

pallets/msa/src/lib.rs
saraswatpuneet: @enddynayn  I combined the two in one where key is deleted from MSAKeys as well as KeyInfo

lets talk about its transactional nature

pallets/msa/src/lib.rs
saraswatpuneet: @enddynayn  I combined the two in one where key is deleted from MSAKeys as well as KeyInfo

lets talk about its transactional nature

pallets/msa/src/lib.rs
saraswatpuneet: @enddynayn  maybe I am mistaken this function is within a transaction (extrinsic)

pallets/msa/src/lib.rs
saraswatpuneet: @enddynayn  maybe I am mistaken this function is within a transaction (extrinsic) if otherwise you mean something else

pallets/msa/src/lib.rs
johnmela: Should the name of this function be refactored to something like `ensure_key_is_available` or `ensure_key_is_not_taken` since it returns an Ok value if the key is not taken?

johnmela: Or, like @saraswatpuneet 's comment says below we could move this logic check out of its own function and into `add_key`

pallets/msa/src/lib.rs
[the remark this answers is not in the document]
enddynayn: Maybe we just rename duplicatedKey to KeyAlreadyRegistered?

pallets/msa/src/lib.rs
johnmela: @enddynayn and I discussed moving the check for `KeyAlreadyRegistered` from `is_key_taken` and placed directly into the `add_key` helper function

pallets/msa/src/lib.rs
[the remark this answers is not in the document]
shannonwells: KeyExists?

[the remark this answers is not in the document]
saraswatpuneet: I differ, we need a clear message whether this key is already added to my msa or this key is added to another msa

Think of range of users and providers and we need to have a clear demarcation of what is what 

KeyExists don't tell me If it exists for me or for someone else

pallets/msa/src/lib.rs
johnmela: After chatting with @enddynayn on this, what I believe his take to be is that both `key_list.remove(index.unwrap());` and the `KeyInfoOf::<T>::remove(key);` lines in `delete_key_for_msa` should happen like a single DB transaction would. Atomic in the sense that if there is an error when removing the index from key_list, we should not continue with removing the key from KeyInfoOf

johnmela: I am not sure if a race condition can cause the above state to occur? Is a substrate node single threaded? Will need to see if remove on StorageMap returns any kind of an error

saraswatpuneet: I am not sure what you mean by this

pallets/msa/src/lib.rs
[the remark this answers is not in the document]
enddynayn: So we put some thought into the renaming and additional error. I like the idea of renaming DuplicatedKey to KeyAlreadyRegistered. However, we found that the added error might only benefit a very small niche of users.
This user is someone who manages multiple MSAs and want to know if he already registered this MSA or if I have assigned this Key other another MSA. Note that you can only register a key if you have proof that it belongs to you.
I think renaming the error KeyAlreadyRegistered will already tell the user that they have registered this key to another MSA and that the additional error does not add much value to this particular user since saying that the KeyAlreadyRegistered implies that they have already assigned this key to another MSA.



Duplicate -
The error will tell the user that some other MSA already registered this key. 

KeyAlreadyRegistered - 
The error will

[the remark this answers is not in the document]
shannonwells: @enddynayn you are saying the opposite of what @saraswatpuneet was saying for the difference between the two errors. This is confusing enough that I think the naming is still not right.  One error is:
* This key is already registered to _this_ MSA.
* This key is already registered to _another_ MSA. 

The plain meaning of `KeyAlreadyRegistered` would apply to both cases; it doesn't make a distinction between them. `DuplicateKey` doesn't help draw the distinction.  So I think both names need improvement, perhaps:

`KeyAlreadyRegisteredToThisMSA`
`KeyAlreadyRegisteredToAnotherMSA`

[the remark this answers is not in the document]
saraswatpuneet: @johnmela  yes I like it we should have this separation between why this key is not getting added and having two set of errors gives correct information back to consumer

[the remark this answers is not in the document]
saraswatpuneet: @shannonwells  I liked this naming, this or something similar will be useful


I don't remember but we do have an issue in icebox or if not we should add to revisit our error names and update or make them more consumer friendly

pallets/msa/src/lib.rs
johnmela: After discussing this in slack, here is what I meant to say more clearly: 
Do we need to ensure that the removal of the value in `MsaKeysOf` inside of `delete_key_for_msa` is successful before proceeding with the `KeyInfoOf::<T>::remove(key);`. If try_mutate is thread safe or if the runtime is singlethreaded we likely don't need to worry about this.

pallets/msa/src/lib.rs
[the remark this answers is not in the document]
johnmela: @enddynayn 's argument is that we don't need to make the distinction between the two cases of where a key has been already registered, This vs Another MSA. A holder of a private key should presumably know where said key has already been registered and must sign proof before an MSA can be created on their behalf. An RPC call to `get_msa_id` could return this information.

That being said, I don't see any detriment in us making this distinction with a clearer error message. I could see it being useful in the case where someone is managing multiple MSAs.  If we do keep the distinction then I also agree that we must make the distinction more clear in better error naming. @shannonwells 's suggestions of `KeyAlreadyRegisteredToThisMSA` vs. `KeyAlreadyRegisteredToAnotherMSA` seem clear enough to me. Since the distinction has already been pushed to this PR, we may as well rename errors.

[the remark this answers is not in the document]
johnmela: In the spirit of un complicating the error response and at @enddynayn 's request, I have merged the two error types into one and named it `KeyAlreadyRegistered`. I think this will likely be sufficient, we can revisit in the future if necessary.

enddynayn (approved):

PR #311, opened 2022-08-09
CHORE: job for specs on release

saraswatpuneet:
# Goal
The goal of this PR is to create a job to release specs and binary on tagged releases for frequency

### DIFF
--- a/.github/workflows/frequency_specs.yml
+++ b/.github/workflows/frequency_specs.yml
@@ -1,0 +1,97 @@
+name: Build Frequency Binary and Specs
+
+on:
+  push:
+    tags:
+      - '[0-9][0-9][0-9][0-9].[0-9][0-9].[0-9][0-9][0-9]'
+  pull_request:
+
+jobs:
+  build_binary_and_specs:
+    name: ${{ matrix.target }}
+    strategy:
+      matrix:
+        os: [ubuntu-latest]
+        target: [build-node]
+        rust_toolchain: [nightly-2022-06-27]
+    runs-on: ${{ matrix.os }}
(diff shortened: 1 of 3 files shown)


.github/workflows/frequency_specs.yml
```diff
@@ -0,0 +1,102 @@
+name: Build Frequency Binary and Specs
+
+on:
+  push:
+    tags:
+      - '[0-9][0-9][0-9][0-9].[0-9][0-9].[0-9][0-9][0-9]'
+  pull_request:
```
wilwade, line 7: This is just for testing right?

.github/workflows/frequency_specs.yml
```diff
@@ -0,0 +1,105 @@
+name: Build Frequency Binary and Specs
+
+on:
+  push:
+    tags:
+      - '[0-9][0-9][0-9][0-9].[0-9][0-9].[0-9][0-9][0-9]'
```
wilwade, line 6: I think we'd want to switch this over to release and then upload the artifacts to release instead of the build.

https://github.com/softprops/action-gh-release

.github/workflows/frequency_specs.yml
saraswatpuneet: Yes

saraswatpuneet:
@wilwade  so the artifacts are getting generated and the step 
https://github.com/softprops/action-gh-release

Looks for branches being pushed to if they have tags
So this job should finish to success on tags creation, though we can test upon merge

wilwade (approved):

PR #323, opened 2022-08-10
Key revocation should be free

### LINKED ISSUE #206: Key Revocation should be free
Should be able to follow the pattern from #205 

As a user, I want to be able to revoke a key, even if I have _zero_ tokens.

## Acceptance Criteria
- [ ] Given that an MSA has **more than one** key, I should be able to revoke all but one of those keys without any tokens.

## Notes
- `revoke_msa_key`
- See: #205 and #229

dopatraman:
# Goal
The goal of this PR is to make sure revoking a key does not cost token.

Closes #206 

# Discussion
This PR adds a `Pays::No` to the `revoke_msa_key` extrinsic. It also adds a clause to the Signed Extension that blocks repeat revocations for a specific key.

# Checklist
- [x] Tests added
- [ ] Weights updated


johnmela (changes requested):
We can make the signed extension more generic by refactoring the struct we are using to implement `SignedExtension`:
`CheckProviderRevocation` -> `CheckRevocation` including code comments to reflect the name change.

johnmela (approved):

pallets/msa/src/lib.rs
```diff
@@ -824,6 +824,14 @@ where
 				)?;
 				return ValidTransaction::with_tag_prefix(TAG_PREFIX).and_provides(who).build()
 			},
+			Some(Call::revoke_msa_key { key, .. }) => {
+				const TAG_PREFIX: &str = "KeyRevocation";
```
enddynayn, line 828: Minor comment but I think it might be a good idea to move the body into to a function like to reduce duplication. Something like:

```
handle_call_priority(who, TAG_PREFIX)
```

enddynayn (approved):
Great. Minor comment.

PR #330, opened 2022-08-11
Removed nonce from KeyInfo

### LINKED ISSUE #193: [Storage Optimization] Remove KeyInfo Struct
Due to the change in how we use MSA keys, we no longer need to store any additional data along with the MSA Id when mapping from Account Id to MSA Id (aka what is currently Account Id -> KeyInfo )

- [ ] Replace KeyInfo with just the MSA Id
- [ ] Update RPCs that returned that information
- [ ] Update js/api-augment
- [ ] Update weights

dopatraman:
# Goal
The goal of this PR is to remove the `nonce` field from the `KeyInfo` class.

Closes #193 

# Discussion
We no longer need the `nonce` field to prevent Replay Attacks on adding keys.

# Checklist
- [x] Tests updated
- [ ] Weights updated


pallets/msa/src/weights.rs
```diff
@@ -81,43 +81,43 @@ impl<T: frame_system::Config> WeightInfo for SubstrateWeight<T> {
...
 			.saturating_add(T::DbWeight::get().reads(5 as Weight))
 			.saturating_add(T::DbWeight::get().writes(4 as Weight))
 	}
 	// Storage: Msa KeyInfoOf (r:1 w:0)
 	// Storage: Msa ProviderInfoOf (r:1 w:1)
 	fn remove_delegation_by_provider(s: u32, ) -> Weight {
-		(44_511_000 as Weight)
-			// Standard Error: 2_000
-			.saturating_add((26_000 as Weight).saturating_mul(s as Weight))
+		(69_036_000 as Weight)
```
shannonwells, line 91: That's weird, I'd think removing data from this struct would lower the weight. I guess it's your hardware.

shannonwells (approved):
Seems legit

enddynayn:
I do not think we want to remove nonce right? Can we hold off till tomorrow before merging.

saraswatpuneet (changes requested):
Keeping it as blocker. I do not think we confirmed removing nonce yet

wilwade:
One other note. If we do remove nonce and expired (#308), then we likely want to remove KeyInfo struct entirely.

wilwade:
Closing as https://github.com/LibertyDSNP/frequency/issues/193 has switched up.

PR #346, opened 2022-08-16
change min schema size and run benchmarks

### LINKED ISSUE #70: Schema constants
Determine min and max schema size, max schema registrations

* Min Schema size: it's an easier check to bail up front on a minimum size and we should. 
* Max schema size: determined by max weight and also @enddynayn suggests this can be limited before the message even gets put in the mempool by restricting the weight for the extrinsic.
* Max Schema registrations:  there was some debate about whether there should be a number below whatever the max of a U16 is.  If the limit is just U16 max value then it would seem we don't need this config?  Perhaps it would be better to leave some wiggle room though.

saraswatpuneet:
# Goal
The goal of this PR is to update minim schema size of 8 bytes

Closes #70 

# Checklist
- [x] Tests runs
- [x] Benchmarks updated
- [x] Weights updated


wilwade (approved):
Looks good.

Issue #234, opened 2022-07-12
Restructure Chain Specs


wilwade:
We need several different chain specs

\* Independent: aka no parachain

- [x] Local development (independent with instant sealing)
- [x] Local development (Parachain, This is what we have now)
- [x] Rococo Testnet (Parachain, temporary slot)
- [x] Mainnet (Parachain, Polkadot)
- [x] Rococo Testnet (Parachain, Long-term slot)
- [x] Readme updates

Examples of how other chains have done it:

- [Acala](https://github.com/AcalaNetwork/Acala/tree/master/node/service/src/chain_spec)
- [Moonbeam](https://github.com/PureStake/moonbeam/tree/master/node/service/src/chain_spec)
- [Litentry](https://github.com/litentry/litentry-parachain/tree/dev/node/src/chain_specs)

saraswatpuneet:
* Related Stories 
#257 #348 #350

PR #352, opened 2022-08-17
Mainnet specs #234

saraswatpuneet:
# Goal
The goal of this PR is to continue for #234 and 
* Add frequency runtime (this is our mainnet runtime)
* Add chain spec for frequency
* Update scripts and jobs to generate mainnet specs/wasm
* refactor existing local runtime/specs to name it frequency-local
* add **TODOs** placeholder in frequency spec for mainnet to be populated with required configuration 

# Checklist
- [x] Chain spec updated
- [x] Run tests for instant sealing as well as local net for correctness


scripts/generate_specs.sh
```diff
@@ -30,7 +30,7 @@ case $build_step in
     ;;
   build-testnet)
     echo "Building Spec for frequency as rococo testnet"
-    $PWD/target/release/frequency build-spec --chain=frequency_rococo --disable-default-bootnode > ./res/genesis/testnet/frequency-spec-rococo-testnet.json
+    $PWD/target/release/frequency build-spec --chain=frequency-rococo --disable-default-bootnode > ./res/genesis/testnet/frequency-spec-rococo-testnet.json
```
wilwade, line 33: I still cannot track when to use `-` and when `_`. So a ping to double check this line.

scripts/generate_specs.sh
```diff
@@ -30,7 +30,7 @@ case $build_step in
     ;;
   build-testnet)
```
wilwade, line 31: Should we rename this to `build-rococo`?

wilwade (changes requested):
Few small things I found. Really nice to have all these setup.

scripts/generate_specs.sh
saraswatpuneet: --chain "{spec-name}" triggers this part of cli, so i am going to change to following

1. "dev" | "rococo-2000" |"" i.e. either of these tags will point to local specs
2. "rococo-4044" | "testnet" will point to testnet specs, and 
3 "frequency" is mainnet 

https://github.com/LibertyDSNP/frequency/blob/1d907b077151c0649f3f6ae7ba56e2b290de09f0/node/cli/src/command.rs#L51

wilwade (approved):
🎉

Makefile
```diff
@@ -35,11 +35,14 @@ install-toolchain:
...
+specs-rococo-2000:
+	./scripts/generate_specs.sh 2000 rococo-2000
 
-specs-testnet:
-	./scripts/generate_specs.sh 4044 build-testnet
+specs-rococo-4044:
+	./scripts/generate_specs.sh 4044 rococo-4044
+
+specs-mainnet:
+	./scripts/generate_specs.sh 999 mainnet
```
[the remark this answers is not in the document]
saraswatpuneet: Good catch @demisx! Will do

Issue #279, opened 2022-07-28
Determine collator reference hardware


shannonwells:
- Q: Are collators expected to use Polkadot's own reference hardware?
   A: Yes, see #354, we are going with the recommended configuration, which uses x86

- Either way it needs to be documented and then used to generate new weights
- [x] @shannonwells to run the current weights on the `c6id.2xlarge` to see what the numbers currently look like.
  - [x] Messages
  - [x] Schema
  - [x] MSA
  - [x] ORML Vesting
  - [x] Scheduler

wilwade:
I propose we just use the Polkadot Validator requirements (most common to use across other parachains as well)

https://wiki.polkadot.network/docs/maintain-guides-how-to-validate-polkadot/#requirements

shannonwells:
We haven't yet; I'm about to check in some weights generated on the Polkadot reference hardware for helping make that decision.

wilwade:
Chosen Reference Hardware (Copied from Polkadot Validator Reference hardware on 2022-08-23)

    CPU
        x86-64 compatible;
        Intel Ice Lake, or newer (Xeon or Core series); AMD Zen3, or newer (EPYC or Ryzen);
        4 physical cores @ 3.4GHz;
        Simultaneous multithreading disabled (Hyper-Threading on Intel, SMT on AMD);
        Prefer single-threaded performance over higher cores count. A comparison of single-threaded performance can be found [here](https://www.cpubenchmark.net/singleThread.html).
    Storage
        An NVMe SSD of 1 TB (As it should be reasonably sized to deal with blockchain growth). An estimation of current chain snapshot sizes can be found [here](https://paranodes.io/DBSize). In general, the latency is more important than the throughput.
    Memory
        16GB DDR4 ECC.
    System
        Linux Kernel 5.16 or newer.
    Network
        The minimum symmetric networking speed is set to 500 Mbit/s (= 62.5 MB/s). This is required to support a large number of parachains and allow for proper congestion control in busy network situations.


Chosen versions on Common Cloud Platforms
- Google Cloud Platform (GCP): `c2d-highcpu-8`
-  Amazon Web Services (AWS): `c6id.2xlarge`

aramikm:
@wilwade RocksDB tries to do sequential writes to storage. So it's optimized for HDD and not SSD. I think there are some rocksdb flags that optimizes it for SSD. Based on this it's hard to tell if we should use HDD or SSD. Maybe we can ask Parity.

[...]

PR #354, opened 2022-08-18
Updated weights using reference hardware

shannonwells:
# Goal
The goal of this PR is to provide new weights for packages listed in the referring Issue.
These weights were generated using an AWS *c6id.2xlarge* instance.

Change summary:
* New weights for orml_vesting, scheduler, pallet_msa, pallet_messages, pallet_schemas
* Use the WeightInfo in config for scheduler pallet.
* Add a script run_all_benchmarks.sh which runs all the benchmarks and puts the weights in one place.
* Add a section to the README detailing how to set up a remote Ubuntu instance for building frequency

References #279 

# Checklist
- [x] Weights updated


pallets/messages/src/weights.rs
```diff
@@ -66,34 +67,34 @@ impl<T: frame_system::Config> WeightInfo for SubstrateWeight<T> {
 	// Storage: Msa KeyInfoOf (r:1 w:0)
 	// Storage: Messages BlockMessages (r:1 w:1)
 	fn add_onchain_message(n: u32, m: u32, ) -> Weight {
-		(33_905_000 as Weight)
+		(0 as Weight)
```
wilwade, line 70: @shannonwells, This is the only one that has zeros here. I haven't looked deep enough into the weights to understand why. Do you know? It feels wrong or perhaps our benchmark setup for the messages pallet is wrong?

runtime/frequency/src/weights/orml_vesting.rs
```diff
@@ -47,7 +49,7 @@ impl<T: frame_system::Config> orml_vesting::WeightInfo for SubstrateWeight<T> {
 	// Storage: System Account (r:2 w:2)
 	// Storage: Balances Locks (r:1 w:1)
 	fn vested_transfer() -> Weight {
```
wilwade, line 51: These came out fairly different from https://github.com/AcalaNetwork/Acala/blob/master/runtime/acala/src/weights/orml_vesting.rs which in theory uses the same reference hardware.

wilwade (commented):
Two comments so far

runtime/frequency/src/weights/orml_vesting.rs
wilwade: Hmmm... Acala does use different parameters for steps and repeat: https://github.com/AcalaNetwork/Acala/blob/7eaccb4163619c5462f99974280d30341297d1cd/Makefile#L250-L251

pallets/messages/src/weights.rs
shannonwells: Thanks for the reminder; I meant to check that out, to see if there had been major changes to messages since the weights were last regenerated.

shannonwells: I talked to @aramikm - it's actually fine, it means that the benchmark run found a good linear relationship when running all the benchmark instances

runtime/frequency/src/weights/orml_vesting.rs
shannonwells: 50 steps would be more accurate but also slower.

shannonwells: update:  I ran the weights with 50.

scripts/run_all_benchmarks.sh
```diff
@@ -0,0 +1,38 @@
...
+if [ "$1"="-h" -o "$1"="--h" -o "$1"="--help" ] ; then
+    echo "Usage:  $0 <project-home-dir>"
+fi
+
+export RUST_LOG=info
+PROJECT=${1:-$HOME/frequency}
+CHAIN=$PROJECT/res/genesis/testnet/frequency-spec-rococo-testnet.json
+RUNTIME=$PROJECT/target/release/frequency
+BENCHMARK="$RUNTIME benchmark pallet "
+OUTPUT_DIR=$PROJECT/runtime/frequency/src/weights
```
shannonwells, line 12: Note this is not really correct for our repo; @enddynayn has a draft PR suggesting a single directory for weights, so treat this as a placeholder. I've been using the root directory for output and copying the weights into the current places.

enddynayn (approved):
Perfect 👍

PR #360, opened 2022-08-18
chore(*): use frame-support constant

enddynayn:
Use frame-support RocksDbWeight constant
and remove file. Both are equivalent.

[RocksDbWeight](https://paritytech.github.io/substrate/master/src/frame_support/weights/rocksdb_weights.rs.html) constant reference.


shannonwells (approved):
Not sure why these were checked in, but I verified that they are both identical, codewise, to what's in Substrate.

wilwade (approved):

PR #361, opened 2022-08-18
chore(*): remove unused constant

enddynayn:
Remove ParityDbWeight constant since
it is not being used. If needed we can use
[frame-support::weights::constant::ParityDbWeight](https://paritytech.github.io/substrate/master/frame_support/weights/constants/struct.ParityDbWeight.html#)
which is equivalent.


shannonwells (approved):
Thanks for splitting this into a separate PR.

PR #362, opened 2022-08-18
chore(*): add common directory

enddynayn:
Create a common runtime directory
to put all weights and commonalities
between runtimes.


runtime/common/Cargo.toml
```diff
@@ -0,0 +1,183 @@
+[package]
```
enddynayn, line 1: clean up

shannonwells (approved):
good idea

aramikm (approved):
Looks good

Issue #282, opened 2022-07-28
Document GitHub release process


shannonwells:

wilwade:
@demisx 

> Is there anything in the current process that we do not want to carry over to the new one?

- We can skip spec and state artifact generation. Binaries and WASM we still need. (once Mainnet and Testnet are live, these are historical given the spec.json file. See #273 for more details)

> Is there anything in the current process that needs to be done differently?

1. Something we are not currently doing, but we should also upload the WASM to IPFS. Likely should pin as well, but not sure what we will use to do that.
2. Context: We should switch to a better versioning system for releases and Docker images.
3. Context: We will eventually have some js integration tests to run: #371 
4. Context: The JS API Augment should be re-written to do all of the code generation at build time (which will require using docker instant (and thus making sure it can retrieve the most recent version tag or running instant or something). Should also align version numbers somehow with Frequency releases. Hmm... Likely it could be folded into the Frequency Releases then...: #382

wilwade:
@demisx 

> What OSs and Rust Toolchains we need to support? Currently I only see ubuntu-latest and nightly-2022-07-23

- Ubuntu is fine for building I think, but it would be nice if we did an arm64 build.
- #356 covers that final version, but we still likely need nightly for clippy and mdbook.

wilwade:
@demisx One other thing.

We should move mdbook deployments to the release tagging instead of main likely. Or deploy to two different locations.

wilwade:
> > One other thing. We should move mdbook deployments to the release tagging instead of main likely. Or deploy to two different locations.
> 
> @wilwade ❓ Can you please clarify what you mean here? I am not seeing any mdbook deployment in the current workflows. Unless I'm missing something.

That's on me. The Frequency generated documentation not mdbook. Sorry about that.

wilwade:
> Re: proposed CI workflow for "commit pushed to PR branch" event

- The JS integration test will need a running frequency in instant sealing mode
- Eventually the same for verify JS API Augmentation. It will need access to the current metadata.
- Another thing that I forgot about #376 which is still a WIP.

saraswatpuneet:
@demisx  I think option 2 is followed by many projects in this space

saraswatpuneet:
We can follow same release as substrate itself 

When we have main on specific version of substrate before we upgrade to say next version.

We can either do a release or branch out

saraswatpuneet:
[...]

PR #367, opened 2022-08-20
[DO NOT MERGE] Chores/benchmarks arm hw 279

shannonwells:
# Goal
The goal of this PR is to show what the weights look like when run on the same exact configuration as the polkadot reference, except it's an ARM cpu - a `c6gd.2xlarge`.

This PR uses the benchmarks-reference-hw-279 branch as a base for comparison.

# Discussion
- [x] Weights updated


wilwade:
@shannonwells Wow. Well I guess that gives us our answer. Thanks for testing that!

I'll also update #282 to skip the arm builds for now.

PR #368, opened 2022-08-20
update readme me and add mkdir to script

saraswatpuneet:
# Goal
The goal of this PR is to update readme and script to be upto date with structure 

# Checklist
- [x] Readme updated
- [x] Scripts updated


saraswatpuneet:
The script change is not needed but some facing issues with rebasing so this will just create directory

PR #383, opened 2022-08-25
[Bug] Store payload length for IPFS messages

### LINKED ISSUE #353: [Bug] Store payload length for IPFS messages
Currently we only store an IPFS message's CID in block storage. We should store the payload length as well.

This is currently blocking https://github.com/LibertyDSNP/frequency/issues/119

johnmela:
# Goal
The goal of this PR is to encode the offchain payload length of IPFS messages in the Message payload along with its CID.

Closes #353

# Checklist
- [x] Tests added
- [x] Update Benchmarking
- [x] Update Weights


pallets/messages/src/benchmarking.rs
```diff
@@ -9,6 +9,7 @@ use frame_system::RawOrigin;
 const MESSAGES: u32 = 499;
 const SCHEMAS: u32 = 50;
 const IPFS_SCHEMA_ID: u16 = 65535;
+const IPFS_PAYLOAD_LENGTH: u32 = 10;
```
johnmela, line 12: I'm not sure how accurately this represents (or should represent) an offchain payload length, but would we want to revisit the benchmarks later?

pallets/messages/src/lib.rs
```diff
@@ -362,7 +367,17 @@ impl<T: Config> Pallet<T> {
...
-				response.content.push(m.map_to_response(block_number));
+
+				let mut payload_length: u32 = 0;
+				let payload = m.payload.clone().into_inner();
+
+				if payload_location == PayloadLocation::OnChain {
+					payload_length = m.payload.len().try_into().unwrap();
+				} else if payload_location == PayloadLocation::IPFS {
+					type PayloadType = (Vec<u8>, u32);
+					(_, payload_length) = PayloadType::decode(&mut &payload[..]).unwrap();
```
wilwade, line 378: Where is the `cid` coming from in the response if not here?

pallets/messages/src/types.rs
```diff
@@ -32,13 +32,15 @@ where
...
 		block_number: BlockNumber,
+		payload_length: u32,
 	) -> MessageResponse<BlockNumber> {
 		MessageResponse {
 			provider_msa_id: self.provider_msa_id,
 			index: self.index,
 			msa_id: self.msa_id,
 			block_number,
 			payload: self.payload.clone().into_inner(),
+			payload_length,
```
wilwade, line 43: This `map_to_response` is missing a few things:

1. Payload is wrong for ipfs messages.
2. Since it is always encoding a single schema, we don't need to have payload_length in here for non-ipfs messages.
3. We can also then get better keys. So instead of payload, we should have `cid` for ipfs schemas.
4. (not this story, but) We actually can do one better still for onchain messages in that the payload should be decoded and passed as a json object under payload as well. Since this is an off-chain interaction.

wilwade (changes requested):
Few small comments

pallets/messages/src/lib.rs
johnmela: This was written under the assumption that the client would decode the CID from the payload, but I think it makes sense to get the cid here and assign it to the payload in `MessageResponse`.

pallets/messages/src/types.rs
johnmela: I will modify for 1-3 in this PR. I like option 4 more

pallets/messages/src/tests.rs
```diff
@@ -11,8 +12,28 @@ use frame_support::{
...
+) {
+	let mut payload = Vec::from("{'fromId': 123, 'content': '232323114432'}".as_bytes());
+
+	if payload_location == PayloadLocation::IPFS {
+		payload = (
+			Vec::from("bafkreidgvpkjawlxz6sffxzwgooowe5yt7i6wsyg236mfoks77nywkptdq".as_bytes()),
+			IPFS_PAYLOAD_LENGTH,
+		)
+			.encode();
+	}
```
wilwade, line 35:
```suggestion
let payload = match payload_location {
    PayloadLocation::IPFS => {
    (
        Vec::from("bafkreidgvpkjawlxz6sffxzwgooowe5yt7i6wsyg236mfoks77nywkptdq".as_bytes()),
        IPFS_PAYLOAD_LENGTH,
	 ).encode()
    }
    _ => {
      Vec::from("{'fromId': 123, 'content': '232323114432'}".as_bytes())
    }
}


```

pallets/messages/src/lib.rs
```diff
@@ -365,7 +370,20 @@ impl<T: Config> Pallet<T> {
 				list.len().try_into().map_err(|_| Error::<T>::TypeConversionOverflow)?;
 			for i in from_index..list_size {
 				let m = list[i as usize].clone();
-				response.content.push(m.map_to_response(block_number));
+
+				let mut payload_length: Option<u32> = None;
+				let mut payload: Vec<u8> = Default::default();
+
+				if payload_location == PayloadLocation::OnChain {
```
wilwade, line 377: match

pallets/messages/src/lib.rs
```diff
@@ -186,13 +186,14 @@ pub mod pallet {
...
-			_payload_length: u32,
+			payload_length: u32,
 		) -> DispatchResultWithPostInfo {
 			let provider_key = ensure_signed(origin)?;
-			let bounded_payload: BoundedVec<u8, T::MaxMessagePayloadSizeBytes> = cid
-				.clone()
-				.try_into()
-				.map_err(|_| Error::<T>::ExceedsMaxMessagePayloadSizeBytes)?;
+			let bounded_payload: BoundedVec<u8, T::MaxMessagePayloadSizeBytes> =
+				(cid.clone(), payload_length)
```
aramikm, line 193: should we assign it to that certain Type first so when you are looking how that type is being used you can see that encode and decode. Other concern is that if this is not first converted to that custom type then this might change and the decode part might not and it would cause issue in decoding.

johnmela: Good idea

pallets/messages/src/lib.rs
```diff
@@ -365,7 +370,17 @@ impl<T: Config> Pallet<T> {
 				list.len().try_into().map_err(|_| Error::<T>::TypeConversionOverflow)?;
 			for i in from_index..list_size {
 				let m = list[i as usize].clone();
-				response.content.push(m.map_to_response(block_number));
+
+				let payload: Vec<u8>;
+				let payload_length: u32;
+				(payload, payload_length) = match payload_location {
```
aramikm, line 376:
```suggestion
				let (payload: Vec<u8>, payload_length: u32) = match payload_location {
```

aramikm (approved):
Looks good! Added one additional improvement

pallets/messages/src/lib.rs
johnmela: I had been trying to do that, not sure why but Rust doesn't like the type constraints in the tuple

johnmela: Ooo this worked! `let (payload, payload_length): OffchainPayloadType = match ...`

wilwade (approved):

PR #388, opened 2022-08-26
Chore/move weights files to common

### LINKED ISSUE #377: Update script that runs all benchmarks
The [current script to run all benchmarks]( https://github.com/LibertyDSNP/frequency/blob/main/scripts/run_all_benchmarks.sh) needs some tweaking.  Per @shannonwells it’s not complete and it doesn’t put the weights in the right place with the right names and it doesn’t have all the needed pallets to run benchmarks for. 

Blocks #280

Tasks:

- [ ] Weights files must be put into `runtime/common/src/weights`
- [ ] Weights file names will need to be updated
- [ ] All pallets for which benchmarks must be run should be in the lists
- [ ] Add a Makefile target for running benchmarks

shannonwells:
# Goal
The goal of this PR is to clean up benchmarks and weights, and to use our generated weights files for external pallets.

**Note 1**: run_all_benchmarks.sh must change if we enable more external pallets or write new frequency pallets.

**Note 2**: I did not add the collective/Council benchmark target + weights, since this PR is already large. This can be added in another PR, the one for creating the Technical collective.

Closes #377

## Change summary
* Corrected the run all benchmarks script so it names things correctly and puts them in the right place. 
* Corrected the runtime template for benchmark output (for external pallets) so it uses the pallet's own WeightInfo.
* Corrected an invalid directory in `.env.sample` after attempting to use it for this work
* Added a Make target for running benchmarks.  You can now run benchmarks with `make benchmarks`.
* Deleted runtime/frequency-rococo/src/weights directory since it's no longer used
* Renamed common weights files/modules for consistency
* Reran benchmarks to verify the correct functioning of the script and to be used as a reference when running them later.

# Discussion

# Checklist
- [x] run_all_benchmarks.sh script run
- [x] Weights updated
- [x] run `make start` to verify that updates to `WeightInfo` for each pallet Config is correct and that parachain runs without error.
- [x] run `cargo test` locally to verify that tests all pass


.maintain/runtime-weight-template.hbs
```diff
@@ -16,23 +16,12 @@
 use frame_support::{traits::Get, weights::{Weight, constants::RocksDbWeight}};
 use sp_std::marker::PhantomData;
 
-/// Weight functions needed for {{pallet}}.
```
shannonwells, line 19: We needed to be using `WeightInfo` from the pallet, not a local version. This was preventing us from actually using the weights we generated for our runtime, for external pallets.

enddynayn (approved):

enddynayn:
Fantastic! 💯

PR #395, opened 2022-08-31
Added asdf support

pfrank13:
Or as much as I could have. Clang proved not to be supportable through
asdf, blegh

# Goal
Add asdf support

Closes
Nothing

# Discussion
Trying to build out the project I noticed I needed some things installed and presumably the project would benefit those things being well versioned.

# Checklist
I would say none of these apply
- [ ] Chain spec updated
- [ ] Design doc(s) updated
- [ ] Tests added
- [ ] Benchmarks added
- [ ] Weights updated

### DIFF
--- a/.tool-versions
+++ b/.tool-versions
@@ -1,0 +1,5 @@
+rust 1.62.0
+make 4.3
+cmake 3.24.1
+# clang 14.0.6 # only plugin I could find for clang and not some subset of clang did not work, it was located here https://github.com/srivathsanmurali/asdf-clang.git
+
--- a/README.md
+++ b/README.md
@@ -54,6 +54,20 @@ We run benchmarks with and recommend the same [reference hardware specified by P
 At this point you should have `./target/release` directory generated locally with compiled
 project files.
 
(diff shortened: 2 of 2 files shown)


.tool-versions
```diff
@@ -0,0 +1,5 @@
+rust 1.62.0
+make 4.3
```
pfrank13, line 2: This just happens to be the current latest version.

.tool-versions
```diff
@@ -0,0 +1,5 @@
...
+cmake 3.24.1
```
pfrank13, line 3: Also happens to be the current latest version

.tool-versions
```diff
@@ -0,0 +1,5 @@
...
+# clang 14.0.6 # only plugin I could find for clang and not some subset of clang did not work, it was located here https://github.com/srivathsanmurali/asdf-clang.git
```
pfrank13, line 4: This is just a bummer

saraswatpuneet (approved):

johnmela (approved):

PR #405, opened 2022-09-06
Removed KeyInfo struct

### LINKED ISSUE #193: [Storage Optimization] Remove KeyInfo Struct

dopatraman:
# Goal
The goal of this PR is to remove all references to the `KeyInfo` struct.

Closes #193 

# Discussion
This PR leaves the `KeyInfoResponse` struct intact. We could rename this class if its current name is confusing.

# Checklist
- [x] Chain spec updated
- [x] Weights updated


pallets/msa/src/lib.rs
```diff
@@ -709,27 +711,27 @@ impl<T: Config> Pallet<T> {
...
 
 	/// Fetches all the keys associated with a message Source Account
 	/// NOTE: This should only be called from RPC due to heavy database reads
 	pub fn fetch_msa_keys(msa_id: MessageSourceId) -> Vec<KeyInfoResponse<T::AccountId>> {
 		let mut response = Vec::new();
 		for key in Self::get_msa_keys(msa_id) {
-			if let Ok(info) = Self::try_get_key_info(&key) {
-				response.push(info.map_to_response(key));
+			if let Ok(_info) = Self::try_get_msa_from_account_id(&key) {
+				response.push(KeyInfoResponse { key, msa_id, nonce: 0 });
```
wilwade, line 723: I don't think this should have a nonce anymore.
```suggestion
				response.push(KeyInfoResponse { key, msa_id });
```

pallets/msa/src/lib.rs
```diff
@@ -151,10 +151,11 @@ pub mod pallet {
 
 	/// Storage type for key to MSA information
 	/// - Key: AccountId
-	/// - Value: [`KeyInfo`](common_primitives::msa::KeyInfo)
+	/// - Value: [`MessageSourceId`]()
 	#[pallet::storage]
 	#[pallet::getter(fn get_key_info)]
-	pub type KeyInfoOf<T: Config> = StorageMap<_, Twox64Concat, T::AccountId, KeyInfo, OptionQuery>;
+	pub type KeyInfoOf<T: Config> =
```
wilwade, line 157: Should this still be named `KeyInfoOf`?

pallets/msa/src/lib.rs
```diff
@@ -151,10 +151,11 @@ pub mod pallet {
 
 	/// Storage type for key to MSA information
 	/// - Key: AccountId
-	/// - Value: [`KeyInfo`](common_primitives::msa::KeyInfo)
+	/// - Value: [`MessageSourceId`]()
```
wilwade, line 154: This will autoconnect it in the docs if you leave off the `()` (that is only sometimes possible, and I'm still not sure all of when you can get away with it)
```suggestion
	/// - Value: [`MessageSourceId`]
```

wilwade (changes requested):
Also getting an error when trying to run `make docs`:

common/primitives/src/msa.rs
142 | /// RPC Response form of [`KeyInfo`]
       | no item named `KeyInfo` in scope

pallets/msa/src/lib.rs
dopatraman: Good question. We could go with something like `MessageSourceIdOf` if it isn't too much of a mouthful.

johnmela (approved):

saraswatpuneet (approved):
lgtm :+1:

Issue #372, opened 2022-08-22
Better Docker Files


wilwade:
We need docker file(s) that are able to do the following:
- Instant sealing
  - `frequency-chain/instant-node`
  - Runs frequency in instant sealing mode for test containers or other testing situations
- Node:
  - `frequency-chain/node` or if we need two `frequency-chain/node` and `frequency-chain/rococo-node`
  - Runs a frequency mainnet/rococo node
  - this is close to what we have now

Optimizations to do:
- [ ] Smaller
    - [ ] Rather have multiple efficient docker images over one that can do it all
- [ ] Readme on Docker Hub
    - [ ] we have a new docker hub to setup for the chain: `frequency-chain`
    - [ ] I or Scott can switch over the GitHub Actions Docker Token env var when ready
- [ ] Better documentation for configuration and how to use it

**Tasks:**

- [x] Create `instant-seal-node` docker image
- [x] Switch to new [Docker Hub location](https://hub.docker.com/r/frequencychain/instant-seal-node)
- [ ] Create and publish `rococo-node` docker image
- [ ] Create and publish `mainnet-node` docker image
- [ ] Populate overview pages on Docker Hub
- [ ] Update README

References:
- [Polkadot Parachains](https://coinmarketcap.com/polkadot-parachains/polkadot/)
- [Multi-arch build and images, the simple way](https://www.docker.com/blog/multi-arch-build-and-images-the-simple-way/)
- [Frequency Docker Hub](https://hub.docker.com/u/dsnp)
---
- [Moonbeam Docker Hub](https://hub.docker.com/u/purestake)
- [Moonbeam Docker files](https://github.com/PureStake/moonbeam/tree/master/docker)
- [Moonbeam publish docker workflow](https://github.com/PureStake/moonbeam/blob/master/.github/workflows/publish-docker.yml)
- [Moonbeam GitHub releases](https://github.com/PureStake/moonbeam/releases)
- [Moonbeam optimized binaries](https://docs.moonbeam.network/node-operators/networks/collators/faq/)
---
- [Astar GitHub releases](https://github.com/AstarNetwork/Astar/releases)
- [Astar Docker Hub](https://hub.docker.com/u/staketechnologies)
- [Astar release workflow](https://github.com/AstarNetwork/Astar/blob/master/.github/workflows/release.yml)
---
- [Acala GitHub releases](https://github.com/AcalaNetwork/Acala/releases)
[...]

PR #408, opened 2022-09-07
add instant seal node docker image

demisx:
# Goal
The goal of this PR is to build and publish docker image for Frequency node in instant sealing mode.

Part of #372

### DIFF
--- a/docker/instant-seal-node.dockerfile
+++ b/docker/instant-seal-node.dockerfile
@@ -1,0 +1,42 @@
+# Requires to run from repository root and to copy the binary in the build folder
+FROM ubuntu:20.04 AS base
+LABEL maintainer="Frequency Team"
+LABEL description="Binary for Frequency parachain"
+
+RUN apt-get update && apt-get install -y ca-certificates && update-ca-certificates
+
+FROM ubuntu:20.04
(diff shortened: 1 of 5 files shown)


scripts/prune_all.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/sh
```
shannonwells, line 1: 🙌

shannonwells (approved):
Works on my machine

docker/instant-seal-node.dockerfile
```diff
@@ -0,0 +1,42 @@
+# Requires to run from repository root and to copy the binary in the build folder
+FROM ubuntu:20.04 AS base
+LABEL maintainer="Frequency Team"
+LABEL description="Binary for Frequency parachain"
+
+RUN apt-get update && apt-get install -y ca-certificates && update-ca-certificates
+
+FROM ubuntu:20.04
```
wilwade, line 8: @demisx I recall that we tried once to switch to the much smaller alpine image, but it didn't work for some reason that I don't recall. Might be worth seeing what others are using as a base image.

Also I think we could bump up to the latest LTS if we are going to stick with Ubuntu

Issue #272, opened 2022-07-28
[Genesis] Distribution of token to Foundation (Treasury) and others


shannonwells:
## Distributions
- [x] Project Liberty (5% Total)
    - [x] Mutlisig: 49_999_700 FRQCY
        - [x] 0xcd0760f6b5ebbcf0181368f7d7258e7e1f3154c43df4a1112b6d7e366937ca42
    - [x] Mutlisig Members: 100 FRQCY Each
        - [x] 0xb6fdaa1aea71c2cabf6fed4c2943d6fcd7c7a594e0bc36c8293220a1971dbc74
        - [x] 0x8e49e6382a3e6fa0e0cecefca2af3c7f9d3eebde95bd9e75893ed179d03cf61c
        - [x] 0x5854d238bff3f7a924eadcb1efa6fb2cce5dbdd340afdbdaaa4586567083883d
- [x] Unfinished Labs (40% Total)
    - [x] Mutlisig: 399_999_700 FRQCY
        - [x] 0xf06e128a8f9b813d07c1212475183c174de545f1705bf9b1ae8ce468a7f0ec0c
    - [x] Mutlisig Members: 100 FRQCY Each
        - [x] 0x46ebddef8cd9bb167dc30878d7113b7e168e6f0646beffd77d69d39bad76b47a
        - [x] 0xb26873df000ddae71bbc035cd907b1d23b06e662d7db11cd5802f845fbae547e
        - [x] 0x2041926697db16c0a27bd1bfe2883b69cd12186f89bc44c2e1cdd2eb69508268
- [x] Foundation Operational (2% Total)
    - [x] Mutlisig: 19_999_700 FRQCY
        - [x] 0xb2bacd25b122a125e0196107e0a13e0505d17399e46b93f339fae9387d9921cf
    - [x] Mutlisig Members: 100 FRQCY Each
        - [x] 0x6a793e27cb1bafc5c241872cafc75e08333b6b95765bf868e47056909cbc5b07
        - [x] 0x202f792d3edc84813c8e89b6a2cbc6dff7d077522f0b8be78e4b0877f2c0b70a
        - [x] 0xf811cbc8611b38f68f4e486284e71eb45523fc5cf685c070bc76331d595cf81c
- [x] Frequency (53% Total)
    - [x] Treasury: 529_988_600 FRQCY
        - [x] `common_runtime::constants::TREASURY_PALLET_ID.into_account_truncating()`
[...]

PR #412, opened 2022-09-09
Ensure sudo has tokens on Rococo and Mainnet

### LINKED ISSUE #271: Genesis: Distribution of token to Sudoer
- [x] Rococo
- [x] Mainnet

wilwade:
# Goal
The goal of this PR is to:
- Make frequency rococo sudo have tokens
- Make it harder to make this mistake with mainnet

Closes #271 

# Discussion

- For the mainnet spec, I wanted it to be closer to the final version. So this follows the pattern from rococo to have the keys above and then just variables below.
- Will we only launch mainnet with 5 collators? Likely not. We will have more. But our minimum is 5.

# Checklist
- [x] Chain spec updated
- [x] Cargo check
- [x] Cargo fmt
- [x] (N/A) Updated the js/api-augment code if a custom RPC added/changed
- [x]  (N/A) Design doc(s) updated
- [x]  (N/A) Tests added
- [x]  (N/A) Benchmarks added
- [x]  (N/A) Weights updated

### DIFF
--- a/node/service/src/chain_spec/frequency.rs
+++ b/node/service/src/chain_spec/frequency.rs
@@ -4,8 +4,9 @@ use frequency_runtime::{
 	AccountId, AuraId, Balance, CouncilConfig, SudoConfig, TechnicalCommitteeConfig,
 	EXISTENTIAL_DEPOSIT,
 };
+use hex::FromHex;
 use sc_service::ChainType;
-use sp_core::sr25519;
+use sp_core::ByteArray;
 
 /// Specialized `ChainSpec` for the normal parachain runtime.
 pub type ChainSpec = sc_service::GenericChainSpec<frequency_runtime::GenesisConfig, Extensions>;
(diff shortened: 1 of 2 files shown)


node/service/src/chain_spec/frequency.rs
```diff
@@ -35,14 +52,129 @@ pub fn frequency() -> ChainSpec {
...
+				Some(frequency_mainnet_keys::MAINNET_FRQ_SUDO.parse::<AccountId>().unwrap().into()),
 				// TODO:: endowed accounts with initial balance.
-				vec![(get_account_id_from_seed::<sr25519::Public>("Alice"), 1 << 60)],
+				vec![
+					(
+						frequency_mainnet_keys::MAINNET_FRQ_SUDO
+							.parse::<AccountId>()
+							.unwrap()
+							.into(),
+						1 << 60,
```
wilwade, line 140: These balances will be changed with the final check-off of #271 and #272

node/service/src/chain_spec/frequency.rs
```diff
@@ -35,14 +52,129 @@ pub fn frequency() -> ChainSpec {
...
+					),
+					(
+						frequency_mainnet_keys::COLLATOR_1_SR25519
```
wilwade, line 143: In the end these may not start with any tokens. TBD #272

saraswatpuneet (approved):
🎉

PR #416, opened 2022-09-09
[Enhancement] Update `add_key_to_msa` `AddKeyData` To Have a Expiration Block

### LINKED ISSUE #336: Update `add_key_to_msa` `AddKeyData` To Have a Expiration Block
As a user, I want to my `add_key_to_msa` proof to expire in the future.

## Acceptance Criteria
- [ ] If the block is too FAR into the future it should be rejected (> 128 blocks)
- [ ] If the block is in the past, it should be rejected

## Arch Notes
- Simple check on the extrinsic
- Does /not/ need to be configurable
- Error: `ProofHasExipred`

## Notes
- Substrate Transaction version of this: https://github.com/paritytech/substrate/blob/master/frame/system/src/extensions/check_mortality.rs
- Also will need to be applied to other extrinsics
    - #337
    - #338

johnmela:
# Goal
The goal of this PR is to add an expiration to the signed proof used to validate the addition of a new key to an MSA

Closes #336

# Checklist
- [x] Tests added
- [x] Update benchmarks
- [x] Update weights
- [x] Updated autogenerated JS api-augmented code


common/primitives/src/msa.rs
```diff
@@ -6,6 +6,9 @@ use serde::{Deserialize, Serialize};
 use sp_runtime::DispatchError;
 use sp_std::prelude::Vec;
 
+/// The gap between the current block and a future expiring block allowed when validating signature proofs.
+pub const EXPIRATION_BLOCK_VALIDITY_GAP: u32 = 128;
+
 /// Message Source Id or msaId is the unique identifier for Message Source Accounts
```
saraswatpuneet, line 12: Question: Will this number be required to be updated via runtime upgrades or via governance, then we should set this number in runtime instead of hardcoding it here
Check for example MaxKeys
https://github.com/LibertyDSNP/frequency/blob/27aa8738785a593a7ee6bd65a600e49783e27487/runtime/frequency/src/lib.rs#L346

saraswatpuneet (approved):
Looks good
Just one comment/question I left here, please confirm and good to go

Also update the js package too

common/primitives/src/msa.rs
wilwade: Hard code is fine I think.

johnmela:
Updated autogenerated JS api-augmented code

common/primitives/src/msa.rs
```diff
@@ -8,6 +8,9 @@ use sp_std::prelude::Vec;
 
 pub use crate::{ds::OrderedSetExt, schema::SchemaId};
 
+/// The gap between the current block and a future expiring block allowed when validating signature proofs.
+pub const EXPIRATION_BLOCK_VALIDITY_GAP: u32 = 128;
```
aramikm, line 12: nit: should this number be selected in a way that we get some round number of minutes considering that each block might be 6 or 12 seconds. it looks like right now with 6 sec per block we get 12.8 minutes which is a bit non round.

aramikm (approved):
Looks good to me

enddynayn:
I think we might need this for this extrinsic.

common/primitives/src/msa.rs
johnmela: I could see this being more useful to be a modifiable value? Perhaps for now we can set it to 150 blocks as per @aramikm 's comment so that at 6 seconds per block, the gap is 15 minutes

common/primitives/src/msa.rs
wilwade: @johnmela What is the default mortality on transactions?

johnmela:
/run-benchmark.*

PR #417, opened 2022-09-12
add parachain node docker image for Rococo and Mainnet

demisx:
# Goal
The goal of this PR is create a docker image for running Frequency parachain full node.

Part of #372


docker/rococo-node.dockerfile
```diff
@@ -0,0 +1,44 @@
...
+# 9933 P2P port
+# 9944 for RPC call
+# 30333 for Websocket
+EXPOSE 9933 9944 30333
+
+VOLUME ["/data"]
+
+ENTRYPOINT ["/frequency/frequency", \
+	"--chain=frequency-rococo", \
+	"--force-authoring", \
```
wilwade, line 32: This is for authoring (aka forming blocks)
```suggestion
```

docker/rococo-node.dockerfile
```diff
@@ -0,0 +1,44 @@
...
+	"--base-path=/data", \
+	"--wasm-execution=compiled", \
+	"--execution=wasm", \
```
wilwade, line 35: We want to just use the default wasms
```suggestion
```

docker/rococo-node.dockerfile
```diff
@@ -0,0 +1,44 @@
...
+	"--port=30333", \
+	"--rpc-port=9933", \
+	"--ws-port=9944", \
+	"--rpc-external", \
+	"--rpc-cors=all", \
```
wilwade, line 40: This is a good default, but I wonder if we want to setup this as a runtime param for docker.

Perhaps that is for a later option. Load balancers can do this for now.

docker/rococo-node.dockerfile
```diff
@@ -0,0 +1,44 @@
...
+	"--ws-external", \
+	"--rpc-methods=Unsafe", \
```
wilwade, line 42: This should be `=safe`

(If you search substrate code for `check_if_safe` you will find all the rpcs that `=safe` will exclude)

docker/rococo-node.dockerfile
```diff
@@ -0,0 +1,44 @@
...
+	"--state-cache-size=0" \
```
wilwade, line 43: Not sure why this is here. I don't think we want it.

docker/rococo-node.dockerfile
```diff
@@ -0,0 +1,44 @@
+# Requires to run from repository root and to copy the binary in the build folder
+FROM --platform=linux/amd64 ubuntu:focal AS base
+LABEL maintainer="Frequency Team"
+LABEL description="Frequency parachain full node for Rococo testnet"
+
+RUN apt-get update && apt-get install -y ca-certificates && update-ca-certificates
```
wilwade, line 6: You can save some disk space with adding this at the end.

```suggestion
RUN apt-get update && apt-get install -y ca-certificates && update-ca-certificates && apt-get clean
```

wilwade (changes requested):
Went through the cli params mostly

docker/parachain-node.dockerfile
```diff
@@ -0,0 +1,47 @@
...
+	"--rpc-port=9933", \
+	"--ws-port=9944", \
+	"--rpc-external", \
+	"--rpc-cors=all", \
+	"--ws-external", \
+	"--rpc-methods=safe" \
+	]
+
+# Params which can be overriden from CLI
+CMD ["--chain=frequency"]
```
wilwade, line 47: We also need to be able to pass through `--name` and `--bootnodes`

docker/collator-node-local.dockerfile
```diff
@@ -30,7 +32,7 @@ RUN chmod +x ./scripts/healthcheck.sh
 ENV Frequency_BINARY_PATH=./target/release/frequency
 
 HEALTHCHECK --interval=300s --timeout=75s --start-period=30s --retries=3 \
-    CMD ["./scripts/healthcheck.sh"]
+	CMD ["./scripts/healthcheck.sh"]
```
wilwade, line 35: Do we want a healthcheck that doesn't require jq?

wilwade: Sounds good.

docker/parachain-node.dockerfile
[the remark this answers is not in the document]
wilwade: overview page for Docker is the most important. I think when you push there is a way to push a md file for that readme... https://github.com/docker/build-push-action/blob/master/docs/advanced/dockerhub-desc.md looks like the GitHub Action way.

wilwade (approved):
🚂

PR #433, opened 2022-09-15
update DockerHub overview pages from actions

demisx:
# Goal
The goal of this PR is to populate DockerHub overview pages from dedicated readme files via GitHub actions.

Part of #372


wilwade (approved):
Nice!

PR #437, opened 2022-09-15
Binary optimization: Support Feature Flags #429

saraswatpuneet:
# Goal
The goal of this PR is  to enable specific flags for different runtimes

part of #429 

# Checklist
- [x] Update Frequency node
- [x] Test cargo check for all 4 features ```--features frequency``` , ```--features frequency-rococo-local``` , ```--features frequency-rococo-test``` and ```--features all-frequency-features```
- [x] Test cargo build for all features 
- [x] Run instant sealing mode with feature flag turned on and test frequency rococo-local chain
- [x] Run unit tests
- [x] Test benchmarks 
- [x] Update makefile and scripts 
- [x] Update spec/binary jobs for rococo and mainnet
- [x] Update main and PR jobs to run all feature

PR #438, opened 2022-09-15
Moved res files to resources directory

### LINKED ISSUE #390: Cleanup: Remove `res` Folder
- Move `res/rococo-local.json` to new `resources` dir
- We want to keep the generated files separate from the spec files we want to save
- Check to see if we need to update `rococo-local.json`
- Once that is done:
    - Move the needed parts of res/genesis/readme to the root readme
    - delete `res`
    - add `res` to the gitignore
    - update generate spec scripts to mkdir as needed.
- Update [Directory Structure Wiki page](https://github.com/LibertyDSNP/frequency/wiki/Directory-Structure) about folders (don't remove res, just note that it is ignored and add the new one if any)
- Open to also renaming `res` for clarity

dopatraman:
# Goal
The goal of this PR is separate vital resource files from generated res files.

Closes #390 

# Discussion
There are empty README files within the `genesis/local`, `genesis/mainnet`, and `genesis/testnet` folders. Those have not been moved over because they didn't seem useful. Please let me know if they should be.


resources/rococo-local.json
```diff
@@ -6,11 +6,10 @@
   "telemetryEndpoints": null,
   "protocolId": "dot",
   "properties": null,
-  "forkBlocks": null,
-  "badBlocks": null,
-  "lightSyncState": null,
   "consensusEngine": null,
   "codeSubstitutes": {},
+  "relay_chain": "rococo-local",
+  "para_id": 2000,
```
wilwade, line 12: This is the spec for the relay chain, not a parachain chain, so it shouldn't have these two fields.

Not sure why it was trying to require them.

wilwade (changes requested):
The genesis/README.md file should just be dropped I think. It is out of date information anyway and the root readme has all the info now.

resources/rococo-local.json
dopatraman: the build actually throws an error if they're not included.

shannonwells (approved):
I don't have anything to add. I'm not sure what the side-effects of the rococo-local chain spec are but since they're required (won't build otherwise) and it's for the local chain, I don't really see a problem.

saraswatpuneet (changes requested):
Putting a hold for https://github.com/LibertyDSNP/frequency/pull/437

above PR has major changes which changes these files too, given we are running extensive test to ensure our features are working, this should go after 432

resources/rococo-local.json
wilwade: @dopatraman Which build fails for you? I pulled this branch and `make build`, `make lint`, and `make start-frequency` all work.

saraswatpuneet: @dopatraman  I dont think this is needed, its not used in any builds

saraswatpuneet:
@dopatraman  few things to try before merge

1. Run frequency locally to confirm it is working
2. Can you check where ./res/genesis folder is consume while generating specs, that might need an update too especially frequency_spec.yml job because make specs-rococo-2000/-4044/-mainnet dumps output in res/genesis/folders, please run those checks and update accordingly

dopatraman:
@saraswatpuneet to my knowledge all the files consuming `rococo-local.json` have been updated. Any files that consume other generated `res/genesis` files should still be able to (we are still generating and writing files to that location, they are just not being committed to version control). Please let me know if I'm understanding this incorrectly.

wilwade (approved):
Really like the new `make specs-rococo-local`

enddynayn (approved):
fabulous!

saraswatpuneet (approved):
Ok cool then this won't break.the jobs

👍

Issue #327, opened 2022-08-11
[Security] Prevent Replay Attacks on MSA Signed Actions


wilwade:
## People to reach out to
- @enddynayn has some ideas around some more structural setups to handle this
- @wilwade has a lot of the context around how to execute these replays

## Mutli PR

- [x] RPC and state (@shannonwells)
    -  [x] Refactor the storage for get_msa_keys and add a struct with the nonce and the keys (that is MSA_Id -> `{ nonce, keys }`)  (@shannonwells)
    - [x] Add a get_msa_nonce RPC call  (@shannonwells)
- [x] Secure add_provider_to_msa (proof)
- [x] Secure add_key_to_msa (Proof)
- [ ] Secure revoke_msa_key (Call) 
- [ ] Secure revoke_msa_delegation_by_delegator (Call)

## Goal

Prevent replay attacks against MSA "signed" (signed by a key on the MSA) actions that are not Token transactions
- add_provider_to_msa (proof)
- add_key_to_msa (Proof)
- Delegate Permissions Update (Future) (Proof)
- revoke_msa_key (Call) 
- revoke_msa_delegation_by_delegator (Call)

Does NOT cover `create_sponsored_account_with_delegation` that is https://github.com/LibertyDSNP/frequency/issues/349

## Current Working Solution
- Add a piece of state that is MSA_Id -> nonce
    -  Refactor the storage for get_msa_keys and add a struct with the nonce and the keys
- Require that the nonce match for the above actions
- Increment the nonce on any of the above actions
- Add `msa_nonce` to the signed payload (for proofs) or the extrinsic (for calls) of each of the above items
- Add a get_msa_nonce RPC call
- Open Question:
    - Should we add a Signed Extension that checks this nonce? No
- General Rule:
  -  This nonce should be used for MSA signatures when that signature is not connected to tokens or capacity.

## Discussion Notes

 Questions
 1. Do we need a nonce?
 2. (assuming yes) Where does the nonce live?
 3. Is it the same nonce for each of our three signed proofs?
     - How can we make sure that this solution assists with solving future
       via a generic or standard structure?

 Problems:
 1. When you sign a payload, that payload should not be able to be used again.
     - If I add someone as a service provider with a signature, that
       signature should not be able to be used to re-add them again.
     - If I sign to allow a key to be added to an MSA, that exact signature should
       not be able to be used to re-added to the MSA
 2. When you sign a payload, the service provider should have to use it within
    a given time frame.
[...]

PR #440, opened 2022-09-16
Invert origin for proof, signers, for add_provider_to_msa

### LINKED ISSUE #428: Invert Origin and Proof signers for `add_provider_to_msa`
Currently the user is responsible for submitting and paying to add a provider to their msa. While that situation may occur, it is not the most common case. The expected common case is that the provider is willing to pay (tokens or capacity) to add themselves to the user's provider list.

## Acceptance Criteria
- [ ] A coinless user can sign a proof that can be passed off to the provider who then submits via the `add_provider_to_msa` extrinsic.

## Related Issues
- #337 
- #327 

## Arch Notes
- Mostly should be just flipping who the key and origin are related to.
- Don't forget to flip the AddProviderPayload as well

shannonwells:
# Goal
The goal of this PR is to swap provider/delegator as origin and signer when adding a provider to an msa.

Closes #428 

# Change summary
* Most of the changes were in tests, and involved replacing the simpler "test_public" account generation for the delegator with a key pair, since the delegator now must sign the `AddPayload` payload for `add_provider_to_msa`.
* Includes two ignored and currently failing replay attack tests that will be addressed in later PRs.  I included these because they were done as part of the original story (#327) for this branch, but backed out once it was found out we needed to do #428.
* Updated the Makefile to call cargo format correctly (so it actually formats files) since this is what my pre-commit hook uses and it didn't DTRT.
* Split up an error test case into multiples because the update was getting confusing.

# Checklist
- [x] N/A Chain spec updated
- [x] Tests added
- [x] N/A Benchmarks added


wilwade (changes requested):
Two small non-code things to fix, but looks great otherwise!

wilwade (approved):
💯

saraswatpuneet (approved):

Issue #418, opened 2022-09-12
RPC to Get the List of Keys Associated with an MSA


wilwade:
Caching services and others checking signatures off-chain want to know what keys are associated with a given MSA Id.

There is no state that maps MSA Ids to Account Ids. Only Account Ids to MSA Ids.

We only need to be able to retrieve the list of Account Ids for a given MSA id off-chain.

## Acceptance Criteria
- [ ] Custom RPC `get_msa_keys`
    - [ ] Param: msaId
    - [ ] Response: Array<AccountId>

## Arch Notes
- Create an off-chain database store that keeps track of MSAId -> Account Id
    - Idea: Off-chain worker (events)
    - Idea: Off-chain worker (state (accountid -> MSAId) for initial + events for upkeep)
    - Idea: Other? (Parity didn't have any)
    -  Options around offchain (regular, offchain indexing)

- Blocked by spike: https://github.com/LibertyDSNP/frequency/issues/474
- Related: #384

shannonwells:
This could use some error requirements and testable conditions for acceptance

wilwade:
@shannonwells / @enddynayn A lot of the suggestions and details are in the spike: https://github.com/LibertyDSNP/frequency/issues/474, but we haven't gotten a chance to discuss the output of the spike yet.

PR #444, opened 2022-09-19
[State Optimization] Refactor MsaKeysOf to store a reference counter

### LINKED ISSUE #384: [State Optimization] Refactor MsaKeysOf to store a reference counter
To save on-chain data, we have found that we do not need to store the mapping of MSA Ids to associated keys. Instead we can just count references

We are still keeping the storage of AccountId to MSA `MessageSourceIdOf`

## Acceptance Criteria
- [x] We no longer store on-chain the connection from MSA Id to Account Id
- [x] We maintain a reference counter for the number of keys attached to an MSA
- [x] Remove: RPC `get_msa_keys` for now (added back in https://github.com/LibertyDSNP/frequency/issues/418 )
- [x] Keep: `MaxKeys` Should still apply at the `add_key_to_msa` extrinsic
- [x] Keep: `delete_key_for_msa` should not be able to delete the last key

## To Do
- [x] JS Update
- [ ] Benchmarks

## Arch Notes
- Rename `MsaKeysOf` `MsaInfoOf` (rejected: `KeyReferenceCounterOf`)
- Reference Counter Change Points (MSA pallet)
    - create_sponsored_account_with_delegation
    - create
    - add_key_to_msa
    - delete_msa_key
- Swap out the storage of MsaKeysOf from mapping to `BoundedVec<T::AccountId, T::MaxKeys>` to just a `u8` that counts references for AccountId -> MSA Id (currently `MessageSourceIdOf`)
- The solution might be useful for other situations such as graph

johnmela:
# Goal
The goal of this PR is to save on-chain data by refactoring the MsaKeysOf storage map to no longer store the mapping of MSA ids to account keys. We have found that we do not need to store the mapping of MSA Ids to associated keys. Instead, we can just count references.

Closes #384

# Discussion
RPC `get_msa_keys` and references to MSA helper function `fetch_msa_keys` have been temporarily removed until #418 at which time they should be reimplemented to retrieve keys associated to an MSA with `MessageSourceIdOf`. 

# Checklist
- [x] Design doc(s) updated
- [x] Tests update


shannonwells (approved):
good to go

saraswatpuneet (approved):

PR #449, opened 2022-09-20
Implemented limit delegation to registered providers check.

### LINKED ISSUE #320: [Provider Registration] Limit delegation to Registered Providers
As the network, we want to limit delegation to providers only, so that delegations are limited.

## Related Work
- #319 

## Acceptance Criteria
- [x] Given that I have a user and a provider, the user should be able to call `add_provider_to_msa` successfully
- [x] Given that I have a user and a provider, the provider should be able to call `create_sponsored_account_with_delegation` successfully
- [x] If I use a non-provider MSA as a provider it should be rejected
    - [x] `add_provider_to_msa`
    - [x] `create_sponsored_account_with_delegation`

## Error

- `ProviderNotRegistered`

## Design Doc
- https://github.com/LibertyDSNP/frequency/blob/main/designdocs/provider_registration.md

rlaferla:
Implemented limit delegation to registered provider check per issue #320.  Fixed unit tests.

# Goal
The goal of this PR is to limit delegation to registered providers.

Closes #320

### DIFF
--- a/pallets/msa/src/lib.rs
+++ b/pallets/msa/src/lib.rs
@@ -274,6 +274,8 @@ pub mod pallet {
 		ExceedsMaxSchemaGrants,
 		/// Provider is not permitted to publish for given schema_id
 		SchemaNotGranted,
+		/// The operation was attempted with a non-provider MSA
+		ProviderNotRegistered,
 		/// The submited proof has expired; the current block is less the expiration block
 		ProofHasExpired,
 		/// The submitted proof expiration block is too far in the future
@@ -311,6 +313,7 @@ pub mod pallet {
 		/// - Returns [`InvalidSignature`](Error::InvalidSignature) if `proof` verification fails; `delegator_key` must have signed `add_provider_payload`
 		/// - Returns [`NoKeyExists`](Error::NoKeyExists) if there is no MSA for `origin`.
 		/// - Returns [`KeyAlreadyRegistered`](Error::KeyAlreadyRegistered) if there is already an MSA for `delegator_key`.
+		/// - Returns [`ProviderNotRegistered`](Error::ProviderNotRegistered) if the a non-provider MSA is used as the provider
 		///
 		#[pallet::weight(T::WeightInfo::create_sponsored_account_with_delegation())]
 		pub fn create_sponsored_account_with_delegation(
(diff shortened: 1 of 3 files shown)


pallets/msa/src/lib.rs
```diff
@@ -329,6 +332,12 @@ pub mod pallet {
 				Error::<T>::UnauthorizedProvider
 			);
 
+			// Verify that the provider is a registered provider
+			ensure!(
+				!Self::get_provider_metadata(Provider(provider_msa_id)).is_none(),
```
wilwade, line 337: Do we want to use `is_registered_provider` here?

pallets/msa/src/lib.rs
```diff
@@ -605,21 +614,31 @@ impl<T: Config> Pallet<T> {
...
 	/// and that a provider has authorized a delegator to create a delegation relationship.
-	pub fn ensure_valid_provider(
+	/// - Returns [`ProviderNotRegistered`](Error::ProviderNotRegistered) if the a non-provider MSA is used as the provider
+	/// - Returns [`InvalidSelfProvider`](Error::InvalidSelfProvider) if the delegator is the provider
+	pub fn ensure_valid_registered_provider(
 		delegator_key: &T::AccountId,
 		provider_key: &T::AccountId,
-		authorized_msa_id: MessageSourceId,
+		// TODO: Move testing payload validity
+		//      authorized_msa_id: MessageSourceId,
```
wilwade, line 631: Left over todo

wilwade (commented):
Two small leftover things that I see, but otherwise looks correct to me.

pallets/msa/src/lib.rs
rlaferla: Yes.  I just pushed the commit with it.

pallets/msa/src/tests.rs
```diff
@@ -591,8 +619,9 @@ pub fn create_sponsored_account_with_delegation_with_valid_input_should_succeed(
 		assert_eq!(provider_info.is_some(), true);
 
 		let events_occured = System::events();
-		let created_event = &events_occured.as_slice()[1];
-		let provider_event = &events_occured.as_slice()[2];
+		// let provider_registered_event = &events_occured.as_slice()[1];
```
johnmela, line 622: Can this comment be removed?

johnmela (changes requested):

pallets/msa/src/tests.rs
rlaferla: It could but I think it's better if it's there.

rlaferla: I can remove it but it’s documenting the event in the array which could prove useful for future modifications to that code.

johnmela: That makes sense, okay I like that

johnmela (approved):

PR #458, opened 2022-09-21
Implemented issue #378 removing a delegation and keys creates a token account

### LINKED ISSUE #378: [bug] removing a delegation and keys creates a token account
Using Pays::No on an extrinsic also runs through all other signed extension. The CheckNonce pre-dispatch creates an account with zero value.

Ensure that a new token account is not created when submitting feeless transactions.

- revoke_msa_delegation_by_delegator
- delete_msa_key

## Reproduce Steps
- Create an MSA account using Alice `create`
- Create a sponsored account using new keys `create_sponsored_account_with_delegation`
- Revoke delegation relation with Alice by calling `revoke_msa_delegation_by_delegator`
- New Token account is created with zero free balance and nonce of 1.
## Notes
- First step is to 100% validate the issue (do a Pays::No transaction and then check the state)
- Fix Ideas (might be more than one)
    - Fork CheckNonce (Replace the Substrate CheckNonce with a [CustomName]CheckNonce to fix the bug)
    - See about submitting an issue or fix to Substrate
    - Reap the account in the extrinsic

rlaferla:
Implemented issue #378.  The check nonce signed extension is in the runtime common.  Three unit tests for MSA have been implemented.

# Goal
The goal of this PR is to not create a token account if the extrinsic is free.

Closes #378


pallets/msa/Cargo.toml
```diff
@@ -26,6 +26,7 @@ sp-std = {default-features = false, git = "https://github.com/paritytech/substra
 
 # Frequency related dependencies
 common-primitives = {default-features = false, path = "../../common/primitives"}
+common-runtime = {path = '../../runtime/common', default-features = false}
```
saraswatpuneet, line 29: Apart from all the good work

I just generally don't like a dependency on runtime code but can't help it 


I wish we had  the capability to modify checked extrinsic but without that this is good

rlaferla: Agree but as you know, we tried several approaches over a long period time with different devs.  If in the future we find a better solution (this is implementation is actually pretty good and works), we can always update it.

runtime/common/src/extensions/check_nonce.rs
```diff
@@ -0,0 +1,141 @@
...
+			} else {
+				InvalidTransaction::Future
+			}
+			.into())
+		}
+
+		// Increment account nonce by 1
+		account.nonce += T::Index::one();
+		#[cfg(test)]
+		println!("\tAfter increment, account.nonce = {}", account.nonce);
```
aramikm, line 103: this should be removed after verification in tests

pallets/msa/src/tests.rs
```diff
@@ -1640,6 +1641,107 @@ fn replaying_add_provider_to_msa_fails() {
...
+				created_token_account = false;
+			},
+		};
+		assert_eq!(created_token_account, false);
+	})
+}
+
+// Assert that check nonce validation does create a token account for a paying call.
+#[test]
+fn signed_ext_check_nonce_creates_token_account_if_paying() {
```
aramikm, line 1714: One other to do it is to write the tests in check nonce file instead of in this specific pallet just like https://github.com/paritytech/substrate/blob/master/frame/system/src/extensions/check_nonce.rs#L126-L167
One other advantage is that you won't need to add that extra dependency

runtime/common/src/extensions/check_nonce.rs
```diff
@@ -0,0 +1,141 @@
...
+use sp_runtime::{
+	traits::{DispatchInfoOf, Dispatchable, One, SignedExtension},
+	transaction_validity::{
+		InvalidTransaction, TransactionLongevity, TransactionValidity, TransactionValidityError,
+		ValidTransaction,
+	},
+};
+use sp_std::vec;
+
+/// Nonce check and increment to give replay protection for transactions.
```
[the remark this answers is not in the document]
rlaferla: I have now included an explanation in a comment at the top of the file.  We can add it to the docs as well if you provide me with which document you think should include it.

runtime/common/src/extensions/check_nonce.rs
rlaferla: Good catch!  It is now removed.

pallets/msa/src/tests.rs
rlaferla: Despite the implementation being broader, the unit tests are actually specific to the MSA pallet as the issue cites two specific MSA functions that need this special override.

pallets/msa/Cargo.toml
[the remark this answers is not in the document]
rlaferla: Yes, I moved it to dev-dependencies.

aramikm (approved):
🚢 it.

saraswatpuneet (approved):

PR #459, opened 2022-09-21
bug in run benchmarks

saraswatpuneet:
# Goal
The goal of this PR is to fix a bug in benchmark script

Build binary while running benchmarks with all features


scripts/run_all_benchmarks.sh
```diff
@@ -26,7 +26,7 @@ function run_benchmark() {
   --template=$3
 }
 
-cargo build --profile production --features runtime-benchmarks --features frequency --workspace || exit_err
+cargo build --profile production --features runtime-benchmarks --features all-frequency-features --workspace || exit_err
```
dopatraman, line 29: why dont we add this line to the Makefile? that way we reference it one place for other build scripts

saraswatpuneet: I need to check with DevOps if they are calling this script as is else they will have to update Jenkins job for the same too

dopatraman:
I see this error when running the msa benchmarks:
```
$ ./target/release/frequency benchmark pallet \
  --pallet pallet_msa \
  --extrinsic "*" \
  --chain="frequency" \
  --execution wasm \
  --wasm-execution compiled \
  --steps 50 \
  --repeat 10 \
  --output=./output \
  --template=./.maintain/frame-weight-template.hbs

Error: Input("Benchmarking wasn't enabled when building the node. You can enable it with `--features runtime-benchmarks`.")
```

Note: This is after running
```
cargo build --profile production --features runtime-benchmarks --features all-frequency-features --workspace
```

saraswatpuneet:
@dopatraman  you got it wrong 

The binary is not in release folder but in production folder
So it is ./target/production/frequency

saraswatpuneet:
Ok this PR has an addition, updated script to run for a specified package

scripts/run_all_benchmarks.sh
wilwade: @saraswatpuneet It uses `make benchmarks`

saraswatpuneet: @wilwade  yes, question should be build all features for benchmarks or just frequency like it was before this PR

PR #461, opened 2022-09-21
Chore/424 remove unused permission field from addprovider

### LINKED ISSUE #424: Remove unused `permission` field from AddProvider
Discussion: Are there other uses for this that are occurring soon enough that we need to keep it?

While in rapid development, we added a `permission` field on `AddProvider` that turned out to be unneeded.

## Acceptance Criteria
- [ ] When signing an Add Provider Payload, I don't need to set a `permission` field
- [ ] Remove dead code around the `permission` field
- [ ]  Updates tests/benchmarks and weights

rlaferla:
# Goal
The goal of this PR is 

Closes #424


wilwade (changes requested):
Looks like the work from `CheckNonce` slipped into this PR

Cargo.lock
```diff
@@ -5805,6 +5805,7 @@ name = "pallet-msa"
 version = "0.1.0"
 dependencies = [
  "common-primitives",
+ "common-runtime",
```
wilwade, line 5808: I think this is extra from something else.

pallets/msa/Cargo.toml
```diff
@@ -26,6 +26,7 @@ sp-std = {default-features = false, git = "https://github.com/paritytech/substra
 
 # Frequency related dependencies
 common-primitives = {default-features = false, path = "../../common/primitives"}
+common-runtime = {path = '../../runtime/common', default-features = false}
```
wilwade, line 29: I think this is also extra.

wilwade (changes requested):
I think the `common-runtime` is extra, but otherwise looks good.

pallets/msa/Cargo.toml
rlaferla: Removed which also removed from Cargo.lock file.

PR #468, opened 2022-09-22
Provide optional block_number for delegation checks

### LINKED ISSUE #430: Check delegations at a given block number without a full node
As a content reader wanting to validate an off-chain claim of delegation at a historical point in time, I want to be able to know that a provider_msa_id was delegated to by one or more delegator_msa_ids at a given block number. I don't want to have to have an archive node to do this.

## Arch Notes
- check_delegations is an existing RPC interface, but it only does current delegation testing, not historical
- "No Archive Node" means that we have to do the lookup and then check the block number and not rely on `at` for a historical state check
- It will require some rework of the functions it calls
- Be careful when working with the functions that are called on-chain vs ones only called off-chain

## Acceptance Criteria
- [x] I should be able to call a custom RPC with a block_number, provider_msa_id, and list of delegator_msa_ids and be able to know which delegations were valid at that point in time.
- [x] I should still be able to call the same or another RPC with the current block number assumed
- [x] I should NOT be able to request a block number greater than the current block number

saraswatpuneet:
# Goal
The goal of this PR is to add additional parameters ```block_number``` while checking for delegation via RPC
Closes #430 

# Checklist
- [x] Tests added


pallets/msa/src/lib.rs
```diff
@@ -837,14 +841,14 @@ impl<T: Config> Pallet<T> {
 	/// # Returns
 	/// * [`DispatchResult`]
 	/// # Errors
-	/// * [`ensure_valid_delegation`] Errors
+	/// * [`ensure_valid_schema_grant`] Errors
```
saraswatpuneet, line 844: @wilwade  this was breaking the doc build, I have fixed it here

wilwade: I thought I fixed all of the breaking parts. Oh well. Thanks!

PR #467, opened 2022-09-22
Expiration parameter for AddProvider struct (#337)

dopatraman:
# Goal
The goal of this PR is to check a block's expiration before adding a new provider.

with @johnmela 

Closes #337 

# Discussion
The logic for checking a block's validity is currently local to the MSA pallet. As discussed, we will table refactoring it to a common module until it is used outside the MSA pallet.

This PR also includes a script for automating JS definition generation. You can invoke it by calling

```
make js
```

# Checklist
- [x] Tests added
- [x] Weights updated

Issue #337, opened 2022-08-14
Update `add_provider_to_msa` `AddProvider` To Have a Expiration Block


wilwade:
As a user, I want my `add_provider_to_msa` proof to expire in the future.

*Note:* Should follow the pattern from #336 

## Acceptance Criteria
- [ ] If the block is too FAR into the future it should be rejected
- [ ] If the block is in the past, it should be rejected

mehtaishita:
#338 stipulates that the future blocks does not need to be configurable. I assume that applies to this as well?

saraswatpuneet:
Notes:
* Refactor/ common lib/ Proxy pallet?
* Or define a macro for generic use?
* Impl as a common package?

saraswatpuneet:
Use the same block count for expiration

PR #469, opened 2022-09-22
Added expiration to create_sponsored_account_with_delegation (#338)

### LINKED ISSUE #338: Update `create_sponsored_account_with_delegation` to have an Expiration Block
As a user, I want to my `create_sponsored_account_with_delegation` proof to expire in the future.

*Note:* Should follow the pattern from #336 #337 

## Acceptance Criteria
- [ ] If the block is too FAR into the future it should be rejected
- [ ] If the block is in the past, it should be rejected

dopatraman:
# Goal
The goal of this PR is to add an expiration check to the `create_sponsored_account_with_delegation` extrinsic, similar to #336 and #337

Closes #338

Blocked on #467 

# Checklist
- [x] Tests added
- [x] Weights updated


scripts/run_benchmark.sh
```diff
@@ -36,5 +36,5 @@ $BENCHMARK \
   --steps 50 \
   --repeat 10 \
   --output=$weights_file \
-  --template=./.maintain/runtime-weight-template.hbs
+  --template=./.maintain/frame-weight-template.hbs
```
aramikm, line 39: any reason why this is changed?

aramikm (approved):
Looks good, just asked a question about the script change.

scripts/run_benchmark.sh
dopatraman: The runtime benchmark template doesn't write a `WeightInfo` trait definition to the benchmark file.

Issue #456, opened 2022-09-21
Remove Unneeded Custom RPC calls


wilwade:
Context: To make the code clearer, we want to have users use state queries instead of Custom RPCs for things that are a direct pass through with no logic or expected logic changes.

Remember: Custom RPCs allow for two things. 1. Off-chain logic and manipulation of data for easier use for the caller, 2. Consistent interface if the storage alters.

## Tasks
- [x] Remove Custom RPC [get_latest_schema_id](https://libertydsnp.github.io/frequency/pallet_schemas_rpc/trait.SchemasApiClient.html#method.get_latest_schema_id)
- [x] Remove Custom RPC [get_msa_id](https://libertydsnp.github.io/frequency/pallet_msa_rpc/trait.MsaApiClient.html#method.get_msa_id)
- [x] Update js/api-augment (custom rpcs are manual!)
- [x] Update example-client-frequency to use state queries
- [x] Notify other teams that might be using those RPCs

rlaferla:
Which other teams are using those calls?  How are they notified?

rlaferla:
Do we want to keep `get_by_schema_id()` and `check_schema_validity()`?  Only `get_latest_schema_id()`?

wilwade:
> Do we want to keep `get_by_schema_id()` and `check_schema_validity()`? Only `get_latest_schema_id()`?

Correct. `get_by_schema_id` may do some additional work and `check_schema_validity` is off-chain and may also do more in the future.

wilwade:
> Which other teams are using those calls? How are they notified?

I think it is just @pfrank13 in Slack and in the private channel in discord

rlaferla:
`get_msa_id` appears to be identical to the api query `messageSourceIdOf`. Do we want to rename it, add an additional interface or change client code to use `messageSourceIdOf`?

rlaferla:
Also, the RPC call `get_latest_schema_id()` takes a parameter: the block number/hash.

rlaferla:
Notified @pfrank13

PR #476, opened 2022-09-27
Removed unneeded RPC calls

rlaferla:
Removed unneeded RPC calls for getMsaId() and getLatestSchemaId() since state queries can be used instead.

Closes: #456


js/api-augment/README.md
```diff
@@ -83,7 +83,7 @@ Caches can also wreck this even if you reorder, so watch out.
 Optional responses are not mapped to `null` and instead return an object with a few properties.
 For more details see the [code for the Option class](https://github.com/polkadot-js/api/blob/master/packages/types-codec/src/base/Option.ts).
 ```javascript
-const optionalExample = await api.rpc.msa.getMsaId(account);
+const optionalExample = await api.rpc.schemas.getBySchemaId(1);
```
wilwade, line 86: Good catch!

js/api-augment/interfaces/augment-api-query.ts
```diff
@@ -292,6 +292,12 @@ declare module '@polkadot/api-base/types/storage' {
       [key: string]: QueryableStorageEntry<ApiType>;
     };
     msa: {
+     /**
+       * Storage type for key to MSA information
+       * - Key: AccountId
+       * - Value: [`MessageSourceId`]
+       **/
+      getMsaId: AugmentedQuery<ApiType, (arg: AccountId32 | string | Uint8Array) => Observable<Option<u64>>, [AccountId32]> & QueryableStorageEntry<ApiType, [AccountId32]>;
```
wilwade, line 300: Where is this coming from?

wilwade (commented):
One question, otherwise looks good

js/api-augment/interfaces/augment-api-query.ts
saraswatpuneet: I think this is a public function

saraswatpuneet: Yes this should not have been added

saraswatpuneet: @rlaferla  I think you also have to update metadeta fetch first and then build with chain running in instant sealing mode

saraswatpuneet (commented):
Need following changes

1. Update runtime to remove the rpc from runtime too
2. Try updating js/api-augment again

js/api-augment/interfaces/augment-api-query.ts
rlaferla: That's some stray code.  It's removed in the next commit.

enddynayn (approved):

saraswatpuneet (approved):

PR #477, opened 2022-09-29
Feat/message from ipfs location

### LINKED ISSUE #119: RPC to get Messages from IPFS Payload Locations
Given that a schema can be created with Payload Location IPFS (#190)

I want to receive in the `get_messages_by_schema` RPC to get back the cid that I used when I called `add_ipfs_message`

# Acceptance Criteria
- [x] I can call an RPC to retrieve previously added IPFS messages with a `cid` field that is the IPFS cid
- [x] When I call an RPC to to retrieve previously added IPFS messages, I should *not* receive the `msa_id` field

## Remember
- [ ] Update api-augment interface

# Examples
- IPFS, Parquet: `{ index, block_number, provider_msa_id, payload_length, cid }` 
- Avro, OnChain: `{index, block_number, provider_msa_id, payload_length, msa_id,  payload: <bytes> }`
- `MessageResponse`
  - Required
    - index
    - block_number
    - provider_msa_id
  - IPFS Only
    - cid
    - payload_length
  - On-chain only
    - msa_id
    - payload

# Notes
- Use `#[cfg_attr(feature = "std", serde(skip_serializing_if = "Option::is_none"))]` on top of optionals

### Arch Notes/Questions
- Do we need to make `MessageResponse` dependent on `Schema`? At least such that the data is dependent on the Payload Location and Model Type? **Yes**
- Do we want to have consistent data in `MessageResponse` even if the payload is a CID instead of the payload? Or should it be custom to that Schema? **Custom to schema**
- Discussion Notes
  - eg Custom
      - IPFS, Parquet: `{ index, block_number, provider_msa_id, payload_length, cid }` 
      - Avro, OnChain: `{index, block_number, msa_id, provider_msa_id, payload: <object from bytes> }`
   - Standard: **_Not using this_**
      - IPFS, Parquet: `{ index, block_number, provider_msa_id, msa_id: <Optional> payload_length, payload: <cid> }` 
      - Avro, OnChain: `{index, block_number, provider_msa_id, msa_id, payload_length: <null | payload bytes length>, payload: <bytes or object> }`
- Should payload be decoded (different story) for Avro? 
  - **No**. We want library developers to be able to encode and decode using bidirectional tools

### Implementation Sub Tasks
- [x] Message Response with the proper optional fields
- [x] Build message response using custom data fields
- [x] Update message response in API across all tests
- [x] Update message response in API possibly across other pallets
- [x] Write unit test: get_messages_by_schema_returns_ipfs_cid_required_fields
- [x] Write more unit test ....(more to come)

mehtaishita:
# Goal

The goal of this PR is return messages’ cid for IPFS payload location

Closes #119

# Discussion

Optional fields were added to Message Response object, while still returning schema specific message response (with only the required relevant fields serialized and returned). Added another serde serializer that accepts Option type

# Checklist

- [x]  Chain spec updated
- [x]  Updated the js/api-augment code if a custom RPC added/changed
- [x]  Design doc(s) updated
- [x]  Tests added
- [x]  ~~Benchmarks added~~
- [x]  ~~Weights updated~~

@wilwade


pallets/messages/src/types.rs
```diff
@@ -31,20 +31,35 @@ impl<MaxDataSize> Message<MaxDataSize>
 where
 	MaxDataSize: Get<u32> + Clone,
 {
-	/// Helper function to handle response type [`MessageResponse`] for RPC.
+	/// Helper function to handle response type [`MessageResponse`] depending on the Payload Location (on chain or IPFS)
 	pub fn map_to_response<BlockNumber>(
 		&self,
 		block_number: BlockNumber,
 		payload: Vec<u8>,
```
wilwade, line 38: This was in the original code, but I think `payload` could be referenced from `self.payload` instead of needing to be passed in.

js/api-augment/definitions/messages.ts
```diff
@@ -24,6 +24,7 @@ export default {
     },
     MessageResponse: {
       payload: "Vec<u8>", //  Serialized data in a user-defined schema format
+      cid: "Vec<u8>", // The content address for an IPFS payload
```
wilwade, line 27: I don't think this matches now right? Several of these should be `Option`?

Also don't forget to run `make js` to update the other parts of the metadata off of this.

common/primitives/src/messages.rs
```diff
@@ -106,6 +112,15 @@ mod tests {
 		message: String,
 	}
 
+	#[test]
+	fn as_hex_option_msg_deserialize_test() {
```
wilwade, line 116: Test is currently failing to build

wilwade (changes requested):
Nice start. Just getting things buttoned up now

common/primitives/src/messages.rs
mehtaishita: I missed committing a couple of imports

mehtaishita:
I need to auto generate types for js/api-augment. The file says I need to run `yarn generate:defs | generate:meta`. But I get the error script `polkadot-types-from-defs` not found

wilwade:
> The file says I need to run `yarn generate:defs | generate:meta`

Where are you seeing that documentation? I thought we updated it all. It should just be as simple as running `make js`.

common/primitives/src/utils.rs
```diff
@@ -19,6 +19,32 @@ pub mod as_hex {
...
+	}
+
+	/// Deserializes a hexadecimal string into a `Vec<u8>`
+	pub fn deserialize<'de, D: Deserializer<'de>>(
+		deserializer: D,
+	) -> Result<Option<Vec<u8>>, D::Error>
+	where
+		D: Deserializer<'de>,
+	{
+		impl_serde::serialize::deserialize(deserializer).map(|r| Some(r))
```
enddynayn, line 45: I think this is suppose to return Ok(Some(r))?

mehtaishita: `.map` takes care of wrapping the option type as a Result
Screen Shot 2022-09-30 at 3 30 57 PM

enddynayn, line 45: How do we get a None result and an Error?

designdocs/batched_messages.md
```diff
@@ -188,15 +188,20 @@ of the `MessageResponse` to be more like the following:
...
-	#[cfg_attr(feature = "std", serde(with = "as_hex"))]
 	/// Serialized data in a the schemas.
-	pub payload: Payload,
-  /// Payload location
-  pub payload_location: PayloadLocation,
-	/// The public key of the provider and the signer of the transaction.
-	pub provider_key: AccountId,
-	/// Message source account id (the original source).
-	pub msa_id: MessageSourceId,
+	#[cfg_attr(feature = "std", serde(with = "as_hex_option", skip_serializing_if = "Option::is_none", default))]
```
enddynayn, line 192: I think I might not be understanding this correctly. But it seems that when Option is None, we do not serialize.
Therefore, can we remove the Option from bytes params as shown here?https://github.com/LibertyDSNP/frequency/pull/477/files#diff-2d880531276d9fc6ad8edfe8cd913d301f52e900e3c5c3ee0c6fee5183276421R28?

pallets/messages/src/types.rs
```diff
@@ -31,20 +31,35 @@ impl<MaxDataSize> Message<MaxDataSize>
...
+		payload_location: &PayloadLocation,
 	) -> MessageResponse<BlockNumber> {
-		MessageResponse {
-			provider_msa_id: self.provider_msa_id,
-			index: self.index,
-			msa_id: self.msa_id.unwrap_or_default(),
-			block_number,
-			payload,
-			payload_length,
+		return match payload_location {
```
enddynayn, line 40: I think you can drop the return value here?

enddynayn:
Looking good.

common/primitives/src/utils.rs
mehtaishita: we really don't, because this deserializer isn't called when using `skip_serializing_if = "Option::is_none"`

enddynayn (approved):
Looks great! Lint Stuff.

common/primitives/src/messages.rs
```diff
@@ -14,24 +14,31 @@ use utils::*;
...
+	pub msa_id: Option<MessageSourceId>,
+	/// Serialized data in a the schemas.
+	#[cfg_attr(
+		feature = "std",
+		serde(with = "as_hex_option", skip_serializing_if = "Option::is_none", default)
+	)]
+	pub payload: Option<Vec<u8>>,
+	/// The content address for an IPFS payload
+	#[cfg_attr(feature = "std", serde(skip_serializing_if = "Option::is_none", default))]
+	pub cid: Option<Vec<u8>>,
```
shannonwells, line 37: The Issue shows the CID as the payload when it's an off-chain, IPFS message.  What motivated the design change?

pallets/messages/src/tests.rs
```diff
@@ -419,12 +420,49 @@ fn get_messages_by_schema_with_ipfs_payload_location_should_return_offchain_payl
 		assert_eq!(
 			pagination_response.content[0],
 			MessageResponse {
-				msa_id: 10,
-				payload: cid,
+				payload: None,
+				index: from_index as u16,
+				provider_msa_id: 1,
+				block_number: 0,
+				payload_length: Some(IPFS_PAYLOAD_LENGTH),
```
shannonwells, line 427: If I recall correctly, this was originally intended to be the size of the batch file referenced by the CID, not a constant, so that we could charge for off-chain messages based on file size.  In that case payload_length may not make sense as a name; it could be `message_size, and then it doesn't need to be None() for on-chain messages; in both cases it can be the length of the on-chain message or batch file.

pallets/messages/src/tests.rs
```diff
@@ -270,12 +270,13 @@ fn get_messages_by_schema_with_valid_request_should_return_paginated() {
...
 			MessageResponse {
-				msa_id: 10,
-				payload: payload.clone(),
+				msa_id: Some(10),
+				payload: Some(payload.clone()),
 				index: from_index as u16,
 				provider_msa_id: 1,
 				block_number: 0,
-				payload_length: payload.len().try_into().unwrap(),
+				payload_length: None,
```
shannonwells, line 278: Also in my opinion this payload length should not be None, it should be the length of the actual payload/message.

shannonwells (changes requested):
This isn't your fault but I think there are problems with the original design, that now I look at the code I think should be addressed.

First of all having four optional fields to be able to smush two types of messages together in a MessageResponse seems like a code smell to me.  Secondly the original design did not have a separate field for CID in the MessageResponse, however this change was approved by others.  

Thirdly I think `payload_length` should always be the message size, whether it's on chain or a file off-chain, and perhaps should be renamed to message_length or message_size.  If we made the CID be the payload instead of a separate field, then message_length/size makes more sense.

common/primitives/src/messages.rs
wilwade: I'm confused. The issue does show responding with the `cid` value if it is an IPFS message. Which is what this is right?

pallets/messages/src/tests.rs
wilwade: We talked about that back and forth in planning. In the end the choice was that payload_length was just for the off-chain payloads as otherwise it was just extra data

pallets/messages/src/tests.rs
wilwade: I could imagine `content_length` to follow the HTTP header, but I think it would need to be a different issue for re-naming or altering so that it can be discussed again.

wilwade:
@shannonwells 
> First of all having four optional fields to be able to smush two types of messages together in a MessageResponse seems like a code smell to me.

In effect this is a union type, but resolved instead of trying to use Rust https://doc.rust-lang.org/reference/items/unions.html. I'm not sure if there is a better way using structs and where statements, but perhaps.

So I agree it is mushing two types together, but I'm not sure if there is a better Rust way to do it for the time being. Perhaps this is a future refactor?

common/primitives/src/messages.rs
```diff
@@ -14,24 +14,31 @@ use utils::*;
...
 	///  Offchain payload length (IPFS).
-	pub payload_length: u32,
+	#[cfg_attr(feature = "std", serde(skip_serializing_if = "Option::is_none", default))]
+	pub payload_length: Option<u32>,
```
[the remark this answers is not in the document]
mehtaishita: okay I can add examples in the spec

pallets/messages/src/tests.rs
shannonwells: I recall, I'm just second guessing.  Is it so much extra data that it's worth the inconsistency and extra logic? I'd prefer the consistency but if it's really outweighed by storage cost then it sounds like we just have to deal with it.

common/primitives/src/messages.rs
shannonwells: No actually, I was confused by the way it was written but after a night's sleep I see how I was mistaken.

shannonwells (approved):

shannonwells:
Let's resolve this later. I don't want to hold up the PR any longer at this late date.

PR #479, opened 2022-09-29
Chores/common runtime config

### LINKED ISSUE #375: [Final Checks] Align frequency and frequency-rococo runtimes
## Open Questions
- Can we use feature flags to handle the differences?
    - ??

## Acceptance Criteria
- [ ] Refactor the runtimes so they are as unified as possible
- [ ] Document such that it known how to work with them and prevent divergence

## Notes
- Can these be the same?
    - No
- What is different between them?
    - Various runtime differences
    - sudo
    - small other differences

shannonwells:
# Goal
The goal of this PR is to refactor common settings between frequency and frequency-rococo chains into their own module and use where needed, so as to prevent copy-paste errors and easily keep the two chains aligned in configuration where they need to be.

Closes #375 

# Discussion
This should not change any behavior. I also made a couple of numerically named constants for easier reading. Some constants/types could not easily be unified and moved into the new common files because they had too many dependencies on the runtime-specific config.

# Checklist
- [x] N/A Chain spec updated
- [x] N/A Tests added
- [x] N/A Benchmarks added

### DIFF
--- a/common/primitives/src/node.rs
+++ b/common/primitives/src/node.rs
@@ -4,12 +4,15 @@ pub use sp_runtime::{
 	traits::{BlakeTwo256, IdentifyAccount, Verify},
 	MultiAddress, MultiSignature, OpaqueExtrinsic,
 };
+/// Some way of identifying an account on the chain. We intentionally make it equivalent
+/// to the public key of our transaction signing scheme.
+pub type AccountId = <<Signature as Verify>::Signer as IdentifyAccount>::AccountId;
+
+/// The address format for describing accounts.
+pub type Address = MultiAddress<AccountId, ()>;
 
 /// Balance is a generic type for the balance of an account.
 pub type Balance = u128;
-
-/// Block header type as expected by this runtime.
-pub type Header = generic::Header<BlockNumber, BlakeTwo256>;
 
 /// Block type as expected by this runtime.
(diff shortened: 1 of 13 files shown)


common/primitives/src/node.rs
```diff
@@ -5,12 +5,16 @@ pub use sp_runtime::{
 	MultiAddress, MultiSignature, OpaqueExtrinsic,
 };
 
+/// Some way of identifying an account on the chain. We intentionally make it equivalent
```
shannonwells, line 8: I brought remaining types in here that were already being used.

runtime/common/src/constants.rs
```diff
@@ -35,3 +46,144 @@ impl sp_std::fmt::Debug for MaxDataSize {
...
+
+pub type AuraMaxAuthorities = ConstU32<100_000>;
+
+pub type CollatorMaxCandidates = ZERO;
+pub type CollatorMinCandidates = ZERO;
+
+parameter_types! {
+	pub const CollatorPotId: PalletId = PalletId(*b"PotStake");
+	// pub const CollatorSessionLength: BlockNumber = 6 * HOURS;
+	pub const CollatorExecutiveBody: BodyId = BodyId::Executive;
```
wilwade, line 179: @shannonwells I confirmed that this is an unused copy-pasta from statemint. So it can go away along with the xcm dependency.

PR #492, opened 2022-10-07
Config Treasury

### LINKED ISSUE #348: [Config][Governance] Mainnet Treasury Config
**Description:**
Add [treasury pallet ](https://paritytech.github.io/substrate/master/pallet_treasury/index.html) so that we can have a treasury for frequency.

## Acceptance Criteria
- [ ] Treasury Pallet Added (Local, Rococo, and Mainnet)
- [ ] Treasury Pallet Configured per Below 
- [ ] For Local
    - [ ] Initial balance like that of the others
- [ ] For Rococo
    - [ ] Initial Balance like that of sudo
- [ ] For Mainnet
    - [ ] Initial Balance of `1 FRQCY` with a todo note pointing at #272 
- [ ] PolkadotJS UI should display treasury pallet dropdown in Rococo, Local, and Mainet

## Config Options and suggested values (Local, Rococo, and Mainnet)
- [x] type ApproveOrigin = 
```
   - EitherOfDiverse<
	EnsureRoot<AccountId>,
	pallet_collective::EnsureProportionAtLeast<AccountId, FrequencyCouncilCollective, 3, 5>,
>;
```
    - Origin from which approvals must come.
- [x] type RejectOrigin = MoreThanHalfCouncil;
```
EitherOfDiverse<
	EnsureRoot<AccountId>,
	pallet_collective::EnsureProportionMoreThan<AccountId, FrequencyCouncilCollective, 1, 2>,
>;
```
    - Origin from which rejections must come.
- [x] type OnSlash = (); We are going to burn // Polkadot uses `Treasury`
    - Handler for the unbalanced decrease when slashing for a rejected proposal or bounty.
- [x] ProposalBond = Permill::from_percent(5);
    - Fraction of a proposal’s value that should be bonded in order to place the proposal. An accepted proposal gets these back. A rejected proposal does not.
- [x] type ProposalBondMinimum = `100 tokens`; // Polkadot 100 DOT
- [x] type ProposalBondMaximum = `1000 tokens`; // Polkadot 500 DOT
- [x] type SpendPeriod = 28 * DAYS; // Polkadot uses `24 * DAYS`
    - Period between successive spends.
- [x] type Burn = Zero for now. Will be something later once the economic system is in place
    - Percentage of spare funds (if any) that are burnt per spend period.
- [x] type BurnDestination = (); // /dev/null
    - May not be able to be (), might be just an unusable account
    - Handler for the unbalanced decrease when treasury funds are burned.
- [x] type SpendFunds = None; // Polakdot uses `Bounties`, but we don't want to do that yet
- [x] type MaxApprovals = 64; // Polkadot uses 100
    - The maximum number of approvals that can wait in the spending queue.
- [x] type SpendOrigin = frame_support::traits::NeverEnsureOrigin<Balance>; // Don't allow
    - The origin required for approving spends from the treasury outside of the proposal process. The Success value is the maximum amount that this origin is allowed to spend at a time.
- [ ] type WeightInfo = weights::pallet_treasury::WeightInfo<Runtime>;
- [x] type PalletId = TreasuryPalletId; // `PalletId(*b"py/trsry")`
- [x] type Currency = Balances;

## Notes
- https://paritytech.github.io/substrate/master/pallet_treasury/index.html
- https://paritytech.github.io/substrate/master/pallet_treasury/pallet/trait.Config.html

wilwade:
# Goal
The goal of this PR is to setup the treasury pallet

Closes #348

# Discussion
- How to Test:
  - Start the chain with the spend period once every 2 blocks: `FRQCY_SPEND_PERIOD=2 make start`
  - Create a proposal
  - Commit it to council
  - Vote on Proposal (2x)
  - Close proposal
  - (Now the proposal is just waiting for the end of the spend period)
  - `make local-block` 2x
  - See the proposal gone, and the money transferred

# Checklist
- [x] Chain spec updated
- [x] Manual Testing. Automated e2e tests tbd
- [x] Benchmarks added
- [ ] Weights updated (triggering Benchmarks after all the config stories are in)


Makefile
```diff
@@ -18,6 +18,15 @@ stop-relay:
 stop-frequency-docker:
 	./scripts/init.sh stop-frequency-docker
 
+.PHONY: local-block
+local-block:
```
wilwade, line 22: Simple make a block script

runtime/common/src/lib.rs
```diff
@@ -3,3 +3,33 @@
...
+		}
+	};
+	($prod:expr, $test:expr, $env:expr) => {
+		if cfg!(feature = "frequency-rococo-local") {
+			core::option_env!($env).map(|s| s.parse().ok()).flatten().unwrap_or($test)
+		} else {
+			$prod
+		}
+	};
+}
```
saraswatpuneet, line 35: :1st_place_medal:

node/service/src/chain_spec/frequency.rs
```diff
@@ -131,12 +132,22 @@ pub fn frequency() -> ChainSpec {
...
+					(
+						frequency_mainnet_keys::MAINNET_FRQ_SUDO
+							.parse::<AccountId>()
+							.unwrap()
+							.into(),
+						1 << 60,
+					),
+					(
+						common_runtime::constants::TREASURY_PALLET_ID.into_account_truncating(),
+						1 * frequency_runtime::UNIT,
```
wilwade, line 148: Mainnet amount of initial treasury will be updated with #272

runtime/common/src/lib.rs
```diff
@@ -3,3 +3,33 @@
...
+/// Usage:
+/// ```Rust
+/// parameter_types! {
+/// 	// Note that the env variable version parameter cannot be const.
+/// 	pub LaunchPeriod: BlockNumber = prod_or_local_or_env!(7 * DAYS, 1 * MINUTES, "FRQCY_LAUNCH_PERIOD");
+/// 	pub const VotingPeriod: BlockNumber = prod_or_local_or_env!(7 * DAYS, 1 * MINUTES);
+/// }
+/// ```
+#[macro_export]
+macro_rules! prod_or_local_or_env {
```
wilwade, line 20: Slightly modified from https://github.com/paritytech/polkadot/blob/3d6b56342e5e1313685aae6f4e222cbf91ba790f/runtime/common/src/lib.rs#L257

saraswatpuneet (approved):
:bullettrain_side:

runtime/common/src/lib.rs
```diff
@@ -3,3 +3,33 @@
...
+/// Macro to set a value (e.g. when using the `parameter_types` macro) to either a production value
+/// or to an environment variable or testing value (in case the `frequency-rococo-local` feature is selected or in instant-sealing mode).
+/// Note that the environment variable is evaluated _at compile time_.
+///
+/// Usage:
+/// ```Rust
+/// parameter_types! {
+/// 	// Note that the env variable version parameter cannot be const.
+/// 	pub LaunchPeriod: BlockNumber = prod_or_local_or_env!(7 * DAYS, 1 * MINUTES, "FRQCY_LAUNCH_PERIOD");
+/// 	pub const VotingPeriod: BlockNumber = prod_or_local_or_env!(7 * DAYS, 1 * MINUTES);
```
wilwade, line 16: Watch out. This `const` MUST be removed to use the env var version.

runtime/common/Cargo.toml
```diff
@@ -32,19 +32,22 @@ sp-std = { git = "https://github.com/paritytech/substrate", default-features = f
...
+pallet-treasury = { git = "https://github.com/paritytech/substrate", default-features = false, branch = "polkadot-v0.9.29" }
 pallet-utility = { git = "https://github.com/paritytech/substrate", default-features = false, branch = "polkadot-v0.9.29" }
 # ORML
 orml-vesting = { git = "https://github.com/open-web3-stack/open-runtime-module-library", default-features = false, branch = "polkadot-v0.9.29" }
 
 [features]
 default = ["std"]
 std = ["frame-support/std"]
+frequency-rococo-local = []
 try-runtime = [
```
saraswatpuneet, line 44: You should import std here, we need std for sure just like default

runtime/common/Cargo.toml
```diff
@@ -32,19 +32,22 @@ sp-std = { git = "https://github.com/paritytech/substrate", default-features = f
...
 pallet-scheduler = { git = "https://github.com/paritytech/substrate", default-features = false, branch = "polkadot-v0.9.29" }
+pallet-treasury = { git = "https://github.com/paritytech/substrate", default-features = false, branch = "polkadot-v0.9.29" }
 pallet-utility = { git = "https://github.com/paritytech/substrate", default-features = false, branch = "polkadot-v0.9.29" }
 # ORML
 orml-vesting = { git = "https://github.com/open-web3-stack/open-runtime-module-library", default-features = false, branch = "polkadot-v0.9.29" }
 
 [features]
 default = ["std"]
 std = ["frame-support/std"]
+frequency-rococo-local = []
```
saraswatpuneet, line 43: You should import std here, we need std for sure just like default

saraswatpuneet: I guess not, std is included in default features, so its good

runtime/common/src/lib.rs
```diff
@@ -3,3 +3,33 @@
...
+	($prod:expr, $test:expr) => {
+		if cfg!(feature = "frequency-rococo-local") {
+			$test
+		} else {
+			$prod
+		}
+	};
+	($prod:expr, $test:expr, $env:expr) => {
+		if cfg!(feature = "frequency-rococo-local") {
+			core::option_env!($env).map(|s| s.parse().ok()).flatten().unwrap_or($test)
```
aramikm, line 30: I like this macro approach.
I think the name or documentation is a bit confusing about the priority of which value will be used in 3 param case. It looks like we first try to get the local env value and if not exists we use provided local value but it's not obvious that the logic is doing that without looking into the macro.

runtime/frequency-rococo/src/lib.rs
```diff
@@ -462,6 +462,73 @@ impl pallet_democracy::Config for Runtime {
 	type OperationalPreimageOrigin = pallet_collective::EnsureMember<AccountId, CouncilCollective>;
 }
 
+parameter_types! {
+	/// The time proposals are collected before payout
+	/// as well as when the burn of left over funds happens (if any)
+	/// Setup on a 1-week scheduled for Rococo
+	pub SpendPeriod: BlockNumber = prod_or_local_or_env!(7 * DAYS, 10 * MINUTES, "FRQCY_SPEND_PERIOD");
```
aramikm, line 469: since we separated we runtimes, it's a bit weird that we are using `prod_or_local_or_env`. It feels like maybe we can combine those runtime s and just use this on every param. It would be easier than duplicate everything in these 2 files

aramikm (approved):
Looks good, Just added 2 comments that might be a bit confusing.

runtime/common/src/lib.rs
shannonwells: So it looks like the only way to get an environment variable value is if it's configured for frequency-rococo-local, is that the intention?

runtime/common/src/lib.rs
enddynayn: why? Can you explain the value add here?

enddynayn (approved):

runtime/common/src/lib.rs
enddynayn: Can I do this by building with an environment variable?
// either a production value or to an environment variable or testing value

runtime/frequency-rococo/src/lib.rs
```diff
@@ -462,6 +462,73 @@ impl pallet_democracy::Config for Runtime {
...
+	/// - Root (sudo or governance)
+	/// - 3/5ths of the Frequency Council
+	type ApproveOrigin = EitherOfDiverse<
+		EnsureRoot<AccountId>,
+		pallet_collective::EnsureProportionAtLeast<AccountId, CouncilCollective, 3, 5>,
+	>;
+
+	/// Who rejects treasury proposals?
+	/// - Root (sudo or governance)
+	/// - 3/5ths of the Frequency Council
```
shannonwells, line 489:
```suggestion
	/// - Simple majority of the Frequency Council
```

runtime/frequency/src/lib.rs
```diff
@@ -493,6 +493,73 @@ impl pallet_democracy::Config for Runtime {
...
+	/// - Root (sudo or governance)
+	/// - 3/5ths of the Frequency Council
+	type ApproveOrigin = EitherOfDiverse<
+		EnsureRoot<AccountId>,
+		pallet_collective::EnsureProportionAtLeast<AccountId, CouncilCollective, 3, 5>,
+	>;
+
+	/// Who rejects treasury proposals?
+	/// - Root (sudo or governance)
+	/// - 3/5ths of the Frequency Council
```
shannonwells, line 520:
```suggestion
	/// - Simple majority of the Frequency Council
```

runtime/frequency-rococo/src/lib.rs
enddynayn: can you please explain why we cant do this with `prod_or_fast`?

shannonwells (approved):
I had the two suggestions to change comments, and a couple of non-blocking questions.

runtime/common/src/lib.rs
saraswatpuneet: This prevent us from purely relying on env variable and have a default path baked in

runtime/common/src/lib.rs
saraswatpuneet: Good question by @shannonwells , how this env variable will be set
Would we want to test different values locally without having to modify runtime code

runtime/common/src/lib.rs
wilwade: Correct. We wanted to make sure that there was no way to adjust the values for a situation where there was more than one authority. (aka public rococo and mainnet)

runtime/common/src/lib.rs
wilwade: @enddynayn It is not building with an env, it is running with an env. (See the example in the PR description)

runtime/common/src/lib.rs
wilwade: That is just a watch out on how to use it. No issue, just how to use the macro. It is there in the documentation, but wanted to highlight it.

runtime/common/src/lib.rs
wilwade: @aramikm I couldn't think of a good way to name the macro to get this grouping `prod OR ( env ELSE local )`. Open to naming thoughts.

runtime/frequency-rococo/src/lib.rs
wilwade: @aramikm It is tempting to do a merge and use a macro like this. I worry that might end up really nasty as well however.

We only use it in `frequency-rococo` however as that's the "shared" one between instant, local rococo, and public rococo. Mainnet should never see this, or if it does, it would never do anything.

wilwade: @enddynayn `prod_or_fast` macro is what this is, just swapping out the feature flag it uses.

runtime/common/src/lib.rs
wilwade: What about `use_prod_unless_local_then_env_check`?

aramikm: I would prefer to move the env to 2nd param and 3rd as local default and rename it as `use_prod_unless_local_env_or_default`

Issue #489, opened 2022-10-05
fix: units of measurement and fees


enddynayn:
**Description**
After updating the denomination of 1 FRQCY to be represented by 1e8 planks. Other units were
not updated to reflect the change. Ensure that all units of measurement are set appropriately and
used to calculate fees appropriately.

resources:
on [computing deposits](https://substrate.stackexchange.com/questions/3932/computing-deposits-in-polkadot)
on [WeightToFee](https://substrate.stackexchange.com/questions/2027/how-is-base-fee-calculated/2083#2083)
on [how fees are calculated](https://docs.substrate.io/build/tx-weights-fees/)

PR #498, opened 2022-10-10
fix(currency): update decimal places

enddynayn:
# Goal
Ensure all currency units of measurements
are updated and fees are calculated based
on new units.

Closes issue #489

### DIFF
--- a/runtime/common/src/constants.rs
+++ b/runtime/common/src/constants.rs
@@ -13,7 +13,9 @@ use frame_support::{
 };
 
 pub const FREQUENCY_ROCOCO_TOKEN: &str = "XRQCY";
+pub const FREQUENCY_LOCAL_TOKEN: &str = "UNIT";
 pub const FREQUENCY_TOKEN: &str = "FRQCY";
+pub const TOKEN_DECIMALS: u8 = 8;
 
 parameter_types! {
 	/// Clone + Debug + Eq  implementation for u32 types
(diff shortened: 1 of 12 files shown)


runtime/common/src/constants.rs
```diff
@@ -66,18 +66,23 @@ pub const HOURS: BlockNumber = MINUTES * 60;
...
-/// Generates an balance based on amount of items and bytes
-/// Items are each worth 20 Tokens
-/// Bytes each cost 1/1_000 of a Token
-pub const fn deposit(items: u32, bytes: u32) -> Balance {
-	items as Balance * 20 * UNIT + (bytes as Balance) * UNIT / 1_000
+pub mod currency {
+	use common_primitives::node::Balance;
+
+	/// The existential deposit. Set to 1/10 of the Connected Relay Chain.
+	pub const EXISTENTIAL_DEPOSIT: Balance = 100 * CENTS;
```
enddynayn, line 73: Do we want to keep it at CENTS or 1 FRQCY?

runtime/common/src/fee.rs
```diff
@@ -0,0 +1,80 @@
...
+		},
+		fee::Balance,
+		weights::ExtrinsicBaseWeight,
+	};
+
+	use frame_support::weights::WeightToFee as WeightToFeeT;
+
+	#[test]
+	// Test that the fee for `MAXIMUM_BLOCK_WEIGHT` of weigh has sane bounds.
+	fn full_block_fee_is_correct() {
```
shannonwells, line 56: Might want to add the note that you have to do a bounds check because the answer isn't entirely predictable based on hardware used, since that was a question for us.

shannonwells (approved):
looks good, nothing blocking, only optional comment suggestions.

runtime/common/src/constants.rs
wilwade: @enddynayn We do want `1 CENT` (or `10 MILLIUNIT` if you prefer).

runtime/common/src/constants.rs
```diff
@@ -66,18 +66,23 @@ pub const HOURS: BlockNumber = MINUTES * 60;
...
+
+	pub const UNITS: Balance = 100_000_000;
+	pub const DOLLARS: Balance = UNITS; // 100_000_000
+	pub const CENTS: Balance = DOLLARS / 100; // 1_000_000
+	pub const MILLICENTS: Balance = CENTS / 1_000; // 1_000
+
+	/// Generates an balance based on amount of items and bytes
+	/// Items are each worth 20 Tokens
```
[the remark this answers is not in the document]
wilwade: That was me. I used Tokens as Mainnet is FRQCY, Rococo is XRQCY, and Local is UNIT. Perhaps just `20 UNITs` is better?

runtime/common/src/constants.rs
```diff
@@ -66,18 +66,23 @@ pub const HOURS: BlockNumber = MINUTES * 60;
...
-/// Bytes each cost 1/1_000 of a Token
-pub const fn deposit(items: u32, bytes: u32) -> Balance {
-	items as Balance * 20 * UNIT + (bytes as Balance) * UNIT / 1_000
+pub mod currency {
+	use common_primitives::node::Balance;
+
+	/// The existential deposit. Set to 1/10 of the Connected Relay Chain.
+	pub const EXISTENTIAL_DEPOSIT: Balance = 100 * CENTS;
+
+	pub const UNITS: Balance = 100_000_000;
```
wilwade, line 75: Is it possible to base this off of a constant that produces the DECIMAL places? That way the everything ties back to decimal places?

wilwade (approved):
🎉

saraswatpuneet (approved):

PR #499, opened 2022-10-11
Changed MSA check delegations to return bool instead of Result

rlaferla:
Changed MSA check delegations to return `bool` instead of `Result`


pallets/msa/src/rpc/src/lib.rs
```diff
@@ -78,14 +78,12 @@ where
 		let provider = Provider(provider_msa_id);
 
 		Ok(delegator_msa_ids
-			.par_iter()
+			.iter() // TODO: Change back to par_iter()
```
saraswatpuneet, line 81: par_iter?

saraswatpuneet (approved):
One comment, lets change it to par_iter

pallets/msa/src/rpc/src/lib.rs
rlaferla: That's a future TODO.  We are intentionally reverting it to .iter().  @wilwade  will be making a more permanent fix which uses par_iter() again.

rlaferla:
The iterator needs to be updated to a parallel iterator but there's a borrow issue when making this change that needs to be addressed.

pallets/msa/src/rpc/src/lib.rs
```diff
@@ -78,14 +77,12 @@ where
 		let provider = Provider(provider_msa_id);
 
 		Ok(delegator_msa_ids
-			.par_iter()
+			.iter() // TODO: Change back to par_iter() which has borrow panic
```
enddynayn, line 80: Please do not add TODO here. Instead, create tickets for them.

PR #500, opened 2022-10-11
Update collator-selection config for mainnet 

### LINKED ISSUE #341: [Config] Collator Mainnet Config
Note @saraswatpuneet worked on the config for Rococo.

Update [pallet_collator_selection](https://github.com/LibertyDSNP/frequency/blob/main/runtime/frequency/src/lib.rs#L556) configurations with the following configurations:

## Config for Mainnet
- [ ] `desired_candidates` = default (aka 0)
- [ ] `candidacy_bond` = `100_000 * UNIT` // Likely to change in the future
    - Statemint does 5*16 DOT if I read it correctly
- [ ] `UpdateOrigin` = Root or FrequencyCouncil
    - Origin that can dictate updating parameters of this pallet.
- [ ] `PotId` = Set this to something that NEVER gets a balance i.e. No block rewards
    - Might be able to just use `()`
    - Account Identifier from which the internal Pot is generated.
- [ ] `MaxCandidates` = 0
- [ ] `MinCandidates` = 0
- [ ] `MaxInvulnerables` = 16
- [ ] `KickThreshold` = `6 * HOURS`
- [ ] `ValidatorId` = `<Self as frame_system::Config>::AccountId`
    - A stable ID for a validator.
- [ ] `ValidatorIdOf` = `pallet_collator_selection::IdentityCollator`
    - A conversion from account ID to validator ID
- [ ] `ValidatorRegistration` = `Session`
    - Validate a user is registered

## Open Questions
- Do `invulnerables` need a bond?
    - No, and we don't want to do one, although they should have a grant likely.
- Is a runtime governance configurable bond supported out of the box?
    - No `invulnerables` bond, so this is not needed for launch.
    - We might not care about this for launch if it is not already supported. Upgrade modification would be fine for day one if it is not easy.
- Can we use the [collator-selection pallet](https://github.com/paritytech/cumulus/blob/master/pallets/collator-selection/src/lib.rs)?
    - Potential Issue: Pot is built in and might not be able to be turned off (without forking)
    - Potential Issue: May want to have a time where we have approved collators, that still have a bond. (Might be possible via call filters?)

johnmela:
# Goal
The goal of this PR is update the collator-selection pallet config to our desired values for mainnet.

Closes #341

# Discussion
- After discussing with @wilwade, most mainnet values for the pallet were pulled from constants (except PotId), to allow mainnet to be more easily independently configurable from Rococo.
- Comments for config types were borrowed from the cumulus pallet config


runtime/frequency/src/lib.rs
```diff
@@ -633,16 +630,46 @@ impl pallet_aura::Config for Runtime {
...
+	// This does not take into account the invulnerables.
+	type MinCandidates = ConstU32<0>;
+
+	// Maximum number of invulnerables. This is enforced in code.
+	type MaxInvulnerables = ConstU32<16>;
+
+	// Will be kicked if block is not produced in threshold.
 	// should be a multiple of session or things will get inconsistent
-	type KickThreshold = SessionPeriod;
+	type KickThreshold = ConstU32<{ 6 * HOURS }>;
```
johnmela, line 660: I removed the type alias references from `common/src/constants.rs` in favor of specifying each `ConstU32` struct with its value. My thinking is that it is easier to glance at the pallet config and immediately know the values of each type. In the spirit of separate individually configurable runtimes, this seems to help?

saraswatpuneet: We might have separate sessionperiods ? 

Not sure the SessionPeriod is set using macro prod_or_local_env

johnmela: After some discussions with Puneet, I will kick back the values to runtime common constants and allow them to be independently configurable with a slightly modified `prod_or_local_env` macro.

saraswatpuneet (approved):

enddynayn (approved):

PR #502, opened 2022-10-12
make releases atomic

demisx:
# Goal
The goal of this PR is to make Frequency releases atomic (or really close to it). In other words, a release should be published only when all build steps have succeed. In order to achieve this, the `release.yml` workflow has been split into 2 parts:

1. **The build** - this is where all builds happen. All built artifacts are uploaded under `artifacts-${{github.run_id}}` name.
2. **The release** - this part of the release workflows kicks in only after all builds have finished and all artifacts have been uploaded successfully. It uses GitHub upload/download artifacts actions to pass built artifacts between them two. More reliable and appropriate than using cache action. 

Screen Shot 2022-10-13 at 1 25 41 PM

❗ Blocked by:
- https://github.com/LibertyDSNP/frequency/pull/507

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -13,8 +13,8 @@ env:
   RELEASE_VERSION: ${{github.ref_name}}
 
 jobs:
-  compile-project:
-    name: Compile ${{matrix.arch}} Binary for ${{matrix.network}}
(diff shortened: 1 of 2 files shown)


saraswatpuneet (approved):
:rocket:

Issue #259, opened 2022-07-27
[Security] Perform Dependency audits


shannonwells:
# Review of Dependencies

## Open issues
- All open security issues are resolved with: https://github.com/LibertyDSNP/frequency/issues/402

## pull requests

- [x] [Upgrades `serde_json`](https://github.com/LibertyDSNP/frequency/pull/514)
- [ ] Update inner dependencies using [`cargo audit fix`](https://github.com/LibertyDSNP/frequency/pull/510)

## Direct Dependencies

List from: `cargo tree --depth 1 --prefix "none" | sed -E 's/ \(.*\)//g' | sort | uniq`

### ORML

From https://github.com/open-web3-stack/open-runtime-module-library

```
orml-benchmarking v0.4.1-dev
orml-utilities v0.4.1-dev
orml-vesting v0.4.1-dev
```

### New External Packages

```
apache-avro v0.14.0
rayon v1.5.3
```

### External But Used By Substrate

```
log v0.4.17
serde v1.0.145
serde_json v1.0.85
smallvec v1.10.0
thiserror v1.0.37
tokio v1.21.2
```

### Polkadot/Substrate/Cumulus/Parity
```
cumulus-pallet-aura-ext v0.1.0
cumulus-pallet-parachain-system v0.1.0
cumulus-pallet-session-benchmarking v3.0.0
cumulus-primitives-core v0.1.0
cumulus-primitives-timestamp v0.1.0
cumulus-primitives-utility v0.1.0
frame-executive v4.0.0-dev
frame-support v4.0.0-dev
frame-system v4.0.0-dev
frame-system-rpc-runtime-api v4.0.0-dev
impl-serde v0.4.0
jsonrpsee v0.15.1
pallet-aura v4.0.0-dev
pallet-authorship v4.0.0-dev
pallet-balances v4.0.0-dev
pallet-collator-selection v3.0.0
pallet-collective v4.0.0-dev
pallet-democracy v4.0.0-dev
pallet-session v4.0.0-dev
pallet-sudo v4.0.0-dev
pallet-timestamp v4.0.0-dev
pallet-transaction-payment v4.0.0-dev
pallet-transaction-payment-rpc-runtime-api v4.0.0-dev
pallet-treasury v4.0.0-dev
pallet-utility v4.0.0-dev
parachain-info v0.1.0
parity-scale-codec v3.2.1
polkadot-parachain v0.9.29
polkadot-runtime-common v0.9.29
[...]

PR #510, opened 2022-10-13
cargo audit fix result

aramikm:
# Goal
To see what can be fixed using `cargo audit fix`

Related to #259


wilwade (approved):
🎉 Looks good. I checked on the questionable updates and they all look good as well. Nice the libc update has some nice updates for arm64 macs.

PR #511, opened 2022-10-14
Feat/nonce buckets against replay attacks

shannonwells:
# Goal
The goal of this PR is to implement and use a replay attack prevention technique in two extrinsics that take externally created signatures of externally provided data:
* `create_sponsored_account_with_delegation`
* `add_provider_to_msa`

The replay attack prevention provides an early exit for both of these extrinsics.

Related to #327 

# Discussion
Sets up a system of rotating virtual buckets to store signatures, using the bucket number and signature instead of a nonce.  Signatures must provide an expiration block. This is used to compute the bucket number it's expected to be in, and find out whether the signature has been recently submitted.

Signatures are rejected if they are found in the registry or the associated payload has an expiration too far in the future (>= MortalityBucketSize * NumberOfBuckets)

Every MortalityBucketSize * NumberOfBuckets blocks, the bucket rotation starts over, the old signatures are cleared, and the new signature is added.

To be able to check whether a virtual bucket needs to be cleared, bucket numbers are mapped to a mortality block.

Signature storage is really a [StorageDoubleMap](https://paritytech.github.io/substrate/master/frame_support/storage/trait.StorageDoubleMap.html) where K1 = bucket number and K2 is the signature.  The value stored is the expiration provided in `add_provider_payload` (currently unused).  The bucket expiration block, being the first key in the `StorageDoubleMap`  makes it extremely easy to "clear" a virtual bucket using [`drain_prefix`](https://paritytech.github.io/substrate/master/frame_support/storage/trait.IterableStorageDoubleMap.html#tymethod.drain_prefix) (via `IterableStorageDoubleMap` trait).

# Checklist
- [x] Refactor `EXPIRATION_BLOCK_VALIDITY_GAP`?
- [x] Set up genesis for buckets in mocks
- [x] New Errors, Constants in pallet
- [x] Values in mock
- [x] Values in frequency and frequency-rococo runtimes
- [x] StorageMap for the bucket --> mortality map
- [x] StorageDoubleMap for BucketSignature --> mortality
- [x] Helpers in pallet
- [x] Unit tests for helper fns
- [x] Replay tests passing
- [x] N/A Chain spec updated
- [x] Tests added
- [x] Benchmarks added
- [x] Weights updated


pallets/msa/src/lib.rs
```diff
@@ -275,6 +319,36 @@ pub mod pallet {
 		ProofHasExpired,
 		/// The submitted proof expiration block is too far in the future
 		ProofNotYetValid,
+		/// Attempted to add a signature when the signature is already in the registry
+		BucketKeyExists,
```
wilwade, line 323: I wonder if a better name is just `PreviouslyUsedSignature`. Or `SignatureReplayDetected` so it is clear to the end user what went wrong.

pallets/msa/src/lib.rs
```diff
@@ -275,6 +319,36 @@ pub mod pallet {
...
+		/// Attempted to add a signature with a mortality too far in the future
+		MortalityTooHigh,
+		/// Attempted to retrieve a bucket outside the bounds of storage
+		NoSuchBucket,
+	}
+
+	#[pallet::genesis_config]
+	pub struct GenesisConfig<T: Config> {
+		phantom: PhantomData<T>,
+	}
```
wilwade, line 333: Not sure why we need this, but I guess it might be for testing?

shannonwells: I'm not sure which "this" you're referring to - if you mean the phantom data, it's because we need to use `T::Blocknumber` when initializing MortalityBlockOf, and rust requires that you use the generic, and Substrate requires the GenesisConfig struct to be defined.

If you mean why we need to initialize `MortalityBlockOf`, It was a decision between using a genesis config to set the very first values, versus including the logic in the extrinsic.  IMO setting genesis is better rather than having logic that will literally only ever be hit once.

pallets/msa/src/lib.rs
shannonwells: Yea the name is left over from the prototype, I was thinking "SignatureAlreadySubmitted"

pallets/msa/src/lib.rs
```diff
@@ -175,6 +196,29 @@ pub mod pallet {
 	pub(super) type MsaInfoOf<T: Config> =
 		StorageMap<_, Twox64Concat, MessageSourceId, u8, ValueQuery>;
 
+	/// Storage type for mapping a "bucket" number to a mortality block
+	/// Used to know when to clear a virtual "bucket" of stored signatures
+	/// in PayloadSignatureRegistry
+	#[pallet::storage]
+	#[pallet::getter(fn get_mortality_block_of)]
+	pub type MortalityBlockOf<T: Config> =
+		StorageMap<_, Twox64Concat, T::BlockNumber, T::BlockNumber, OptionQuery>;
```
aramikm, line 205: I'm obviously lacking some context but it seems like if the `MortalityBucketSize` is fixed then you can always calculate MortalityBlockOf certain block. Unless something dynamically modifies it

pallets/msa/src/lib.rs
```diff
@@ -883,6 +976,90 @@ impl<T: Config> Pallet<T> {
...
+		// https://paritytech.github.io/substrate/master/frame_support/pallet_prelude/struct.StorageDoubleMap.html#method.drain_prefix
+		// we dont' need the value; it's just to finalize the drain.
+		<PayloadSignatureRegistry<T>>::drain_prefix(bucket_num).count();
+		Ok(())
+	}
+
+	// Virtual bucket mortalities are increasing multiples of MortalityBucketSize.
+	fn bucket_mortality_block(for_block: T::BlockNumber) -> T::BlockNumber {
+		let bucket_size: T::BlockNumber = T::MortalityBucketSize::get().into();
+		(for_block / bucket_size + T::BlockNumber::from(1u32)) * bucket_size
```
aramikm, line 1045:
```suggestion
		(for_block / bucket_size + T::BlockNumber::one()) * bucket_size
```

pallets/msa/src/lib.rs
```diff
@@ -883,6 +976,90 @@ impl<T: Config> Pallet<T> {
...
+		})
+	}
+
+	// delete signatures with the given bucket number since they have existed past the mortality
+	// block limit.
+	fn delete_signatures_for_bucket(bucket_num: T::BlockNumber) -> Result<(), DispatchError> {
+		// use drain_prefix
+		// https://paritytech.github.io/substrate/master/frame_support/pallet_prelude/struct.StorageDoubleMap.html#method.drain_prefix
+		// we dont' need the value; it's just to finalize the drain.
+		<PayloadSignatureRegistry<T>>::drain_prefix(bucket_num).count();
```
aramikm, line 1038: I'm not sure how `drain_prefix` is implemented but if it uses the underlying logic as `remove_prefix` it should be fine, if not maybe switching with `remove_prefix` would be more efficient  https://substrate.recipes/double.html

pallets/msa/src/lib.rs
```diff
@@ -883,6 +976,90 @@ impl<T: Config> Pallet<T> {
...
+			Err(Error::<T>::MortalityTooHigh.into())
+		} else {
+			let bucket_num = Self::bucket_for(mortality_block.into());
+			<PayloadSignatureRegistry<T>>::try_mutate(
+				bucket_num,
+				signature,
+				|maybe_mortality_block| -> DispatchResult {
+					ensure!(maybe_mortality_block.is_none(), Error::<T>::BucketKeyExists);
+
+					Self::reset_virtual_bucket_if_needed(mortality_block, bucket_num)?;
```
aramikm, line 997: It does seem like you can also do this `on_initialize`. Having this process here will cause some transactions to have higher weight which doesn't seem to be fair or consistent

pallets/msa/src/nonce_bucket.rs
```diff
@@ -0,0 +1,294 @@
...
+		let mut bucket_num: usize = bucket_for(current_block) as usize;
+		for _i in 0..NUM_BUCKETS {
+			let bucket = buckets.get_mut(bucket_num).unwrap();
+			bucket.mortality_block = bucket_mortality_block;
+			bucket_mortality_block += MORTALITY_SIZE;
+			if bucket_num == (NUM_BUCKETS - 1) as usize {
+				bucket_num = 0;
+			} else {
+				bucket_num += 1;
+			}
```
aramikm, line 97:
```suggestion
			bucket_num = (bucket_num + 1) % NUM_BUCKETS as usize;
```

pallets/msa/src/lib.rs
```diff
@@ -175,6 +196,29 @@ pub mod pallet {
...
+
+	/// PayloadSignatureRegistry is used to prevent replay attacks for extrinsics
+	/// that take an externally-signed payload.
+	/// For this to work, the payload must include a mortality block number, which
+	/// is used in lieu of a monotonically increasing nonce.
+	#[pallet::storage]
+	pub(super) type PayloadSignatureRegistry<T: Config> = StorageDoubleMap<
+		_,
+		Twox64Concat,
+		T::BlockNumber, // Bucket number. Stored as BlockNumber because I'm done arguing with rust about it.
```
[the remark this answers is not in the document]
shannonwells: that was part of it, converting back and forth. I was attempting to create limits using data types and gave up

pallets/msa/src/lib.rs
shannonwells: That is basically a registry for each bucket's current mortality block.  It gets changed every MortalityBucketSize * NumberOfBuckets (or longer if the extrinsics that check it are not called).

pallets/msa/src/nonce_bucket.rs
shannonwells: This was fixed in the cleanup stage.

pallets/msa/src/lib.rs
shannonwells: I tried that already, unfortunately:
```
 996 |         (for_block / bucket_size + T::BlockNumber::one() * bucket_size
      |                                                    ^^^ function or associated item not found in `<T as frame_system::Config>::BlockNumber`
```

pallets/msa/src/lib.rs
shannonwells: Good point

pallets/msa/src/lib.rs
shannonwells: thanks I'll look at that some more.

common/primitives/src/msa.rs
```diff
@@ -8,10 +8,6 @@ use sp_std::prelude::Vec;
 
 pub use crate::{ds::OrderedSetExt, schema::SchemaId};
 
-/// The gap between the current block and a future expiring block allowed when validating signature proofs.
-/// 150 blocks at 6 seconds per block would equate to a 15 minute gap.
-pub const EXPIRATION_BLOCK_VALIDITY_GAP: u32 = 150;
```
shannonwells, line 13: @dopatraman  confirmed this was a placeholder for when this change set was done

pallets/msa/src/lib.rs
```diff
@@ -659,33 +727,6 @@ impl<T: Config> Pallet<T> {
...
-	/// # Arguments
-	/// * `expiration` - A block number that which validity or expiration is being checked.
-	///
-	/// # Returns
-	/// * [`DispatchResult`]
-	///
-	/// # Errors
-	/// * [Error::ProofHasExpired] - If the current block is less than the `expired` bock number set in `AddKeyData`.
-	/// * [Error::ProofNotYetValid] - If the `expired` bock number set in `AddKeyData` is greater than the current block number plus EXPIRATION_BLOCK_VALIDITY_GAP.
-	pub fn ensure_block_is_valid(expiration: BlockNumber) -> DispatchResult {
```
shannonwells, line 673: No longer needed with the signature/proof bucket registry

pallets/msa/src/lib.rs
wilwade: @shannonwells I wonder if instead you can operate with u32 all the way to the end and then do an into(). Instead of doing the math on the `BlockNumber` type. Just a thought.

pallets/msa/src/lib.rs
shannonwells: Looks like `remove_prefix` is deprecated. They tell you to use `clear_prefix` instead, which I had overlooked, so I'll use that.

pallets/msa/src/lib.rs
shannonwells: Not without other type gymnastics, no.  And we also actually want some of these to be BlockNumbers.  I was only willing to spend so much time arguing with rust.

pallets/msa/src/lib.rs
```diff
@@ -955,6 +995,94 @@ impl<T: Config> Pallet<T> {
...
+			let bucket_mortality_block = Self::bucket_mortality_block(mortality_block);
+
+			match maybe_mortality_block {
+				None => Err(Error::<T>::NoSuchBucket.into()),
+				Some(mortality_block_for_bucket) => {
+					if mortality_block > *mortality_block_for_bucket {
+						Self::delete_signatures_for_bucket(bucket_num)?;
+						*maybe_mortality_block = Some(bucket_mortality_block);
+					} else if *mortality_block_for_bucket == T::BlockNumber::default() {
+						*maybe_mortality_block = Some(bucket_mortality_block);
```
saraswatpuneet, line 1047: maybe_mortality_block can be out of if ,they look the same Some(bucket_mortality_block) only a delete step in taken when conditions meet

pallets/msa/src/lib.rs
saraswatpuneet: Good point, if we mutate Blocknumber to do an add , saturating_add, not sure if its a type trait

saraswatpuneet: read the while code the whole output can be T::BlockNumber::from({/// some value u32}) or an into might work too

pallets/msa/src/lib.rs
saraswatpuneet: @shannonwells  I have noticed when it is the first time write in MortalityBlockOf while registering msa or adding provider etc. it will always have ```None``` for ```maybe_mortality_block```  for a given ```bucket_num```

pallets/msa/src/lib.rs
```diff
@@ -479,7 +547,7 @@ pub mod pallet {
 		/// - Returns [`NoKeyExists`](Error::NoKeyExists) if the MSA id for the account in `add_key_payload` does not exist.
 		/// - Returns ['NotMsaOwner'](Error::NotMsaOwner) if Origin's MSA is not the same as 'add_key_payload` MSA. Essentially you can only add a key to your own MSA.
 		/// - Returns ['ProofHasExpired'](Error::ProofHasExpired) if the current block is less than the `expired` bock number set in `AddKeyData`.
-		/// - Returns ['ProofNotYetValid'](Error::ProofNotYetValid) if the `expired` bock number set in `AddKeyData` is greater than the current block number plus EXPIRATION_BLOCK_VALIDITY_GAP.
+		/// - Returns ['ProofNotYetValid'](Error::ProofNotYetValid) if the `expired` block number set in `AddKeyData` is greater than the current block number plus mortality_block_limit().
 		#[pallet::weight(T::WeightInfo::add_key_to_msa())]
```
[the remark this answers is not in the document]
shannonwells: it's the lifetime of a captured signature to stay in the registry to prevent replay

pallets/msa/src/lib.rs
```diff
@@ -957,6 +972,74 @@ impl<T: Config> Pallet<T> {
...
+		T::WeightInfo::on_initialize(multi_removal_result.unique)
+	}
+
+	// The furthest in the future a mortality_block value is allowed
+	// to be for current_block
+	// This is calculated to be past the risk of a replay attack
+	fn mortality_block_limit(current_block: T::BlockNumber) -> T::BlockNumber {
+		let bucket_size: T::BlockNumber = T::MortalityBucketSize::get().into();
+		let num_buckets: T::BlockNumber = T::NumberOfBuckets::get().into();
+		current_block + (bucket_size * (num_buckets - T::BlockNumber::one()))
```
aramikm, line 1034: Just want to mentioned that this seems to be based on assumption that always `num_buckets > 1`

pallets/msa/src/weights.rs
```diff
@@ -141,14 +143,20 @@ impl<T: frame_system::Config> WeightInfo for SubstrateWeight<T> {
 			.saturating_add(T::DbWeight::get().reads(2 as u64))
 			.saturating_add(T::DbWeight::get().writes(1 as u64))
 	}
+
+	fn on_initialize(s: u32) -> Weight {
+		Weight::from_ref_time(15_000_000 as u64)
+			.saturating_add(RocksDbWeight::get().reads(s as u64))
```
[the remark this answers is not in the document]
saraswatpuneet: i think this is generated because benchmark.rs does not have benchmark of on_initialize

pallets/msa/src/lib.rs
shannonwells: it is always > 1

pallets/msa/src/lib.rs
```diff
@@ -957,6 +972,74 @@ impl<T: Config> Pallet<T> {
...
+
+	/// Adds a signature to the PayloadSignatureRegistry based on a virtual "bucket" grouping.
+	/// Check that mortality_block is within bounds. If so, proceed and add the new entry.
+	/// Raises `SignatureAlreadySubmitted` if the bucket-signature double key exists in the
+	/// registry.
+	pub fn register_signature(
+		signature: &MultiSignature,
+		mortality_block: T::BlockNumber,
+	) -> DispatchResult {
+		let current_block = frame_system::Pallet::<T>::block_number();
```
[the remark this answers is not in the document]
shannonwells: It was before, but that means that every function that uses it has to get the current block first, so it made more sense to get it in one place.

pallets/msa/src/lib.rs
```diff
@@ -957,6 +972,74 @@ impl<T: Config> Pallet<T> {
...
+
+		let max_lifetime = Self::mortality_block_limit(current_block);
+		if max_lifetime < mortality_block {
```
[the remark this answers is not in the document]
shannonwells: Thanks, yes I should fix that, they were both < and I only changed it in the first spot.

pallets/msa/src/weights.rs
```diff
@@ -218,4 +226,10 @@ impl WeightInfo for () {
 			.saturating_add(RocksDbWeight::get().reads(2 as u64))
 			.saturating_add(RocksDbWeight::get().writes(1 as u64))
 	}
+
+	fn on_initialize(s: u32) -> Weight {
+		Weight::from_ref_time(15_000_000 as u64)
+			.saturating_add(RocksDbWeight::get().reads(s as u64))
```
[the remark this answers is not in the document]
shannonwells: it's a generated file and was just a placeholder to let me run some tests

common/primitives/src/msa.rs
shannonwells: They can call `mortality_block_limit(current_block)`

pallets/msa/src/lib.rs
shannonwells: it's already using `.one()`, I just needed to use the `One` trait at the top.

pallets/msa/src/lib.rs
```diff
@@ -957,6 +972,75 @@ impl<T: Config> Pallet<T> {
...
+		let bucket_size: T::BlockNumber = T::MortalityBucketSize::get().into();
+
+		// Now that we're in a "new bucket block set"
+		if current_block % bucket_size != T::BlockNumber::zero() {
+			return T::WeightInfo::on_initialize(0 as u32)
+		}
+
+		// Clear the previous bucket block set
+		let bucket_num = Self::bucket_for(current_block - T::BlockNumber::one());
+		let multi_removal_result = <PayloadSignatureRegistry<T>>::clear_prefix(
```
enddynayn, line 1020: multi_removal_result is this number of deleted items?

shannonwells: Yes, see [MultiRemovalResults](https://paritytech.github.io/substrate/master/sp_io/struct.MultiRemovalResults.html). We might want to include loops in that somehow, because it reflects the number of reads, however I would want to do some more testing first.  This is a good baseline for now though.

enddynayn: For my understanding of `clear_prefix `, what is the cost of deleting 1 item vs 1000? Is this an O(1) operation?

enddynayn (approved):

enddynayn:
Fantastic! 👍 You made something so difficult look so easy.

enddynayn (approved):
Fantastic! 👍 You made something so difficult look so easy.

PR #515, opened 2022-10-14
Ensure production binary is built before running benchmarks

rlaferla:
# Goal
The goal of this PR is to ensure production binary is built before running benchmarks

Closes N/A


saraswatpuneet (approved):

saraswatpuneet:
@wilwade  this is used for dev testing so they don't have to run all benchmarks if changing only one pallet or the other 

Jenkins run full suite using make benchmarks

wilwade (approved):

PR #517, opened 2022-10-14
44 as a user i should be able to retire my msa

### LINKED ISSUE #44: As a user, I should be able to retire my MSA

rlaferla:
Closes #44

- [ x] Tests added
- [x ] Benchmarks added
- [ x] Weights updated


pallets/msa/src/lib.rs
```diff
@@ -556,6 +571,51 @@ pub mod pallet {
...
+		/// - Returns [`NoKeyExists`](Error::NoKeyExists) if `delegator` does not have an MSA key.
+		/// - Returns [`MoreThanOneKeyExists`](Error::MoreThanOneKeyExists) if the MSA has more than one account key.
+		/// - Returns [`RegisteredProviderCannotBeRetired`](Error::RegisteredProviderCannotBeRetired) if the MSA id is a registered provider
+
+		#[pallet::weight((T::WeightInfo::retire_msa(), DispatchClass::Normal, Pays::No))]
+		pub fn retire_msa(origin: OriginFor<T>) -> DispatchResult {
+			// Check and get the account id from the origin
+			let who = ensure_signed(origin)?;
+
+			// Get the MSA id of the origin
```
enddynayn, line 591: please remove comments

pallets/msa/src/benchmarking.rs
```diff
@@ -115,6 +115,22 @@ benchmarks! {
 
 	}: _(RawOrigin::Signed(caller), key)
 
+	retire_msa {
+		let caller: T::AccountId = whitelisted_caller();
+		// let (caller_key_pair, _) = sr25519::Pair::generate();
```
enddynayn, line 120: Some left over stuff.

pallets/msa/src/lib.rs
```diff
@@ -556,6 +571,51 @@ pub mod pallet {
...
+		/// ### Events
+		/// - Deposits [`MsaRetired`](Event::MsaRetired) when MSA is retired
+		///
+		/// ### Errors
+		///
+		/// - Returns [`NoKeyExists`](Error::NoKeyExists) if `delegator` does not have an MSA key.
+		/// - Returns [`MoreThanOneKeyExists`](Error::MoreThanOneKeyExists) if the MSA has more than one account key.
+		/// - Returns [`RegisteredProviderCannotBeRetired`](Error::RegisteredProviderCannotBeRetired) if the MSA id is a registered provider
+
+		#[pallet::weight((T::WeightInfo::retire_msa(), DispatchClass::Normal, Pays::No))]
```
enddynayn, line 586: Do we need to update or create a SignedExtension for this? If so perhaps make this a  Pays::Yes until we add it.

pallets/msa/src/lib.rs
```diff
@@ -556,6 +571,51 @@ pub mod pallet {
...
+		pub fn retire_msa(origin: OriginFor<T>) -> DispatchResult {
+			// Check and get the account id from the origin
+			let who = ensure_signed(origin)?;
+
+			// Get the MSA id of the origin
+			let msa_id = Self::get_owner_of(&who).ok_or(Error::<T>::NoKeyExists)?;
```
enddynayn, line 592: I think there is a function for this `ensure_msa_owner`.

pallets/msa/src/lib.rs
```diff
@@ -556,6 +571,51 @@ pub mod pallet {
...
+				!Self::is_registered_provider(msa_id),
+				Error::<T>::RegisteredProviderCannotBeRetired,
+			);
+
+			// Dispatches error "MoreThanOneKeyExists" if the MSA has more than one account key.
+			let key_count = Self::get_msa_key_count(msa_id);
+			ensure!(key_count == 1, Error::<T>::MoreThanOneKeyExists);
+
+			// Remove delegator from all delegator<->provider delegations
+			Self::remove_delegator(delegator)?;
```
enddynayn, line 609: What are your thoughts on maybe calling it `delete_delegator`?

pallets/msa/src/lib.rs
```diff
@@ -556,6 +571,51 @@ pub mod pallet {
...
+
+			// Delete the last and only account key and deposit the "KeyRemoved" event
+			Self::delete_key_for_msa(msa_id, &who)?;
+			Self::deposit_event(Event::KeyRemoved { key: who });
```
enddynayn, line 613: Maybe we do not have to emit this event. Emitting the` MsaRetired` event tells me all the keys have been emitted. I think doing so saves one less DB writes.

enddynayn (commented):
Looking fabulous! 💅

pallets/msa/src/lib.rs
rlaferla: I will make it Pays::Yes

pallets/msa/src/lib.rs
rlaferla: I think **remove** is appropriate here because there are other delegations in the map but it could work either way.   It's an internal function so it doesn't matter so much.

pallets/msa/src/lib.rs
rlaferla: We need for the off-chain worker.

enddynayn: I think if something is not being used than no reason to add it.

pallets/msa/src/lib.rs
```diff
@@ -556,6 +571,51 @@ pub mod pallet {
...
+		/// ### Events
+		/// - Deposits [`MsaRetired`](Event::MsaRetired) when MSA is retired
+		///
+		/// ### Errors
+		///
+		/// - Returns [`NoKeyExists`](Error::NoKeyExists) if `delegator` does not have an MSA key.
+		/// - Returns [`MoreThanOneKeyExists`](Error::MoreThanOneKeyExists) if the MSA has more than one account key.
+		/// - Returns [`RegisteredProviderCannotBeRetired`](Error::RegisteredProviderCannotBeRetired) if the MSA id is a registered provider
+
+		#[pallet::weight((T::WeightInfo::retire_msa(), DispatchClass::Normal, Pays::Yes))]
```
enddynayn, line 586: these are the default settings. 
You can remove these 
```
 DispatchClass::Normal, Pays::Yes
 ```

File: pallets/msa/src/lib.rs [hunk 8]
```diff
@@ -556,6 +571,51 @@ pub mod pallet {
...
+				!Self::is_registered_provider(msa_id),
+				Error::<T>::RegisteredProviderCannotBeRetired,
+			);
+
+			// Dispatches error "MoreThanOneKeyExists" if the MSA has more than one account key.
+			let key_count = Self::get_msa_key_count(msa_id);
+			ensure!(key_count == 1, Error::<T>::MoreThanOneKeyExists);
+
+			// Remove delegator from all delegator<->provider delegations
+			Self::remove_delegator(delegator)?;
```
enddynayn, line 609: I wonder if we can also remove this event?

pallets/msa/src/lib.rs
```diff
@@ -556,6 +571,51 @@ pub mod pallet {
 
 			Ok(())
 		}
+
+		/// Retire a MSA
+		///
```
enddynayn, line 576: maybe add a longer description and remove some of the comments.

pallets/msa/src/lib.rs
```diff
@@ -556,6 +571,51 @@ pub mod pallet {
...
+		/// ### Errors
+		///
+		/// - Returns [`NoKeyExists`](Error::NoKeyExists) if `delegator` does not have an MSA key.
+		/// - Returns [`MoreThanOneKeyExists`](Error::MoreThanOneKeyExists) if the MSA has more than one account key.
+		/// - Returns [`RegisteredProviderCannotBeRetired`](Error::RegisteredProviderCannotBeRetired) if the MSA id is a registered provider
+
+		#[pallet::weight((T::WeightInfo::retire_msa(), DispatchClass::Normal, Pays::Yes))]
+		pub fn retire_msa(origin: OriginFor<T>) -> DispatchResult {
+			// Check and get the account id from the origin
+			let who = ensure_signed(origin)?;
```
enddynayn, line 589: this should be probably be called key.

pallets/msa/src/lib.rs
```diff
@@ -556,6 +571,51 @@ pub mod pallet {
...
+		///
+		/// ### Errors
+		///
+		/// - Returns [`NoKeyExists`](Error::NoKeyExists) if `delegator` does not have an MSA key.
+		/// - Returns [`MoreThanOneKeyExists`](Error::MoreThanOneKeyExists) if the MSA has more than one account key.
+		/// - Returns [`RegisteredProviderCannotBeRetired`](Error::RegisteredProviderCannotBeRetired) if the MSA id is a registered provider
+
+		#[pallet::weight((T::WeightInfo::retire_msa(), DispatchClass::Normal, Pays::Yes))]
+		pub fn retire_msa(origin: OriginFor<T>) -> DispatchResult {
+			// Check and get the account id from the origin
```
enddynayn, line 588: this comment is not completely accurate. I can see why this is tricky. This actually represents an account key in our system and not an account id.

pallets/msa/src/lib.rs
enddynayn: We are doing an additional db read here.
I think its already taken care of here : `ensure_msa_owner`.

enddynayn: I think you accidentally missed this comment.

pallets/msa/src/lib.rs
```diff
@@ -556,6 +571,51 @@ pub mod pallet {
...
+			ensure!(key_count == 1, Error::<T>::MoreThanOneKeyExists);
+
+			// Remove delegator from all delegator<->provider delegations
+			Self::remove_delegator(delegator)?;
+
+			// Delete the last and only account key and deposit the "KeyRemoved" event
+			Self::delete_key_for_msa(msa_id, &who)?;
+			Self::deposit_event(Event::KeyRemoved { key: who });
+
+			// Deposit the "MsaRetired" event
```
enddynayn, line 615: this comment does not add value because it is clear from that we are emitting an MSARetired event.

PR #518, opened 2022-10-14
Chore/account provider mocks

### LINKED ISSUE #452: Better or no AccountProvider mocks for  in messages pallet tests
# Problem 
The mock for AccountProvider currently used by the messages pallet is not working for us any more because the test needs are too diverse for a one-size-fits-all AccountProvider mock.  This can cause false test passes/failures.

# Solution?
Consider possibilities, evaluate and implement one (not all of these may actually be viable):
1. Get rid of the mock altogether and simply use MsaPallet in mocks.rs for messages testing.  This will require that you use and configure the pallet.  This may cause issues with benchmarks.
2. Allow the mock to be configured on a per test basis, for example, by being able to provide test function(s) for the AccountProvider function(s) needed.
3. Require each test to set up its own test_ext configuration with an AccountProvider mock.
4. Some other solution

# Acceptance Criteria
- [ ] all tests pass
- [ ] if we don't get rid of mocks then somehow there is a way to mock AccountProvider in each test suite or each test case that does not bloat the mock or render the test useless
- [ ] ?

dopatraman:
# Goal
The goal of this PR is to divide the responsibilities of the `AccountProvider` trait into smaller, more manageable components.

Closes #452 

# Discussion
This PR splits `AccountProvider` into 5 traits:
- `MsaLookup`
- `MsaValidator`
- `ProviderLookup`
- `DelegationValidator`
- `SchemaGrantValidator`

The traits are then bundled into 3 structs:
- `MsaInfoProvider` (`MsaLookup` + `MsaValidator`)
- `DelegationInfoProvider` (`ProviderLookup` + `DelegationValidator`)
- `SchemaGrantValidator`

This bundling was done to avoid having 5 additional members on any given implementation of the MSA pallet config.

An alternative is to have a single struct implement all 5 traits (`AccountProvider: MsaLookup + MsaValidator + ProviderLookup + DelegationValidator + SchemaGrantValidator`). I don't think this is the right way to go, but I'm including it here as an option for completeness and because it would make implementation convenient.

# Checklist
- [x] Tests added


pallets/messages/src/lib.rs
```diff
@@ -76,8 +82,14 @@ pub mod pallet {
...
-		/// A type that will supply account related information.
-		type AccountProvider: AccountProvider<AccountId = Self::AccountId>;
+		/// A type that will supply MSA related information
+		type MsaInfoProvider: MsaLookup + MsaValidator<AccountId = Self::AccountId>;
+
+		/// A type that will supply delegation related information
+		type DelegationInfoProvider: ProviderLookup + DelegationValidator;
+
+		/// A type that will validate schema grants
+		type SchemaGrantValidator: SchemaGrantValidator;
```
dopatraman, line 92: In theory we could combine all 5 into one config member:
```
`AccountProvider: MsaLookup + MsaValidator + ProviderLookup + DelegationValidator + SchemaGrantValidator
```

But I think it would be unwieldy.

shannonwells (approved):
I really like this split. I was tempted to suggest that AccountId and BlockNumber associated types have additional constraint traits like AtLeast32BitUnsigned, but I don't think it's needed.

I think at least one more review is warranted though

enddynayn (approved):

enddynayn:
Fabulous! Some make me think we should combine but let's give this a shot. ⛹️‍♂️

PR #522, opened 2022-10-17
44 as a user i should be able to retire my msa

rlaferla:
As per a technical discussion with Enddy, the MsaRetired event already implies a KeyRemoved event so we can remove the latter.


pallets/msa/src/lib.rs
```diff
@@ -610,7 +610,6 @@ pub mod pallet {
 
 			// Delete the last and only account key and deposit the "KeyRemoved" event
 			Self::delete_key_for_msa(msa_id, &who)?;
-			Self::deposit_event(Event::KeyRemoved { key: who });
```
enddynayn, line 613: If this passes, it tells me that maybe we are not testing that we are emitting events. Do we need to add test?

rlaferla:
```
		// Check if MsaRetired event was dispatched.
		System::assert_last_event(Event::MsaRetired { msa_id }.into());
```

wilwade (approved):

PR #523, opened 2022-10-18
```add_key_to_msa```  accept owner signatures to add new key

### LINKED ISSUE #478: `add_key_to_msa` Figure out cost shifting
Blocked by #327 as if the signature solution doesn't work, this might need a different solution as well.

## Acceptance Criteria
- [x] As a user who wants to add a key to my MSA I want anyone to be able to execute that request on chain, even without a delegation
- [x] As a wallet who wants to provide basic key management services to the user. I want to be able to relay an add key to msa request to the chain.
- [x] As a user who's key is going to be added to the MSA, I want to be required to "approve" the adding of the key to that specific MSA so that it cannot be added to a different MSA

## Notes
- Three signatures, some of these might be the same actor
    - Payee (aka `Origin`)
    - Current MSA key holder
    - New key holder for the MSA
- The Payee only needs to pay for the transaction
- The current and new key holders each can sign the same data
- Right now we assume that the origin and the current MSA owner are the same actor. We want to make them able to be separate.
- https://github.com/LibertyDSNP/frequency/blob/main/pallets/msa/src/lib.rs#L476
  - Expected Todos
  - Add a new key and proof for the current msa owner instead of using the origin account

```rust
pub fn add_key_to_msa(
	origin: OriginFor<T>,
	msa_owner_key: T::AccountId,
	msa_owner_proof: MultiSignature,
	new_key: T::AccountId,
	new_proof: MultiSignature,
	add_key_payload: AddKeyData,
) -> DispatchResult {
```

saraswatpuneet:
# Goal

The goal of this PR is to extend ```add_key_to_msa``` to accept msa owner key and signature proof along with new key with signature proof to ownership of new account. This in turn will 

* Providers can add keys on behalf of users : Example use case is to have an account to handle offchain data
* Independent frequency users can also leverage it to add new accounts to their msa without having to be delegated: Blockchain pros or data pipelines branching out etc. could be some use cases

Closes #478 

# Checklist
- [x] N/A Chain spec updated
- [x] Tests added
- [x] Benchmarks updated
- [x] Weights updated


pallets/msa/src/tests.rs
```diff
@@ -1968,40 +1999,6 @@ pub fn delegation_expired_long_back() {
 	})
 }
 
-#[test]
-pub fn ensure_all_schema_ids_are_valid_errors() {
```
wilwade, line 1972: Were these tests supposed to be dropped?

pallets/msa/src/tests.rs
```diff
@@ -2021,6 +2018,31 @@ fn register_signature_and_validate(
 	assert_eq!(Some(mortality_block as u64), actual);
 }
 
+#[test]
+// for illustration purposes
+pub fn bucket_for() {
```
wilwade, line 2023: I don't think this should have moved. Rebase issue?

pallets/msa/src/tests.rs
```diff
@@ -1,25 +1,26 @@
+use frame_support::{
```
wilwade, line 1: We should have an explicit replay rejection test for `add_key_to_msa`

pallets/msa/src/mock.rs
```diff
@@ -98,6 +98,8 @@ impl sp_std::fmt::Debug for MaxSchemaGrantsPerDelegation {
 	}
 }
 
+pub const NUMBER_OF_BUCKETS: u32 = 2; // so this can be used in genesis
```
wilwade, line 101: Not sure we need any of the changes to the mock. Rebase issue?

pallets/msa/src/tests.rs
saraswatpuneet: Yes plenty rebase, let me move them around

pallets/msa/src/lib.rs
```diff
@@ -533,23 +538,31 @@ pub mod pallet {
...
 		) -> DispatchResult {
-			let who = ensure_signed(origin)?;
+			let _ = ensure_signed(origin)?;
 
-			Self::verify_signature(proof.clone(), key.clone(), add_key_payload.encode())
-				.map_err(|_| Error::<T>::AddKeySignatureVerificationFailed)?;
+			Self::verify_signature(
+				vec![
+					(msa_owner_proof.clone(), msa_owner_key.clone()),
+					(new_proof, new_key.clone()),
```
enddynayn, line 552: what new_proof?

pallets/msa/src/lib.rs
```diff
@@ -533,23 +538,31 @@ pub mod pallet {
 		#[pallet::weight(T::WeightInfo::add_key_to_msa())]
 		pub fn add_key_to_msa(
 			origin: OriginFor<T>,
-			key: T::AccountId,
-			proof: MultiSignature,
+			msa_owner_key: T::AccountId,
+			msa_owner_proof: MultiSignature,
+			new_key: T::AccountId,
+			new_proof: MultiSignature,
 			add_key_payload: AddKeyData,
```
enddynayn, line 545: I think I just noticed that we could remove the nonce field from AddKeyData. Maybe on a different task.

pallets/msa/src/lib.rs
```diff
@@ -770,15 +783,14 @@ impl<T: Config> Pallet<T> {
...
-		signer: T::AccountId,
+		signer_signature: Vec<(MultiSignature, T::AccountId)>,
 		payload: Vec<u8>,
 	) -> DispatchResult {
-		let key = T::ConvertIntoAccountId32::convert(signer);
-		let wrapped_payload = wrap_binary_data(payload);
-
-		ensure!(signature.verify(&wrapped_payload[..], &key), Error::<T>::InvalidSignature);
-
+		let wrapped_payload = wrap_binary_data(payload.clone());
```
enddynayn, line 789: are we validating multiple signatures?

enddynayn (commented):
Looking great! 👍

pallets/msa/src/lib.rs
saraswatpuneet: Yep another task

pallets/msa/src/tests.rs
saraswatpuneet: Rebase issue, brought PR back to normal

pallets/msa/src/lib.rs
```diff
@@ -533,23 +538,28 @@ pub mod pallet {
...
-			let who = ensure_signed(origin)?;
+			let _ = ensure_signed(origin)?;
 
-			Self::verify_signature(proof.clone(), key.clone(), add_key_payload.encode())
-				.map_err(|_| Error::<T>::AddKeySignatureVerificationFailed)?;
+			Self::register_signature(&msa_owner_proof, add_key_payload.expiration.into())?;
 
-			Self::register_signature(&proof, add_key_payload.expiration.into())?;
+			Self::verify_signature(
+				vec![(msa_owner_proof, msa_owner_key.clone()), (new_proof, new_key.clone())],
```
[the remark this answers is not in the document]
saraswatpuneet: Good point, fixed in other places where it was getting cloned and copied

pallets/msa/src/lib.rs
saraswatpuneet: Renamed , also reverted to verify_signature and call it twice in add_key_to_msa

pallets/msa/src/lib.rs
```diff
@@ -467,12 +463,8 @@ pub mod pallet {
...
 
 			// delegator must have signed the payload.
-			Self::verify_signature(
-				proof.clone(),
-				delegator_key.clone(),
-				add_provider_payload.encode(),
-			)
-			.map_err(|_| Error::<T>::AddProviderSignatureVerificationFailed)?;
+			Self::verify_signature(&proof, &delegator_key, add_provider_payload.encode())
+				.map_err(|_| Error::<T>::AddProviderSignatureVerificationFailed)?;
```
saraswatpuneet, line 467: borrow than moving

pallets/msa/src/tests.rs
shannonwells: I didn't think too hard about this but should we register both signatures?  Is there an attack where I replay one signature with a different second signature?

shannonwells (approved):
I don't see anything glaring, I just had the one question so if there's an attack vector from not registering both `msa_owner_proof` and `new_key_owner_proof`, naturally it needs a replay test and a fix.

enddynayn (approved):
Nice work! ⛹️‍♂️

pallets/msa/src/mock.rs
shannonwells: Yes, this one should go away

PR #525, opened 2022-10-18
Fixed docker cli arguments to map correct port for http

rlaferla:


johnmela (approved):

README.md
```diff
@@ -142,7 +142,7 @@ make start
 ### In Docker Container
 
 ```sh
-docker run --rm -p 9944:9944 -p 9933:9944 -p 30333:30333 frequencychain/instant-seal-node
+docker run --rm -p 9944:9944 -p 9933:9933 -p 30333:30333 frequencychain/instant-seal-node
 ```
```
saraswatpuneet, line 146: So the outside port was the same the for both I think it was done intentionally? @wilwade 
And inside ports are different so it should work though

rlaferla: It was not working.  It broke `make local-block` for example which uses `cURL` to perform a `POST` request to the node.

Issue #508, opened 2022-10-13
External Naming Consistency Check


wilwade:
We want to have consistent naming, and external naming in particular.

## Outside Naming Assistance
- https://rust-lang.github.io/api-guidelines/naming.html
- `rust fmt`

## Hard Rules
- Careful when adding code that is indexed. Should ALWAYS be the last one. Do NOT mess with prior indexes
  - Pallet Indexes
  - Events
  - Errors
  - Extrinsics
  - Argument order
-  (after this issue is closed) Changes in name or type to an external interface MUST occur in a two step process
  - 1. Create a new one and deprecate the old
  - 2. ~6 months later, remove the deprecated one. 

## Rules of Thumb
- Avoid single words
- Governance managed variables should be prefaced by `governance`
- State `StorageMap` Naming
  - [map key]To[map value]
- State `DoubleMap` Naming
  - [map key]And[map key 2]To[map value]

## General Naming Question
- `key` vs `account`?
  - Current
    - `key`: This refers to public keys that are attached to an MSA (internally as `AccountId` Type)
    - `address` The "public key" in ss58 format
    - `account`: The refers to the keypair or multisig account when holding a Balance

# State Storage
- [x] Messages `blockMessages`, `get_block_messages`
- [x] Messages `messages`, `get_messages`
- [x] Schemas `governanceSchemaModelMaxBytes`, `get_schema_model_max_bytes`
- [x] Schemas `schemas`, `get_schema`
- [x] Schemas `SchemaCount`, `schema_count`
  - `CurrentSchemaIdentifierMaximum`, `get_current_schema_identifier_maximum`
    - MUST make sure that we increment BEFORE using in a clear single function  `get_next_schema_id`
- [x] https://github.com/LibertyDSNP/frequency/pull/531 MSA `MsaIdentifier`, `get_identifier`
    - `CurrentMsaIdentifierMaximum`, `get_current_msa_identifier_maximum`
    - MUST make sure that we increment BEFORE using in a clear single function `get_next_msa_id`
    - **In Code Review**
- [x] https://github.com/LibertyDSNP/frequency/pull/532 MSA `providerRegistry`, `get_provider_metadata`
  - MSA `providerToRegistryEntry`, `get_provider_registry_entry`
    - Update Type: `ProviderMetadata`, ProviderRegistryEntry`
    - **In Code Review**
- [x] https://github.com/LibertyDSNP/frequency/pull/533 MSA `providerInfoOf`, `get_provider_info`
[...]

PR #531, opened 2022-10-19
Renaming: MSA Identifier rename

dopatraman:
# Goal
The goal of this PR is to rename the `MsaIdentifier` Storage.

# Discussion
This partially completes #508

# Checklist
- [x] Tests added
- [x] Benchmarks added


shannonwells (commented):
I understand why Msa pallet weights need to be changed, but it seems like you should be able to skip checking in changes for external pallet weights (runtime/common/src/weights)

pallets/msa/src/lib.rs
```diff
@@ -133,11 +133,13 @@ pub mod pallet {
...
-	/// Storage type for MSA identifier
+	/// Storage type for the current MSA identifier maximum.
+	/// We need to track this value because the identifier maximum
+	/// is incremented each time a new identifier is created.
 	/// - Value: The current maximum MSA Id
 	#[pallet::storage]
-	#[pallet::getter(fn get_identifier)]
-	pub type MsaIdentifier<T> = StorageValue<_, MessageSourceId, ValueQuery>;
+	#[pallet::getter(fn get_current_msa_identifier_maximum)]
+	pub type CurrentMsaIdentifierMaximum<T> = StorageValue<_, MessageSourceId, ValueQuery>;
```
enddynayn, line 142: CurrentMsaIdentifierMaximum I do not know what this means.

dopatraman: The idea (as discussed with @wilwade) is that since this value is used to create all MSA ids, it can be thought of as the maximum value at any given time (since all MSA ids in the wild are guaranteed to be lower).

enddynayn: maybe flight the words...it is not a big deal.

```
get_current_msa_maximum_identifier
```

enddynayn (approved):

PR #539, opened 2022-10-20
Rename: register_provider -> create_provider

dopatraman:
# Goal
The goal of this PR is to rename the `register_provider` extrinsic.

# Discussion
This PR partially completes #508 

# Checklist
- [x] Design doc(s) updated
- [x] Tests added
- [x] Benchmarks added

### DIFF
--- a/pallets/msa/src/tests.rs
+++ b/pallets/msa/src/tests.rs
@@ -1500,7 +1500,7 @@ fn create_provider() {
 }
 
 #[test]
-fn register_provider_max_size_exceeded() {
+fn create_provider_max_size_exceeded() {
 	new_test_ext().execute_with(|| {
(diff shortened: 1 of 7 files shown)


pallets/msa/src/tests.rs
```diff
@@ -1533,7 +1506,7 @@ fn register_provider_max_size_exceeded() {
 		let (_new_msa_id, _) =
```
shannonwells, line 1506: I think this test should be renamed to create_provider_max_size_exceeded, but unless you are going to change something else, I would not hold up the PR.

shannonwells (approved):
Just the test name nitpick, but I don't think the PR should be held up for that. Looks good as is

Issue #443, opened 2022-09-16
Make some test helpers for pallets (MSA mostly)


shannonwells:
# Problem
There is a lot of copy-paste code that could be abstracted into helper functions, particularly in the msa pallet.  We already have a good start with test_public and test_origin_signed, and create_account.  

It would make the code more readable, easier to maintain and easier for writing new tests if we could find a good abstraction for some of the keypair, account, MSA creation and any other things that are repeated in numerous (4+ ?) places.  Also see if current tests can use existing test functions.

## Details
There are several places in the Msa pallet tests where we generate two key pairs, set an account variable to the public key of one of them, and use that to call `Msa::create_account`, [like this test](https://github.com/LibertyDSNP/frequency/blob/84aa8d7593f9db87ec8d91c1c5fecd67728b121c/pallets/msa/src/tests.rs#L262). 3 of them begin with "it_throws_error".

Tests like these might be good to convert to a Table Test pattern, useful for concisely defining and running several very similar tests.  Define a test case struct with one or two setup parameters, and an expected result parameter, declare a vec with your cases in it, then iterate over each and call the same code with different parameters.   Here is a simple example (note the use of generics):
```rust
struct TestCase<T> {
	schema: Vec<u8>,
	expected: T,
}

#[test]
fn test_require_validated_schema_size_errors() {
  new_test_ext().execute_with(|| {
    let test_cases: [TestCase<Error<Test>>; 3] = [
      TestCase { schema: vec![], expected: Error::TooShortSchema },
      TestCase { schema: Vec::from("foo".as_bytes()), expected: Error::TooShortSchema },
      TestCase { 
        schema: 
      Vec::from("foo,bar,bazz,way,wayway,foo,bar,bazz,thisiswaywaywaywaytoolong".as_bytes()),
        expected: Error::TooLongSchema,
      },
    ];
    for tc in test_cases {
      let err = AnnouncementPallet::require_validated_schema(tc.schema).unwrap_err();
        assert_eq!(tc.expected, err)
    }
  })
}

```

Other tests may not be similar enough to work with a Table Test pattern, but it's possible this type of code in the tests could be encapsulated into a function by itself, depending on where any of these variables are reused:
```rust
    let (key_pair, _) = sr25519::Pair::generate();
    let (key_pair_2, _) = sr25519::Pair::generate();

    let account = key_pair.public();
    let (new_msa_id, _) = Msa::create_account(account.into(), EMPTY_FUNCTION).unwrap();

[...]

PR #552, opened 2022-10-20
WIP #443 helper tests for MSA pallet

allanperlee:
# Goal
The goal of this PR is 
Change some tests to use the table testing pattern; reduce redundant code.
Closes 
#443 
# Discussion

Separate tests were added as an example of the possible reduced code. These are titled with `using_mock_create_account` and they use mock.rs's `create_account` function. A brief table test was introduced that verifies the length of schemas and another test for the iteration of schemas that should pass with the given provider, delegator, and testcase schema. Need to learn the pull-request workflow for Frequency as the previous pull-request was closed due to an error in my fork. Also, the testcase struct  was accidentally removed from mock.rs, will be implemented like Puneet's changes in line 2057 in pallet MSA's tests.rs.
# Checklist
- [ ] Chain spec updated
- [ ] Design doc(s) updated
- [ X] Tests added
- [ ] Benchmarks added
- [ ] Weights updated

### DIFF
--- a/pallets/msa/src/tests.rs
+++ b/pallets/msa/src/tests.rs
@@ -989,7 +989,7 @@ pub fn revoke_delegation_by_delegatoris_successful() {
 		));
 
 		System::assert_last_event(
-			Event::DelegatorRevokedDelegation { delegator: 1.into(), provider: 2.into() }.into(),
+			Event::DelegationRevoked { delegator: 1.into(), provider: 2.into() }.into(),
 		);
 	});
 }
@@ -1218,8 +1218,7 @@ pub fn revoke_delegation_by_provider_happy_path() {
 
 		// 7. verify the event
 		System::assert_last_event(
-			Event::ProviderRevokedDelegation { provider: Provider(1), delegator: Delegator(2) }
-				.into(),
+			Event::DelegationRevoked { provider: Provider(1), delegator: Delegator(2) }.into(),
 		);
 	})
 }
--- a/pallets/msa/src/lib.rs
+++ b/pallets/msa/src/lib.rs
@@ -255,14 +255,7 @@ pub mod pallet {
 			provider_msa_id: MessageSourceId,
 		},
 		/// The Delegator revoked its delegation to the Provider
-		DelegatorRevokedDelegation {
-			/// The Provider MSA Id
-			provider: Provider,
-			/// The Delegator MSA Id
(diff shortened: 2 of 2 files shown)


pallets/msa/src/tests.rs
```diff
@@ -96,6 +96,31 @@ fn it_throws_error_when_key_verification_fails() {
 	});
 }
 
+#[test]
+fn it_throws_error_when_key_verification_fails_using_mock_create_account() {
```
enddynayn, line 100: Looking good.

pallets/msa/src/tests.rs
```diff
@@ -218,6 +337,48 @@ fn add_key_with_valid_request_should_store_value_and_event() {
 		let encode_data_new_key_data = wrap_binary_data(add_new_key_data.encode());
 		let signature_owner: MultiSignature = key_pair.sign(&encode_data_new_key_data).into();
 		let signature_new_key: MultiSignature = key_pair_2.sign(&encode_data_new_key_data).into();
+<<<<<<< HEAD
```
enddynayn, line 340: Some merge conflicts.

pallets/msa/src/tests.rs
```diff
@@ -129,6 +154,36 @@ fn it_throws_error_when_not_msa_owner() {
 	});
 }
 
+
+#[test]
+fn it_throws_error_when_not_msa_owner_using_mock_create_account() {
```
enddynayn, line 159: nice

pallets/msa/src/tests.rs
allanperlee: Should I replace my tests with the tests that use `Msa::create_account`?

enddynayn: Not 100% sure what you are asking.

allanperlee: The issue was that there was repeated code with the two key pairs. In mock.rs, `create_account` creates an MSA id, and this function is used in my tests. These tests could replace the tests that use `Msa::create_acount` right?

shannonwells: Assuming that the tests are otherwise the same, yes they could replace the old ones.  Please delete the `_using_mock_create_account` part of the test name though, otherwise the test name almost sounds like it's not a "real" test, when in fact it does call Msa::create_account.

PR #556, opened 2022-10-21
bash

saraswatpuneet:
# Goal
The goal of this PR is to update script to use bash env for generate_specs, all our instances are bash


shannonwells (approved):
Yep.  We should be doing this

PR #560, opened 2022-10-21
feat(schemas): update schema permission storage 

enddynayn:
# Goal
Update the schema-permission storage to use a
BoundedBTreeMap instead of an OrderSet.

This update increase performance when writing and
updating schema permissions.

Additionally, each schema permission
stores a field, "revoked_at," that indicates the
block number for which the permission expires.
If set to "None" type, it means that
the permission is not revoked and is valid.

issue-455

### DIFF
--- a/common/primitives/src/msa.rs
+++ b/common/primitives/src/msa.rs
@@ -1,5 +1,7 @@
 use codec::{Decode, Encode, EncodeLike, Error, MaxEncodedLen};
-use frame_support::{dispatch::DispatchResult, traits::Get, BoundedVec, RuntimeDebug};
+use frame_support::{
+	dispatch::DispatchResult, traits::Get, BoundedBTreeMap, BoundedVec, RuntimeDebug,
+};
 use scale_info::TypeInfo;
 #[cfg(feature = "std")]
 use serde::{Deserialize, Serialize};
@@ -7,7 +9,6 @@ use sp_runtime::DispatchError;
 use sp_std::prelude::Vec;
 
 pub use crate::schema::SchemaId;
-pub use orml_utilities::OrderedSet;
 
 /// Message Source Id or msaId is the unique identifier for Message Source Accounts
 /// Message Source Id or msaId is the unique identifier for Message Source Accounts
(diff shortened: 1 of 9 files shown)


pallets/msa/src/lib.rs
```diff
@@ -799,20 +803,11 @@ impl<T: Config> Pallet<T> {
 		delegator: Delegator,
 		schemas: Vec<SchemaId>,
 	) -> DispatchResult {
-		let granted_schemas: BoundedVec<SchemaId, T::MaxSchemaGrantsPerDelegation> = schemas
```
enddynayn, line 802: add test

pallets/msa/src/lib.rs
```diff
@@ -821,6 +816,30 @@ impl<T: Config> Pallet<T> {
 		Ok(())
 	}
 
+	/// Add anotation
+	pub fn initialize_schema_permissions(
```
enddynayn, line 820: add test

pallets/msa/src/lib.rs
```diff
@@ -825,6 +820,30 @@ impl<T: Config> Pallet<T> {
 		Ok(())
 	}
 
+	/// Initializes the BoundedBTreeMap used to store schema permissions and validates schema ids.
+	pub fn initialize_schema_permissions(
```
enddynayn, line 824: I think I prefer `build_schema_permissions`.

dopatraman (approved):

common/primitives/src/msa.rs
```diff
@@ -48,16 +49,17 @@ impl From<Delegator> for MessageSourceId {
...
+#[derive(TypeInfo, RuntimeDebug, Clone, Decode, Encode, PartialEq, MaxEncodedLen, Eq)]
 #[scale_info(skip_type_params(MaxSchemaGrantsPerDelegation))]
-pub struct Delegation<BlockNumber, MaxSchemaGrantsPerDelegation>
+pub struct Delegation<SchemaId, BlockNumber, MaxSchemaGrantsPerDelegation>
 where
 	MaxSchemaGrantsPerDelegation: Get<u32>,
 {
 	/// Block number the grant will be revoked.
-	pub expired: BlockNumber,
+	pub revoked_at: BlockNumber,
```
enddynayn, line 59: make this Optional instead of zero?

PR #562, opened 2022-10-24
Redesign of MSA retirement to use signed extension

### LINKED ISSUE #44: As a user, I should be able to retire my MSA

rlaferla:
Implemented redesign of MSA retirement to use signed extension Implemented unit tests for MSA retirement

Closes #44

### DIFF
--- a/pallets/msa/src/lib.rs
+++ b/pallets/msa/src/lib.rs
@@ -63,12 +63,14 @@
 use codec::{Decode, Encode};
 use common_primitives::{
 	msa::{
-		Delegation, DelegationValidator, Delegator, MsaLookup, MsaValidator, OrderedSet, Provider,
+		Delegation, DelegationValidator, Delegator, MsaLookup, MsaValidator, Provider,
 		ProviderLookup, ProviderRegistryEntry, SchemaGrantValidator,
 	},
 	schema::{SchemaId, SchemaValidator},
 };
-use frame_support::{dispatch::DispatchResult, ensure, traits::IsSubType, weights::DispatchInfo};
+use frame_support::{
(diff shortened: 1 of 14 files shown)


pallets/msa/src/lib.rs
```diff
@@ -646,33 +646,29 @@ pub mod pallet {
 
 		/// Retire a MSA
 		///
+		/// When a user wants to disassociate themselves from Frequency, they can retire their MSA for free provided that:
+		///  (1) They own the MSA
+		///  (2) There is only one account key
+		///  (3) The MSA is not a registered provider.
+		///
+		/// This does not currently remove any messages related to the MSA.
```
saraswatpuneet, line 654: Can we add a documentation for #Arguments that this extrinsic takes, you can check other examples of adding arguments like ``add_key_to_msa```

saraswatpuneet (approved):

enddynayn (approved):
Fantastic! 👍

pallets/msa/src/lib.rs
rlaferla: Definitely.  I see many other functions that are missing documentation for #Arguments as well.  We should document all of them but I'll add them for this one function.

rlaferla: So for the `retire_msa()` function, it just takes in the `origin` as an argument.  I will document it.

PR #577, opened 2022-10-25
#443 helper tests for MSA pallet

### LINKED ISSUE #443: Make some test helpers for pallets (MSA mostly)

allanperlee:
# Goal
The goal of this PR is 
Change some tests to use the table testing pattern; reduce redundant code.
Closes #443
 
# Discussion

Separate tests were added as an example of the possible reduced code. These are titled with using_mock_create_account and they use mock.rs's create_account function. A brief table test was introduced that verifies the length of schemas and another test for the iteration of schemas that should pass with the given provider, delegator, and testcase schema. Will consider further suggestions.
# Checklist
- [ ] Chain spec updated
- [ ] Design doc(s) updated
- [X] Tests added
- [ ] Benchmarks added
- [ ] Weights updated


wilwade:
@allanperlee This looks great! I'll do a rebase and we can merge it! Thanks so much for your work on this.

allanperlee:
Hi @wilwade sounds great! And thank you all for your patience.

enddynayn (approved):

PR #584, opened 2022-10-26
Chores/review unwrap #561

### LINKED ISSUE #561: Review use of unwrap() in non-test and non-benchmark code.
The use of Rust's `unwrap() `on `Result<T,E>` and `Option<T>` will return `T` in the case of a `Ok` for `Result()` and `Some()` for `Option` but will **panic** otherwise.  A panic causes the process to abort.  This is fine for _unit tests_ and _benchmarks_ but not generally in other code.  i.e.  exceptions exist perhaps for CLI code or possibly chain specs.

Need to review these files:

```
common/helpers/src/avro.rs
pallets/messages/src/lib.rs
pallets/messages/src/types.rs
pallets/messages/src/rpc/src/lib.rs
pallets/msa/src/lib.rs
pallets/msa/src/rpc/src/lib.rs
runtime/frequency/src/lib.rs
runtime/frequency-rococo/src/lib.rs
```

shannonwells:
# Goal
The goal of this PR is to ensure that there are no hidden possible panics in production code.

Closes #561 

# Discussion

- [x] common/helpers/src/avro.rs converted to use `match`
- [x] pallets/messages/src/lib.rs - Two check first for `is_some()` (and converting them to matches makes the code less readable) so they're fine. The other is now `unwrap_or_default` with an empty check.
- [x] pallets/messages/src/types.rs (completed by @saraswatpuneet )
- [x] pallets/messages/src/rpc/src/lib.rs - converted to `unwrap_or_default()`
- [x] pallets/msa/src/lib.rs - this one is only for benchmarks
- [x] pallets/msa/src/rpc/src/lib.rs - replace with `unwrap_or(false)`
- [x] runtime/frequency/src/lib.rs -  - this one is only for `try-runtime` and not production
- [x] runtime/frequency-rococo/src/lib.rs - this one is only for `try-runtime` and not production

## Additionally
Code coverage was reduced in `avro.rs` so I had to write/rewrite some tests. In the process I reorganized the tests so that valid schemas and invalid schemas are completely separate, and tests for schema errors are separate from happy path tests.

# Checklist
- [x] N/A Chain spec updated
- [x] Tests added
- [X] N/A Benchmarks added
- [X] N/A Weights updated


pallets/messages/src/lib.rs
```diff
@@ -376,7 +376,11 @@ impl<T: Config> Pallet<T> {
 		// insert into storage and create events
 		for (schema_id, messages) in map {
 			let count = messages.len() as u16;
-			let bounded_vec: BoundedVec<_, _> = messages.try_into().unwrap();
+			let bounded_vec: BoundedVec<_, _> = messages.try_into().unwrap_or_default();
+
+			if bounded_vec.is_empty() {
```
shannonwells, line 381: I set this to the same return value as previous, because there were no additional db accesses at this point.

pallets/messages/src/tests.rs
```diff
@@ -119,7 +119,7 @@ fn add_message_with_too_large_message_should_panic() {
 }
 
 #[test]
-fn add_message_with_invalid_msa_account_should_panic() {
+fn add_message_with_invalid_msa_account_errors() {
```
shannonwells, line 122: it doesn't panic.

saraswatpuneet (approved):
:1st_place_medal:

pallets/messages/src/lib.rs
```diff
@@ -376,7 +376,11 @@ impl<T: Config> Pallet<T> {
 		// insert into storage and create events
 		for (schema_id, messages) in map {
 			let count = messages.len() as u16;
-			let bounded_vec: BoundedVec<_, _> = messages.try_into().unwrap();
+			let bounded_vec: BoundedVec<_, _> = messages.try_into().unwrap_or_default();
```
enddynayn, line 379: This is probably where we want to add a log.

enddynayn: Maybe we add something along this line so that we can log and also return the proper weight in the case of an error. Something along this lines:
```
match message.try_into() {
  Ok() => {
    if bounded_vec.is_empty() { return ...}
  },
  Error() => {
log("something")
   return T::DbWeight::get().reads(1)
     }
   
}
```

enddynayn (approved):

pallets/messages/src/lib.rs
enddynayn: Actually, probably don't want this either.

saraswatpuneet: Me likey, question is do we have any holes that this case may occur, "terrible"

enddynayn: lol...iI think it autocorrected to terrible and I just rolled with it. Terrible is probably not the right word.

common/helpers/src/avro_tests.rs
```diff
@@ -1,99 +1,139 @@
...
+		assert!(
+			schema_result.is_err(),
+			"schema {} was supposed to be invalid; error: {:?}",
+			invalid_schema,
+			schema_result.err()
+		);
+	}
+}
+
+// const VALID_EXAMPLES: &[(&str, bool)] = &[
```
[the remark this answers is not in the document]
shannonwells: I'd like to actually check for the actual error, however most are InvalidSchema and I feel like I have to draw the line somewhere for the scope of this PR.

pallets/messages/src/lib.rs
shannonwells: Yeah, I was kind of on the fence about this anyway, a log is a better idea.

Issue #255, opened 2022-07-27
[Security] General code audit


shannonwells:
- [x] Outside Audit
- [x] (This) Internal Audit
- [x] Move Custom RPC only code out of the WASM

Divide up code base and audit for such things as:

* glaring security issues
* outdated or incorrect descriptions/comments
* tests that should be added

- [x] inconsistencies in naming
- [x] Collect notes & make suitably sized issues out of them
- [x] Assign priority to issues - especially if any of these should be pre-release
- [x] Documentation Improvements
  - [x] Methods
  - [x] Inline comments specifically for help with those auditing the code
  - [x] Module Documentation
- [x] Review Benchmarking
  - [x] Comment why those parameters were chosen
  - [x] Shifted to https://github.com/LibertyDSNP/frequency/issues/666
- [x] Places we should be using `#[pallet::compact]`
  -  https://github.com/LibertyDSNP/frequency/pull/588 
- [x] Address every TODO

All dependency security issues dealt with or listed with comments here:
https://github.com/LibertyDSNP/frequency/blob/1038293e1763a3ca98067496a5cb827a6c1b74cd/.cargo-deny.toml#L76-L92

rlaferla:
See #561

rlaferla:
How can we check for _index out of bounds_ conditions?  i.e.  find lines of code that could be subject to it and write unit tests to ensure the code is safe

PR #586, opened 2022-10-27
Remove unused and duplicate Error: TooManySchemas

wilwade:
# Goal
The goal of this PR is remove an unused and duplicate error `TooManySchemas` with the one that was actually used `SchemaCountOverflow`

Part of #255


shannonwells (approved):
looks good to go

PR #587, opened 2022-10-27
Method Documentation Updates

wilwade:
# Goal
The goal of this PR is to update and standardize our method documentation

Part of #255 

# Note: Rebased on #586 (so it has that code in it as well)

# Discussion
- Removed `Arguments`
- Updated with Errors and Events
- Moved one piece of code as it was splitting some impls up
- Generally followed suggestions from https://www.tangramvision.com/blog/making-great-docs-with-rustdoc

# Checklist
- [x] No Code changes

### DIFF
--- a/common/primitives/src/msa.rs
+++ b/common/primitives/src/msa.rs
@@ -14,11 +14,11 @@ use sp_std::prelude::Vec;
 pub use crate::schema::SchemaId;
 
 /// Message Source Id or msaId is the unique identifier for Message Source Accounts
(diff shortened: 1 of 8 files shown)


common/primitives/src/msa.rs
```diff
@@ -84,18 +85,6 @@ impl Decode for Provider {
 	}
 }
 
-/// This is the metadata associated with a provider. As of now it is just a
```
wilwade, line 87: This code just moved down as it was splitting up some connected impls.

saraswatpuneet (approved):

PR #588, opened 2022-10-27
Code Audit: Compact Encoding for Extrinsic Params

### LINKED ISSUE #582: Use compact integer encoding where possible
Substrate allows for compact integer encoding via `#[pallet::compact]`. We should use this where possible.

dopatraman:
# Goal
The goal of this PR is to continue the general code audit by adding compact integer encoding to all extrinsic parameters.

Closes #582 

# Discussion
It looks like pallet compact encoding can only be used on extrinsic code. There may be a way to add encoding for integer types used in RPC but it will need some investigation.

# Checklist
- [ ] Weights updated


wilwade (approved):
💯

shannonwells (approved):
:shipit:

PR #595, opened 2022-10-28
feat(schemas): add the ability to revoke schemas

### LINKED ISSUE #455: Schema grants need expiration
# Description
Currently, a user can only revoke permissions for all schemas (by revoking the delegation entirely) after a given block or remove a schema which makes all previously posted messages for that schema invalid.

As a user, I want to be able to add and/or revoke some (1 or more) of my schema grants without making my previously posted messages for that schema and provider invalid.

## Acceptance Criteria:
- [ ] As a consumer of a DSNP Batch Publication, I want to be able to validate each announcement (row) based on these criteria: 
    -  [ ] Did a delegation from the `fromId` (aka an MSA Id) exist to the provider with the given schema id at the time of publication?
 - [ ] A user with an **existing delegation** should be able to add and/or revoke some (1 or more) of my schema grants without making my previously posted messages for that schema and provider invalid.
 - [ ] As a user with a **previously revoked delegation**, I want to be able to re-delegate to a provider with some new and re-granting some old schema grants without making my previously posted messages for schemas that I am *not* adding back valid.
 - [ ] As a user with **no delegation**, I want to be able to add a *new* delegation with *new* schema grants.
  
## Architecture

- [ ] Optimization: where ever there is a check for a valid schema grant, first check if the delegation is expired for an early exit, before checking if the schema itself is expired.
- [ ] Add block number parameter to `ensure_valid_schema_grant`.
- [ ] Inside `ProviderInfo`, instead of storing just a schema ID, store a SchemaGrantInfo : 

```rust
struct SchemaPermission {
   schema_id: SchemaId,
   schema_expires_at: BlockNumber
}
```

## Notes
- Just as we don't track gaps with delegations, we also do not track gaps with schema grants. So if a user grants, revokes, and then grants again, it is as if the user never revoked the grant.
- Technically, we only care about revoked schema grants when they are off-chain payload location grants
- We never read on-chain off-chain payload grants, only on-chain payload grants (however, knowing that a schema id is one or the other takes a read)
- Technically, we could store grants and revoked grants separately, but there is additional read and write costs to that.

## Questions
How should `add_provider_to_msa` work?
  - Situations:
  - A. New Delegation, New Grants
  - B. Existing Delegation, Different Grants
  - C. Revoked Delegation, Same Grants
  - D. Revoked Delegation, Different Grants

## Related Issues
- Custom RPC `get_granted_schemas` should accept an optional block number to filter out expired grants so that I can get schema grants at a point in time historically without an archive node
- As a consumer of a DSNP Batch Publication, I want to be able to validate each announcement (row) based on these criteria:
  - Did a delegation from the `fromId` (aka an MSA Id) exist to the provider with the given schema id at the time of publication?
  - aka Custom RPC `check_delegations` should also take an optional schema_id

## TODO
- [ ] Refactor Schemas data structure https://github.com/LibertyDSNP/frequency/pull/516/files
- [ ] extrinsic to add a schema ID

enddynayn:
# Goal
Add the ability to revoke schemas. Attempting
to revoke a schema that does not exist
returns with an error.

Update benchmarks.

issue-455
Closes #455


pallets/msa/src/tests.rs
```diff
@@ -2365,3 +2365,93 @@ fn try_mutate_delegation_success() {
...
+			SchemaId,
+			<Test as frame_system::Config>::BlockNumber,
+			<Test as Config>::MaxSchemaGrantsPerDelegation,
+		>::new();
+
+		expected.try_insert(1, Default::default()).expect("testing expected");
+
+		assert_eq!(delegation_relationship.schema_permissions, expected);
+
+		// Add new schema ids
```
enddynayn, line 2391: remove

enddynayn:
Waiting for benchmarks to complete so I can update the code.

pallets/msa/src/lib.rs
```diff
@@ -701,6 +701,35 @@ pub mod pallet {
 			Ok(())
 		}
 
+		/// Revokes a list of schema permissions to a provider. Attempting to revoke a Schemas that have already
+		/// been revoked are ignored.
+		///
+		/// ### Errors
+		/// - Returns [`NoKeyExists`](Error::NoKeyExists) - If there is not MSA for `origin`.
+		/// - Returns [`DelegationNotFound`](Error::DelegationNotFound) - If there is not delegation relationship between Origin and Delegator or Origin and Delegator are the same.
+		/// - Returns [`SchemaNotGranted`](Error::SchemaNotGranted) - If attempting to revoke a schema that has not previously been granted.
```
enddynayn, line 710: add extra line
///

pallets/msa/src/tests.rs
```diff
@@ -2365,3 +2365,175 @@ fn try_mutate_delegation_success() {
...
+				provider.into(),
+				schema_ids,
+			),
+			Error::<Test>::NoKeyExists
+		);
+	});
+}
+
+#[test]
+fn schema_permissions_trait_impl_try_get_mut_schema_success() {
```
enddynayn, line 2511: add test for :: try_insert_schema[s]

pallets/msa/src/tests.rs
```diff
@@ -2365,3 +2365,175 @@ fn try_mutate_delegation_success() {
...
+		assert_ok!(Msa::add_provider(provider, delegator, schema_grants));
+
+		let additional_grants = (2..32 as u16).collect::<Vec<_>>();
+		let result = Msa::revoke_permissions_for_schemas(delegator, provider, additional_grants);
+
+		assert_noop!(result, Error::<Test>::SchemaNotGranted);
+
+		let result = Msa::get_delegation(delegator, provider);
+
+		let mut expected = Delegation {
```
shannonwells, line 2441: Is this testing that the grants were not modified?  Seems like line 2437 fulfills the test requirement though, since it was a no-op.

pallets/msa/src/tests.rs
```diff
@@ -2365,3 +2365,175 @@ fn try_mutate_delegation_success() {
...
+		assert_eq!(result.unwrap(), expected);
+	});
+}
+
+#[test]
+fn revoke_schema_permissions_success() {
+	new_test_ext().execute_with(|| {
+		set_schema_count::<Test>(3);
+
+		let (key_pair, _) = sr25519::Pair::generate();
```
shannonwells, line 2464: You might check and see if `test_create_delegator_msa_with_provider` in mocks will work to save some repeated code.  Even modify it if it would help.

pallets/msa/src/lib.rs
```diff
@@ -798,6 +827,28 @@ impl<T: Config> Pallet<T> {
 		})
 	}
 
+	/// Revokes a list of schema permissions from a delegation relationship.
+	pub fn revoke_permissions_for_schemas(
+		delegator: Delegator,
+		provider: Provider,
```
enddynayn, line 833: provider_msa_id

pallets/msa/src/lib.rs
```diff
@@ -701,6 +701,35 @@ pub mod pallet {
...
+		/// been revoked are ignored.
+		///
+		/// ### Errors
+		/// - Returns [`NoKeyExists`](Error::NoKeyExists) - If there is not MSA for `origin`.
+		/// - Returns [`DelegationNotFound`](Error::DelegationNotFound) - If there is not delegation relationship between Origin and Delegator or Origin and Delegator are the same.
+		/// - Returns [`SchemaNotGranted`](Error::SchemaNotGranted) - If attempting to revoke a schema that has not previously been granted.
+		#[pallet::weight(1_000)]
+		pub fn revoke_schema_permissions(
+			origin: OriginFor<T>,
+			provider: MessageSourceId,
```
enddynayn, line 714: provider_msa_id

pallets/msa/src/tests.rs
```diff
@@ -2365,3 +2365,175 @@ fn try_mutate_delegation_success() {
...
+	new_test_ext().execute_with(|| {
+		set_schema_count::<Test>(31);
+
+		let delegator = Delegator(2);
+		let provider = Provider(1);
+		let schema_grants = vec![1];
+
+		assert_ok!(Msa::add_provider(provider, delegator, schema_grants));
+
+		let additional_grants = (2..32 as u16).collect::<Vec<_>>();
```
shannonwells, line 2434: spiffy!

pallets/msa/src/types.rs
```diff
@@ -66,6 +66,25 @@ pub trait PermittedDelegationSchemas<T: Config> {
 
 		Ok(())
 	}
+
+	/// Attempt get and mutate a collection of schemas. Dispatches error when a schema cannot be found.
+	fn try_get_mut_schemas(
```
shannonwells, line 71: I like that you're following the substrate/rust naming conventions.

shannonwells (approved):
A couple of nit-questions but nothing blocking.

pallets/msa/src/tests.rs
enddynayn: That is correct. I updated the test to include more than one schema grant. It was just a sanity check to ensure that even if the first two got modified that it would throw those modifications away if an error occurred.

PR #603, opened 2022-10-31
Implemented issue #598 small cleanups.

### LINKED ISSUE #598: Small Cleanups
Here are some things we can potentially improve:

- [x] 1. Rename `try_get_msa_from_account_id` to `try_get_msa_from_public_key` since we are not getting an MSA from an account id but rather a public key.
- [x] 2. Rename `remove_delegator` to `delete_delegation_relationship` since it is not removing the delegator but instead removing the relationship.
- [x] 3. Update `remove_delegator` to return nothing:

```
pub fn remove_delegator(delegator_id: Delegator) {
    DelegatorAndProviderToDelegation::<T>::clear_prefix(delegator, u32::max_value(), None);
}
```
The following line makes me think that we can fail when I think we can't:

https://github.com/LibertyDSNP/frequency/blob/377bbe37fbe6e226cfa8361dba623e0c07a54db0/pallets/msa/src/lib.rs#L1102

rlaferla:
Renamed try_get_msa_from_account_id() to try_get_msa_from_public_key(), remove_delegator() to delete_delegation_relationship()  and updated remove_delegator() to return nothing

Closes #598

### DIFF
--- a/pallets/msa/src/benchmarking.rs
+++ b/pallets/msa/src/benchmarking.rs
@@ -60,7 +60,7 @@ fn create_account_with_msa_id<T: Config>(n: u32) -> (T::AccountId, MessageSource
 
 	assert_ok!(Msa::<T>::create(RawOrigin::Signed(provider.clone()).into()));
 
-	let msa_id = Msa::<T>::try_get_msa_from_account_id(&provider).unwrap();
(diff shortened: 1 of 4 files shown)


enddynayn (approved):
Fabulous! 👍

PR #606, opened 2022-10-31
generate and use some more weights

### LINKED ISSUE #604: Use our weights for substrate pallets
We aren't using, or we don't have generated weights to use for Substrate pallets.

Relates to #255 

Go through runtime configs and make sure that

- [ ] There are weights generated for all substrate pallets that we use.
- [ ] That we use those generated weights in **frequency** runtime
- [ ] That we use those generated weights in **frequency-rococo** runtime

shannonwells:
# Goal
The goal of this PR is to use our generated weights on the reference hardware for substrate pallet weights.

Closes #604 

# Checklist
- [x] There are weights generated for all substrate pallets that we use
- [x] That we use those generated weights in **frequency** runtime
- [x] That we use those generated weights in **frequency-rococo** runtime


runtime/frequency-rococo/src/lib.rs
```diff
@@ -81,14 +81,21 @@ pub type SignedExtra = (
...
 
-/// Extrinsic type that has already been checked.
-pub type CheckedExtrinsic = generic::CheckedExtrinsic<AccountId, Call, SignedExtra>;
 /// A Block signed with a Justification
 pub type SignedBlock = generic::SignedBlock<Block>;
 
 /// BlockId type as expected by this runtime.
 pub type BlockId = generic::BlockId<Block>;
 
+/// Block type as expected by this runtime.
```
shannonwells, line 90: I moved these to match frequency/src/lib.rs for easier comparison when configs are changed.

runtime/frequency-rococo/src/lib.rs
```diff
@@ -287,7 +289,7 @@ impl pallet_timestamp::Config for Runtime {
 	type Moment = u64;
 	type OnTimestampSet = ();
 	type MinimumPeriod = MinimumPeriod;
-	type WeightInfo = ();
+	type WeightInfo = weights::pallet_timestamp::SubstrateWeight<Runtime>;
```
wilwade, line 292: I've see a few chains with this set as `()`. Are we sure about this one?

shannonwells: Yes.  In Moonbeam (for timestamp) it's explicitly `pallet_timestamp::weights::SubstrateWeight<Runtime>`.
In Acala they are using defaults `()`.
However, Kilt has run benchmarks , including timestamp and session, and uses these weights in their runtime configs, such as [peregrine](https://github.com/KILTprotocol/kilt-node/blob/develop/runtimes/peregrine/src/lib.rs).
 
Parity says "Benchmarking the Pallets is always necessary since the results depend on your runtime configuration. Each variable that you change in the runtime config can lead to different results for that Pallet."

This is why we have been running those benchmarks for the others, such as orml and democracy, and this PR simply adds the ones that we hadn't run benchmarks for yet, generally because they were added later. Moonbeam and Acala simply have not done this yet.

This [Question/Answer on Substrate StackExchange may help](https://substrate.stackexchange.com/questions/3229/benchmark-frame-pallets) for future reference.

wilwade: Sounds good. Thanks for checking in and the great explanation.

runtime/frequency-rococo/src/lib.rs
```diff
@@ -558,7 +560,7 @@ impl pallet_session::Config for Runtime {
 	// Essentially just Aura, but lets be pedantic.
 	type SessionHandler = <SessionKeys as sp_runtime::traits::OpaqueKeys>::KeyTypeIdProviders;
 	type Keys = SessionKeys;
-	type WeightInfo = ();
+	type WeightInfo = weights::pallet_session::SubstrateWeight<Runtime>;
```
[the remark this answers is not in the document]
shannonwells: Yes, see above

wilwade (approved):

Issue #605, opened 2022-10-31
Validity checks for Pays::No extrinsics should go into SignedExtension


shannonwells:
Free extrinsics can not be able to fail, or else this provides an attack vector for filling up a block with bad transactions that cost nothing to the attacker.

Follow the pattern with `retire_msa` where any type of ensure! or other validation logic is moved to the `CheckFreeExtrinsicUse` SignedExtension.

- [x] `revoke_delegation_by_provider`  (@shannonwells)
  - [x]  ensure_valid_msa_key
  
- [x] `revoke_delegation_by_delegator` (@enddynayn )
  - [x]  ensure_valid_msa_key

- [ ] `delete_msa_public_key` ( @enddynayn )
  - [ ]  ensure!(who != key, Error::<T>::InvalidSelfRemoval);
  - [ ]  try_get_msa_from_account_id
  - [ ]  try_get_msa_from_account_id
  - [ ]  ensure!(who_msa_id == account_to_remove_msa_id, Error::<T>::NotKeyOwner);

- [x] `retire_msa` (@shannonwells )
  - [x]  try_get_msa_from_account_id
  - [x]  remove_delegator
  - [x]  delete_key_for_msa

aramikm:
Question: Can a malicious node ignore this signed extension and create a block with these free transactions?

If yes, should we limit the number or percentage of free transactions inside a block?

PR #611, opened 2022-11-01
retire_msa check and update for dos attack #605

shannonwells:
# Goal
The goal of this PR is to address the retire_msa section of #605 : ensure that the extrinsic cannot fail and instead, any potential failures happen in the SignedExtension.

# Discussion
The main change is to move all db access inside a match that fetches the MSA, mainly for clarity.

A check for an MSA is already made in the SignedExtension so it fails if the Origin (actually the AccountId) does not have an MSA.  It also verifies that there is only one key, the AccountId, associated with it, and that it's not a registered provider.  This means that we should never get into the extrinsic unless there is an MSA. We're required to provide a None option in our match so if we somehow strangely wind up in there, it logs a warning and exits.

- [ ]  `ensure_signed` - this must remain inside the extrinsic to guarantee having an AccountId for looking up an MSA.
- [ ]  `try_get_msa_from_account_id` - this has been removed, as it is identical to another function. We know already we should have an MSA, so we just fetch it instead of doing a try_get type of function call.
- [ ]  `remove_delegator` - similarly, this being in the match it doesn't happen unless there is an MSA. If there is no delegator - which is allowed, the MSA may already have removed a provider - then there is no failure; this is harmless.
- [ ]  `delete_key_for_msa` - also in the Some(msa_id) clause as above.

## Also
I refactored away a function that wasn't really doing anything extra.

## Verified
Attempting to `retire_msa` fails the SignedExtension validation - and so does not add a txn to the block - under the following conditions:
1. with an account with no MSA ID
2. with an account that has an MSA ID but is a registered provider
3. by submitting an unsigned transaction

# Checklist
## Not applicable:
- Chain spec updated
- Updated the js/api-augment code if a custom RPC added/changed
- N/A Design doc(s) updated
- N/A Tests added
- Benchmarks added
- Weights updated

### DIFF
--- a/pallets/msa/src/benchmarking.rs
+++ b/pallets/msa/src/benchmarking.rs
@@ -60,7 +60,7 @@ fn create_account_with_msa_id<T: Config>(n: u32) -> (T::AccountId, MessageSource
 
 	assert_ok!(Msa::<T>::create(RawOrigin::Signed(provider.clone()).into()));
 
-	let msa_id = Msa::<T>::try_get_msa_from_public_key(&provider).unwrap();
+	let msa_id = Msa::<T>::ensure_valid_msa_key(&provider).unwrap();
 
 	(provider.clone(), msa_id)
 }
@@ -138,7 +138,7 @@ benchmarks! {
 
 		// Create a MSA account
 		assert_ok!(Msa::<T>::create(RawOrigin::Signed(caller.clone()).into()));
-		let msa_id = Msa::<T>::try_get_msa_from_public_key(&caller).unwrap();
+		let msa_id = Msa::<T>::ensure_valid_msa_key(&caller).unwrap();
 
 		assert_eq!(Msa::<T>::is_registered_provider(msa_id),false);
 
--- a/pallets/msa/src/lib.rs
+++ b/pallets/msa/src/lib.rs
@@ -55,6 +55,19 @@
 )]
 
 use codec::{Decode, Encode};
+use frame_support::{
+	dispatch::DispatchResult, ensure, pallet_prelude::*, traits::IsSubType, weights::DispatchInfo,
+};
+use frame_system::pallet_prelude::*;
+use scale_info::TypeInfo;
+use sp_core::crypto::AccountId32;
+use sp_runtime::{
+	traits::{Convert, DispatchInfoOf, Dispatchable, One, SignedExtension, Verify, Zero},
+	DispatchError, MultiSignature,
+};
+use sp_std::prelude::*;
+
+pub use common_primitives::{msa::MessageSourceId, utils::wrap_binary_data};
 use common_primitives::{
 	msa::{
 		Delegation, DelegationValidator, Delegator, MsaLookup, MsaValidator, Provider,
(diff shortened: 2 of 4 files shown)


pallets/msa/src/benchmarking.rs
```diff
@@ -60,7 +60,7 @@ fn create_account_with_msa_id<T: Config>(n: u32) -> (T::AccountId, MessageSource
 
 	assert_ok!(Msa::<T>::create(RawOrigin::Signed(provider.clone()).into()));
 
-	let msa_id = Msa::<T>::try_get_msa_from_public_key(&provider).unwrap();
+	let msa_id = Msa::<T>::ensure_valid_msa_key(&provider).unwrap();
```
shannonwells, line 63: `try_get_msa_from_public_key` was identical to `ensure_valid_msa_key` and used in the same way.

pallets/msa/src/lib.rs
```diff
@@ -1125,18 +1117,17 @@ impl<T: Config> Pallet<T> {
...
 
 	// 	response
 	// }
 
-	/// Checks that a key is associated to an MSA and has not been revoked.
+	/// Retrieve MSA Id associated with `key` or return `NoKeyExists`
 	pub fn ensure_valid_msa_key(key: &T::AccountId) -> Result<MessageSourceId, DispatchError> {
-		let msa_id = Self::try_get_msa_from_public_key(key)?;
-
+		let msa_id = Self::get_msa_by_public_key(key).ok_or(Error::<T>::NoKeyExists)?;
```
shannonwells, line 1130: Even now it could use some simplification, but later.

enddynayn (approved):

pallets/msa/src/lib.rs
```diff
@@ -1125,18 +1113,17 @@ impl<T: Config> Pallet<T> {
...
+	// 		if let Ok(_info) = Self::ensure_valid_msa_key(&key) {
 	// 			response.push(KeyInfoResponse { key, msa_id });
 	// 		}
 	// 	}
 
 	// 	response
 	// }
 
-	/// Checks that a key is associated to an MSA and has not been revoked.
+	/// Retrieve MSA Id associated with `key` or return `NoKeyExists`
```
wilwade, line 1124: This is the new pattern for noting errors from a function:
```suggestion
	/// Retrieve MSA Id associated with `key`
	///
	/// # Errors
	/// * [`Error::NoKeyExists`]
	///
```

pallets/msa/src/lib.rs
```diff
@@ -1125,18 +1113,17 @@ impl<T: Config> Pallet<T> {
...
 	pub fn ensure_valid_msa_key(key: &T::AccountId) -> Result<MessageSourceId, DispatchError> {
-		let msa_id = Self::try_get_msa_from_public_key(key)?;
-
+		let msa_id = Self::get_msa_by_public_key(key).ok_or(Error::<T>::NoKeyExists)?;
```
wilwade, line 1126: Adding the new error here, it would be nice to bubble up the error listing for methods that call this one.

- `create_provider`
- `revoke_delegation_by_delegator`

wilwade (approved):
Two notes on documentation to update, but not blocking for merging

pallets/msa/src/lib.rs
wilwade: Moved to #613

pallets/msa/src/lib.rs
wilwade: Moved to #613

PR #617, opened 2022-11-02
fix(extrinsic): ensure signature includes key 

### LINKED ISSUE #620: Security Issues with add_public_key_to_msa
In review, `add_public_key_to_msa` was found to have two issues:
- [ ] The signature of the `new_key_owner_proof` was not being registered against replay attacks
- [ ] `add_key_payload` did not include the new key owner, allowing the `msa_owner_proof` to be swapped to a different `new_public_key` by the party submitting the transaction

enddynayn:
# Goal
Ensure the AddKeyData struct includes the new
public-key being added. Including the public-key
in the signature prevents a provider from
being able to add any key.

Additionally, register the proof for the new
public-key to prevent replay attacks.

Update benchmarks for add-public-key-to-msa
and delete-msa-public-key.

Closes #620

### DIFF
--- a/pallets/msa/src/benchmarking.rs
+++ b/pallets/msa/src/benchmarking.rs
@@ -2,6 +2,7 @@
 
 use super::*;
 
+use crate::types::EMPTY_FUNCTION;
 #[allow(unused)]
 use crate::Pallet as Msa;
 use frame_benchmarking::{account, benchmarks, whitelisted_caller};
@@ -45,13 +46,21 @@ fn create_payload_and_signature<T: Config>() -> (AddProvider, MultiSignature, T:
 	(add_provider_payload, MultiSignature::Sr25519(signature.into()), acc.into())
 }
 
-fn add_key_payload_and_signature<T: Config>() -> (AddKeyData, MultiSignature, T::AccountId) {
-	let account = SignerId::generate_pair(None);
-	let add_key_payload = AddKeyData { msa_id: 1u64.into(), expiration: 10 };
-	let encode_add_provider_data = wrap_binary_data(add_key_payload.encode());
+fn add_key_payload_and_signature<T: Config>(
+	msa_id: u64,
+) -> (AddKeyData<T>, MultiSignature, T::AccountId) {
+	let new_keys = SignerId::generate_pair(None);
+	let public_key = T::AccountId::decode(&mut &new_keys.encode()[..]).unwrap();
+	let add_key_payload = AddKeyData::<T> {
+		msa_id: msa_id.into(),
+		expiration: 10u32.into(),
+		new_public_key: public_key.into(),
+	};
 
-	let signature = account.sign(&encode_add_provider_data).unwrap();
-	let acc = T::AccountId::decode(&mut &account.encode()[..]).unwrap();
+	let encoded_add_key_payload = wrap_binary_data(add_key_payload.encode());
+
+	let signature = new_keys.sign(&encoded_add_key_payload).unwrap();
+	let acc = T::AccountId::decode(&mut &new_keys.encode()[..]).unwrap();
 	(add_key_payload, MultiSignature::Sr25519(signature.into()), acc.into())
 }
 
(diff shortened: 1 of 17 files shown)


pallets/msa/src/lib.rs
```diff
@@ -562,37 +562,43 @@ pub mod pallet {
 		#[pallet::weight(T::WeightInfo::add_public_key_to_msa())]
 		pub fn add_public_key_to_msa(
 			origin: OriginFor<T>,
-			msa_owner_public_key: T::AccountId,
+			delegator_key: T::AccountId,
```
wilwade, line 565: I don't think we want to confuse adding keys with delegating. I would suggest keeping the original name.

pallets/msa/src/lib.rs
```diff
@@ -562,37 +562,43 @@ pub mod pallet {
...
-				.map_err(|_| Error::<T>::AddKeySignatureVerificationFailed)?;
-
+			Self::register_signature(&msa_owner_proof, add_key_payload.expiration.into())?;
 			Self::register_signature(&new_key_owner_proof, add_key_payload.expiration.into())?;
 
-			let msa_id = add_key_payload.msa_id;
+			let delegator_msa_id = add_key_payload.msa_id;
 
-			Self::ensure_msa_owner(&msa_owner_public_key, msa_id)?;
+			Self::ensure_msa_owner(&delegator_key, delegator_msa_id)?;
```
wilwade, line 589: Same here. I think using delegator terminology here is confusing.

pallets/msa/src/types.rs
```diff
@@ -17,12 +17,15 @@ use scale_info::TypeInfo;
...
+#[derive(TypeInfo, RuntimeDebugNoBound, Clone, Decode, Encode, PartialEq, Eq)]
+#[scale_info(skip_type_params(T))]
+pub struct AddKeyData<T: Config> {
 	/// Message Source Account identifier
 	pub msa_id: MessageSourceId,
 	/// The block number at which the signed proof for add_public_key_to_msa expires.
-	pub expiration: BlockNumber,
+	pub expiration: T::BlockNumber,
+	/// The public key to be added.
+	pub new_public_key: T::AccountId,
```
wilwade, line 28: Could we skip the `T` by using `AccountId32` here instead?

enddynayn: how about we pass the generic params instead like so:

```
pub struct AddKeyData<BlockNumber, AccountId> {...}
```
This allows us to be consistent with grabbing the type from system.

wilwade (approved):
💯

pallets/messages/src/lib.rs
```diff
@@ -224,6 +224,7 @@ pub mod pallet {
 
 			Ok(Some(T::WeightInfo::add_ipfs_message(cid.len() as u32, message.index as u32)).into())
 		}
+
```
shannonwells, line 227: if there are no other changes, maybe revert this file

pallets/msa/src/benchmarking.rs
```diff
@@ -65,6 +74,15 @@ fn create_account_with_msa_id<T: Config>(n: u32) -> (T::AccountId, MessageSource
 	(provider.clone(), msa_id)
 }
 
+fn create_msa_account_and_keys<T: Config>() -> (T::AccountId, SignerId, MessageSourceId) {
```
shannonwells, line 77: ❤️

pallets/msa/src/replay_tests.rs
```diff
@@ -35,17 +35,17 @@ pub fn user_adds_key_to_msa(
 	delegator_pair: sp_core::sr25519::Pair,
 	new_pair: sp_core::sr25519::Pair,
 ) {
```
shannonwells, line 37: I thought there was a new replay test for the other signature, does there need to be one?

shannonwells (approved):
Just comments/questions, assuming there is adequate replay test coverage, nothing blocking

pallets/msa/src/replay_tests.rs
enddynayn: You mean this: https://github.com/LibertyDSNP/frequency/pull/617/files#diff-376fd47e95f7a80f67ffe3cf86e50bfd6d4d4d04dbdf54ed76c5a4b06b9c5b36R1609

PR #621, opened 2022-11-03
Chore/update api augment

### LINKED ISSUE #614: Bug: Api Augment Node Issue
API Augment doesn't work with node since we moved to the new build system. We need to add back in API Augment cjs support.

### LINKED ISSUE #615: Bug: Api Augment Custom RPC error
We missed in the renaming `grantedSchemaIds` -> `grantedSchemaIdsByMsaId` for api augment.

Should also double check the types.

wilwade:
# Goal
The goal of this PR is 

Closes #614
Closes #615 

# Discussion
- We don't yet have a way to catch broken rpc calls, but we should try to add that at some point.
- We also don't have any import tests for cjs, also something we should consider for the future.

# Checklist


enddynayn (approved):
💯

PR #638, opened 2022-11-04
disallow failure of extrinsic due to no msa

shannonwells:
# Goal
The goal of this PR is to ensure that `revoke_delegation_by_delegator` does nothing if there is no MSA. There is still a `?` inside the match but as with the other free extrinsics, these checks are performed in `CheckFreeExtensionUse` signed extension.

Relates to #605 

# Discussion
Simply follows the same pattern as the other extrinsics, with a match clause.  Also doesn't bother with `ensure_valid_msa_key`.

Additionally, updates the comments to follow previous PRs, in accordance with [Great Docs with Rustdoc](https://www.tangramvision.com/blog/making-great-docs-with-rustdoc)

# Checklist
## Not applicable:
- Chain spec updated
- Updated the js/api-augment code if a custom RPC added/changed
- Design doc(s) updated
- Tests added
- Benchmarks added
- Weights updated

### DIFF
--- a/pallets/msa/src/lib.rs
+++ b/pallets/msa/src/lib.rs
@@ -527,8 +527,7 @@ pub mod pallet {
 		) -> DispatchResult {
 			let who = ensure_signed(origin)?;
 
-			let res = Self::get_msa_by_public_key(&who);
-			match res {
+			match Self::get_msa_by_public_key(&who) {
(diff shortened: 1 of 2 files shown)


enddynayn (approved):
💯 Nice!

PR #639, opened 2022-11-04
sign binaries

### LINKED ISSUE #594: Frequency Versioning Scheme
Determine versioning scheme and filename conventions for Frequency releases.

- Production releases
- Test releases including release candidates
- Metadata versioning for portal

Semantic Versioning
https://semver.org

demisx:
# Goal
The goal of this PR is to sign released binaries and post `<binary>.asc` signatures separately next to each binary in a GitHub release.

Closes #594

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -86,7 +86,8 @@ jobs:
           sha256sum ${{env.RELEASE_BIN_FILENAME}} > ${{env.RELEASE_BIN_FILENAME}}.sha256
       - name: Import GPG key
         id: import-gpg
-        uses: crazy-max/ghaction-import-gpg@v5
+        # Pinning it to the exact version since this touches our private key
+        uses: crazy-max/ghaction-import-gpg@v5.2.0
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -84,13 +84,29 @@ jobs:
         run: |
           cd ${{env.BIN_DIR}}
           sha256sum ${{env.RELEASE_BIN_FILENAME}} > ${{env.RELEASE_BIN_FILENAME}}.sha256
-      - name: Upload Artifact
+      - name: Import GPG key
+        id: import-gpg
+        uses: crazy-max/ghaction-import-gpg@v5
```
wilwade, line 89: Since this touches our private key, we should likely pin it to the exact version.

wilwade (commented):
Just one note on pinning a version for security

wilwade (approved):

aramikm:
If rust builds are non deterministic how can a user check these signatures? Unless we are building and signing using that deterministic docker image?

PR #653, opened 2022-11-08
messages: minor fix #561

### LINKED ISSUE #561: Review use of unwrap() in non-test and non-benchmark code.

aramikm:
# Goal
related to #561

# Discussion
Fixes a minor edge case

### DIFF
--- a/pallets/messages/src/lib.rs
+++ b/pallets/messages/src/lib.rs
@@ -386,7 +386,8 @@ impl<T: Config> Pallet<T> {
 			let bounded_vec: BoundedVec<_, _> = messages.try_into().unwrap_or_default();
 
 			if bounded_vec.is_empty() {
-				return T::DbWeight::get().reads(1)
+				log::warn!("empty bounded_vec for schema id {}", schema_id);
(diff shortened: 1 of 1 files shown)


shannonwells (approved):
ty for catching this

enddynayn (approved):
Nice catch!

PR #628, opened 2022-11-03
fix(schemas): ensure schema permission 

rlaferla:
Ensure that schema grant validation works.
A schema permission is valid for a block-number if:
- a delegation relationship has not been revoked.
- a schema permission has not been revoked.

If a schema is revoked at block-number X:
- the schema is valid for the block-number X
  and prior.
- the schema is invalid after block-number X.

Add benchmarking helper to messages-pallet.
Adding a benchmarking helper facilitated the
removal of duplicating functionality for
benchmarks.

Update benchmarks.

issue-627

PR #656, opened 2022-11-08
600 - Add optional schema check to MSA check_delegations() RPC

### LINKED ISSUE #600: Add Optional Schema Check to `msa_checkDelegations`
As a consumer, I want to be able to check not just that a delegation exists, but that a specific schema id is delegated.

## Acceptance Criteria
Given: MSA A has granted 1,2,3 and MSA B has granted 3. Each to Provider X. MSA C hasn't delegated at all.
- [ ] I should be able to call the RPC `msa_checkDelegations` _with_ a schema id and it test down to the schema id (`X, [A,B,C], block, 3` -> `(A,true), (B,true), (C,false)`)
- [ ] I should be able to call the RPC `msa_checkDelegations` _without_ a schema id to test general delegation existence (`X, [A,B,C], block` -> `(A,true), (B,true), (C,false)`)

## Remember
- [x] Update the Custom RPC definition in js/api-augment
- [ ] Keep as much logic out of the runtime as possible
- [x] Block number is now required

## Updated RPC Interface
```
fn check_delegations(
  &self,
  delegator_msa_ids: Vec<MessageSourceId>,
  provider_msa_id: MessageSourceId,
  block_number:BlockNumber,
  schema_id: Option<SchemaId>,
) -> RpcResult<Vec<(MessageSourceId, bool)>>;
```

rlaferla:
# Goal
The goal of this PR is to add schema id validity checks to the check_delegations() RPC method.

Closes #600


pallets/msa/src/lib.rs
```diff
@@ -1128,18 +1130,30 @@ impl<T: Config> Pallet<T> {
...
+
+		if *schema_permission_revoked_at_block_number == T::BlockNumber::zero() {
+			return Ok(())
+		}
 
 		ensure!(
-			provider_info.schema_permissions.contains_key(&schema_id),
+			block_number <= *schema_permission_revoked_at_block_number,
 			Error::<T>::SchemaNotGranted
 		);
```
wilwade, line 1155: Suggestion: This feels like a perfect place for a match statement to make it clear the three outcomes.

```
match schema_permission_revoked_at_block_number {
	x if *x == T::BlockNumber::zero() => Ok(()),
	x if *x < block_number => Ok(()),
	_ => Err(Error::<T>::SchemaNotGranted.into())
} 
```

Bit sad we cannot just use ranges (`1..=block_number => Ok`), but BlockNumber would have to be converted first and that is more of a pain.

runtime/frequency-rococo/src/lib.rs
```diff
@@ -812,10 +812,20 @@ impl_runtime_apis! {
...
+						Ok(_) => true,
+						Err(_) => false,
+					}
+				},
+				None => {
+					match Msa::ensure_valid_delegation(provider, delegator, Some(block_number)) {
+						Ok(_) => true,
+						Err(_) => false,
+					}
+				},
```
wilwade, line 828: More code golf than anything else. Feel free to ignore
```suggestion
let result = match schema_id {
  Some(sid) => match Msa::ensure_valid_schema_grant(provider, delegator, sid, block_number),
  None => match Msa::ensure_valid_delegation(provider, delegator, Some(block_number))
};
match result {
  Ok(_) => true,
  Err(_) => false,
}
```

pallets/msa/src/runtime-api/src/lib.rs
```diff
@@ -31,7 +31,7 @@ sp_api::decl_runtime_apis! {
 		// fn get_msa_keys(msa_id: MessageSourceId) ->	Vec<KeyInfoResponse<AccountId>>;
 
 		/// Check to see if a delegation existed between the given delegator and provider at a given block
-		fn has_delegation(delegator: Delegator, provider: Provider, block_number: Option<BlockNumber>) -> bool;
+		fn has_delegation(delegator: Delegator, provider: Provider, block_number: BlockNumber, schema_id: Option<SchemaId>) -> bool;
```
wilwade, line 34: This needs to be reflected in `MsaRuntimeApi` under `js/api-augment/definitions/msa.ts` as well.

wilwade (changes requested):
Two ignorable suggestions and one fix for js/api-augment/definitions/msa.ts

Also we'll want to re-run the benchmarks before merging.

pallets/msa/src/runtime-api/src/lib.rs
rlaferla: Strange.  I do not see this in my code.  My code has:

`fn has_delegation(delegator: Delegator, provider: Provider, block_number: BlockNumber, schema_id: Option<SchemaId>) -> bool;`

pallets/msa/src/lib.rs
rlaferla: Interesting.

enddynayn (changes requested):
Just putting a blocker until PR [628](https://github.com/LibertyDSNP/frequency/pull/628) is merged.

enddynayn (approved):

PR #659, opened 2022-11-08
[Part1] Remove generic ```BlockNumber``` types to concrete types for client side 

saraswatpuneet:
# Goal
The goal of this PR is to Remove generic ```BlockNumber``` types to concrete types for client side 

Part of: 

# Checklist
- [ ] Chain spec updated
- [ ] Design doc(s) updated
- [ ] Tests added
- [ ] Benchmarks added
- [ ] Weights updated

### DIFF
--- a/pallets/messages/src/lib.rs
+++ b/pallets/messages/src/lib.rs
@@ -350,11 +350,12 @@ impl<T: Config> Pallet<T> {
 		schema_id: SchemaId,
 		schema_payload_location: PayloadLocation,
 		block_number: T::BlockNumber,
-	) -> Vec<MessageResponse<T::BlockNumber>> {
+	) -> Vec<MessageResponse> {
+		let block_number_value: u32 = block_number.try_into().unwrap_or_default();
 		<Messages<T>>::get(block_number, schema_id)
 			.into_inner()
 			.iter()
-			.map(|msg| msg.map_to_response(block_number, schema_payload_location))
+			.map(|msg| msg.map_to_response(block_number_value, schema_payload_location))
 			.collect()
 	}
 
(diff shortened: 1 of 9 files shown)


pallets/messages/src/lib.rs
```diff
@@ -350,11 +350,16 @@ impl<T: Config> Pallet<T> {
...
 		block_number: T::BlockNumber,
-	) -> Vec<MessageResponse<T::BlockNumber>> {
+	) -> Vec<MessageResponse> {
 		<Messages<T>>::get(block_number, schema_id)
 			.into_inner()
 			.iter()
-			.map(|msg| msg.map_to_response(block_number, schema_payload_location))
+			.map(|msg| {
+				msg.map_to_response(
+					block_number.try_into().unwrap_or_default(),
```
wilwade, line 359: Worried about unwrap or default being in the runtime. I wonder if this map function should be on the client side. Shouldn't change in this PR.

I know it is "impossible" for unwrap_or_default here to fail, but it is hiding an overflow error.

I've got an idea. Perhaps instead we move this out of the iter and do a check at the top and return an empty vec if that happens?

Really we should just move this mapping to the client (where it can also par_iter as this is just a mapping).

Thoughts?

wilwade (approved):
One separate thing that we should fix in a follow up.

pallets/messages/src/lib.rs
saraswatpuneet: Good idea just resolve it before the call and error if overflow

Good catch

saraswatpuneet: @wilwade  I have resolved the number outside however this is a tricky place, it will always resolve to default if error occurs which keep us safe

Still i moved it out

Issue #264, opened 2022-07-27
List of initial collators


shannonwells:
Blocked Pending Final Collator Aura and Session Keys

For launch on mainnet we'll need to create collator accounts.

- [x] Generate list of potential partners
- [x] Create Node list for Genesis
- [x] Set Mainnet aura keys for `invulnerables`

Related: #341 

- Unfinished Labs Collator 1
    - Controller: `0x40c0cf083ba1d081fce7c1f38f5344a4eba4ee5c13e1dddcb9cc6acf74559710`
    - Session: `0x7c63b545dc15066a4867f5d127b554d2bad49eab73303f550e99ea11d9ab123c`
- Unfinished Labs Collator 2
    - Controller: `0x1e5dc06d8418a0dd695bf3a23ecdc219103a149a863c1977396bf0697a474015`
    - Session: `0xa0e9a263e02d8b504808a67abbab811cb14fb376c0f79f5b7d6f7c2e7dcb585d`
- Unfinished Labs Collator 3
    - Controller: `0x7eb4de08f44e6077ed072e2fc357aac007150e3b73373434c47ec56323b52a75`
    - Session: `0x8062ffd04e84549b541668e4974f37f79f490ea846321e8a613c3406fceff565`
- OnFinality Collator 1
    - Controller: `0x3c9ea854901c1b36bd203400c90c31a69367316f7deaeeb14037f2be44bcb118`
    - Session: `0xb615416b0c34c5f3d1451a5d44390325832154ea20196b152ea5fa49346f5a30`
- External Collator `I`
    - Controller: `0x3c33f8c4c1155958d60d4d52a5181677da96052493d876b747272311b52cfe61`
    - Session: `0xbe16ebba3525a83e5f5e49cea331d5bc15c723d1dff7319b42837524b40c5970`
- External Collator `O`
    - Controller: `0x36523fba06e0f31f35dd904386a0ebe25d505ef09af4d39041afa1d994a23f5d`
    - Session: `0x8415832dcab4bcc60023222bbef2e222082ba69d53c9a870129e1c858a9ff87a`

wilwade:
Closed by #665

PR #665, opened 2022-11-11
add collator account and aura keys

saraswatpuneet:
# Goal
The goal of this PR is to update mainnet spec and  include invulnerable keys and their session keys 

Part of: #264 

# Checklist
- [x] Chain spec updated
- [ ] Design doc(s) updated
- [ ] Tests added
- [ ] Benchmarks added
- [ ] Weights updated

### DIFF
--- a/node/service/src/chain_spec/frequency.rs
+++ b/node/service/src/chain_spec/frequency.rs
@@ -24,17 +24,32 @@ pub mod frequency_mainnet_keys {
 	pub const MAINNET_FRQ_SUDO: &str =
 		"0xd43593c715fdd31c61141abd04a99fd6822c8558854ccde39a5684e7a56da27d"; // Alice
 
-	//TODO: final collator key(s) for mainnet
-	pub const COLLATOR_1_SR25519: &str =
-		"0xd43593c715fdd31c61141abd04a99fd6822c8558854ccde39a5684e7a56da27d"; // Alice
-	pub const COLLATOR_2_SR25519: &str =
-		"0x8eaf04151687736326c9fea17e25fc5287613693c912909cb226aa4794f26a48"; // Bob
-	pub const COLLATOR_3_SR25519: &str =
-		"0x90b5ab205c6974c9ea841be688864633dc9ca8a357843eeacf2314649965fe22"; // Charlie
-	pub const COLLATOR_4_SR25519: &str =
-		"0x306721211d5404bd9da88e0204360a1a9ab8b87c66c1bc2fcdd37f3c2222cc20"; // Dave
-	pub const COLLATOR_5_SR25519: &str =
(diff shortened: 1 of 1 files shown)


node/service/src/chain_spec/frequency.rs
```diff
@@ -56,13 +71,28 @@ pub fn frequency() -> ChainSpec {
 				// TODO: initial collators.
```
wilwade, line 71:
```suggestion
```

node/service/src/chain_spec/frequency.rs
```diff
@@ -24,17 +24,32 @@ pub mod frequency_mainnet_keys {
...
-		"0xd43593c715fdd31c61141abd04a99fd6822c8558854ccde39a5684e7a56da27d"; // Alice
-	pub const COLLATOR_2_SR25519: &str =
-		"0x8eaf04151687736326c9fea17e25fc5287613693c912909cb226aa4794f26a48"; // Bob
-	pub const COLLATOR_3_SR25519: &str =
-		"0x90b5ab205c6974c9ea841be688864633dc9ca8a357843eeacf2314649965fe22"; // Charlie
-	pub const COLLATOR_4_SR25519: &str =
-		"0x306721211d5404bd9da88e0204360a1a9ab8b87c66c1bc2fcdd37f3c2222cc20"; // Dave
-	pub const COLLATOR_5_SR25519: &str =
-		"0xe659a7a1628cdd93febc04a4e0646ea20e9f5f0ce097d9a05290d4a9e054df4e"; // Eve
+	//TODO: update collator key naming convention
```
wilwade, line 27: I think this is fine. Perhaps move the session key next to the corresponding controller key, but otherwise good enough to produce the genesis. Then it is removed eventually anyway.

wilwade (changes requested):
Two todos to clean up, otherwise good to go.

node/service/src/chain_spec/frequency.rs
saraswatpuneet: I think I like that one

node/service/src/chain_spec/frequency.rs
```diff
@@ -24,17 +24,44 @@ pub mod frequency_mainnet_keys {
...
+
+	// Unfinished Collator 1 public key (sr25519) and session key
+	pub const UNFINISHED_COLLATOR_1_SR25519: &str =
+		"0x40c0cf083ba1d081fce7c1f38f5344a4eba4ee5c13e1dddcb9cc6acf74559710"; // Unfinished Labs Collator 1
+	pub const UNFINISHED_COLLATOR_1_SESSION_KEY: &str =
+		"0x7c63b545dc15066a4867f5d127b554d2bad49eab73303f550e99ea11d9ab123c"; // Unfinished Labs Collator 1
+
```
saraswatpuneet, line 34: @wilwade  how about like this, keeping similarly related ones together

PR #671, opened 2022-11-11
660 correct benchmarking for retire msa

### LINKED ISSUE #660: Correct benchmarking for retire_msa()
The retire MSA functionality is not properly accounting for resources used.  Furthermore, the MSA pallet is not placing limits on the number of providers a delegator can have.

Acceptance criteria
- [x] Create a constant for the maximum # of providers for a given delegator
- [x] Change `delete_delegation_relationship()` to return the number of providers deleted as returned by the `MultiRemovalResults` from `clear_prefix()`
- [ ] Change the `retire_msa` benchmark to account for the number of delegations

See also:

https://docs.substrate.io/test/benchmark
https://www.youtube.com/watch?v=i3zW4wGexAc
https://drive.google.com/file/d/1iSzxov5qrLaLPI2g3wMNdpmF-wdsQDP9/view?usp=sharing

rlaferla:
# Goal
The goal of this PR is to improve benchmarking for retire_msa()

Closes #660


enddynayn (approved):
Nice weight updates! 💯

pallets/msa/src/benchmarking.rs
```diff
@@ -157,16 +157,20 @@ benchmarks! {
...
 		let caller: T::AccountId = whitelisted_caller();
-
-		// Create a MSA account
-		assert_ok!(Msa::<T>::create(RawOrigin::Signed(caller.clone()).into()));
-		let msa_id = Msa::<T>::ensure_valid_msa_key(&caller).unwrap();
-
-		assert_eq!(Msa::<T>::is_registered_provider(msa_id),false);
-
-	}: _(RawOrigin::Signed(caller))
+		assert_ok!(Msa::<T>::add_key(ProviderId(1).into(), &caller.clone(), EMPTY_FUNCTION));
```
aramikm, line 161: I think you will need to have something like
```
let n in 2 .. MAX_NUMBER_OF_PROVIDERS_PER_DELEGATOR;
for j in 2..=n {
    assert_ok!(Msa::<T>::add_provider(ProviderId(j.into()), DelegatorId(1), vec![1, 2]));
}
```

aramikm: this will allow the benchmark function to find the multiplier correctly instead of using the maximum as a constant

aramikm (changes requested):
Added a comment which is very important

pallets/msa/src/benchmarking.rs
enddynayn: We were talking about something like this but we were not certain about it. Can you explain what you mean?
Why not use a constant since it is the worst-case scenario?

>You can then use the results of the benchmarking tests to establish an approximate worst case weight to represent the resources required to execute each function call and each code path

https://docs.substrate.io/test/benchmark/

aramikm: Because this worst case is not accurate. Calculated weights are substantially bigger than the correct one and it is unnecessarily consuming a huge chunk of block size.

pallets/msa/src/weights.rs
```diff
@@ -88,87 +88,88 @@ impl<T: frame_system::Config> WeightInfo for SubstrateWeight<T> {
...
 	}
 	// Storage: Msa PublicKeyToMsaId (r:1 w:1)
 	// Storage: Msa PublicKeyCountForMsaId (r:1 w:1)
+	// Storage: Msa DelegatorAndProviderToDelegation (r:0 w:126)
 	fn retire_msa() -> Weight {
-		Weight::from_ref_time(31_300_000 as u64)
+		Weight::from_ref_time(186_146_000 as u64)
 			.saturating_add(T::DbWeight::get().reads(2 as u64))
-			.saturating_add(T::DbWeight::get().writes(2 as u64))
+			.saturating_add(T::DbWeight::get().writes(128 as u64))
```
aramikm, line 125: @enddynayn writes here should be dependent on a param not being `128` on every call

pallets/msa/src/benchmarking.rs
enddynayn: I think that weight gets refunded after the extrinsic runs.

aramikm: I don't think it is.

enddynayn: https://github.com/paritytech/substrate/blob/master/frame/system/src/extensions/check_weight.rs#L246

aramikm: I don't think that is related for this case

pallets/msa/src/weights.rs
enddynayn: Yeah. we probably do not want this.

pallets/msa/src/benchmarking.rs
aramikm: `calc_unspent` Calculate how much (if any) weight was not used by the `Dispatchable`. But we are not specifying how much actually we used in our return

enddynayn: You are right maybe we should be returning in this case. Also, perhaps if we knew how many delegations needed to be deleted it would make for better estimates.

enddynayn: I think this is a good argument to keep count of how many delegation one has.

pallets/msa/src/lib.rs
```diff
@@ -1090,8 +1091,13 @@ impl<T: Config> Pallet<T> {
...
 	/// Removes all delegations from the specified delegator MSA id to providers
-	pub fn delete_delegation_relationship(delegator: DelegatorId) {
-		_ = DelegatorAndProviderToDelegation::<T>::clear_prefix(delegator, u32::max_value(), None);
+	pub fn delete_delegation_relationship(delegator: DelegatorId) -> u32 {
+		let result = DelegatorAndProviderToDelegation::<T>::clear_prefix(
+			delegator,
+			MAX_NUMBER_OF_PROVIDERS_PER_DELEGATOR,
+			None,
+		);
+		result.backend
```
aramikm, line 1100: I'm not sure this should be backend. I think it should be unique but I'm not certain

enddynayn:
Can you please update `retire_msa` to return the actual weight consumed.

pallets/msa/src/benchmarking.rs
enddynayn: I am thinking. Maybe we ask the user to provide how many delegations they have + a signed extension to validate the provided count vs the actual count. We drop the transaction if it does not match.

pallets/msa/src/lib.rs
rlaferla: I've modified to return the number of unique so let's re-evaluate it.

pallets/msa/src/benchmarking.rs
aramikm: What would be the benefit of doing that?

pallets/msa/src/lib.rs
```diff
@@ -780,16 +781,17 @@ pub mod pallet {
 		/// # Errors
 		/// * [`Error::NoKeyExists`] - `delegator` does not have an MSA key.
 		///
-		#[pallet::weight((T::WeightInfo::retire_msa(), DispatchClass::Normal, Pays::No))]
-		pub fn retire_msa(origin: OriginFor<T>) -> DispatchResult {
+		#[pallet::weight((T::WeightInfo::retire_msa(10_000), DispatchClass::Normal, Pays::No))]
```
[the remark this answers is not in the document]
rlaferla: Updated.

pallets/msa/src/benchmarking.rs
```diff
@@ -157,16 +157,21 @@ benchmarks! {
 	}: _(RawOrigin::Signed(caller_and_delegator_public_key), new_public_key)
 
 	retire_msa {
-
+		let s in 5 .. MAX_NUMBER_OF_PROVIDERS_PER_DELEGATOR;
 		let caller: T::AccountId = whitelisted_caller();
-
-		// Create a MSA account
-		assert_ok!(Msa::<T>::create(RawOrigin::Signed(caller.clone()).into()));
```
aramikm, line 164: don't you need to create an MSA first to be able to retire it?

pallets/msa/src/lib.rs
```diff
@@ -780,16 +781,17 @@ pub mod pallet {
 		/// # Errors
 		/// * [`Error::NoKeyExists`] - `delegator` does not have an MSA key.
 		///
-		#[pallet::weight((T::WeightInfo::retire_msa(), DispatchClass::Normal, Pays::No))]
-		pub fn retire_msa(origin: OriginFor<T>) -> DispatchResult {
+		#[pallet::weight((T::WeightInfo::retire_msa(128), DispatchClass::Normal, Pays::No))]
```
aramikm, line 784:
```suggestion
		#[pallet::weight((T::WeightInfo::retire_msa(MAX_NUMBER_OF_PROVIDERS_PER_DELEGATOR), DispatchClass::Normal, Pays::No))]
```

aramikm (commented):
Added 2 comments

pallets/msa/src/benchmarking.rs
rlaferla: I think we were taking a shortcut here by using `Delegator(1)` (MSA id 1) in the `add_provider` call.  I could create it with `create_account_with_msa_id()` if you think that is better.

common/primitives/src/msa.rs
```diff
@@ -16,6 +16,9 @@ pub use crate::schema::SchemaId;
 /// Message Source Id or msaId is the unique identifier for Message Source Accounts
 pub type MessageSourceId = u64;
 
+/// Maximum # of providers per delegator for deletion/benchmarking use
+pub const MAX_NUMBER_OF_PROVIDERS_PER_DELEGATOR: u32 = 128;
```
enddynayn, line 20: maybe we pass this as a config into msa

pallets/msa/src/benchmarking.rs
enddynayn: This could potentially block other transactions from making it into the block. 128 writes a lot.

common/primitives/src/msa.rs
rlaferla: I had considered that but Wil and I decided to use a constant.  I can easily change it if you want.

rlaferla: The rationale for using a constant is that there is no logic on `add_provider` to enforce it.  It’s just used for benchmarking.

rlaferla:
```
		/// The maximum number of providers per delegator
		#[pallet::constant]
		type MaxNumberOfProvidersPerDelegator: Get<u32>;
```

aramikm (approved):
Looks great!

enddynayn (approved):

rlaferla:
Merging but updated version with Config will be in a new PR.

PR #676, opened 2022-11-14
660 correct benchmarking for retire msa

rlaferla:
# Goal
The goal of this PR is to improve benchmarking for retire MSA by making MaxNumberOfProvidersPerDelegator a pallet constant.


pallets/msa/src/lib.rs
```diff
@@ -145,6 +145,10 @@ pub mod pallet {
 		/// There are exactly NumberOfBuckets first-key values in PayloadSignatureRegistry.
 		#[pallet::constant]
 		type NumberOfBuckets: Get<u32>;
+
+		/// The maximum number of providers per delegator
+		#[pallet::constant]
+		type MaxNumberOfProvidersPerDelegator: Get<u16>;
```
saraswatpuneet, line 151: Are we restricting every msa user to only be able to delegate to n number of providers?

Why?

rlaferla: No

rlaferla: It's for benchmarking primarily.

PR #684, opened 2022-11-16
messages: seed schema for benchmarks

### LINKED ISSUE #657: Remove `#[cfg(not(feature = "runtime-benchmarks"))]` from Schema Pallet
We use `#[cfg(not(feature = "runtime-benchmarks"))]` in three places to mock the code for benchmarks, but when we run with the `--all-features` flag, it turns it on breaking tests.

We must remove `#[cfg(not(feature = "runtime-benchmarks"))]` from the code base:

- [ ] #628 by @enddynayn  due to it being required for testing there: pallet_msa `ensure_valid_schema_grant`
- [ ] #628 by @enddynayn  pallet_msa `ensure_valid_msa_key`
- [ ] THIS ISSUE: pallet_schemas `get_schema_by_id`

Reference: https://doc.rust-lang.org/cargo/reference/features.html#mutually-exclusive-features

aramikm:
# Goal
The goal of this PR is remove `#[cfg(not(feature = "runtime-benchmarks"))]` and seeds schema related storage for benchmarks

Closes #657 

# Discussion

- removes `#[cfg(not(feature = "runtime-benchmarks"))]`
- modify benchmarks on messages to use seeding utility

# Checklist
- [x] Benchmarks added
- [x] Weights updated

### DIFF
--- a/pallets/messages/src/lib.rs
+++ b/pallets/messages/src/lib.rs
@@ -69,7 +69,7 @@ use common_primitives::{
 };
 
 #[cfg(feature = "runtime-benchmarks")]
-use common_primitives::benchmarks::BenchmarkHelper;
+use common_primitives::benchmarks::{MsaBenchmarkHelper, SchemaBenchmarkHelper};
 
 pub use pallet::*;
 pub use types::*;
@@ -108,7 +108,11 @@ pub mod pallet {
 
 		#[cfg(feature = "runtime-benchmarks")]
 		/// A set of helper functions for benchmarking.
-		type Helper: BenchmarkHelper<Self::AccountId>;
+		type MsaBenchmarkHelper: MsaBenchmarkHelper<Self::AccountId>;
+
+		#[cfg(feature = "runtime-benchmarks")]
+		/// A set of helper functions for benchmarking.
+		type SchemaBenchmarkHelper: SchemaBenchmarkHelper;
 	}
 
 	#[pallet::pallet]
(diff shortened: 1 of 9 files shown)


pallets/messages/src/lib.rs
```diff
@@ -108,7 +108,11 @@ pub mod pallet {
 
 		#[cfg(feature = "runtime-benchmarks")]
 		/// A set of helper functions for benchmarking.
-		type Helper: BenchmarkHelper<Self::AccountId>;
```
enddynayn, line 111: I was hoping we only needed to have one helper here.

I wonder if we could do something like
```
pub struct BenchmarkHelperThing<SomeMSAThing, SomeSchemaThing>(PhantomData<(SomeMSAThing, SomeSchemaThing)>);

impl<SomeMSAGen, SomeSchemaGen, AccountId> BenchmarkHelper<AccountId> where
where SomeMSAGen:  ...,

{

}
```




```

enddynayn (commented):
Looking good 👍

pallets/messages/src/lib.rs
aramikm: @enddynayn  I went that route but got stuck on an error and it got too complicated at that point, so I think this is simple enough. Your are welcome if can make it work without making it too complicated.

enddynayn: Sounds good. Rust is not easy. This is an improvement from what we have now.

enddynayn (approved):
💯 Nice improvements!

shannonwells (approved):
Looks good to me

PR #687, opened 2022-11-16
Generate mainnet genesis spec

### LINKED ISSUE #273: [Final] Generate Genesis spec
## Blocked by

- [x] #263
- [x] #264 
- [x] #270 
- [x] #271 
- [x] #347  
- [x] #272
- [x] #365 

## Tasks
- [x] Use srtooll to generate the mainnet production wasm
    - [x] Do this with a pair on two different machines to verify deterministic success
- [x] Generate the mainnet spec and use the wasm as the code value
- [x] Commit the spec file to `/resources/mainnet-frequency-genesis-spec.json`
- [x] Use [`load_frequency_spec`](https://github.com/LibertyDSNP/frequency/blob/d70911165395f71d17c0d6eb4e572daa5782d242/node/service/src/chain_spec/frequency.rs#L38-L40) to load up `mainnet-frequency-genesis-spec.json` and replace the contents of `pub fn frequency()` to use `load_frequency_spec`
- [x] Remove mainet spec generation (but not the WASM) from GitHub Actions
- [x] Remove mainnet as an option from `generate_specs.sh`

shannonwells:
# Goal
The goal of this PR is 

Closes #273 

# Checklist
- [x] Chain spec updated
- [x] bump cargo versions
- [x] add generated chain spec in mainnet
- [x] use the generated chain spec in mainnet


saraswatpuneet (approved):
🪙

node/service/src/chain_spec/frequency.rs
```diff
@@ -111,9 +111,9 @@ pub mod foundation_keys {
...
 }
 
-// pub fn load_frequency_spec() -> Result<ChainSpec, String> {
-// 	ChainSpec::from_json_bytes(&include_bytes!("../../specs/frequency.json")[..])
-// }
+pub fn load_frequency_spec() -> ChainSpec {
+	ChainSpec::from_json_bytes(&include_bytes!("../../../../resources/frequency.json")[..]).unwrap()
+}
 
 pub fn frequency() -> ChainSpec {
```
saraswatpuneet, line 118: Want to delete this function?

shannonwells: @wilwade decided not yet, we will take it out in an update even though it's not being used.

wilwade: Yes. We will take it out in a future PR. I wanted to make sure there was a commit that we could reference where both existed.

PR #691, opened 2022-11-16
chore(benchmarks): update msa create benchmark

### LINKED ISSUE #668: Review Benchmarks: Msa pallet
We need to review and improve our benchmarks on Msa pallet. Currently in case of failures we are allocating the maximum possible weight but we need to calculate and reduce only the processed amount.

- [x] Review the happy path
- [x] Remove all magic numbers
- [x] Review the failure paths and add more accurate benchmarks for different failures
- [x] Speed up slow benchmarks (ex: `on_initialize`)

Update:
The following has been pushed to a [different](https://github.com/LibertyDSNP/frequency/issues/812) issue:
> Currently in case of failures we are allocating the maximum possible weight but we need to calculate and reduce only the processed amount.

enddynayn:
# Goal
Update Msa pallet "create" extrinsic benchmark.
Remove the complexity parameter since it has no
parameters that can affect weights. Additionally,
there are no bounds to the number of Msa accounts
that can be created.

Update benchmark files.

References #668

### DIFF
--- a/pallets/msa/src/benchmarking.rs
+++ b/pallets/msa/src/benchmarking.rs
@@ -26,11 +26,6 @@ fn create_account<T: Config>(name: &'static str, index: u32) -> T::AccountId {
 	account(name, index, SEED)
 }
 
-fn create_msa<T: Config>(n: u32) -> DispatchResult {
-	let acc = create_account::<T>("account", n);
-	Msa::<T>::create(RawOrigin::Signed(acc.clone()).into())
-}
-
 fn create_payload_and_signature<T: Config>() -> (AddProvider, MultiSignature, T::AccountId) {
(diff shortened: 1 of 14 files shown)


aramikm (approved):
Looks good!

pallets/msa/src/benchmarking.rs
```diff
@@ -101,12 +96,8 @@ fn register_signature<T: Config>(mortality_block: u32) {
 
 benchmarks! {
 	create {
-		let s in 1 .. 1000;
```
enddynayn, line 104: add verify {}

shannonwells (approved):
💯 percent pure code

shannonwells (approved):

PR #707, opened 2022-11-17
fix(cli): ensure benchmarks run

enddynayn:
# Goal
When running benchmarks, the release
chain specs are used and cause an error because
the specs do not match the build. New code changes drive the mismatch.

Ensure that the latest chain specs are used
when running benchmarks.


wilwade (approved):

shannonwells (approved):
shipit

PR #625, opened 2022-11-03
Fix/delete msa public key free tx issue 605

enddynayn:
# Goal
The goal of this PR is 

Closes 

# Discussion

# Checklist
- [ ] Chain spec updated
- [ ] Updated the js/api-augment code if a custom RPC added/changed
- [ ] Design doc(s) updated
- [ ] Tests added
- [ ] Benchmarks added
- [ ] Weights updated

PR #709, opened 2022-11-18
Add a cli command to extract runtime version to be used in important release checks

saraswatpuneet:
# Goal
The goal of this PR is Add a cli command to extract runtime version to be used in important release checks
``` 
 ./frequency export-runtime-version
```

part of: #625 

# Checklist
- [ ] Chain spec updated
- [ ] Design doc(s) updated
- [ ] Tests added
- [ ] Benchmarks added
- [ ] Weights updated


node/cli/src/runtime_version_cmd.rs
```diff
@@ -0,0 +1,48 @@
...
+		} else if cfg!(feature = "frequency-rococo-testnet") ||
+			cfg!(feature = "frequency-rococo-local")
+		{
+			frequency_service::service::frequency_rococo_runtime::VERSION
+		} else {
+			return Err(Error::from("No runtime version found"))
+		};
+		Ok(version)
+	}
+}
```
saraswatpuneet, line 42: Based on the activated feature in cli , runtime version will be dumped out

saraswatpuneet:
image

node/cli/src/runtime_version_cmd.rs
```diff
@@ -0,0 +1,60 @@
...
+		let result = json!(runtime_version);
+		let file: Box<dyn io::Write> = match &self.output {
+			Some(filename) => Box::new(fs::File::create(filename)?),
+			None => Box::new(io::stdout()),
+		};
+		to_writer(file, &result).map_err(|_| Error::from("Failed Encoding"))
+	}
+
+	pub fn read_runtime_version(&self) -> Result<RuntimeVersion, Error> {
+		if cfg!(feature = "frequency") {
```
enddynayn, line 31: those something like this workd

```
	#[cfg(feature = "frequency")]
			{
				return Ok(frequency_service::service::frequency_runtime::VERSION)
			}
			
	#[cfg(any(feature = "frequency-rococo-testnet", feature = "frequency-rococo-local"))]
			{
				return Ok(frequency_service::service::frequency_rococo_runtime::VERSION)
			}
		#[cfg(not(any(
				feature = "frequency-rococo-testnet",
				feature = "frequency-rococo-local"
			)))]
			panic!("Frequency rococo feature is not enabled")
	```

File: node/cli/src/command.rs [hunk 3]
```diff
@@ -470,6 +470,22 @@ pub fn run() -> Result<()> {
...
+			let runner = cli.create_runner(cmd)?;
+			let chain_spec = &runner.config().chain_spec;
+			with_runtime_or_err!(chain_spec, {
+				{
+					runner.async_run(|config| {
+						// grab the task manager.
+						let registry = config.prometheus_config.as_ref().map(|cfg| &cfg.registry);
+						let task_manager =
+							sc_service::TaskManager::new(config.tokio_handle.clone(), registry)
+								.map_err(|e| format!("Error: {:?}", e))?;
```
enddynayn, line 483: Maybe a little more description of the error. What kinds of errors can occur?

enddynayn (commented):
Looking good! 👍

node/cli/src/command.rs
```diff
@@ -470,6 +470,22 @@ pub fn run() -> Result<()> {
 				Err("Try-runtime must be enabled by `--features try-runtime`.".into())
 			}
 		},
+		Some(Subcommand::ExportRuntimeVersion(cmd)) => {
+			let runner = cli.create_runner(cmd)?;
+			let chain_spec = &runner.config().chain_spec;
+			with_runtime_or_err!(chain_spec, {
+				{
+					runner.async_run(|config| {
+						// grab the task manager.
```
enddynayn, line 479: doesn't the change spec already have the runtime version?

node/cli/src/runtime_version_cmd.rs
saraswatpuneet: Found a shorted one

node/cli/src/command.rs
saraswatpuneet: We are returning that error from command function

node/cli/src/command.rs
saraswatpuneet: Not sure what it means, with_runtime_or_err macro requires chain spec

saraswatpuneet: We cant do that as we are not invoking runtime_api but get it directly from file, to access native_runtime_api, i have to create client etc which is not needed for this case as we want to dump out runtime version

Chain spec is used by the macro to identity the current chain identity and import correct runtime code for the same, given we hide different runtimes behind different flags

node/cli/src/runtime_version_cmd.rs
```diff
@@ -0,0 +1,48 @@
...
+		let result = json!(runtime_version);
+		let file: Box<dyn io::Write> = match &self.output {
+			Some(filename) => Box::new(fs::File::create(filename)?),
+			None => Box::new(io::stdout()),
+		};
+		to_writer(file, &result).map_err(|_| Error::from("Failed Encoding"))
+	}
+
+	#[allow(unreachable_code)]
+	pub fn read_runtime_version(&self) -> Result<RuntimeVersion, Error> {
```
enddynayn, line 31: can we use?
https://github.com/LibertyDSNP/frequency/pull/709/files#diff-ced5116f173c1d100d5f2d15b398a258b5c129b3eb8acbc9fb6e6ea876e9de79L205

saraswatpuneet: @enddynayn  cant seem to find that diff

saraswatpuneet: But I believe you are referring to native_runtime_version?

We cant access that without invoking runtime api which gives access to runtime via api which we dont need for this case, as we can read directly from runtime gives those values are static

saraswatpuneet: I think i found another approach

saraswatpuneet: Check now

node/cli/src/command.rs
saraswatpuneet: @enddynayn  i was refactoring some of it, let me know if all looks good now

node/cli/src/runtime_version_cmd.rs
```diff
@@ -0,0 +1,34 @@
...
+
+impl ExportRuntimeVersionCmd {
+	/// Run the get-runtime-version command.
+	pub async fn run(&self, runtime_version: &RuntimeVersion) -> Result<(), Error> {
+		let result = json!(runtime_version);
+		let file: Box<dyn io::Write> = match &self.output {
+			Some(filename) => Box::new(fs::File::create(filename)?),
+			None => Box::new(io::stdout()),
+		};
+		to_writer(file, &result).map_err(|_| Error::from("Failed Encoding"))
```
enddynayn, line 26: maybe this is ok to_writer(file, &result)?

enddynayn (approved):
Fantastic! ⛹️‍♂️

aramikm (approved):
Looks good to me!

PR #713, opened 2022-11-23
chore(msa): update benchmarks

enddynayn:
# Goal
Update benchmarks for grant_delegation
to use max schema grants per delegation
as a complexity parameter.

Refactor revoke_delegation_by_delegator
benchmark to not use a helper function.

References #668


pallets/msa/src/benchmarking.rs
```diff
@@ -156,6 +156,9 @@ benchmarks! {
 		assert_ok!(Msa::<T>::add_public_key_to_msa(RawOrigin::Signed(provider_public_key).into(), caller_and_delegator_public_key.clone(), owner_signature,  new_public_key_signature, add_key_payload));
 
 	}: _(RawOrigin::Signed(caller_and_delegator_public_key), new_public_key)
+	verify {
+		assert_eq!(frame_system::Pallet::<T>::events().len(), 1);
```
aramikm, line 160: Is this to verify if the extrinsic is successful?  If that's the case this should be fine for now but it will break if we add more events later. It seems the more correct thing to do is to assert the call, but that seems to be not compatible with the shorten macro.

pallets/msa/src/lib.rs
```diff
@@ -519,7 +519,7 @@ pub mod pallet {
 		/// * [`Error::ProviderNotRegistered`] - the a non-provider MSA is used as the provider
 		/// * [`Error::UnauthorizedDelegator`] - Origin attempted to add a delegate for someone else's MSA
 		///
-		#[pallet::weight(T::WeightInfo::grant_delegation())]
+		#[pallet::weight(T::WeightInfo::grant_delegation(T::MaxSchemaGrantsPerDelegation::get()))]
```
aramikm, line 522:
```suggestion
		#[pallet::weight(T::WeightInfo::grant_delegation(add_provider_payload.schema_ids.len() as u32))]
```

aramikm (commented):
Overall looks good but the parameter passed in to weight will need to change

shannonwells (approved):
💯

pallets/msa/src/benchmarking.rs
enddynayn: Yup. It is to check that we are not accidentally erroring out early and, as a result, not benchmarking what we want. I think this can be updated with a specific log message when we update.

aramikm (approved):
Looks great. Ship it.

saraswatpuneet (approved):
🫴

Issue #667, opened 2022-11-11
Review benchmarks: Messages pallet


aramikm:
We need to review and improve our benchmarks on Messages pallet. Currently in case of failures we are allocating the maximum possible weight but we need to calculate and reduce only the processed amount.

- [x] Review the happy path
- [x] Remove all magic numbers
- [x] Speed up slow benchmarks

PR #717, opened 2022-11-23
[Experiment] Messages: added on chain failure benchmarks

aramikm:
# Goal
The goal of this PR is introduce a pattern to do failure benchmarks for extrinsic with dynamic params

Related to Issue #667 

# Discussion
- Only updates `add_onchain_message` weights
- Instead of returning mapped errors with weights when it happens, tries to map them in the end of execution to not cause distraction in the execution logic

# Checklist
- [x] Tests updated
- [x] Benchmarks added
- [x] Weights updated


pallets/messages/src/lib.rs
```diff
@@ -252,57 +255,61 @@ pub mod pallet {
...
 			origin: OriginFor<T>,
 			on_behalf_of: Option<MessageSourceId>,
 			#[pallet::compact] schema_id: SchemaId,
 			payload: Vec<u8>,
 		) -> DispatchResultWithPostInfo {
-			let provider_key = ensure_signed(origin)?;
-
-			let bounded_payload: BoundedVec<u8, T::MaxMessagePayloadSizeBytes> =
-				payload.try_into().map_err(|_| Error::<T>::ExceedsMaxMessagePayloadSizeBytes)?;
+			let result = || -> DispatchResultWithPostInfo {
```
aramikm, line 265: encloses the logic in a closure so we can map errors in the end

pallets/messages/src/lib.rs
```diff
@@ -417,4 +424,43 @@ impl<T: Config> Pallet<T> {
...
+			error,
+		}
+	}
+
+	fn map_on_chain_errors(result: DispatchResultWithPostInfo) -> DispatchResultWithPostInfo {
+		match result {
+			Ok(p) => Ok(p),
+			Err(e) => {
+				if let DispatchError::Module(ModuleError { message: Some(err_msg), .. }) = e.error {
+					if err_msg == Error::<T>::ExceedsMaxMessagePayloadSizeBytes.as_str() {
```
aramikm, line 443: I wished I could replace these if elses with match but couldn't find an easy way.

aramikm: These errors are not exhaustive but takes care of most of use cases and we can add more later

pallets/messages/src/mock.rs
```diff
@@ -26,7 +28,8 @@ type Block = frame_system::mocking::MockBlock<Test>;
 
 pub const INVALID_SCHEMA_ID: SchemaId = 65534;
 pub const IPFS_SCHEMA_ID: SchemaId = 50;
-
+// following number is not random it is the account_id of INVALID_CALLER_NAME in benchmarking.rs
+pub const INVALID_CALLER_ACCOUNT: u64 = 15032055221142339925;
```
aramikm, line 32: not sure how substrate maps string account names to account ids, but it seems like there is a 1-1 relationship between them. Found this by logging what was coming back

pallets/messages/src/benchmarking.rs
```diff
@@ -77,6 +80,45 @@ benchmarks! {
...
+		let m in 1 .. MESSAGES;
+		let message_source_id = DelegatorId(2);
+		let caller: T::AccountId = whitelisted_caller();
+		let schema_id = 65534;
+		let payload = vec![1; n as usize];
+	}: {
+		assert_err_ignore_postinfo!(MessagesPallet::<T>::add_onchain_message(RawOrigin::Signed(caller).into(), Some(message_source_id.into()), schema_id, payload), Error::<T>::InvalidSchemaId);
+	}
+
+	add_onchain_message_invalid_account {
```
enddynayn, line 105: I like this idea. However, I think it would be ok to add one additional error and not 3.

PR #720, opened 2022-11-24
chore(msa): update benchmarks

enddynayn:
# Goal
Update MSA benchmarks:
- revoke_schema_permissions
- grant_schema_permissions

Ensure that both use schema_ids vector length
as complexity parameters.

Remove helper methods create_account_with_msa_id,
add_delegation, and register_signature so that that
they are called inline with function.

Update weights for all benchmarks.

References #668


enddynayn:
commit message needs to be updated but ready for review.

pallets/msa/src/benchmarking.rs
```diff
@@ -196,64 +178,76 @@ benchmarks! {
...
 		assert_eq!(frame_system::Pallet::<T>::events().len(), 1);
+		assert_eq!(Msa::<T>::get_delegation(DelegatorId(delegator_msa_id), ProviderId(provider_msa_id)).unwrap().revoked_at, T::BlockNumber::from(1u32));
 	}
 
 	create_provider {
-		let (provider, _provider_msa_id) = create_account_with_msa_id::<T>(1);
-	}: _ (RawOrigin::Signed(provider), Vec::from("Foo"))
+		let account = create_account::<T>("account", 0);
+		let (provider_msa_id, provider_public_key) = Msa::<T>::create_account(account.into(), EMPTY_FUNCTION).unwrap();
+	}: _ (RawOrigin::Signed(provider_public_key), Vec::from("Foo"))
```
aramikm, line 197: Size of provider name should be a weight param.

pallets/msa/src/benchmarking.rs
```diff
@@ -196,64 +178,76 @@ benchmarks! {
...
+	verify {
+		assert!(Msa::<T>::get_provider_registry_entry(ProviderId(provider_msa_id)).is_some());
+	}
 
 	on_initialize {
```
aramikm, line 202: Since you are touching this here I think you should fix this. This benchmark runs very slowly and wastes a lot of time. I think we need to speed it up

aramikm (commented):
Added some comments and improvements

pallets/msa/src/benchmarking.rs
enddynayn: Yeah I think I am going to do this one in a different PR.

pallets/msa/src/benchmarking.rs
enddynayn: good catch.

shannonwells (approved):
🥚 -cellent

aramikm (approved):
Looks good!

PR #726, opened 2022-11-29
[WIP]Chore/unify runtimes

### LINKED ISSUE #714: Unify versions across runtimes
Move Runtime VERSIONS to common runtime

* [x] We would want a unified version release for both rococo/mainnet runtimes
* [x] For the same move RuntimeVersion from individual runtimes to common-runtime 
* [ ] Update PR template to including check point for version update

saraswatpuneet:
# Goal
The goal of this PR is  to unify runtimes and separate mainnet "limited" time filters via flag

Closes #714 

# Checklist
- [ ] Chain spec updated
- [ ] Design doc(s) updated
- [ ] Tests added
- [ ] Benchmarks added
- [ ] Weights updated


node/cli/src/command.rs
```diff
@@ -202,42 +182,8 @@ impl SubstrateCli for Cli {
...
-				}
-				#[cfg(not(feature = "frequency-rococo-local"))]
-				panic!("Frequency Local runtime is not compiled!");
-			} else {
-				panic!("Frequency Local runtime is not compiled!");
-			}
-		}
-		panic!("Unknown spec id: {}", spec.id());
+	fn native_runtime_version(_spec: &Box<dyn ChainSpec>) -> &'static RuntimeVersion {
+		&VERSION
```
saraswatpuneet, line 186: Frequency is the only runtime

node/service/src/chain_spec/frequency.rs
```diff
@@ -117,7 +119,7 @@ pub fn load_frequency_spec() -> ChainSpec {
 
 pub fn frequency() -> ChainSpec {
 	let properties =
-		get_properties(FREQUENCY_TOKEN, TOKEN_DECIMALS as u32, SS58Prefix::get().into());
+		get_properties(FREQUENCY_TOKEN, TOKEN_DECIMALS as u32, SS58Prefix90::get().into());
 	let para_id: ParaId = 2091.into();
```
saraswatpuneet, line 123: SS58Prefix90 for mainnet

node/service/src/chain_spec/frequency_local.rs
```diff
@@ -6,29 +6,28 @@ use common_runtime::constants::{
 };
 
 use cumulus_primitives_core::ParaId;
-use frequency_rococo_runtime::{
-	AuraId, CouncilConfig, SS58Prefix, SudoConfig, TechnicalCommitteeConfig,
```
saraswatpuneet, line 10: SS58Prefix42 for rococo

node/service/src/service.rs
```diff
@@ -61,55 +60,9 @@ pub mod frequency_executor {
...
-		fn dispatch(method: &str, data: &[u8]) -> Option<Vec<u8>> {
-			frequency_rococo_runtime::api::dispatch(method, data)
-		}
-
-		fn native_version() -> sc_executor::NativeVersion {
-			frequency_rococo_runtime::native_version()
-		}
-	}
-}
-#[cfg(feature = "frequency")]
```
saraswatpuneet, line 104: No need for separate executors

runtime/frequency/src/lib.rs
```diff
@@ -65,9 +66,14 @@ pub use common_runtime::{
...
 	weights::{BlockExecutionWeight, ExtrinsicBaseWeight},
 };
+#[cfg(feature = "frequency")]
+use frame_support::traits::Contains;
 
 /// Basefilter to only allow specified transactions call to be executed
+#[cfg(feature = "frequency")]
 pub struct BaseCallFilter;
+
+#[cfg(feature = "frequency")]
```
saraswatpuneet, line 76: Only enabled when mainnet is built

saraswatpuneet:
@demisx only major change or rather update in CI is in rust doc running for -all-features as we don't default the rococo runtime but have a single one

So I am waiting for that step to pass

runtime/frequency/src/lib.rs
```diff
@@ -151,7 +157,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	spec_name: create_runtime_str!("frequency"),
 	impl_name: create_runtime_str!("frequency"),
 	authoring_version: 1,
-	spec_version: 1,
+	spec_version: 3,
```
wilwade, line 160: ✅

runtime/frequency/src/lib.rs
wilwade: I wish this were more clear what it applied to instead of being duplicated, but it works.

runtime/frequency/src/lib.rs
```diff
@@ -202,6 +206,10 @@ impl frame_system::Config for Runtime {
 	/// The identifier used to distinguish between accounts.
 	type AccountId = AccountId;
 	/// Base call filter to use in dispatchable.
+	// enable for cfg feature "frequency" only
+	#[cfg(not(feature = "frequency"))]
```
wilwade, line 210: This worries me `not(feature)` can be an issue for `--all-features` in testing, benchmarking, and checks.

Any ideas about other ways to handle it? I wonder if this code needs to use a macro like `prod_or_testnet_or_local_or_env`

saraswatpuneet: So the order in which flags are defined are ```not frequency ```and then ``` frequency``` . This entails when a binary is built with all-features flag it will always choose first over the other

And mainnet binary will not see ```not part``` that way we default rococo(local, sealing, testnet) if binary is built with all-features 

When we release mainnet we build using frequency flag and it will only keep mainnet related definitions

saraswatpuneet: I will put a macro and check

runtime/common/src/constants.rs
```diff
@@ -275,3 +275,8 @@ impl Clone for MessagesMaxPayloadSizeBytes {
 pub type MSAMortalityWindowSize = ConstU32<100>;
 pub type MSAMaxSignaturesPerBucket = ConstU32<50_000>;
 pub type MSANumberOfBuckets = ConstU32<2>;
+
+parameter_types! {
+	/// SS58 Prefix for the for Frequency Network
+	pub const Ss58Prefix: u16 = prod_or_testnet_or_local_or_env!(90, 42, 42);
+}
```
saraswatpuneet, line 282: SS58 prefix 90 for prod  42 and 42 for rococo local and testnet

saraswatpuneet: Using the macro to select

runtime/frequency/src/lib.rs
saraswatpuneet: Applied a simpler and much intuitive pattern for this now

runtime/frequency/src/lib.rs
```diff
@@ -65,25 +66,30 @@ pub use common_runtime::{
...
+		{
+			matches!(
+				_call,
+				Call::System(..) |
+					Call::Timestamp(..) | Call::ParachainSystem(..) |
+					Call::Sudo(..) | Call::TechnicalCommittee(..) |
+					Call::Council(..) | Call::Democracy(..) |
+					Call::Session(..) | Call::Preimage(..) |
+					Call::Scheduler(..)
+			)
```
saraswatpuneet, line 91: when we build -all-features it will turn off not side of function scope and if we build with frequency flag it will activate the call filter side of scope

pretty much works like if-else but does not carry the non needed code to binary and wasm, hence its nice that if else but not opposed to having conditional separation rather


I have tested this though

aramikm:
I get this error when running `cargo check`

Screen Shot 2022-11-30 at 3 15 53 PM

runtime/frequency/src/lib.rs
aramikm: why 3 instead of 2?

node/service/Cargo.toml
```diff
@@ -96,23 +95,21 @@ pallet-schemas-runtime-api = { path = "../../pallets/schemas/src/runtime-api" }
 substrate-build-script-utils = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.29" }
 
 [features]
-default = ["std"]
+default = ["std", "frequency"]
```
saraswatpuneet, line 98: @aramikm  cargo check will work now, since I am used to make file, adding a default runtime was needed

runtime/frequency/src/lib.rs
saraswatpuneet: Rococo is current 2, so we push all 2 to 3 next upgrade and keep them in sync

saraswatpuneet: [Rococo Live Version](https://github.com/LibertyDSNP/frequency/blob/440ad4781f78fe1b3e8f6d64898d86921c734ba7/runtime/frequency-rococo/src/lib.rs#L134)

[Mainnet Live Version](https://github.com/LibertyDSNP/frequency/blob/440ad4781f78fe1b3e8f6d64898d86921c734ba7/runtime/frequency/src/lib.rs#L154)

enddynayn:
> I get this error when running `cargo check`
> 
> Screen Shot 2022-11-30 at 3 15 53 PM

Is this something we should add to the CI? Does the CI miss this check?

node/cli/src/command.rs
```diff
@@ -57,38 +62,6 @@ impl<T: sc_service::ChainSpec + 'static> IdentifyChain for T {
 	}
 }
 
-macro_rules! with_runtime_or_err {
```
enddynayn, line 60: 🥇

runtime/frequency/src/lib.rs
```diff
@@ -202,6 +206,7 @@ impl frame_system::Config for Runtime {
 	/// The identifier used to distinguish between accounts.
 	type AccountId = AccountId;
 	/// Base call filter to use in dispatchable.
+	// enable for cfg feature "frequency" only
```
enddynayn, line 209: Is this a note to add a feature flag here?

saraswatpuneet: Also it's a dev comment // only in case anyone adds new features added to frequency, mainnet should still be filtered until we remove it when we remove sudo

runtime/common/src/constants.rs
shannonwells: works for me

runtime/frequency/src/lib.rs
```diff
@@ -242,7 +247,7 @@ impl frame_system::Config for Runtime {
 	/// The maximum length of a block (in bytes).
 	type BlockLength = RuntimeBlockLength;
 	/// This is used as an identifier of the chain. 42 is the generic substrate prefix.
-	type SS58Prefix = SS58Prefix;
+	type SS58Prefix = Ss58Prefix;
```
shannonwells, line 250: I built mainnet locally and attempted to do a balance transfer just to check that the call filters were acting correctly, but got a SS58Prefix error. I did see that you tested the call filters though. The system constant does show that the prefix is 90 and other configs are as expected.

shannonwells (approved):
It works me as a local testnet relay/parachain. It builds and runs for mainnet too, but I'm assuming I have to do something else to fake a local mainnet and get transactions to work.

runtime/frequency/src/lib.rs
saraswatpuneet: Yes because now that our mainnet specs are files, you will have to change at quite a few places in scripts/onobard etc to make mainnet run locally most importantly paraid to be changed to 2000 and network from polkadot to rococo-local. These changes will make mainnet run locally like rococo 

I did all that and ran a full node locally for mainnet and calls are getting filtered

Issue #350, opened 2022-08-16
[Config] Democracy Config


wilwade:
Note @shannon worked on the config from https://github.com/LibertyDSNP/frequency/pull/313.

## Notes
- Should use [`prod_or_fast!`](https://paritytech.github.io/polkadot/doc/polkadot_runtime_common/macro.prod_or_fast.html) with a fast version a reasonable value
- Each of these should have a comment with a description of what it is and means

## Config
- [x] `EnactmentPeriod` = `8 * DAYS`
    - The period between a proposal being approved and enacted.
    - Polkadot uses `28 * DAYS`
- [x] `LaunchPeriod` = `7 * DAYS`
    - How often (in blocks) new public referenda are launched.
    - Polkadot uses `28 * DAYS`
- [x] `VotingPeriod` = `7 * DAYS`
    - How often (in blocks) to check for new votes.
    - Polkadot uses `28 * DAYS`
- [x] `VoteLockingPeriod` = `EnactmentPeriod`
    - The minimum period of vote locking.
    - Polkadot uses `EnactmentPeriod` aka `28 * DAYS`
- [x] `MinimumDeposit` = `10 * DOLLARS`
    - The minimum amount to be used as a deposit for a public referendum proposal.
    - Polkadot uses `100 DOT`
- [x] `ExternalOrigin` = 
```
EitherOfDiverse<
		pallet_collective::EnsureProportionAtLeast<AccountId, FrequencyCouncilCollective, 1, 2>,
		frame_system::EnsureRoot<AccountId>,
	>;
```
    - Origin from which the next tabled referendum may be forced. This is a normal "super-majority-required" referendum.
    - Polkadot uses `50% of Council`
- [x] `ExternalMajorityOrigin` = 
```
EitherOfDiverse<
		pallet_collective::EnsureProportionAtLeast<AccountId, FrequencyCouncilCollective, 3, 5>,
		frame_system::EnsureRoot<AccountId>,
	>;
```
    - Origin from which the next tabled referendum may be forced; this allows for the tabling of a majority-carries referendum.
    - Polkadot uses `60% of Council`
- [x] `ExternalDefaultOrigin` = 
```
EitherOfDiverse<
		pallet_collective::EnsureProportionAtLeast<AccountId, FrequencyCouncilCollective, 1, 1>,
		frame_system::EnsureRoot<AccountId>,
	>;
```
[...]

PR #731, opened 2022-12-02
set TC as cancelproposalorigin and as instantorigin

### LINKED ISSUE #730: InstantOrigin and CancelProposalOrigin should be TC not Council
A couple of Configs in democracy pallet seemed to have suffered from copypasta. These origins are for things like emergency upgrades or cancelling dangerous proposals for technical reasons, so it can't be left up to a unanimous Council vote.

As shown in issue #350, pallet_democracy's `InstantOrigin` and `CancelProposalOrigin` should be  unanimous vote from `TechnicalCommitteeInstance`. In other words just replace CouncilCollective with TechnicalCommitteeInstance for those two origins in the frequency runtime.

NB: all the rest of the democracy configs are as specified in the issue.

shannonwells:
# Goal
The goal of this PR is to fix a bug in pallet_democracy config where we had the Council as the InstantOrigin and CancelProposalOrigin when it should have been TechnicalCommittee.

Closes #730

# Checklist
- [x] TC substituted for Council in InstantOrigin and CancelProposalOrigin
- [x] TC instance renamed to be TechnicalCommitteeCollective
- [x] Verified by running local instant seal

## Not applicable:
- [x] Chain spec updated
- [x] Design doc(s) updated
- [x] Tests added
- [x] Benchmarks added
- [x] Weights updated

### DIFF
--- a/runtime/frequency/src/lib.rs
+++ b/runtime/frequency/src/lib.rs
@@ -459,7 +459,7 @@ impl pallet_democracy::Config for Runtime {
 	/// vote immediately and asynchronously in a similar manner to the emergency origin.
 	/// Requires TechnicalCommittee to be unanimous.
 	type InstantOrigin = EitherOfDiverse<
-		pallet_collective::EnsureProportionAtLeast<AccountId, CouncilCollective, 1, 1>,
+		pallet_collective::EnsureProportionAtLeast<AccountId, TechnicalCommitteeInstance, 1, 1>,
 		frame_system::EnsureRoot<AccountId>,
 	>;
 	/// Overarching type of all pallets origins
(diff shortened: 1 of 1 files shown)


saraswatpuneet (approved):
:raised_hands:

runtime/frequency/src/lib.rs
```diff
@@ -459,7 +459,7 @@ impl pallet_democracy::Config for Runtime {
 	/// vote immediately and asynchronously in a similar manner to the emergency origin.
 	/// Requires TechnicalCommittee to be unanimous.
 	type InstantOrigin = EitherOfDiverse<
-		pallet_collective::EnsureProportionAtLeast<AccountId, CouncilCollective, 1, 1>,
+		pallet_collective::EnsureProportionAtLeast<AccountId, TechnicalCommitteeInstance, 1, 1>,
```
saraswatpuneet, line 462: Can we name them to something "collective" which would be the reason it got missed from the following story 
TechnicalCouncilCollective

https://github.com/LibertyDSNP/frequency/issues/350

Probably believe the confusion between Instance and Collective naming

shannonwells: Yes, I will rename to make them consistent with each other.

aramikm (approved):

PR #733, opened 2022-12-02
```prod_or_testnet_or_local_or_env``` requires wasm be built with flags

### LINKED ISSUE #732: Fix build runtime errors
I just started seeing this new error while building deterministic WASM with SRTool (Rococo) in the Release workflow. I believe it started happening after [this PR was merged](https://github.com/LibertyDSNP/frequency/pull/726).

```sh
# Rococo
error: Package `frequency-runtime v0.9.29 (/build/runtime/frequency)` does not have the feature `frequency-rococo-testnet`
Error: Process completed with exit code 1.
```

```sh
# Mainnet
error: Package `frequency-runtime v0.9.29 (/build/runtime/frequency)` does not have the feature `frequency`
Error: Process completed with exit code 1.
```

Here is an example of the failed CI job: https://github.com/LibertyDSNP/frequency/actions/runs/3599018706/jobs/6062290997

saraswatpuneet:
# Goal
The goal of this PR is to fix a bug discovered by @shannonwells that wasm based constants such as LaunchPeriod that are derived from ```prod_or_testnet_or_local_or_env``` defaults to prod values for rococo

This just makes our wasm to have prod values in all such constants because SRTool wasm is generated with default features and it overrides what value chain sees in metadata 

# Resolution:

* Generate wasm artifacts for rococo and mainnet enabling their respective feature flags i.e. ```frequency-rococo-testnet``` and ```frequency``` (mainnet) and update CI job to build feature based wasms
* The latter part got covered when unifying runtime #714 

Closes #732 

# Checklist
- [ ] Chain spec updated
- [ ] Design doc(s) updated
- [ ] Tests added
- [ ] Benchmarks added
- [ ] Weights updated

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -3,11 +3,11 @@ concurrency:
   group: ${{github.workflow}}-${{github.ref}}
   cancel-in-progress: true
 on:
-  push:
-    tags:
-      - "v[0-9]+.[0-9]+.[0-9]+" # ex. v0.29.9
-      - "v[0-9]+.[0-9]+.[0-9]+-[0-9]+" # ex. v0.29.9-1
-      - "v[0-9]+.[0-9]+.[0-9]+-rc[0-9]+" # ex. v0.29.9-1-rc1
+  pull_request:
+    # tags:
+    #   - "v[0-9]+.[0-9]+.[0-9]+" # ex. v0.29.9
+    #   - "v[0-9]+.[0-9]+.[0-9]+-[0-9]+" # ex. v0.29.9-1
+    #   - "v[0-9]+.[0-9]+.[0-9]+-rc[0-9]+" # ex. v0.29.9-1-rc1
 env:
   RUST_TOOLCHAIN: nightly-2022-09-22 # Update this when updating the Rust toolchain
   NEW_RELEASE_VERSION: ${{github.ref_name}}
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -3,11 +3,11 @@ concurrency:
   group: ${{github.workflow}}-${{github.ref}}
   cancel-in-progress: true
 on:
-  pull_request:
-    # tags:
-    #   - "v[0-9]+.[0-9]+.[0-9]+" # ex. v0.29.9
-    #   - "v[0-9]+.[0-9]+.[0-9]+-[0-9]+" # ex. v0.29.9-1
-    #   - "v[0-9]+.[0-9]+.[0-9]+-rc[0-9]+" # ex. v0.29.9-1-rc1
+  push:
+    tags:
+      - "v[0-9]+.[0-9]+.[0-9]+" # ex. v0.29.9
+      - "v[0-9]+.[0-9]+.[0-9]+-[0-9]+" # ex. v0.29.9-1
+      - "v[0-9]+.[0-9]+.[0-9]+-rc[0-9]+" # ex. v0.29.9-1-rc1
 env:
   RUST_TOOLCHAIN: nightly-2022-09-22 # Update this when updating the Rust toolchain
   NEW_RELEASE_VERSION: ${{github.ref_name}}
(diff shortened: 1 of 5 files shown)


.github/workflows/release.yml
```diff
@@ -117,7 +117,7 @@ jobs:
             build-profile: production
             package: frequency-runtime
             runtime-dir: runtime/frequency
-            built-wasm-file-name-prefix: frequency_rococo_runtime
+            built-wasm-file-name-prefix: frequency_runtime
```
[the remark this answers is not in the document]
saraswatpuneet: this is the file generated by srtool for runtime specified so it will never be anything but frequency_runtime.x

Its just you will rename that to either rococo/mainnet if  am correct

check the error here https://github.com/LibertyDSNP/frequency/actions/runs/3600119600/jobs/6064429748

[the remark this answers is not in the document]
saraswatpuneet: Example
https://github.com/LibertyDSNP/frequency/actions/runs/3600119600/jobs/6064429748#step:6:1528

Given we have only one runtime srtool will spit out same named files for both of rococo and mainnet as it takes "name" of the file == {runtime_name}.compress.wasm.x.x.etc

runtime/common/Cargo.toml
```diff
@@ -43,6 +43,8 @@ orml-vesting = {git = "https://github.com/open-web3-stack/open-runtime-module-li
 
 [features]
 default = ["std"]
+frequency = []
+frequency-rococo-testnet = []
```
saraswatpuneet, line 47: because feature flags are used in runtime package of frequency we need to add these to enable on runtime build which generates wasm

.github/workflows/release.yml
saraswatpuneet, line 120: For anyone reading this, basically in our releases we should be releasing rococo/mainnet wasms with their respective feature flags on such as ```frequency-rococo-testnet``` ```frequency``` 
To be compatible with how we generate our binaries

runtime/common/src/lib.rs
```diff
@@ -28,13 +28,4 @@ macro_rules! prod_or_testnet_or_local_or_env {
 			$prod
 		}
 	};
-	($prod:expr, $test:expr, $local:expr, $env:expr) => {
-		if cfg!(feature = "frequency-rococo-local") {
-			core::option_env!($env).map(|s| s.parse().ok()).flatten().unwrap_or($local)
```
saraswatpuneet, line 33: This probably always fails due to wasm restrictions but we should remove this from our code and let dev make changes locally for testing

runtime/common/src/lib.rs
```diff
@@ -13,12 +13,12 @@ pub mod weights;
...
 /// 	// Note that the env variable version parameter cannot be const.
-/// 	pub LaunchPeriod: BlockNumber = prod_or_testnet_or_local_or_env!(7 * DAYS, 28 * DAYS, 1 * MINUTES, "FRQCY_LAUNCH_PERIOD");
-/// 	pub const VotingPeriod: BlockNumber = prod_or_testnet_or_local_or_env!(7 * DAYS, 28 * DAYS, 1 * MINUTES);
+/// 	pub LaunchPeriod: BlockNumber = prod_or_testnet_or_local!(7 * DAYS, 28 * DAYS, 1 * MINUTES);
+/// 	pub const VotingPeriod: BlockNumber = prod_or_testnet_or_local!(7 * DAYS, 28 * DAYS, 1 * MINUTES);
 /// }
 /// ```
 #[macro_export]
-macro_rules! prod_or_testnet_or_local_or_env {
+macro_rules! prod_or_testnet_or_local {
```
saraswatpuneet, line 21: renaming it for correctness as we remove reading env variable form runtime code

shannonwells (approved):
💯

runtime/common/src/lib.rs
aramikm: so we are not going to use any env variables for any of these?

aramikm (approved):
Looks good. Do we need any version changes? I think it would be great if there is guideline when we need a version change and for which one, since there are multiple versions.

saraswatpuneet:
@aramikm we have to keep one first of frequency runtime across different specs to prevent managing multiple runtimes 

That will be enforced with our next release which will be used for runtime upgrade on rococo and mainnet .


Regarding env variable yes we don't need that unless we really want to run out nodes as compiled binary over wasm execution 

and sys calls like reading env variable are prohibited from within wasm code ( runtime )

So this will simplify all that

Issue #739, opened 2022-12-05
Potental Overflow


wilwade:
`pallets/msa/src/lib.rs`

The function add_key() adds a key to an MSA ID and increments the count of keys for that MSA ID. However, the incrementation of the count of keys is vulnerable to overflow if the number of keys is already `2**8 -1`.

Note: May be tested already via other functions

saraswatpuneet:
Add a regression test

PR #711, opened 2022-11-21
chore(msa): update benchmarks

enddynayn:
Update benchmarks.

issue-668

# Goal
The goal of this PR is 

Closes 

# Discussion

# Checklist
- [ ] Chain spec updated
- [ ] Custom RPC OR Runtime API added/changed? Updated js/api-augment.
- [ ] Design doc(s) updated
- [ ] Tests added
- [ ] Benchmarks added
- [ ] Weights updated

Issue #669, opened 2022-11-11
Review Benchmarks: Schema pallet


aramikm:
We need to review and improve our benchmarks on Schema pallet in response to auditing.

- [ ] Review the happy path
- [ ] Remove all magic numbers
- [ ] Speed up slow benchmarks

PR #753, opened 2022-12-06
Parity - Code review

NachoPal:
I based my review on your last release. Most of the code review comments are part of this PR. Just a few other points to mention:
### Runtime
- I see that some `OnUnbalanced` pallet’s Config types are set to `()` that will reduce the initial total issuance having a deflationary system. Are you ok with that? For instance, you are burning the fees. Wouldn't you prefer to send those fees to the collators to keep the total issuance?
### Pallet Messages
- What is the point of saving write time in `BlockMessages` if then some block time will be used anyway to move the data to `Messages` ? Did you compare weight times of writing directly in `Messages` vs moving them from `BlockMessages` in the `on_initialize`?
- Schemas Grants do not apply for IPFS messages? There is not any kind of validation for off-chain messages as long as the provider exists?

### DIFF
--- a/pallets/messages/src/lib.rs
+++ b/pallets/messages/src/lib.rs
@@ -208,10 +208,15 @@ pub mod pallet {
 		/// * [`Error::TooManyMessagesInBlock`] - Block is full of messages already
 		/// * [`Error::TypeConversionOverflow`] - Failed to add the message to storage as it is very full
 		///
+		// Wondering how the total number of messages affects in this case to calculate the weight
+		// Also why 1_000 is used. Shouldn't be MaxMessagesPerBlock as worst case?
 		#[pallet::weight(T::WeightInfo::add_ipfs_message(cid.len() as u32, 1_000))]
 		pub fn add_ipfs_message(
 			origin: OriginFor<T>,
 			#[pallet::compact] schema_id: SchemaId,
+			// CID format doesn't have a expected format with a fixed length?
+			// Consider creating a new type for it instead of a Vec<u8> of arbitrary length
+			// (Maybe not, apparently there are 2 different versions)
 			cid: Vec<u8>,
 			#[pallet::compact] payload_length: u32,
 		) -> DispatchResultWithPostInfo {
@@ -229,7 +234,7 @@ pub mod pallet {
 				Error::<T>::InvalidPayloadLocation
 			);
 
-			let provider_msa_id = Self::find_msa_id(&provider_key)?;
+			let provider_msa_id = Self::find_msa_id(&provider_key)?; // Probably check this first rigt after getting the provider_key ?
 			let message = Self::add_message(provider_msa_id, None, bounded_payload, schema_id)?;
 
 			Ok(Some(T::WeightInfo::add_ipfs_message(cid.len() as u32, message.index as u32)).into())
@@ -248,6 +253,8 @@ pub mod pallet {
 		/// * [`Error::TooManyMessagesInBlock`] - Block is full of messages already
 		/// * [`Error::TypeConversionOverflow`] - Failed to add the message to storage as it is very full
 		///
+		// Wondering how the total number of messages affects in this case to calculate the weight
+		// Also why 1_000 is used. Shouldn't be MaxMessagesPerBlock as worst case?
 		#[pallet::weight(T::WeightInfo::add_onchain_message(payload.len() as u32, 1_000))]
 		pub fn add_onchain_message(
 			origin: OriginFor<T>,
@@ -267,7 +274,7 @@ pub mod pallet {
 				Error::<T>::InvalidPayloadLocation
 			);
 
-			let provider_msa_id = Self::find_msa_id(&provider_key)?;
+			let provider_msa_id = Self::find_msa_id(&provider_key)?; // Same, should it be the first thing to check?
 			let provider_id = ProviderId(provider_msa_id);
 
 			let current_block = frame_system::Pallet::<T>::block_number();
--- a/pallets/messages/src/types.rs
+++ b/pallets/messages/src/types.rs
@@ -26,6 +26,8 @@ where
 	///  Message source account id (the original source).
 	pub msa_id: Option<MessageSourceId>,
 	///  Stores index of message in block to keep total order.
+	// Is this index really needed, messages don't end up ordered in the Messages strorage Item (form on_initialize)?
+	// Maybe necessary for reading purposes from the StorageDoubleMap I guess?
 	pub index: u16,
 }
 
--- a/pallets/msa/src/lib.rs
+++ b/pallets/msa/src/lib.rs
@@ -381,7 +381,7 @@ pub mod pallet {
 		///
 		/// * [`Error::KeyAlreadyRegistered`] - MSA is already registered to the Origin.
 		///
-		#[pallet::weight(T::WeightInfo::create(10_000))]
+		#[pallet::weight(T::WeightInfo::create(10_000))] // Number of accounts is not part of the extrinsic complexity
 		pub fn create(origin: OriginFor<T>) -> DispatchResult {
 			let who = ensure_signed(origin)?;
 
@@ -427,7 +427,7 @@ pub mod pallet {
 
 			Self::register_signature(&proof, add_provider_payload.expiration.into())?;
 
-			let provider_msa_id = Self::ensure_valid_msa_key(&provider_key)?;
+			let provider_msa_id = Self::ensure_valid_msa_key(&provider_key)?; // Shouldn't it be the first thing to check?
 			ensure!(
 				add_provider_payload.authorized_msa_id == provider_msa_id,
 				Error::<T>::UnauthorizedProvider
@@ -474,7 +474,7 @@ pub mod pallet {
 			let bounded_name: BoundedVec<u8, T::MaxProviderNameSize> =
 				provider_name.try_into().map_err(|_| Error::<T>::ExceedsMaxProviderNameSize)?;
 
-			let provider_msa_id = Self::ensure_valid_msa_key(&provider_key)?;
+			let provider_msa_id = Self::ensure_valid_msa_key(&provider_key)?; // check first?
 			ProviderToRegistryEntry::<T>::try_mutate(
 				ProviderId(provider_msa_id),
 				|maybe_metadata| -> DispatchResult {
@@ -621,7 +621,7 @@ pub mod pallet {
 
 			let msa_id = add_key_payload.msa_id;
 
-			Self::ensure_msa_owner(&msa_owner_public_key, msa_id)?;
+			Self::ensure_msa_owner(&msa_owner_public_key, msa_id)?; // Shouldn't be the first thing to check?
 
 			Self::add_key(
 				msa_id,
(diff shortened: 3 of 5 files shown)


wilwade:
> I see that some OnUnbalanced pallet’s Config types are set to () that will reduce the initial total issuance having a deflationary system. Are you ok with that? For instance, you are burning the fees. Wouldn't you prefer to send those fees to the collators to keep the total issuance?

Good question! Part of the coming capacity system will provide an inflationary offset. Also because of capacity, fees are not going to be that stable, so the collator rewards will be minted instead. Still working out the exact numbers, but the expectation is to be (on average) inflationary once it is all in place by about Q3.

> What is the point of saving write time in BlockMessages if then some block time will be used anyway to move the data to Messages ? Did you compare weight times of writing directly in Messages vs moving them from BlockMessages in the on_initialize?

@aramikm I think you can answer this one better than I can.

> Schemas Grants do not apply for IPFS messages? There is not any kind of validation for off-chain messages as long as the provider exists?

They do, but they have to be checked at read time. It's expected that IPFS messages could contain more than one user's payload. So the validation of off-chain data is left to read time.

Here's an example of how DSNP suggests validation (although other protocol schemas on Frequency could do it differently): https://spec.dsnp.org/Frequency/Validation.html#batch-publication-frequency-message-validation

pallets/messages/src/types.rs
```diff
@@ -26,6 +26,8 @@ where
 	///  Message source account id (the original source).
 	pub msa_id: Option<MessageSourceId>,
 	///  Stores index of message in block to keep total order.
+	// Is this index really needed, messages don't end up ordered in the Messages strorage Item (form on_initialize)?
+	// Maybe necessary for reading purposes from the StorageDoubleMap I guess?
```
wilwade, line 30: While the final storage is ordered, it is nested by `SchemaId`, so to be able to tell the ordering difference between different schemas, we need to keep this value.

Also if we modify the temp block, it requires modifying that memory.

@aramikm to double check. We might be able to remove it now.

pallets/msa/src/lib.rs
```diff
@@ -720,6 +720,8 @@ pub mod pallet {
 		/// * [`Error::DelegationNotFound`] no delegation relationship between Origin and Delegator or Origin and Delegator are the same.
 		/// * [`Error::ExceedsMaxSchemaGrantsPerDelegation`] the limit of maximum allowed grants per delegation relationship has been exceeded.
 		///
+		// Number of delegations is not part of the extrinsic complexity
+		// I could see a dependancy on the length of `schema_ids` as they should be checked as valid. However, considering its small number the difference can be neglected
```
wilwade, line 724: @enddynayn

pallets/msa/src/lib.rs
```diff
@@ -748,7 +750,7 @@ pub mod pallet {
 		/// - [`DelegationNotFound`](Error::DelegationNotFound) - If there is not delegation relationship between Origin and Delegator or Origin and Delegator are the same.
 		/// - [`SchemaNotGranted`](Error::SchemaNotGranted) - If attempting to revoke a schema that has not previously been granted.
 		///
-		#[pallet::weight(T::WeightInfo::revoke_schema_permissions(20_000))]
+		#[pallet::weight(T::WeightInfo::revoke_schema_permissions(20_000))] // Number of delegations is not part of the extrinsic complexity
```
wilwade, line 753: @enddynayn

pallets/msa/src/lib.rs
```diff
@@ -902,7 +904,7 @@ impl<T: Config> Pallet<T> {
 			// Increment the key counter
 			<PublicKeyCountForMsaId<T>>::try_mutate(msa_id, |key_count| {
 				// key_count:u8 should default to 0 if it does not exist
-				let incremented_key_count: u8 = *key_count + 1;
+				let incremented_key_count: u8 = *key_count + 1; // Not checking overflow here?
```
wilwade, line 907: Tracking in #739

pallets/msa/src/lib.rs
```diff
@@ -1206,8 +1209,8 @@ impl<T: Config> Pallet<T> {
 		// Clear the previous bucket block set
 		let multi_removal_result = <PayloadSignatureRegistry<T>>::clear_prefix(
 			prior_bucket_num,
-			T::MaxSignaturesPerBucket::get(),
-			None,
+			T::MaxSignaturesPerBucket::get(), // I couldn't find any limitation for the number of PayloadSignatures that can be registered
+			None, // If there are more than 'MaxSignaturesPerBucket' items, you should check 'multi_removal_result.maybe_cursor' and handle the situation.
```
wilwade, line 1213: @shannonwells Could you check on these potential overflows in the signature buckets?

pallets/schemas/src/lib.rs
```diff
@@ -198,21 +198,26 @@ pub mod pallet {
 		/// * [`Error::InvalidSchema`] - Schema is malformed in some way
 		/// * [`Error::SchemaCountOverflow`] - The schema count has exceeded its bounds
 		///
+		// Wondering how the total number of Schemas affects in this case to calculate the weight
+		// Also why 1_000 is used. Shouldn't be MaxSchemaRegistrations as worst case?
```
wilwade, line 202: Tracking in #669

pallets/schemas/src/lib.rs
```diff
@@ -198,21 +198,26 @@ pub mod pallet {
...
+			// You could for example lock/resrve some tokens (deposit for creating the Schame) or add a especial origin (Governance and Providers?)
 			let sender = ensure_signed(origin)?;
 
 			ensure!(
 				model.len() > T::MinSchemaModelSizeBytes::get() as usize,
 				Error::<T>::LessThanMinSchemaModelBytes
 			);
 			ensure!(
-				model.len() < Self::get_schema_model_max_bytes() as usize,
+				model.len() < Self::get_schema_model_max_bytes() as usize, // < or <= ?
```
wilwade, line 220: Should likely be inclusive instead of exclusive. @wilwade to add an issue for it.

pallets/schemas/src/lib.rs
```diff
@@ -235,7 +240,7 @@ pub mod pallet {
 		/// # Errors
 		/// * [`Error::ExceedsMaxSchemaModelBytes`] - Cannot set to above the hard coded maximum [`Config::SchemaModelMaxBytesBoundedVecLimit`]
 		///
-		#[pallet::weight(30_000)]
+		#[pallet::weight(30_000)] // Set as DispatchClass::Operational
```
wilwade, line 243: ✅ @wilwade to add an issue.

runtime/frequency/src/lib.rs
```diff
@@ -69,7 +69,7 @@ pub use common_runtime::{
 /// Basefilter to only allow specified transactions call to be executed
 pub struct BaseCallFilter;
 impl Contains<Call> for BaseCallFilter {
-	fn contains(call: &Call) -> bool {
+	fn contains(call: &Call) -> bool { // I see some missing pallets here, did you double check you have all you want?
```
wilwade, line 72: We'll be opening those up as we get the audits fixes in.

runtime/frequency/src/lib.rs
```diff
@@ -386,6 +386,7 @@ impl pallet_collective::Config<CouncilCollective> for Runtime {
 	type Event = Event;
 	type MotionDuration = CouncilMotionDuration;
 	type MaxProposals = CouncilMaxProposals;
+	 // Only one member? I guess it will be just sudo for now? To be consistent, probably move the value to constants.rs
 	type MaxMembers = ConstU32<1>;
```
wilwade, line 390: That is wrong, but it is just for weight calculations so it works that we have 5 members currently. @wilwade to add an issue for it to update it.

runtime/frequency/src/lib.rs
```diff
@@ -420,7 +421,7 @@ impl pallet_democracy::Config for Runtime {
 	type PreimageByteDeposit = PreimageByteDeposit;
 	type Proposal = Call;
 	type Scheduler = Scheduler;
-	type Slash = (); // Treasury;
+	type Slash = (); // Treasury; // Null for now?
```
wilwade, line 424: Correct

pallets/schemas/src/lib.rs
wilwade: https://github.com/LibertyDSNP/frequency/issues/755

pallets/schemas/src/lib.rs
wilwade: https://github.com/LibertyDSNP/frequency/issues/756

runtime/frequency/src/lib.rs
wilwade: https://github.com/LibertyDSNP/frequency/issues/757

pallets/msa/src/lib.rs
```diff
@@ -684,7 +684,7 @@ pub mod pallet {
 		/// * [`Error::DelegationRevoked`] - delegation is already revoked
 		/// * [`Error::DelegationNotFound`] - no Delegation found between origin MSA and delegator MSA.
 		///
-		#[pallet::weight((T::WeightInfo::revoke_delegation_by_provider(20_000), DispatchClass::Normal, Pays::No))]
+		#[pallet::weight((T::WeightInfo::revoke_delegation_by_provider(20_000), DispatchClass::Normal, Pays::No))]  // Number of delegations is not part of the extrinsic complexity
```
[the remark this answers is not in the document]
enddynayn: This was addressed in this PR:
https://github.com/LibertyDSNP/frequency/pull/711

pallets/msa/src/lib.rs
enddynayn: This one is being addressed in this outstanding PR:
https://github.com/LibertyDSNP/frequency/pull/720

pallets/messages/src/types.rs
aramikm: @NachoPal @wilwade 
The decision to use the temporary storage comes exactly from having canonical order. If we don't use the temporary storage , to be able to keep the total index we will need **1** additional `DB_READ` and **1** additional `DB_WRITE` per `add_message`. This is going to be very expensive since we are expecting that this extrinsic is going to be used a lot. By using the temporary storage we are getting around that additional read and write but we are going to have additional S (number of unique schemas used) DB_WRITES. Since S is always expected to be smaller than total number of added messages in a block, from weights perspective it makes sense to do this process `on_initialize`.

If we don't need total order then we can get rid of this temporary storage and directly add to `Message` storage

aramikm: Based on my comparisons between these two different implementations we are getting the following equation

M = number of messages in a block
S = Number of unique schemas used in block
DB_READ = 25_000_000
DB_WRITE = 100_000_000

`
 121_000_000 M > 185_619_000 S`

so if we assume S = 5 then M > 7
which means after adding the 7's message in block the weights of regular implementation will become bigger than our current implementation
and for S = 20 we get M > 30
and ...

pallets/messages/src/lib.rs
```diff
@@ -208,10 +208,15 @@ pub mod pallet {
...
 		///
+		// Wondering how the total number of messages affects in this case to calculate the weight
+		// Also why 1_000 is used. Shouldn't be MaxMessagesPerBlock as worst case?
 		#[pallet::weight(T::WeightInfo::add_ipfs_message(cid.len() as u32, 1_000))]
 		pub fn add_ipfs_message(
 			origin: OriginFor<T>,
 			#[pallet::compact] schema_id: SchemaId,
+			// CID format doesn't have a expected format with a fixed length?
+			// Consider creating a new type for it instead of a Vec<u8> of arbitrary length
+			// (Maybe not, apparently there are 2 different versions)
```
wilwade, line 219: Tracking in:
- #786 
- #787

pallets/msa/src/lib.rs
wilwade: Tracking in #797

wilwade:
All issues resolved or tracking in separate issues. Thanks @NachoPal!

PR #765, opened 2022-12-07
Audit fix - schema model min/max size should be inclusive

### LINKED ISSUE #755: schema_model max/min bytes should be inclusive
`create_schema` ensure! for schema model bytes are both exclusive when they should be inclusive (`<` and `>` instead of `<=` and `>=`).

_Originally posted by @wilwade in https://github.com/LibertyDSNP/frequency/pull/753#discussion_r1041326286_

rlaferla:
Fixes https://github.com/LibertyDSNP/frequency/issues/755

# Goal
The goal of this PR is to fix an issue raised by the audit where the min/max schema model size should be inclusive.

Closes #755

### DIFF
--- a/pallets/schemas/src/lib.rs
+++ b/pallets/schemas/src/lib.rs
@@ -211,11 +211,11 @@ pub mod pallet {
 			let sender = ensure_signed(origin)?;
 
 			ensure!(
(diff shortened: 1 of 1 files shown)


enddynayn (approved):
💯 awesome!

saraswatpuneet (approved):

PR #766, opened 2022-12-07
Fix: Kill Freq node even if integration tests fail

dopatraman:
This command makes the script exit on any error. That means that if any tests fail, the frequency node will continue to run in the background. Removing it allows the script to kill the bootstrap Freq node even if a test fails.


rlaferla (approved):

saraswatpuneet (approved):

enddynayn (approved):
💯 spectacular!

PR #769, opened 2022-12-08
messages: simplify adding messages

### LINKED ISSUE #768: Audit: Simpler message storage
# Problem
After rechecking the weights it became apparent that keeping `canonical order` between messages is the only reason that justifies having the temporary storage. So after digging a bit in substrate code base we were able to find about 
`<frame_system::Pallet<T>>::extrinsic_index()`
which will allow us to get the canonical order without having to do an extra `DB_WRITE`

# Acceptance Criteria
- used extrinsic_index() for canonical order
- remove on_initialize
- Modify add_message to directly add to storage
- Re calculate weights

aramikm:
# Goal
The goal of this PR is simplify adding messages without increasing the weights.

Closes #768 

# Discussion
- By utilizing `extrinsic_index()` I was able to directly store the messages without loosing canonical order and optimized weights.
- Increases weights of each `add_message` by 1 `DB_READ` but since it removes the `on_initialize` completely, overall used weights should be less or equal to previous implementation
- Did not removed the unused `BlockMessage` storage yet since it required runtime migration and would increase the complexity of this PR

# Checklist
- [ ] Chain spec updated
- [ ] Design doc(s) updated
- [ ] Tests added
- [ ] Benchmarks added
- [ ] Weights updated


docker/docker-compose.yml
```diff
@@ -30,6 +30,7 @@ services:
       --rpc-methods=Unsafe
       --alice
       --log="main"
+    restart: unless-stopped
```
aramikm, line 33: Automatically restarts stopped relay nodes

pallets/messages/src/lib.rs
```diff
@@ -123,6 +124,7 @@ pub mod pallet {
 	/// A temporary storage of messages, given a schema id, for a duration of block period.
 	/// At the start of the next block this storage is cleared and moved into Messages storage.
 	/// - Value: List of Messages
+	/// Obsolete storage (will be removed in next storage migration)
```
aramikm, line 127: Migration not included to reduce the scope of this PR

pallets/messages/src/lib.rs
```diff
@@ -179,16 +184,13 @@ pub mod pallet {
...
-			count: u16,
 		},
 	}
 
 	#[pallet::hooks]
 	impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {
-		fn on_initialize(current: T::BlockNumber) -> Weight {
-			let prev_block = current - T::BlockNumber::one();
-			Self::move_messages_into_final_storage(prev_block)
+		fn on_initialize(_current: T::BlockNumber) -> Weight {
```
aramikm, line 192: Keeping this around for potential retention policy implementation but open to remove it in case of objections

pallets/messages/src/lib.rs
```diff
@@ -315,28 +319,26 @@ impl<T: Config> Pallet<T> {
...
+	) -> DispatchResult {
+		<Messages<T>>::try_mutate(current_block, schema_id, |existing_messages| -> DispatchResult {
+			let extrinsic_index =
+				<frame_system::Pallet<T>>::extrinsic_index().ok_or(Error::<T>::BadContext)?;
+			let msg = Message {
+				payload, // size is checked on top of extrinsic
+				provider_msa_id,
+				msa_id,
+				// converting u32 to u16 and defaulting to u16:MAX if overflowed
+				index: extrinsic_index.try_into().unwrap_or(u16::MAX),
```
aramikm, line 332: the probability of having more than 65,536 extrinsics in a block is slim to none but just in case something happens we will use max value for our sorting

pallets/messages/src/lib.rs
```diff
@@ -315,28 +319,26 @@ impl<T: Config> Pallet<T> {
...
+			};
+
+			existing_messages
+				.try_push(msg)
+				.map_err(|_| Error::<T>::TooManyMessagesInBlock)?;
+
+			Self::deposit_event(Event::MessagesStored { schema_id, block_number: current_block });
```
aramikm, line 339: Since we are removing `on_initialize` it does seem like we will need to add an event per message. Open to discuss about this and what we should include in here.

pallets/messages/src/lib.rs
```diff
@@ -369,48 +371,4 @@ impl<T: Config> Pallet<T> {
...
-	/// and calculates execution weight
-	///
-	/// Called inside of `on_initialize`
-	///
-	/// # Events
-	/// * [`Event::MessagesStored`]
-	///
-	/// Returns execution weights
-	///
-	fn move_messages_into_final_storage(block_number: T::BlockNumber) -> Weight {
```
aramikm, line 383: If BlockMessage was in use we would need to drain it first before removing this. But since we don't have any messages in there directing removing this should be fine

pallets/messages/src/tests.rs
```diff
@@ -294,64 +289,6 @@ fn get_messages_by_schema_with_ipfs_payload_location_should_fail_bad_schema() {
 	});
 }
 
-#[test]
-fn add_message_via_valid_delegate_should_pass() {
```
aramikm, line 298: Duplicated test

pallets/messages/src/weights.rs
```diff
@@ -65,69 +64,51 @@ impl<T: frame_system::Config> WeightInfo for SubstrateWeight<T> {
...
-	// Storage: Messages BlockMessages (r:1 w:1)
+	// Storage: Messages Messages (r:1 w:1)
+	// Storage: unknown [0x3a65787472696e7369635f696e646578] (r:1 w:0)
 	fn add_onchain_message(n: u32, ) -> Weight {
-		Weight::from_ref_time(88_911_000 as u64)
+		Weight::from_ref_time(80_238_000 as u64)
 			// Standard Error: 0
 			.saturating_add(Weight::from_ref_time(1_000 as u64).saturating_mul(n as u64))
-			.saturating_add(T::DbWeight::get().reads(4 as u64))
+			.saturating_add(T::DbWeight::get().reads(5 as u64))
```
aramikm, line 73: adds read by 1

pallets/messages/src/weights.rs
```diff
@@ -65,69 +64,51 @@ impl<T: frame_system::Config> WeightInfo for SubstrateWeight<T> {
...
-	// Storage: Messages BlockMessages (r:1 w:1)
+	// Storage: Messages Messages (r:1 w:1)
+	// Storage: unknown [0x3a65787472696e7369635f696e646578] (r:1 w:0)
 	fn add_ipfs_message(n: u32, ) -> Weight {
-		Weight::from_ref_time(74_858_000 as u64)
+		Weight::from_ref_time(69_077_000 as u64)
 			// Standard Error: 0
 			.saturating_add(Weight::from_ref_time(1_000 as u64).saturating_mul(n as u64))
-			.saturating_add(T::DbWeight::get().reads(3 as u64))
+			.saturating_add(T::DbWeight::get().reads(4 as u64))
```
aramikm, line 84: same here

aramikm (commented):
I will need a reminder, do I need to re generate spec for this beside `rococo-local-json` or we are doing that in CI? and I assume we don't need to bump up any version changes since no interface or storage is changed

pallets/messages/src/lib.rs
wilwade: @aramikm as we don't have any values stored in it (since they are cleared on the next block), do we need a storage migration at all? Perhaps one to be safe that no additional information is in it?

pallets/messages/src/lib.rs
wilwade: I'm up for removing it. I think we might end up using the scheduler or something different, so best to clear it out.

pallets/messages/src/lib.rs
wilwade: I recall some cheaper single block storage options that are cheap as they are not in the tree. If so, perhaps a flag so we emit only on the first one for each schema?

We lose count however. Perhaps a single block storage and then an emit `on_finalize` as a simple set of event emissions would be fine for an `on_finalize`?

pallets/messages/src/weights.rs
```diff
@@ -18,11 +18,11 @@
 //! Autogenerated weights for pallet_messages
 //!
 //! THIS FILE WAS AUTO-GENERATED USING THE SUBSTRATE BENCHMARK CLI VERSION 4.0.0-dev
-//! DATE: 2022-12-07, STEPS: `20`, REPEAT: 10, LOW RANGE: `[]`, HIGH RANGE: `[]`
```
wilwade, line 21: Make sure to run this branch through the actual benchmarking before merging.

wilwade (commented):
Few questions, but really like that we are able to clean this up so much.

Re: rococo-local-json. That's just for running the local relay chain, so it shouldn't need re-
generating unless we are updating the version of the the relay chain.

pallets/messages/src/lib.rs
enddynayn: Seems that there is some deciding to do here. However, let's try to move the event out of this function.

enddynayn (commented):
💯 Looking fabulous! 💯

pallets/messages/src/benchmarking.rs
```diff
@@ -42,7 +43,13 @@ fn ipfs_message<T: Config>(schema_id: SchemaId) -> DispatchResult {
...
 		payload.try_into().expect("Invalid payload");
 
-	MessagesPallet::<T>::add_message(provider_id.into(), None, bounded_payload, schema_id)?;
+	MessagesPallet::<T>::add_message(
+		provider_id.into(),
+		None,
+		bounded_payload,
+		schema_id,
+		<T as frame_system::Config>::BlockNumber::one(),
+	)?;
```
enddynayn, line 52: Maybe out of scope, but if not, let's consider adding an assertion to the benchmarks: 
- add_onchain_message
- add_ipfs_message

docker/docker-compose.yml
enddynayn: Maybe make a different pr for this?

pallets/messages/src/lib.rs
saraswatpuneet: We might now need storage migration, its a good time to clean it up

saraswatpuneet (approved):
Looks good! just that discussion item seems to be only blocker

pallets/messages/src/benchmarking.rs
enddynayn: It seems good practice to keep this assertion in the assertion section. I notice this update throughout substrates benchmarks.

aramikm:
Closed in favor of https://github.com/LibertyDSNP/frequency/pull/796

PR #771, opened 2022-12-08
Fixes to code documentation from audit

### LINKED ISSUE #742: Code Documentation Fixes
- [x] `runtime/common/src/constants.rs`: a typo in: "Generates 'an' balance based on amount of items and bytes".
- [x] `designdocs/capacity.md`: "The is no limit on the amount that you can schedule" should be "There is no limit on the amount that you can schedule" instead.
- [x] There are several functions, such as try_mutate_delegation() that take another function as a parameter. It would improve code readability if code comments were added to describe the expectations of what the function parameter should do.
- [x] Check Runtime configuration files for missing documentation (may have already been improved)

rlaferla:
https://github.com/LibertyDSNP/frequency/issues/742

# Goal
The goal of this PR is 

Closes #742


enddynayn (approved):
🥇 excellent updates!

saraswatpuneet (approved):
Lgtm

runtime/common/src/constants.rs
```diff
@@ -107,15 +107,20 @@ pub type FrameSystemMaxConsumers = ConstU32<16>;
 // -end- Frame System Pallet ---
 
 // --- MSA Pallet ---
+// The maximum number of public keys per MSA
```
shannonwells, line 110: Just in this file, let's make these all `///` comments so they are made part of the rust docs (I just verified that they're included in the common_runtime generated documentation if so).  The others are already documented in their own pallets. 

IntelliJ's hover-help (and maybe VSCode's too) also will pick the comments up that way, and it's handy to have so you don't have to open the file to look.

shannonwells (commented):
I just have the one request

runtime/common/src/constants.rs
rlaferla: Great suggestion, Shannon!

PR #776, opened 2022-12-09
Return KeyLimitExceeded if incrementing key_count overflows

### LINKED ISSUE #739: Potental Overflow

johnmela:
# Goal
The goal of this PR is to prevent an overflow

Closes #739

# Discussion
Prevents a potential integer overflow  in `add_key` if `MaxPublicKeysPerMsa: u8` is set to `255`.

### DIFF
--- a/pallets/msa/src/lib.rs
+++ b/pallets/msa/src/lib.rs
@@ -911,13 +911,20 @@ impl<T: Config> Pallet<T> {
 			// Increment the key counter
 			<PublicKeyCountForMsaId<T>>::try_mutate(msa_id, |key_count| {
 				// key_count:u8 should default to 0 if it does not exist
-				let incremented_key_count: u8 = *key_count + 1;
-				ensure!(
-					incremented_key_count <= T::MaxPublicKeysPerMsa::get(),
-					Error::<T>::KeyLimitExceeded
-				);
+				let checked_key_count = key_count.checked_add(1);
 
-				*key_count = incremented_key_count;
+				match checked_key_count {
+					// A value of None indicates that key_count would have overflowed
+					None => return Err(Error::<T>::KeyLimitExceeded.into()),
+					Some(incremented_key_count) => {
+						ensure!(
+							incremented_key_count <= T::MaxPublicKeysPerMsa::get(),
+							Error::<T>::KeyLimitExceeded
+						);
+
+						*key_count = incremented_key_count;
+					},
+				}
 				on_success(msa_id)
 			})
 		})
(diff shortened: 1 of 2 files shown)


pallets/msa/src/mock.rs
```diff
@@ -67,7 +67,7 @@ impl pallet_schemas::Config for Test {
 }
 
 parameter_types! {
-	pub const MaxPublicKeysPerMsa: u8 = 10;
+	pub const MaxPublicKeysPerMsa: u8 = 255;
```
johnmela, line 70: Setting the mocked `MaxPublicKeysPerMsa` to `255` allows the existing unit test `add_key_with_more_than_allowed_should_panic` to hit the overflow case. If an increment on `key_count` would overflow, `Error::<T>::KeyLimitExceeded` is returned.

pallets/msa/src/lib.rs
```diff
@@ -911,13 +911,20 @@ impl<T: Config> Pallet<T> {
...
 			<PublicKeyCountForMsaId<T>>::try_mutate(msa_id, |key_count| {
 				// key_count:u8 should default to 0 if it does not exist
-				let incremented_key_count: u8 = *key_count + 1;
-				ensure!(
-					incremented_key_count <= T::MaxPublicKeysPerMsa::get(),
-					Error::<T>::KeyLimitExceeded
-				);
-
-				*key_count = incremented_key_count;
+				let checked_key_count = key_count.checked_add(1);
```
enddynayn, line 914: use sp_runtime::ArithmeticError;

maybe:

```
key_count.checked_add(1).ok_or(ArithmeticError::Overflow)?
```

enddynayn (commented):
Looking good 👍

pallets/msa/src/lib.rs
johnmela: Since the error gets propagated to the end user of the chain and the key limit is being exceeded in the overflow case, I'd like to preserve the error message to `Error::<T>::KeyLimitExceeded`.  

How about:
`key_count.checked_add(1).ok_or(ArithmeticError::Overflow)?`

enddynayn (approved):
fabulous! 👍

saraswatpuneet (approved):

pallets/msa/src/mock.rs
rlaferla: Looks good!

rlaferla (approved):

PR #777, opened 2022-12-09
chore(msa): update benchmarks

enddynayn:
# Goal
Update on_initialize benchmarks
so that it verifies that the
payload-signature-registry storage
is cleared for bucket zero.

Update the mock runtime Msa config,
MortalityWindowSize, so that it is
the same as the frequency runtime.
This update allows running benchmarks
for test and production have the same
configurations. By not having them the same,
the test benchmarks fail to run.

Update test to pass failing test after updating
runtime config MortalityWindowSize.
Update benchmarks.

References #668


pallets/msa/src/benchmarking.rs
```diff
@@ -211,8 +211,14 @@ benchmarks! {
 			let sig = generate_test_signature();
 			assert_ok!(Msa::<T>::register_signature(&sig, T::BlockNumber::from(mortality_block)));
 		}
+
```
enddynayn, line 214: We might have to do some rethinking here.

WEIGHT_PER_SECOND = 1_000_000_000_000 = 10^12

We have two seconds of computation time:
- 2 * 10^12

Cost of 1 write:
1 write = 100_000 * constants::WEIGHT_PER_NANOS = 100_000 * 1_000 = 10 ^ 8

Msa configuration:
- MSAMaxSignaturesPerBucket = 50_000

approx on_initialize weights:
```
 Weight::from_ref_time(0 as u64)
        // Standard Error: 5_000
        .saturating_add(Weight::from_ref_time(1_242_000 as u64).saturating_mul(m as u64))
        .saturating_add(T::DbWeight::get().writes((1 as u64).saturating_mul(m as u64)))
```
an extreme case, if we had 50_000 signatures
10^8 * 50_000 = 5_000_000_000_000 

This exceeds the block space by more than two times without including any other operational-related weights.

There are few solutions to this posted on [stackexchange](https://substrate.stackexchange.com/questions/5842/on-initialize-best-practice-for-executing-a-large-for-loop)
- queue
- using on_idle to slowly drain on each block...ect

pallets/msa/src/weights.rs
```diff
@@ -130,48 +130,52 @@ impl<T: frame_system::Config> WeightInfo for SubstrateWeight<T> {
...
-		Weight::from_ref_time(21_502_000 as u64)
+	fn create_provider(s: u32, ) -> Weight {
+		Weight::from_ref_time(21_273_000 as u64)
+			// Standard Error: 6_000
+			.saturating_add(Weight::from_ref_time(30_000 as u64).saturating_mul(s as u64))
 			.saturating_add(T::DbWeight::get().reads(2 as u64))
 			.saturating_add(T::DbWeight::get().writes(1 as u64))
 	}
+	// Storage: Msa PayloadSignatureRegistry (r:0 w:1)
 	fn on_initialize(m: u32, ) -> Weight {
```
enddynayn, line 156: marking.

shannonwells (approved):
Looks good to go, all the changes to the numbers make sense given what's in the mock.

pallets/msa/src/weights.rs
wilwade: Tracking in issue https://github.com/LibertyDSNP/frequency/issues/785

pallets/msa/src/mock.rs
```diff
@@ -106,7 +106,7 @@ impl pallet_msa::Config for Test {
 	type MaxSchemaGrantsPerDelegation = MaxSchemaGrantsPerDelegation;
 	type MaxProviderNameSize = MaxProviderNameSize;
 	type SchemaValidator = Schemas;
-	type MortalityWindowSize = ConstU32<200>;
+	type MortalityWindowSize = ConstU32<100>;
```
wilwade, line 109: Bringing into alignment with production 👍

wilwade (approved):
🚄

PR #788, opened 2022-12-12
Version bump 0.9.29-2

wilwade:
# Goal
The goal of this PR is to bump the version to 0.9.29-2 for Frequency.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1018,7 +1018,7 @@ dependencies = [
 
 [[package]]
 name = "common-helpers"
-version = "0.9.29"
+version = "0.9.29-2"
 dependencies = [
  "apache-avro",
  "common-primitives",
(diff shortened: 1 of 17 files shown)


shannonwells (approved):

aramikm (approved):
Looks good

aramikm:
@wilwade  Does https://github.com/LibertyDSNP/frequency/blob/main/integration-tests/package.json#L16 or https://github.com/LibertyDSNP/frequency/blob/main/js/api-augment/package.json#L3 need to be changed?

wilwade:
@aramikm 
> @wilwade Does https://github.com/LibertyDSNP/frequency/blob/main/integration-tests/package.json#L16 or https://github.com/LibertyDSNP/frequency/blob/main/js/api-augment/package.json#L3 need to be changed?

- The api-augment one does not. It is automatically set.
- The integration one is odd as the npm package doesn't exist yet. When we get this running in CI, it will use a local build of api-augment, but not sure what to do right now.

PR #789, opened 2022-12-12
Audit test and test comments

### LINKED ISSUE #735: Potential Replay Attack: Unprevented Edge Case
`pallets/msa/src/lib.rs`

Currently a signature registry is implemented to prevent replay attacks. The registry would delete all the signatures registered in a virtual bucket N using reset_virtual_bucket_if_needed(), when switching from the virtual bucket N to the next virtual bucket (i.e., the bucket N+1). Replay attack could happen if a signature is used near the end of the virtual bucket N, and the signature_expires_at falls within the bucket N+1.

## Scenario
1. Assume that MortalityWindowSize == 10, NumberOfBuckets == 10, and current_block == 9.
2. Alice signed a signature sig where signature_expires_at == 13 for Bob to transfer funds from Alice to Bob.
3. Bob successfully used sig when current_block == 9, and the chain registered sig.
4. Later when current_block == 10, the chain called reset_virtual_bucket_if_needed() and removed the previously registered sig.
5. Bob successfully used sig when current_block == 11, and the chain registered sig again.

shannonwells:
add clarifying comments, some rearranging and removal of duplicate tests, and demonstrate that a particular boundary case does not result in a replay

# Goal
The goal of this PR is 1) to add some comments on test cases as to what the point of each is, and b) to write up a test case proposed by the auditors to see if the test fails, and if so to fix it.

Closes #735 

# Discussion
1. is done
b. it doesn't fail

# Checklist
- [x] Tests added

### DIFF
--- a/pallets/msa/src/replay_tests.rs
+++ b/pallets/msa/src/replay_tests.rs
@@ -178,22 +178,34 @@ fn replaying_grant_delegation_fails() {
 }
 
 #[test]
-pub fn add_signature_replay_fails() {
+pub fn add_signature_replay_boundary_checks() {
 	struct TestCase {
 		current: u64,
 		mortality: u64,
 		run_to: u64,
 	}
 	new_test_ext().execute_with(|| {
-		// these should all fail replay
+		// This tests signature replay attacks for mortality window size = 100 and 2 buckets,
+		// .by looking at different boundary cases.  It checks that they can't
+		// be resubmitted at the expiration block. We assume if they cannot be replayed at a
+		// boundary that they can't be replayed earlier than that boundary.
 		let test_cases: Vec<TestCase> = vec![
-			// fails because
-			TestCase { current: 10_949u64, mortality: 11_001u64, run_to: 11_000u64 }, // fails test
+			// 1-block expiration for bucket 0, mortality crosses no boundary
 			TestCase { current: 1u64, mortality: 3u64, run_to: 2u64 },
+			// expiration for bucket 1, mortality is fast, crosses a 100-block boundary, we check at the boundary.
+			// at block 100, on_initialize's bucket-clearing has happened by the time register_extrinsic is called,
+			// so this should make sure that the signature is still there and the wrong bucket was not cleared.
 			TestCase { current: 99u64, mortality: 101u64, run_to: 100u64 },
+			// This does the same as above, only it's a different bucket.
+			TestCase { current: 11_149u64, mortality: 11_201u64, run_to: 11_200u64 },
+			// This sets mortality at a boundary
+			TestCase { current: 11_149u64, mortality: 11_200u64, run_to: 11_199u64 },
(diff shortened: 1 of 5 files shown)


pallets/msa/src/tests.rs
```diff
@@ -2206,36 +2203,6 @@ pub fn clears_stale_signatures_after_mortality_limit() {
 	})
 }
 
-#[test]
-pub fn add_signature_replay_fails() {
```
shannonwells, line 2210: These were duplicates of ones in replay_tests.

pallets/msa/src/replay_tests_2.rs
```diff
@@ -0,0 +1,171 @@
+use crate::{self as pallet_msa, types::EMPTY_FUNCTION, AddKeyData, AddProvider, Config, Error};
```
shannonwells, line 1: Not sure how else to do proposed boundary checks which require different configs.  Maybe the filename could be renamed though.

pallets/msa/src/replay_tests.rs
```diff
@@ -198,6 +210,11 @@ pub fn add_signature_replay_fails() {
 				Msa::register_signature(sig1, tc.mortality),
 				Error::<Test>::SignatureAlreadySubmitted,
 			);
+			run_to_block(tc.run_to + 1);
```
wilwade, line 213: Should this be just:
```suggestion
			run_to_block(tc.mortality);
```

pallets/msa/src/replay_tests.rs
```diff
@@ -172,22 +172,34 @@ fn replaying_grant_delegation_fails() {
...
+			// This sets mortality at a boundary
+			TestCase { current: 11_149u64, mortality: 11_200u64, run_to: 11_199u64 },
+			// This does the same as above, but for a different bucket.
+			TestCase { current: 11_249u64, mortality: 11_300u64, run_to: 11_299u64 },
+			// This case sets current block at a boundary, and sets the mortality to the very last block before the boundary
 			TestCase { current: 1_100u64, mortality: 1_199u64, run_to: 1_198u64 },
-			TestCase { current: 1_102u64, mortality: 1_201u64, run_to: 1_200u64 },
-			TestCase { current: 1_099u64, mortality: 1_148u64, run_to: 1_101u64 },
+			// This case has current block right before a boundary, and sets expiration to the very last possible block
+			TestCase { current: 1_699u64, mortality: 1_798u64, run_to: 1_797u64 },
```
wilwade, line 202: All of the `run_to`s are `mortality - 1`. I guess that makes sense, but do we want `run_to` if that is the case?

pallets/msa/src/replay_tests.rs
shannonwells: Yeah I think that's better.

enddynayn (approved):

PR #791, opened 2022-12-13
chore(msa): update on-initialize benchmarks

enddynayn:
# Goal
Update on-initialize benchmarks to use
T::MaxSignaturesPerBucket for calculating
benchmarks. This update allows benchmarks
to be calculated based on production configuration.

Update how signatures are calculated to decrease
the time that it takes to run benchmarks.

References #668

### DIFF
--- a/pallets/msa/src/benchmarking.rs
+++ b/pallets/msa/src/benchmarking.rs
@@ -67,13 +67,6 @@ fn create_msa_account_and_keys<T: Config>() -> (T::AccountId, SignerId, MessageS
 	(account_id, key_pair, msa_id)
 }
 
-pub fn generate_test_signature() -> MultiSignature {
-	let account = SignerId::generate_pair(None);
-	let fake_data = vec![4u8; 32];
-	let signature = account.sign(&fake_data).unwrap();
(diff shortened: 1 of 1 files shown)


shannonwells (approved):
Awesome

pallets/msa/src/benchmarking.rs
```diff
@@ -205,11 +198,13 @@ benchmarks! {
 	on_initialize {
 		// we should not need to max out storage for this benchmark, see:
 		// https://substrate.stackexchange.com/a/4430/2060
-		let m in 1 .. 3_000;
+		let m in 1 .. T::MaxSignaturesPerBucket::get();
 		for j in 0 .. m {
 			let mortality_block = 49 as u32;
```
aramikm, line 203: what is the reason behind keeping `mortality_block` fixed?

aramikm (approved):
Looks good!

pallets/msa/src/benchmarking.rs
enddynayn: Keeping it constant makes it easy since making mortality_block more dynamic would not impact the benchmarks.

PR #796, opened 2022-12-13
Messages whitelist storage

### LINKED ISSUE #768: Audit: Simpler message storage

aramikm:
# Goal
The goal of this PR is simplify adding messages using `whitelist_storage` macro

Closes #768 

# Checklist
- [x] Design doc(s) updated
- [x] Tests added
- [x] Benchmarks added
- [x] Weights updated


pallets/messages/src/lib.rs
```diff
@@ -315,28 +345,25 @@ impl<T: Config> Pallet<T> {
...
+				provider_msa_id,
+				msa_id,
+				index,
+			};
+
+			existing_messages
+				.try_push(msg)
+				.map_err(|_| Error::<T>::TooManyMessagesInBlock)?;
+
+			Self::track_schema_id(schema_id)?;
```
saraswatpuneet, line 363: you can avoid a `try_mutate` on ```BlockSchema``` if ```existing_messages count``` is ==1 is when you need to add schema id to  ```BlockSchema``` otherwise we know can safely assume it exists and avoid calls to storage on each message add?

saraswatpuneet: I see you are adding a counter too, probably there is a need

pallets/messages/src/weights.rs
```diff
@@ -64,35 +64,34 @@ impl<T: frame_system::Config> WeightInfo for SubstrateWeight<T> {
...
+	// Storage: Messages Messages (r:1 w:1)
+	// Storage: Messages MessageIndex (r:1 w:1)
+	// Storage: Messages BlockSchemas (r:1 w:1)
 	fn add_onchain_message(n: u32, ) -> Weight {
-		Weight::from_ref_time(127_540_000 as u64)
-			// Standard Error: 14
-			.saturating_add(Weight::from_ref_time(1_079 as u64).saturating_mul(n as u64))
-			.saturating_add(T::DbWeight::get().reads(4 as u64))
-			.saturating_add(T::DbWeight::get().writes(1 as u64))
+		Weight::from_ref_time(132_900_000 as u64)
```
aramikm, line 71: DB read and writes should not increase.

pallets/messages/src/lib.rs
aramikm: To Avoid changing event contract I'm counting how many times it's used. Nevermind since it seems like `whitelist_storage` is not working as expected and we have to figure out what is going on there

pallets/messages/src/lib.rs
```diff
@@ -146,6 +148,23 @@ pub mod pallet {
 		ValueQuery,
 	>;
 
+	/// A temporary storage of unique schema ids and counts, for a duration of block period.
+	/// Main use for this storage is to propagate events
+	/// At the start of the next block this storage is cleared
+	/// - Value: Sorted List(Set) of SchemaIds and their counts
+	#[pallet::storage]
```
enddynayn, line 155: we discusses how to get rid of this.

pallets/messages/src/lib.rs
```diff
@@ -186,11 +208,17 @@ pub mod pallet {
...
-			let prev_block = current - T::BlockNumber::one();
-			Self::move_messages_into_final_storage(prev_block)
+		fn on_initialize(_current: T::BlockNumber) -> Weight {
+			T::WeightInfo::on_finalize()
 			// TODO: add retention policy execution GitHub Issue: #126 and #25
 		}
+
+		fn on_finalize(block_number: T::BlockNumber) {
+			<MessageIndex<T>>::set(0u16);
+			for (schema_id, count) in <BlockSchemas<T>>::take() {
```
enddynayn, line 218: We can get rid of this as discussed.

pallets/messages/src/lib.rs
```diff
@@ -179,16 +191,15 @@ pub mod pallet {
 			schema_id: SchemaId,
 			/// The block number for these messages
 			block_number: T::BlockNumber,
-			/// Number of messages in this block for this schema
```
aramikm, line 182: This might cause some backwards compatibility issues for Peter, but I think it will worth it due to simplifying a lot of event related problems

pallets/messages/src/lib.rs
```diff
@@ -179,16 +191,15 @@ pub mod pallet {
...
 		},
 	}
 
 	#[pallet::hooks]
 	impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {
-		fn on_initialize(current: T::BlockNumber) -> Weight {
-			let prev_block = current - T::BlockNumber::one();
-			Self::move_messages_into_final_storage(prev_block)
+		fn on_initialize(_current: T::BlockNumber) -> Weight {
+			<MessageIndex<T>>::set(0u16);
```
aramikm, line 200: reseting temporary to 0

runtime/frequency/src/lib.rs
```diff
@@ -989,18 +989,8 @@ impl_runtime_apis! {
 			impl cumulus_pallet_session_benchmarking::Config for Runtime {}
 			use orml_benchmarking::{add_benchmark as orml_add_benchmark};
 
-			let whitelist: Vec<TrackedStorageKey> = vec![
```
aramikm, line 992: Automatically getting all the whitelisted storages instead of hardcoding. Based on https://github.com/paritytech/substrate/pull/12205/files

pallets/messages/src/lib.rs
aramikm: Outdated

pallets/messages/src/lib.rs
```diff
@@ -146,6 +148,13 @@ pub mod pallet {
 		ValueQuery,
 	>;
 
+	/// A temporary storage for getting the index for messages
+	/// At the start of the next block this storage is set to 0
+	#[pallet::storage]
+	#[pallet::whitelist_storage]
```
aramikm, line 154: Temporary storage for indices

pallets/messages/src/lib.rs
```diff
@@ -369,48 +385,4 @@ impl<T: Config> Pallet<T> {
...
-	/// and calculates execution weight
-	///
-	/// Called inside of `on_initialize`
-	///
-	/// # Events
-	/// * [`Event::MessagesStored`]
-	///
-	/// Returns execution weights
-	///
-	fn move_messages_into_final_storage(block_number: T::BlockNumber) -> Weight {
```
enddynayn, line 383: 💯

runtime/frequency/src/lib.rs
```diff
@@ -1041,3 +1031,40 @@ cumulus_pallet_parachain_system::register_validate_block! {
 	BlockExecutor = cumulus_pallet_aura_ext::BlockExecutor::<Runtime, Executive>,
 	CheckInherents = CheckInherents,
 }
+
+#[cfg(test)]
+mod tests {
```
enddynayn, line 1036: 💯

pallets/messages/src/lib.rs
enddynayn: What if we do not clear it and keep the index?

pallets/messages/src/lib.rs
```diff
@@ -315,28 +328,31 @@ impl<T: Config> Pallet<T> {
...
+				index,
+			};
+
+			existing_messages
+				.try_push(msg)
+				.map_err(|_| Error::<T>::TooManyMessagesInBlock)?;
+
+			if existing_messages.len() == 1 {
+				// first message for any schema_id is going to trigger an event
+				Self::deposit_event(Event::MessagesStored {
```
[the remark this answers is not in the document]
aramikm: which function? `add_message` ?

pallets/messages/src/benchmarking.rs
```diff
@@ -92,21 +99,5 @@ benchmarks! {
 		}
 	}: _ (RawOrigin::Signed(caller),IPFS_SCHEMA_ID, cid, IPFS_PAYLOAD_LENGTH)
 
-	on_initialize {
```
enddynayn, line 95: Let's consider adding a final check of the expected end state to the above benchmarks.

pallets/messages/src/lib.rs
aramikm: It will overflow. We just need order in a block to be able to get canonical order. To keep each message size small I would not like to increase existing u16 to `u32` or `u64`

pallets/messages/src/lib.rs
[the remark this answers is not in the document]
enddynayn: Self:: deposit_event event. Also, let's consider renaming this function to create_message.

pallets/messages/src/lib.rs
enddynayn: Ahh thats right, each message has an index.

pallets/messages/src/lib.rs
```diff
@@ -123,6 +124,7 @@ pub mod pallet {
 	/// A temporary storage of messages, given a schema id, for a duration of block period.
 	/// At the start of the next block this storage is cleared and moved into Messages storage.
 	/// - Value: List of Messages
+	/// Obsolete storage (will be removed in next storage migration)
```
enddynayn, line 127: I think we can remove this here, right - it's empty.

pallets/messages/src/lib.rs
```diff
@@ -168,6 +177,9 @@ pub mod pallet {
 
 		/// Invalid payload location
 		InvalidPayloadLocation,
+
+		/// Attempted to call `store` outside of block execution.
+		BadContext,
```
enddynayn, line 182: Do we need this?

runtime/frequency/src/lib.rs
```diff
@@ -989,18 +989,8 @@ impl_runtime_apis! {
...
-				hex_literal::hex!("c2261276cc9d1f8598ea4b6a74b15c2f57c875e4cff74148e4628f264b974c80").to_vec().into(),
-				// Execution Phase
-				hex_literal::hex!("26aa394eea5630e07c48ae0c9558cef7ff553b5a9862a516939d82b3d3d8661a").to_vec().into(),
-				// Event Count
-				hex_literal::hex!("26aa394eea5630e07c48ae0c9558cef70a98fdbe9ce6c55837576c60c7af3850").to_vec().into(),
-				// System Events
-				hex_literal::hex!("26aa394eea5630e07c48ae0c9558cef780d41e5e16056765bc8461851072c9d7").to_vec().into(),
-			];
+			use frame_support::traits::WhitelistedStorageKeys;
+			let whitelist: Vec<TrackedStorageKey> = AllPalletsWithSystem::whitelisted_storage_keys();
```
wilwade, line 993: Does this require a bump of the runtime version?

pallets/messages/src/lib.rs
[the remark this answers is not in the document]
wilwade: It's not a super long function, so I'd say we should keep the event in here. It just looks long due to the formatting of the code.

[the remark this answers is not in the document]
aramikm: Would you please elaborate more on this.

pallets/messages/src/lib.rs
aramikm: Not anymore, Good find!

wilwade (commented):
Looking good. One suggestion to limit breaking changes.

pallets/messages/src/lib.rs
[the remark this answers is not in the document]
enddynayn: Yeah. Single responsible principle. Also, we do not do this anywhere else. It is not difficult to do.

pallets/messages/src/lib.rs
enddynayn: Test it on rococo and development.

pallets/messages/src/benchmarking.rs
```diff
@@ -75,6 +82,13 @@ benchmarks! {
 			assert_ok!(onchain_message::<T>(schema_id));
 		}
 	}: _ (RawOrigin::Signed(caller), Some(message_source_id.into()), schema_id, payload)
+	verify {
```
enddynayn, line 85: thanks!

pallets/messages/src/lib.rs
enddynayn: I do not understand the reasoning we are leaving it. It feels like we are worrying about a problem we do not have.

wilwade: True we have a limited number of integrations, but setting I want to set the patterns across the board.

This event in particular is being used by at least Peter. Let's keep the churn low as possible.

pallets/messages/src/lib.rs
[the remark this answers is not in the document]
wilwade: @enddynayn So moving it to the end of the function? I think that is reasonable to wait to trigger it until the end of the actions.

pallets/messages/src/lib.rs
wilwade: We cannot just test it on frequency-rococo now as that's a testnet and should always be either in line with production, or pre-release for production.

If there isn't a storage migration needed, I'm fine with just removing it.

pallets/messages/src/lib.rs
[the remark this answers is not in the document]
enddynayn: Yes. @aramikm, I realize you cannot read my mind. I will make a better effort to elaborate more when I leave a comment. Thanks for taking the time to address my comments.

runtime/frequency/src/lib.rs
aramikm: Bumped the spec version. I realized we still don't have an easy rule of thumb method to figure out when exactly we need to bump the versions or at least I don't. And there are multiple version also like author_version

runtime/frequency/src/lib.rs
```diff
@@ -164,7 +164,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	spec_name: create_runtime_str!("frequency"),
 	impl_name: create_runtime_str!("frequency"),
 	authoring_version: 1,
-	spec_version: 4,
+	spec_version: 5,
```
aramikm, line 167: bumping the spec version

runtime/frequency/src/lib.rs
wilwade: We are working on some CI to help with some of that. Mostly it is the spec version we change.

pallets/messages/src/lib.rs
```diff
@@ -120,17 +121,6 @@ pub mod pallet {
 	#[pallet::without_storage_info]
 	pub struct Pallet<T>(_);
 
-	/// A temporary storage of messages, given a schema id, for a duration of block period.
```
aramikm, line 123: this is removed after some local testing, message add worked and the old storage did not appear after upgrade as expected!

Screen Shot 2022-12-14 at 3 39 14 PM
Screen Shot 2022-12-14 at 3 39 33 PM

pallets/messages/src/lib.rs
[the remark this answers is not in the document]
aramikm: This is resolved

pallets/messages/src/lib.rs
aramikm: removed after local upgrade test

enddynayn (approved):
⛹️‍♂️  amazing! So many great improvements!

PR #797, opened 2022-12-13
Bugs/signature storage oflo #753

shannonwells:
# Goal
The goal of this PR is to add limits to signature registration storage.

Closes #753

# Discussion
Adds new Config values for the bucket storage:  one for how many signatures per bucket and one for how many total are allowed. Adds error checks and enums for when a bucket is full or `checked_add` fails.

The limit for how many signatures in storage is used for configuring the signature registration `StorageDoubleMap`.  To constrain the size, it takes a type and cannot accept a calculation. That's why there is a separate config for it, even though that value may be calculated with `NumberOfBuckets * MaxSignaturesPerBucket`.  I've added comments in the configs to this extent, and since we're controlling the config values, I think the risk is very low for the values being off.

Some signature registry tests were moved to their own file with their own pallet config, because they needed small limits to be efficient, whereas the benchmarks need much bigger limits.

Update: instead of adding a new failure enum, I changed our `checked_add` calls to return `ArithmeticError::Overflow` just as substrate does when it fails.

# Checklist
- [x] Tests added
- [x] Weights updated

## Not applicable:
- Chain spec updated
- Design doc(s) updated
- Benchmarks added


pallets/msa/src/audit_replay_tests.rs
```diff
@@ -105,32 +108,10 @@ impl pallet_msa::Config for Test {
 	type MaxProviderNameSize = MaxProviderNameSize;
 	type SchemaValidator = Schemas;
 	type MortalityWindowSize = ConstU32<10>;
-	type MaxSignaturesPerBucket = ConstU32<10>;
+	type MaxSignaturesPerBucket = ConstU32<20>;
 	type NumberOfBuckets = ConstU32<10>;
-}
-
-pub fn new_test_ext() -> sp_io::TestExternalities {
```
shannonwells, line 112: importing from mock instead

pallets/msa/src/mock.rs
```diff
@@ -107,8 +107,10 @@ impl pallet_msa::Config for Test {
 	type MaxProviderNameSize = MaxProviderNameSize;
 	type SchemaValidator = Schemas;
 	type MortalityWindowSize = ConstU32<100>;
-	type MaxSignaturesPerBucket = ConstU32<10>;
+	type MaxSignaturesPerBucket = ConstU32<4000>;
```
shannonwells, line 110: This had to be increased - because of the new limit on signatures, the benchmark test was failing.

pallets/msa/src/signature_registry_tests.rs
```diff
@@ -0,0 +1,336 @@
+use crate::{
```
shannonwells, line 1: Moved signature registry tests in here (except for replays) because of the need for a different config (mainly smaller signature registry limits) compared to what benchmarks needs.  This allowed me to undo changes in more or less unrelated tests to get them to pass as well.

pallets/msa/src/tests.rs
```diff
@@ -2142,201 +2142,6 @@ pub fn is_registered_provider_is_true() {
 	});
 }
 
-fn register_signature_and_validate(
```
shannonwells, line 2145: moved to own file

pallets/msa/src/lib.rs
```diff
@@ -1197,12 +1228,43 @@ impl<T: Config> Pallet<T> {
...
+						bucket_signatures.get_mut(bucket_index).ok_or(Error::<T>::NoSuchBucket)?;
+
+					let limit = T::MaxSignaturesPerBucket::get();
+					ensure!(
+						*bucket_signature_count < limit,
+						Error::<T>::SignatureRegistryLimitExceeded
+					);
+					let new_count = bucket_signature_count
+						.checked_add(1)
+						.ok_or(Error::<T>::FailedBucketSignatureCountOverflowCheck)?;
```
enddynayn, line 1251: substrate does something like:
```
use sp_runtime::ArithmeticError;
checked_add(1).ok_or(ArithmeticError::Overflow)?
```

shannonwells: Now using that error.

pallets/msa/src/signature_registry_tests.rs
```diff
@@ -0,0 +1,356 @@
...
+			TestCase { block: 1_010, expected_bucket: 0 },
+			TestCase { block: 1_110, expected_bucket: 1 },
+			TestCase { block: 1_201, expected_bucket: 0 },
+			TestCase { block: 1_301, expected_bucket: 1 },
+			TestCase { block: 1_401, expected_bucket: 0 },
+			TestCase { block: 1_501, expected_bucket: 1 },
+			TestCase { block: 1_601, expected_bucket: 0 },
+			TestCase { block: 1_701, expected_bucket: 1 },
+			TestCase { block: 1_801, expected_bucket: 0 },
+			TestCase { block: 1_901, expected_bucket: 1 },
```
saraswatpuneet, line 214: :dolphin:

pallets/msa/src/lib.rs
```diff
@@ -218,13 +226,25 @@ pub mod pallet {
...
-		OptionQuery,
+		OptionQuery,    // The type for the query
+		GetDefault,     // OnEmpty return type, defaults to None
+		T::MaxSignaturesStored, // Maximum total signatures to store
+	>;
+
+	/// Records how many signatures are currently stored in each virtual signature registration bucket
+	#[pallet::storage]
+	pub(super) type PayloadSignatureBucketStorageCount<T: Config> = StorageMap<
+		_,
```
enddynayn, line 243: PayloadSignatureBucketStorageCount > I think we can drop storage from the name since the macro above implies its storage.

enddynayn (approved):
Fabulous! 👍

pallets/msa/src/lib.rs
```diff
@@ -366,8 +386,15 @@ pub mod pallet {
 
 		/// Cryptographic signature verification failed for proving ownership of new public-key.
 		NewKeyOwnershipInvalidSignature,
+
 		/// Attempted to request validity of schema permission or delegation in the future.
 		CannotPredictValidityPastCurrentBlock,
+
+		/// Attempted to access a virtual signature registration bucket that is out of bounds
+		NoSuchBucket,
```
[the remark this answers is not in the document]
shannonwells: Good catch, and  it's not in the new release so out it goes.

saraswatpuneet (approved):

pallets/msa/src/signature_registry_tests.rs
aramikm: I wish there was a way to have some kind of runtime builder that you could pass in the config params and then you could use that runtime to run against tests. So far I have't seen anything like that if we can make one we can get rid of all these setup runtime duplicates

pallets/msa/src/mock.rs
```diff
@@ -107,8 +107,10 @@ impl pallet_msa::Config for Test {
...
 	type NumberOfBuckets = ConstU32<2>;
+	/// This MUST ALWAYS be MaxSignaturesPerBucket * NumberOfBuckets.
+	type MaxSignaturesStored = ConstU32<8000>;
```
aramikm, line 113: nit: is it possible to define it by multiplying those two?

runtime/common/src/constants.rs
```diff
@@ -283,12 +283,17 @@ impl Clone for MessagesMaxPayloadSizeBytes {
...
+// --- MSA Pallet ---
 /// The number of blocks per virtual bucket
 pub type MSAMortalityWindowSize = ConstU32<100>;
 /// The maximum number of signatures per virtual bucket
 pub type MSAMaxSignaturesPerBucket = ConstU32<50_000>;
 /// The total number of virtual buckets
 pub type MSANumberOfBuckets = ConstU32<2>;
+/// The upper limit on total stored signatures.
+/// This MUST be MaxSignaturesPerBucket * NumberOfBuckets.
+pub type MSAMaxSignaturesStored = ConstU32<100_000>;
```
aramikm, line 295: nit: do we really need this, since it is completely derivable from other configs?

aramikm (approved):
Looks great!

PR #804, opened 2022-12-15
Update the `main` development branch to have a version of 0.0.0

wilwade:
# Goal
The goal of this PR is to align with the https://github.com/LibertyDSNP/frequency/wiki/Release documentation

# Discussion
- Updated cargo toml versions to 0.0.0
- Added `make version v=X.X.X` to make updating versions easy.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -134,3 +134,17 @@ test:
 
 integration-test:
 	./scripts/run_integration_tests.sh
+
+PHONY: version
+version:
+ifndef v
+	@echo "Please set the version with v=X.X.X-X"
+	@exit 1
+endif
+ifneq (,$(findstring v,  $(v)))
+	@echo "Please don't prefix with a 'v'. Use: v=X.X.X-X"
+	@exit 1
+endif
+	find ./ -type f -name 'Cargo.toml' -exec sed -i '' 's/^version = \"0\.0\.0\"/version = \"$(v)\"/g' {} \;
+	cargo check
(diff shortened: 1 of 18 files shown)


Makefile
```diff
@@ -133,4 +133,17 @@ test:
...
+ifndef v
+	@echo "Please set the version with v=X.X.X-X"
+	@exit 1
+endif
+ifneq (,$(findstring v,  $(v)))
+	@echo "Please don't prefix with a 'v'. Use: v=X.X.X-X"
+	@exit 1
+endif
+	find ./ -type f -name 'Cargo.toml' -exec sed -i '' 's/^version = \"[0-9\.]*\"/version = \"$(v)\"/g' {} \;
+	cargo check
```
saraswatpuneet, line 149: :wheel_of_dharma:

saraswatpuneet (approved):

johnmela (approved):

Makefile
aramikm: Is there any chance that this might match with multiline dependency version? (not that we have any now but I remember we had something like that before) ex:

https://github.com/Liberty30/polkadot-experiments/blob/main/pallets/announcement/Cargo.toml#L24

aramikm (approved):

wilwade:
@aramikm 

> Is there any chance that this might match with multiline dependency version?

Hmm... Yes there is. I'll update it to only search for the 0.0.0. That would be safer.

PR #810, opened 2022-12-16
add metadata compare artifact to Release workflow

### LINKED ISSUE #624: [CI] Setup extrinsic ordering checks prior releases compared to mainnet
## Goal

Once Frequency is on mainnet, we should setup an [extrinsic ordering check](https://github.com/paritytech/polkadot/blob/master/doc/release-checklist.md#extrinsic-ordering) compared to live chain metadata.

TODO:
-  [ ] Setup a github action similar to [this](https://github.com/paritytech/polkadot/actions/workflows/release-20_extrinsic-ordering-check-from-bin.yml)

- [ ] Tie it with release process?

demisx:
# Goal
The goal of this PR is to generate metadata compare diff between the binary in the previous full release and the binary in the PR. The metadata diff is saved as `metadata-compare-mainnet.txt` and and attached to the release artifacts. 

See example file below:

- [metadata-compare-mainnet.txt](https://github.com/LibertyDSNP/frequency/files/10522538/metadata-compare-mainnet.txt) ⬇️ 

This is where it's attached in the release artifacts:

Screenshot 2023-01-27 at 11 36 07 AM

Closes #624


.github/workflows/release.yml
```diff
@@ -344,12 +335,137 @@ jobs:
...
+        uses: actions/upload-artifact@v3
+        with:
+          name: artifacts-${{github.run_id}}
+          path: ${{env.OUTPUT_DIR}}/${{env.OUTPUT_FILENAME}}
+          if-no-files-found: error
+      - name: Stop Local Nodes
+        if: always()
+        run: |
+          echo "Attempting to kill all frequency processes..."
+          pkill -9 ${{env.FREQUENCY_PROCESS_NAME}}
```
wilwade, line 460: @demisx Are we sure this is safe on the shared runners?

.github/workflows/release.yml
```diff
@@ -344,12 +335,137 @@ jobs:
...
+          ./${{env.REF_BIN_FILENAME}} \
+            --rpc-external \
+            --rpc-cors=all \
+            --ws-external \
+            --rpc-methods=Unsafe \
+            --no-telemetry \
+            --no-prometheus \
+            --port $((30334)) \
+            --rpc-port $((9934)) \
+            --ws-port $((9945)) \
```
wilwade, line 420: @demisx These are the default ports that the relay part of the test node uses. And (for reasons I am not sure of why) substrate binaries do not complain when they use a port that conflicts with another open process.

Also for this you only really care about the `--ws-port`.

Options:
- Update the ref node ws port to `9946`
- Add `--ws-port` after the `--` for the test node and reference nodes to non-conflicting ports



```suggestion
            --ws-port $((9946)) \
```

.github/workflows/release.yml
```diff
@@ -344,12 +335,137 @@ jobs:
...
+          REF_VERSION=$(./${{env.REF_BIN_FILENAME}} --version)
+          BIN_VERSION=$(./${{env.TEST_BIN_FILENAME}} --version)
+          echo "Metadata comparison against previous full release:" >> ${{env.OUTPUT_FILENAME}}
+          echo "Date: $(date)" >> ${{env.OUTPUT_FILENAME}}
+          echo "Ref. version: $REF_VERSION" >> ${{env.OUTPUT_FILENAME}}
+          echo "Test version: $BIN_VERSION" >> ${{env.OUTPUT_FILENAME}}
+          echo "----------------------------------------------------------------------" >> ${{env.OUTPUT_FILENAME}}
+      - name: Compare Metadata
+        run: |
+          CMD="docker run --pull always --network host jacogr/polkadot-js-tools metadata ws://localhost:9945 ws://localhost:9944"
```
wilwade, line 438: Here's the other side of that port problem.
```suggestion
          CMD="docker run --pull always --network host jacogr/polkadot-js-tools metadata ws://localhost:9946 ws://localhost:9944"
```

wilwade (changes requested):
One fix needed

wilwade (approved):

PR #822, opened 2022-12-20
schema: more accurate benchmarks

### LINKED ISSUE #669: Review Benchmarks: Schema pallet

aramikm:
# Goal
The goal of this PR is review benchmarks in schema pallet and modify them if necessary

Closes #669 

# Discussion
- Removed extra parameter in `create_schema` benchmarks since it was not necessary
- Added benchmark for `set_max_schema_model_bytes`

# Checklist
- [x] Benchmarks added
- [x] Weights updated


pallets/schemas/src/benchmarking.rs
```diff
@@ -32,36 +30,28 @@ fn generate_schema<T: Config>(
...
 		ensure!(SchemasPallet::<T>::get_current_schema_identifier_maximum() > 0, "Registered schema count should be > 0");
+		ensure!(SchemasPallet::<T>::get_schema(1).is_some(), "Registered schema should exist");
 	}
+
+	set_max_schema_model_bytes {
+		let sender = RawOrigin::Root;
+		let max_size = T::SchemaModelMaxBytesBoundedVecLimit::get();
+	}: _(sender, max_size)
+	verify {
+		ensure!(SchemasPallet::<T>::get_schema_model_max_bytes() == T::SchemaModelMaxBytesBoundedVecLimit::get(), "Schema model max should be updated!");
```
rlaferla, line 52: L👀ks good.  BTW: I'm wondering why in the pallet we are using `MinSchemaModelSizeBytes` and not using `MaxSchemaModelSizeBytes` instead of `SchemaModelMaxBytesBoundedVecLimit`?

rlaferla (approved):

pallets/schemas/src/benchmarking.rs
aramikm: I guess the reason was to be more flexible and be able to update the max dynamically by governance

Issue #786, opened 2022-12-12
Requiring IPFS CIDs as multibase-encoded strings with CIDv1


wilwade:
To make it more likely that our IPFS CIDs are correct and stored efficiently, we should require IPFS CIDv1s for `add_ipfs_message`.

Note: Expected to expand to CIDv2+ as they come out with the only change being the ensure statement assuming v2+ still follow `base32` encoding.

Related:
- #738 
- #787
- CID Spec: https://github.com/multiformats/cid
- CID Inspector: https://cid.ipfs.tech
- JS for multiformats https://github.com/multiformats/js-multiformats/blob/master/src/cid.js

## Acceptance Criteria
- [x] `add_ipfs_message` should ensure CIDv1
- [x] NEW Error: `Unsupported CID version` Supplied CID is v0
- [x] NEW Error: `Invalid CID` Supplied CID value cannot be decoded as a CID
- [x] Calling `add_ipfs_message` with CIDv1 should succeed:
    - SUCCESS (string)`bafybeihdwdcefgh4dqkjv67uzcmw7ojee6xedzdetojuzjevtenxquvyku`
    - FAIL: hex: `0x01701220e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855`
    - FAIL (utf-8 to hex) `0x516d64665462427142505137564e785a4559456a3134566d52755a426b714662697752656f674a6753317a52316e0x626166796265696864776463656667683464716b6a763637757a636d77376f6a6565367865647a6465746f6a757a6a657674656e78717576796b75`
- [x] Calling `add_ipfs_message` with CIDv0 should always fail:
    - `QmdfTbBqBPQ7VNxZEYEj14VmRuZBkqFbiwReogJgS1zR1n`
    - Failing UTF-8 Hex encoding of: `0x516d64665462427142505137564e785a4559456a3134566d52755a426b714662697752656f674a6753317a52316e`
    - Failing v0 Bytes: `0x1220e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855`
- [x] Better method documentation around the format required for IPFS:
    - [x] https://libertydsnp.github.io/frequency/common_primitives/schema/enum.PayloadLocation.html#variant.IPFS
    - [x] https://libertydsnp.github.io/frequency/pallet_messages/pallet/struct.Pallet.html#method.add_ipfs_message

JoeCap08055:
[...]

PR #824, opened 2022-12-20
Moderate rework of integration test suites. (#798)

### LINKED ISSUE #798: Add error handling to integration test scaffolding
Currently, integration tests know if an extrinsic call failed but not why.  The integration test scaffolding needs to be enhanced to support errors returned from failed extrinsic calls.

e.g.
```
api.tx.balances
  .transfer(recipient, 123)
  .signAndSend(sender, ({ status, events, dispatchError }) => {
    // status would still be set, but in the case of error we can shortcut
    // to just check it (so an error would indicate InBlock or Finalized)
    if (dispatchError) {
      if (dispatchError.isModule) {
        // for module errors, we have the section indexed, lookup
        const decoded = api.registry.findMetaError(dispatchError.asModule);
        const { docs, name, section } = decoded;

        console.log(`${section}.${name}: ${docs.join(' ')}`);
      } else {
        // Other, CannotLookup, BadOrigin, no extra info
        console.log(dispatchError.toString());
      }
    }
  });
```

See also:
https://polkadot.js.org/docs/api/cookbook/tx/#how-do-i-get-the-decoded-enum-for-an-extrinsicfailed-event

JoeCap08055:
# Goal
Add a standard mechanism to integration tests to parse and re-throw errors from extrinsic calls. Also other general enhancements/cleanup to integration testing framework as follows:
* Eliminate checking for `ExtrinsicFailed` event is `undefined`; instead, simply expect call not to throw (or just ignore--a thrown exception will fail the current test)
* Change test callbacks to normal functions rather than arrow functions (per [Moca documentation](https://mochajs.org/#arrow-functions))
* To avoid test suites stepping on each other's MSAs, instead of using the well-known development accounts in Substrate, create unique accounts/keys.
    * Note, this requires funding the new accounts--currently done by transferring tokens from the well-know dev accounts. Created a helper method to encapsulate this.
    * To assist with tracking appropriate funding, added helper functions and a `after()` hook to display the total cost of all chain operations for each unique account. This can assist if the initial funding amount is insufficient as tests are added.\
* Added some additional non-happy-path tests

Closes #798 

# Discussion

# Checklist
- [ ] Chain spec updated
- [ ] Design doc(s) updated
- [x] Tests added
- [ ] Benchmarks added
- [ ] Weights updated


integration-tests/msa/createMsa.test.ts
```diff
@@ -1,69 +1,70 @@
...
+import { createAccount, createAndFundAccount, DevAccounts, INITIAL_FUNDING, showTotalCost, signPayloadSr25519 } from "../scaffolding/helpers";
 import { KeyringPair } from "@polkadot/keyring/types";
-import { addPublicKeyToMsa, createMsa, createSchema, deletePublicKey } from "../scaffolding/extrinsicHelpers";
+import { addPublicKeyToMsa, createMsa, createSchema, deletePublicKey, signAndSend } from "../scaffolding/extrinsicHelpers";
 import { AVRO_GRAPH_CHANGE } from "../schemas/fixtures/avroGraphChangeSchemaType";
-import { filter, firstValueFrom } from "rxjs";
 
-describe("Create Accounts", () => {
+describe("Create Accounts", function () {
+    this.timeout(150000);
```
rlaferla, line 11: Is this 150 seconds?  Is that too long?  Is there a default timeout value constant for all tests (that can be overridden if needed)?

integration-tests/msa/createMsa.test.ts
```diff
@@ -1,69 +1,70 @@
...
         assert.notEqual(chainEvents["system.ExtrinsicSuccess"], undefined);
         assert.notEqual(chainEvents["msa.MsaCreated"], undefined);
         assert.notEqual(chainEvents["transactionPayment.TransactionFeePaid"], undefined);
-    }).timeout(15000)
+    });
 
-    it("should successfully mimic a user's path using tokens", async () => {
-        keys = createKeys("//Charlie")
+    it("should successfully mimic a user's path using tokens", async function () {
+        const { newAccount: keys } = await createAndFundAccount({ api, amount, source, context });
```
rlaferla, line 41: Is there a way to simplify this to what a developer would most often use?  i.e.  just the `amount`. Also, it may be better to give the accounts a real name (like _henryAccount_ or _johnAccount_) to make it easier to read.

integration-tests/messages/addIPFSMessage.test.ts
```diff
@@ -1,52 +1,75 @@
...
+    // % ipfs add frequency_test
+    // added QmYzm8KGxRHr7nGn5g5Z9Zv9r8nN5WNn7Ajya6x7RxmAB1 frequency_test
+    // % wc -c frequency_test
+    // 29 frequency_test
+    const ipfs_cid = "QmYzm8KGxRHr7nGn5g5Z9Zv9r8nN5WNn7Ajya6x7RxmAB1";
+    const ipfs_payload_data = "This is a test of Frequency.";
+    const ipfs_payload_len = ipfs_payload_data.length;
 
-    before(async () => {
+    before(async function () {
```
rlaferla, line 33: Good catch!  🎣

integration-tests/scaffolding/extrinsicHelpers.ts
```diff
@@ -24,81 +29,76 @@ type AddProviderPayload = { authorizedMsaId?: any; schemaIds?: any; expiration?:
  * up in the test.
  */
 
+
+/** Generic wrapper **/
+export async function signAndSend<T extends ISubmittableResult>(
```
rlaferla, line 34: 🏅 excellent!

integration-tests/scaffolding/helpers.ts
```diff
@@ -1,22 +1,143 @@
...
+import { pipe, map, tap, of } from "rxjs";
+import { createKeys } from "./apiConnection";
+import { getAccountInfo, transferFunds } from "./extrinsicHelpers";
 
-export type EventMap = {[key: string]: Event}
-export type Sr25519Signature = {Sr25519: `0x${string}`}
+export const INITIAL_FUNDING = 100000000n;
 
-export function groupEventsByKey() {
+export enum DevAccounts {
```
rlaferla, line 14: This level of abstraction is probably not needed and makes the code harder to read.

integration-tests/scenarios/grantDelegation.test.ts
```diff
@@ -1,29 +1,29 @@
...
+    const context = this.title;
+    const source = DevAccounts.Alice;
+    const amount = INITIAL_FUNDING;
 
-describe("Delegation Scenario Tests", () => {
     let api: ApiRx;
     let keys: KeyringPair;
+    let schemaId: Codec;
 
     before(async () => {
```
rlaferla, line 21: Does this need a `function` keyword?

integration-tests/scenarios/grantDelegation.test.ts
```diff
@@ -1,29 +1,29 @@
...
         let connectApi = await connect(process.env.WS_PROVIDER_URL);
-        keys = createKeys("//Alice")
         api = connectApi
-    })
-
-    after(() => {
-        api.disconnect()
-    })
-
-    it("should grant a delegation to a provider", async () => {
```
rlaferla, line 23: Does this need a `function` keyword?

rlaferla (changes requested):

integration-tests/msa/createMsa.test.ts
JoeCap08055: Oops... that should be `15000`, not `150000`. Fixing.

integration-tests/scaffolding/helpers.ts
JoeCap08055: I disagree. Some form of constant or enum, I feel, is always better than the proliferation of string literals in code. Consider the following code:
```
createKeys("//Al1ce");
```

This would compile fine, but likely result in a runtime error because the typo in the account name would fail to map to the well-known Substrate dev account `//Alice`.

Now consider the following code:
```
createKeys(DevAccounts.Al1ce);
```

This would fail to compile, alerting the developer of the typo.

Also, the `DevAccounts` enum could be useful in other use cases, say where we want to iterate over dev accounts or something.

Anyway, I don't see a down side to the enum... and since the enum maps exactly to the string values lexically, I don't think it makes the code harder to read at all.

My $0.02.

integration-tests/scenarios/grantDelegation.test.ts
JoeCap08055: Good catch.

integration-tests/scenarios/grantDelegation.test.ts
JoeCap08055: Does what? That's the diff from the old version, not the new code.

integration-tests/msa/createMsa.test.ts
JoeCap08055: Not clear on what you mean here... give the accounts a name how? Or you just mean the variable name?
And for "developer would most often use"? You mean for the arguments to `createAndFundAccount`? I suppose I could create an object containing the standard options that could simply be passed in, is that what you're after?

rlaferla: The variable names.

rlaferla: Reduce the number of inputs to that function.

JoeCap08055: Can't really eliminate any of the inputs to that function, but I defined an interface to encapsulate it & set up a default instance of it at test suite initialization to simplify things. See what you think.

rlaferla (approved):

integration-tests/messages/addIPFSMessage.test.ts
wilwade: Do arrow functions not work here?

integration-tests/messages/addIPFSMessage.test.ts
```diff
@@ -1,52 +1,73 @@
...
+    let schemaId: Codec;
+
+    // This is how the IPFS data was created:
+    //
+    // echo "This is a test of Frequency." > frequency_test
+    // % ipfs add frequency_test
+    // added QmYzm8KGxRHr7nGn5g5Z9Zv9r8nN5WNn7Ajya6x7RxmAB1 frequency_test
+    // % wc -c frequency_test
+    // 29 frequency_test
+    const ipfs_cid = "QmYzm8KGxRHr7nGn5g5Z9Zv9r8nN5WNn7Ajya6x7RxmAB1";
```
wilwade, line 27: @JoeCap08055 If you work on #786 it will fail this test. Not an issue for this test (as it is correct for the test), but heads up for that issue.

integration-tests/msa/createMsa.test.ts
```diff
@@ -1,69 +1,69 @@
...
-        assert.equal(chainEvents["system.ExtrinsicFailed"], undefined);
         assert.notEqual(chainEvents["system.ExtrinsicSuccess"], undefined);
         assert.notEqual(chainEvents["msa.MsaCreated"], undefined);
         assert.notEqual(chainEvents["transactionPayment.TransactionFeePaid"], undefined);
-    }).timeout(15000)
+    });
 
-    it("should successfully mimic a user's path using tokens", async () => {
-        keys = createKeys("//Charlie")
+    it("should successfully mimic a user's path using tokens", async function () {
```
enddynayn, line 39: Curious. Does this only work if we switch it from async () => to async function()?

integration-tests/scaffolding/helpers.ts
```diff
@@ -1,22 +1,166 @@
...
+                        break;
+                    }
+
+                    default: {
+                        err = new EventError(result.dispatchError);
+                        break;
+                    }
+                }
+
+                console.log(err.toString());
```
wilwade, line 101: Do we want this console.log?

integration-tests/scaffolding/helpers.ts
```diff
@@ -1,22 +1,166 @@
...
+                throw err;
+            }
+        }),
+        map((result: ISubmittableResult) => result.events.reduce((acc, { event }) => { acc[eventKey(event)] = event.data; return acc }, {})),
+        // tap((events) => console.log(events)),
```
wilwade, line 106: Might be worth a comment that un-commenting this line will output all events, but fairly clear. Does make me wonder if we want something like a debug flag on these tests...

integration-tests/scaffolding/helpers.ts
```diff
@@ -1,22 +1,166 @@
...
+
+interface BeginEndBalances {
+    beginBalance: bigint;
+    endBalance?: bigint;
+}
+let balancesMap = new Map<string, Map<string, BeginEndBalances>>;
+let doTxAccounting = false;
+export let txAccountingHook = async (api, context?): Promise<void> => { };
+
+if (process.env.ENABLE_TX_ACCOUNTING === 'true' || process.env.ENABLE_TX_ACCOUNTING === '1') {
```
wilwade, line 78: We should add this to the integration-tests README... which doesn't exist currently

Also good example perhaps for an event logging and other logging things.

integration-tests/scaffolding/extrinsicHelpers.ts
```diff
@@ -24,81 +29,76 @@ type AddProviderPayload = { authorizedMsaId?: any; schemaIds?: any; expiration?:
...
+    f: () => SubmittableExtrinsic<"rxjs", T>,
+    keys: KeyringPair,
+    customErrorHandler: (err: RpcError, caught: Observable<T>) => Observable<T> = (err, caught) => caught): Promise<EventMap> {
+    return firstValueFrom(f().signAndSend(keys).pipe(
+        filter(({ status }) => status.isInBlock || status.isFinalized),
+        catchError((err: RpcError, caught) => {
+            console.log(`code: ${err.code}, name: ${err.name}, message: ${err.message}`);
```
wilwade, line 41: This log feels wrong as we get it even when we are expecting it. Perhaps something else we might want on a flag?

integration-tests/scaffolding/helpers.ts
```diff
@@ -1,22 +1,166 @@
...
 }
 
 export function signPayloadSr25519(keys: KeyringPair, data: Codec): Sr25519Signature {
-    return {Sr25519: u8aToHex(keys.sign(u8aWrapBytes(data.toU8a())))}
-}
\ No newline at end of file
+    return { Sr25519: u8aToHex(keys.sign(u8aWrapBytes(data.toU8a()))) }
+}
+
+export function createAccount(name: string = 'first pair') {
```
enddynayn, line 118: Maybe we can call this`createKeypair` so as not to be confused with creating an MSA account or a token account.
No strong opinion.

wilwade (approved):
Overall great and nothing I see that would be blocking a merge.

Few thoughts for this or another PR:

1. We should add some wrappers around our console logs so they don't pollute the output.
2. We should document things like the TX_Accounting flag in a (needs to exist) README file for the integration tests.

integration-tests/msa/createMsa.test.ts
wilwade: I had to look this up as well. It's based on https://mochajs.org/#arrow-functions

That said @JoeCap08055 we might want to add a linter to enforce and autofix it.

integration-tests/scaffolding/helpers.ts
JoeCap08055: I like it. After discussing some more with @rlaferla , I'd like to get this in as an initial overhaul & create a followup ticket for some more enhancement/standardization to the integration test framework. I intend to clean up the names some more there.

integration-tests/scaffolding/helpers.ts
```diff
@@ -1,22 +1,166 @@
...
+    }
+    console.log(`${titleStr}Operation costs were: `);
+    for (const [key, value] of map) {
+        const acct = await getAccountInfo(api, key);
+        value.endBalance = acct.data.free.toBigInt();
+        const cost = value.beginBalance - value.endBalance;
+        cum += cost;
+        console.log(`${titleStr}${key}: ${cost}`);
+    }
+    console.log(`${titleStr}Total cost of all operations was (excluding standard dev accounts): ${cum}`);
```
wilwade, line 165: Nicer numbers:

```suggestion
        console.log(`${titleStr}${key}: ${cost.toLocaleString()}`);
    }
    console.log(`${titleStr}Total cost of all operations was (excluding standard dev accounts): ${cum.toLocaleString()}`);
```

> Operation costs were:
5FU2Mxp6AqU5BcCizsD7w4ENqzECfYxd8PVKd1WAHkXEvm1s: 10,913,296
5DkQGNFiU6jzc2MTwsDRWJn7j1QzSvgPtfZXRmcY7p3UPG8t: 1,486,738
5DFieHsHhn5kgXaHfXYCagjDzLVuicCTMPmrSaYScj1Sn8Ay: 9,407,504
5HKRtZEYhrsDVkVKxunRqErjDaFTdfAtxVzSMgfyYGdjXRVY: 7,080,948

integration-tests/scaffolding/helpers.ts
```diff
@@ -1,22 +1,166 @@
...
+    const keypair = createAccount(name);
+
+    // Get keypair for pre-funded dev account
+    const devKeys = createKeys(source);
+
+    // Transfer funds from pre-funded dev account to new account
+    await transferFunds(api, devKeys, keypair, amount);
+
+    if (doTxAccounting) {
+        let map = balancesMap.get(context || 'undefined');
```
enddynayn, line 138: Code golf:

let map = balancesMap.get(context) ||  new Map<string, BeginEndBalances>();

integration-tests/messages/addIPFSMessage.test.ts
JoeCap08055: Cool. Will make a good test case.

integration-tests/msa/createMsa.test.ts
JoeCap08055: I'll look into adding a linter for the follow-on issue (see my comment below)

integration-tests/scaffolding/helpers.ts
```diff
@@ -1,22 +1,166 @@
...
+        if (!map) {
+            map = new Map<string, BeginEndBalances>();
+        }
+
+        map.set(keypair.address, { beginBalance: amount });
+        balancesMap.set(context || 'undefined', map);
```
enddynayn, line 144: Are we setting the string 'undefined' to a map key? Is this to get a default?

integration-tests/scaffolding/helpers.ts
JoeCap08055: Couldn't decide.
More cleanup & README to come in follow-on ticket. Or I can push & get re-review/re-approval. I've already got quite a bit queued up for followup, though, and I'd like to get this in because it enables the possibility of someone starting to integrate this into the CI pipeline.

integration-tests/scaffolding/helpers.ts
```diff
@@ -1,22 +1,166 @@
...
+    }
+
+    return { newAccount: keypair, devAccount: devKeys };
+}
+
+export async function showTotalCost(api: ApiRx, context?: string) {
```
enddynayn, line 150: how do you feel about renaming showTotalCost to logTotalCost or printTotalCost?

integration-tests/scaffolding/helpers.ts
```diff
@@ -1,22 +1,166 @@
...
+    name: string = '';
+    message: string = '';
+    stack?: string = '';
+    section?: string = '';
+    rawError: DispatchError;
+
+    constructor(source: DispatchError) {
+        super();
+
+        if (source) {
```
enddynayn, line 47: Can we remove this condition since the source param is not an optional parameter?

integration-tests/scaffolding/helpers.ts
JoeCap08055: When formatting, should I scale down 10^-8 to show actual fractionalized token values? Or simply format with commas?

integration-tests/scaffolding/helpers.ts
JoeCap08055: Yes. In cases where a context was not set, we'll display it as 'undefined'.

integration-tests/scaffolding/helpers.ts
JoeCap08055: Works for me.

enddynayn (approved):
Awesome! ⛹️‍♂️ 🎉

integration-tests/scaffolding/helpers.ts
JoeCap08055: Sure.

enddynayn:
Maybe for another story, but it seems we are logging a lot. Is this something we can use for logging? https://github.com/winstonjs/winston

shannonwells (approved):
The changes I suggested aren't blocking, more "nice to have"

PR #837, opened 2022-12-21
chore(capacity): add storage

enddynayn:
# Goal
Add storage for staking and capacity:

`StakingAccountLedger`
Storage for keeping a ledger for staking

`StakingTargetLedger`
Storage to record how many tokens were targeted to an MSA.

`CapacityOf`
Storage for target Capacity usage.

These storage items are based on the [Capacity design docs](https://github.com/LibertyDSNP/frequency/blob/main/designdocs/capacity.md).

Adding storage makes it so staking, unstaking, and withdraw_unstake features can be worked on in parallel.

### DIFF
--- a/pallets/capacity/src/lib.rs
+++ b/pallets/capacity/src/lib.rs
@@ -51,7 +51,10 @@ use frame_support::{
 
 pub use common_primitives::{msa::MessageSourceId, utils::wrap_binary_data};
 pub use pallet::*;
+pub use types::*;
 pub use weights::*;
+
+pub mod types;
 
 #[cfg(feature = "runtime-benchmarks")]
 mod benchmarking;
@@ -69,7 +72,7 @@ type BalanceOf<T> =
 #[frame_support::pallet]
 pub mod pallet {
 	use super::*;
-	use frame_support::pallet_prelude::*;
+	use frame_support::{pallet_prelude::*, Twox64Concat};
 	use frame_system::pallet_prelude::*;
 
 	#[pallet::config]
@@ -82,7 +85,38 @@ pub mod pallet {
 
 		/// Function that allows a balance to be locked.
 		type Currency: LockableCurrency<Self::AccountId, Moment = Self::BlockNumber>;
+
+		/// The minimum required token amount to stake. It facilitates cleaning dust when unstaking.
+		#[pallet::constant]
+		type MinimumStakingAmount: Get<BalanceOf<Self>>;
+
+		/// The maximum number of unlocking chunks a StakingAccountLedger can have. It determines how many concurrent unstaked chunks may exist.
+		#[pallet::constant]
+		type MaxUnlockingChunks: Get<u32>;
 	}
+
+	/// Storage for keeping a ledger of staked token amounts for accounts.
+	#[pallet::storage]
+	#[pallet::getter(fn get_staking_ledger)]
+	pub type StakingAccountLedger<T: Config> =
+		StorageMap<_, Twox64Concat, T::AccountId, StakingAccountDetails<T>>;
+
+	/// Storage to record how many tokens were targeted to an MSA.
+	#[pallet::storage]
+	#[pallet::getter(fn get_target_ledger)]
+	pub type StakingTargetLedger<T: Config> = StorageDoubleMap<
+		_,
+		Twox64Concat,
+		T::AccountId,
+		Twox64Concat,
+		MessageSourceId,
+		StakingTargetDetails<BalanceOf<T>>,
+	>;
+
+	/// Storage for target Capacity usage.
+	#[pallet::storage]
+	pub type CapacityOf<T: Config> =
+		StorageMap<_, Twox64Concat, MessageSourceId, CapacityDetails<BalanceOf<T>, T::BlockNumber>>;
 
 	// Simple declaration of the `Pallet` type. It is placeholder we use to implement traits and
 	// method.
--- a/pallets/capacity/src/types.rs
+++ b/pallets/capacity/src/types.rs
@@ -1,0 +1,53 @@
+//! Types for the Capacity Pallet
+use super::*;
+use codec::{Decode, Encode, MaxEncodedLen};
+use frame_support::{BoundedVec, RuntimeDebug, RuntimeDebugNoBound};
+use scale_info::TypeInfo;
+use sp_runtime::traits::Zero;
+
+/// Details about the total token amount targeted to an MSA.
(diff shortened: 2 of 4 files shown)


pallets/capacity/src/lib.rs
```diff
@@ -82,8 +85,39 @@ pub mod pallet {
...
+		/// The minimum required token amount to stake. It facilitates cleaning dust when unstaking.
+		#[pallet::constant]
+		type MinimumStakingAmount: Get<BalanceOf<Self>>;
+
+		/// The maximum number of `unlocking` chunks a `StakingAccountLedger` can have. It determines how many unique Epochs a Staker may be unstaked in.
+		#[pallet::constant]
+		type MaxUnlockingChunks: Get<u32>;
 	}
 
+	/// Storage for keeping a ldeger of staked token amounts for accounts.
```
johnmela, line 98:
```suggestion
	/// Storage for keeping a ledger of staked token amounts for accounts.
```

pallets/capacity/src/lib.rs
```diff
@@ -82,8 +85,39 @@ pub mod pallet {
 
 		/// Function that allows a balance to be locked.
 		type Currency: LockableCurrency<Self::AccountId, Moment = Self::BlockNumber>;
+
+		/// The minimum required token amount to stake. It facilitates cleaning dust when unstaking.
+		#[pallet::constant]
+		type MinimumStakingAmount: Get<BalanceOf<Self>>;
+
+		/// The maximum number of `unlocking` chunks a `StakingAccountLedger` can have. It determines how many unique Epochs a Staker may be unstaked in.
```
johnmela, line 93: Should this state something like:
/// The maximum number of `unlocking` chunks a `StakingAccountLedger` can have. It determines how many concurrent unstaked chunks may exist.

pallets/capacity/src/types.rs
```diff
@@ -0,0 +1,57 @@
...
+}
+
+/// The type for storing Registered Provider Capacity balance:
+#[derive(PartialEq, Eq, Clone, Default, Encode, Decode, RuntimeDebug, TypeInfo, MaxEncodedLen)]
+pub struct CapacityDetails<Balance> {
+	/// The Capacity remaining for the `last_replenished_epoch`.
+	pub remaining: Balance,
+	/// The total Capacity issued to an MSA.
+	pub total_available: Balance,
+	/// TODO: update this
```
enddynayn, line 54: rm

pallets/capacity/src/types.rs
```diff
@@ -0,0 +1,57 @@
...
+use scale_info::TypeInfo;
+use sp_runtime::traits::Zero;
+
+/// Details about the total token amount targeted to an MSA.
+/// The Capacity that the target will receive.
+#[derive(PartialEq, Eq, Default, Clone, Encode, Decode, RuntimeDebug, TypeInfo, MaxEncodedLen)]
+pub struct StakingTargetDetails<Balance> {
+	/// The total amount of tokens that have been targeted to the MSA.
+	pub amount: Balance,
+	/// TODO: This can change to something else.
```
enddynayn, line 16: rm

pallets/capacity/src/types.rs
```diff
@@ -0,0 +1,57 @@
...
+/// The type for storing Registered Provider Capacity balance:
+#[derive(PartialEq, Eq, Clone, Default, Encode, Decode, RuntimeDebug, TypeInfo, MaxEncodedLen)]
+pub struct CapacityDetails<Balance> {
+	/// The Capacity remaining for the `last_replenished_epoch`.
+	pub remaining: Balance,
+	/// The total Capacity issued to an MSA.
+	pub total_available: Balance,
+	/// TODO: update this
+	/// The last Epoch that an MSA was replenished.
+	pub last_replenished_epoch: u32,
```
johnmela, line 56: The epoch be a type on config? Or at least a hardwired type that we can ensure consistency on everywhere?

johnmela (changes requested):

shannonwells (approved):
No need to block merging but the type for last_replenished_epoch type should be changed.

pallets/capacity/src/types.rs
enddynayn: I think you are right. We can use a block number for now.

runtime/frequency/src/lib.rs
```diff
@@ -290,6 +290,8 @@ impl pallet_capacity::Config for Runtime {
 	type RuntimeEvent = RuntimeEvent;
 	type WeightInfo = pallet_capacity::weights::SubstrateWeight<Runtime>;
 	type Currency = Balances;
+	type MinimumStakingAmount = ConstU128<1>;
```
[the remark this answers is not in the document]
enddynayn: Yes. These are placeholders. I will add a ticket so that we update these.

johnmela (approved):

pallets/capacity/src/types.rs
```diff
@@ -0,0 +1,53 @@
...
+use sp_runtime::traits::Zero;
+
+/// Details about the total token amount targeted to an MSA.
+/// The Capacity that the target will receive.
+#[derive(PartialEq, Eq, Default, Clone, Encode, Decode, RuntimeDebug, TypeInfo, MaxEncodedLen)]
+pub struct StakingTargetDetails<Balance> {
+	/// The total amount of tokens that have been targeted to the MSA.
+	pub amount: Balance,
+	/// The total Capacity that an MSA received.
+	pub capacity: Balance,
```
aramikm, line 15: Should capacity has it's own type to not being confused with Token amount?

pallets/capacity/src/lib.rs
```diff
@@ -82,8 +85,39 @@ pub mod pallet {
...
+	/// Storage to record how many tokens were targeted to an MSA.
+	#[pallet::storage]
+	#[pallet::getter(fn get_target_ledger)]
+	pub type StakingTargetLedger<T: Config> = StorageDoubleMap<
+		_,
+		Twox64Concat,
+		T::AccountId,
+		Twox64Concat,
+		MessageSourceId,
+		StakingTargetDetails<BalanceOf<T>>,
```
aramikm, line 113: what happens if there are multiple staking happens from same account to the same MSA? 
My assumption is that probably they add up. 

One thing to consider is that what if the formula that we are using to convert token to capacity changes over time? If that happens then we are going to loose the context of which formula (version) was used for each staking conversion and is it important to track it?

PR #841, opened 2022-12-22
treat version with iteration suffix as full release

### LINKED ISSUE #829: Fix latest tag not being assigned to subsequent releases
Currently, the logic in the release workflow doesn't treat `vN.N.N-N` as full release, so the `latest` tag doesn't get moved from `v09.29` to `v0.9.29-1` for example. Need to adjust regex here:

https://github.com/LibertyDSNP/frequency/blob/1f5b9e7424d2874442cd243677cbf34f5b5b322b/.github/workflows/release.yml#L316 and retest the release process.

demisx:
# Goal
The goal of this PR is recognize subsequent release with numeric suffix as a full release, i.e. these both should trigger update of the `latest` tag:

- v0.9.29
- v0.9.29-2

Closes #829

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -313,7 +313,7 @@ jobs:
           fetch-depth: 0
       - name: Full Release?
         run: |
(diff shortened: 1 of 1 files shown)


aramikm (approved):
Looks good!

rlaferla (approved):
Bravo 🇮🇹

Issue #839, opened 2022-12-21
Unified runtime causes Metadata Portal to create QR bar code with same name


rlaferla:
The WASM files for both `frequency` and `frequency-rococo` both use the name `frequency` in the metadata.  As a result, the Metadata QR code generation process is creating a `frequency` bar code for the `frequency-rococo` chain.

Consider this workflow run:
https://github.com/LibertyDSNP/metadata-portal/actions/runs/3753278560/jobs/6376380563

[2022-12-21T22:30:22Z INFO  metadata_cli::autosigner] :mag: **Checking for updates for frequency-rococo**
[2022-12-21T22:30:22Z INFO  metadata_cli::updater::wasm] FOUND RUNTIME INFO!
[2022-12-21T22:30:22Z DEBUG metadata_cli::updater::github] wasm.chain=frequency-rococo chain=frequency-rococo
[2022-12-21T22:30:22Z INFO  metadata_cli::autosigner] :date: Found version 3
[2022-12-21T22:30:22Z INFO  metadata_cli::updater::wasm] :arrow_down:  Downloading frequency-rococo runtime... 
:gear: Generating /home/runner/work/metadata-portal/metadata-portal/public/qr/**frequency_metadata_3.apng**...

rlaferla:
We could modify the Metadata Portal autosigner to not use the chain name from the metadata but the one that was used for the version check.   This would fix the filename collision BUT it's unclear if this would result in invalid metadata bar codes.

rlaferla:
I modified the autosigner as described above BUT we will need to verify that the bar code for Rococo is valid to use.

PR #846, opened 2022-12-22
Set the spec name to be different for mainnet vs rococo

wilwade:
# Goal
The goal of this PR is to have a separate `spec_name` for mainnet vs other setups (specifically rococo).

Part of #839 

# Discussion
- Because of the const and macro layers, the only way to really make this work ended up being the set you see.

# Checklist
- Tested
- `make rococo-build`
- `subwasm info target/release/wbuild/frequency-runtime/frequency_runtime.compact.compressed.wasm`

## Before:

image

## After

image

### DIFF
--- a/runtime/frequency/src/lib.rs
+++ b/runtime/frequency/src/lib.rs
@@ -159,9 +159,27 @@ impl_opaque_keys! {
 	}
 }
 
+// The duplicate macros are annoying, but #[sp_version::runtime_version]
+// has fairly string limits on what can go in there.
+
+// Override the spec name when not mainnet to be frequency-rococo
+#[cfg(not(feature = "frequency"))]
+macro_rules! spec_name {
+	( $y:expr ) => {{
+		create_runtime_str!("frequency-rococo")
+	}};
+}
+
+#[cfg(feature = "frequency")]
+macro_rules! spec_name {
+	( $y:expr ) => {{
+		create_runtime_str!($y)
+	}};
+}
+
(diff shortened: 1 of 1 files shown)


runtime/frequency/src/lib.rs
```diff
@@ -159,9 +159,27 @@ impl_opaque_keys! {
 	}
 }
 
+// The duplicate macros are annoying, but #[sp_version::runtime_version]
+// has fairly string limits on what can go in there.
+
+// Override the spec name when not mainnet to be frequency-rococo
+#[cfg(not(feature = "frequency"))]
```
wilwade, line 166: Yes. Everything except `frequency` is `frequency-rococo`.

Simple, but works.

runtime/frequency/src/lib.rs
```diff
@@ -159,9 +159,27 @@ impl_opaque_keys! {
...
+macro_rules! spec_name {
+	( $y:expr ) => {{
+		create_runtime_str!($y)
+	}};
+}
+
 #[sp_version::runtime_version]
 pub const VERSION: RuntimeVersion = RuntimeVersion {
-	spec_name: create_runtime_str!("frequency"),
+	spec_name: spec_name!("frequency"),
```
wilwade, line 182: Has to be a macro and has to have one param based on `#[sp_version::runtime_version]`

rlaferla (approved):
Looks g👀d!

johnmela (approved):

PR #850, opened 2022-12-28
Simplify integration test connection setup and tests (#845)

### LINKED ISSUE #845: Simplify integration test connection setup and tests
Currently, each integration test creates a new connection and has to pass the connection and user's keys as parameters for every call.   

Eliminate having to code `before()` and `after()` for each test and having to pass the `api` and `keys` parameters on every call.  Each test should have a main `user` with a set of keys.

Current:
`        const createMsaEvents = await createMsa(api, keys);
`
New:
`        const createMsaEvents = await createMsa();
`

JoeCap08055:
# Goal
The goal of this PR is to simplify integration test & suite setup, eliminate tests interfering with each other, and increase test coverage.

Closes #845

# Discussion

# Checklist
- [ ] Chain spec updated
- [ ] Design doc(s) updated
- [x] Tests added
- [ ] Benchmarks added
- [ ] Weights updated


integration-tests/README-TESTING.md
```diff
@@ -0,0 +1,31 @@
+Notes on Integration Testing
+============================
+
```
enddynayn, line 3: 💯

integration-tests/messages/addIPFSMessage.test.ts
```diff
@@ -29,45 +22,48 @@ describe("Add Offchain Message", function () {
     const ipfs_payload_len = ipfs_payload_data.length;
 
     before(async function () {
-        let connectApi = await connect(process.env.WS_PROVIDER_URL);
```
enddynayn, line 32: nice!

integration-tests/messages/addIPFSMessage.test.ts
```diff
@@ -29,45 +22,48 @@ describe("Add Offchain Message", function () {
...
+        keys = await createAndFundKeypair();
 
         // Create a new MSA
-        await createMsa(api, keys);
+        const createMsa = ExtrinsicHelper.createMsa(keys);
+        await createMsa.fundAndSend();
 
         // Create a schema
-        const createSchemaEvents = await createSchema(api, keys, PARQUET_BROADCAST, "Parquet", "IPFS");
+        const createSchema = ExtrinsicHelper.createSchema(devAccounts[0].keys, PARQUET_BROADCAST, "Parquet", "IPFS");
```
dopatraman, line 32: Maybe we should have a method for accessing the dev account instead of using an arbitrary int.

integration-tests/scaffolding/extrinsicHelpers.ts
```diff
@@ -29,76 +30,114 @@ type AddProviderPayload = { authorizedMsaId?: any; schemaIds?: any; expiration?:
...
+
+    public async getEstimatedTxFee(): Promise<bigint> {
+        return firstValueFrom(this.extrinsic().paymentInfo(this.keys).pipe(
+            map((info) => info.partialFee.toBigInt())
+        ));
+    }
+
+    private async fundOperation(source?: KeyringPair): Promise<void> {
+        const amount = await this.getEstimatedTxFee();
+        await ExtrinsicHelper.transferFunds(source || devAccounts[0].keys, this.keys, amount).signAndSend();
```
dopatraman, line 64: It looks like most extrinsics are using the default case ("//Alice"), but I don't think it is guaranteed that this account will be funded each time. We should have a method that encapsulates the account returned here.

integration-tests/scaffolding/helpers.ts
```diff
@@ -125,42 +82,21 @@ export function createAccount(name: string = 'first pair') {
...
-            map = new Map<string, BeginEndBalances>();
-        }
+export async function createAndFundKeypair(amount = EXISTENTIAL_DEPOSIT, keyName?: string): Promise<KeyringPair> {
+    const keypair = createKeys(keyName);
 
-        map.set(keypair.address, { beginBalance: amount });
-        balancesMap.set(context || 'undefined', map);
-    }
+    // Transfer funds from source (usually pre-funded dev account) to new account
+    await fundKeypair(devAccounts[0].keys, keypair, amount);
```
dopatraman, line 93: Same as above.

integration-tests/msa/createMsa.test.ts
```diff
@@ -1,69 +1,244 @@
...
+    describe("addPublicKeyToMsa", function () {
+        let badKeys: KeyringPair;
+        let payload: AddKeyData = {};
+        let ownerSig: Sr25519Signature;
+        let newSig: Sr25519Signature;
+        let badSig: Sr25519Signature;
+        let addKeyData: Codec;
+
+        before(async function () {
+            authorizedKeys.push(await createAndFundKeypair());
```
dopatraman, line 51: It doesn't seem like we more than one of these keypairs. Why do we need this array?

integration-tests/msa/createMsa.test.ts
```diff
@@ -1,69 +1,244 @@
...
+            await assert.rejects(op.fundAndSend(), {
+                name: 'RpcError',
+                message: /Custom error: 5/,
+            });
+        });
+
+        it("should test for 'NoKeyExists' error");
+
+        it("should delete all other authorized keys", async function () {
+            for (const key of authorizedKeys) {
```
dopatraman, line 227: If we remove the `authorizedKeys` array, we don't need to loop here. Not sure it serves a purpose.

integration-tests/messages/addIPFSMessage.test.ts
JoeCap08055: We actually don't need dev account keys here, changed to use the keys we just created.

integration-tests/scaffolding/extrinsicHelpers.ts
JoeCap08055: This method allows the caller to specify a funding source; if one is not provided it falls back to the default '//Alice' account. From my understanding, this (as well as the other dev accounts) is guaranteed to be funded in a development chain. The only real reason to provide an alternate funding source is if we suspect the amount of tests in the suite would drain the '//Alice' account if all operations were funded from there; in that case we might revisit this. But since the idea is that the integration test pipeline will always be run against a new local Frequency chain, funding should not be an issue at this point.

integration-tests/scaffolding/helpers.ts
JoeCap08055: See reply above.

integration-tests/msa/createMsa.test.ts
JoeCap08055: Any test that (successfully) adds an authorized keypair pushes it to this array. This way, later on when we test that we cannot retire an MSA until all authorized keys have been deleted, we can simply go through the array and delete any keys that were authorized, thus allowing us to successfully retire the MSA.

integration-tests/msa/createMsa.test.ts
JoeCap08055: In earlier tests, we explicitly authorize at least 2 additional keys beyond the "creator" keypair. We need to delete those 2 additional keys before the next test will allow us to retire the MSA. Rather than keeping 2 specific variables for the keypairs added, I think an array is a better/more generic method.

integration-tests/msa/createMsa.test.ts
dopatraman: it doesn't look like we access any keys apart from the first. am i mistaken?

integration-tests/scaffolding/extrinsicHelpers.ts
dopatraman: Right, but there is no guarantee that `devAccounts[0]` is Alice. I'm proposing wrapping the account we use with a method. 

Granted, a method would not guarantee it either, but makes it would make it easier to decipher for the next developer.

integration-tests/msa/createMsa.test.ts
JoeCap08055: Look on line 176... we push a second keypair

JoeCap08055: Also any future test cases that add keypairs to that MSA can push to that array and the later tests will not need to be modified.

integration-tests/msa/createMsa.test.ts
dopatraman: If I'm not mistaken, it sounds like we've create the need for this loop by having an array in the first place. If we don't need that array, we won't need this loop either. Is that correct?

integration-tests/msa/createMsa.test.ts
dopatraman: > Look on line 176... we push a second keypair

I see that, but I don't see us accessing it anywhere. If we're not accessing it.... do we need it?

integration-tests/scaffolding/extrinsicHelpers.ts
JoeCap08055: It doesn't matter which `devAccounts` item we use. We don't care if it's Alice. It could be Bob, Charlie, etc. The point is we just want to use one of the default dev accounts for funding purposes. It doesn't matter which one.

integration-tests/msa/createMsa.test.ts
```diff
@@ -1,69 +1,244 @@
...
+        let newSig: Sr25519Signature;
+        let badSig: Sr25519Signature;
+        let addKeyData: Codec;
+
+        before(async function () {
+            authorizedKeys.push(await createAndFundKeypair());
+            badKeys = createKeys();
+            payload.msaId = msaId;
+            payload.newPublicKey = authorizedKeys[0].publicKey;
+            payload.expiration = (await ExtrinsicHelper.getLastBlock()).block.header.number.toNumber() + 30;
```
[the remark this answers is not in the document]
JoeCap08055: Ah, this was inherited from the existing test. Removed; not necessary. We do expirations more generically now for the tests.

integration-tests/msa/createMsa.test.ts
JoeCap08055: We somehow need to track all of the keys added to this MSA. Right now there are only 2. But future tests could add more, maybe. An array seemed like the easiest/most generic way to keep track of them. This way, no matter how many keys are added to the MSA, as long as they're added to the array as well, cleanup will take place & subsequent tests will pass.

JoeCap08055: Here's the flow:
1. Create keypair & MSA
2. Add a 2nd keypair to MSA, push to array [array size: 1]
3. Use 2nd keypair to add a 3rd keypair to MSA, push 3rd keypair to aray [array size: 2]
4. Attempt to retire MSA, rejected due to multiple keypairs authorized
5. Loop through array and delete additional keypairs from MSA
6. Now able to successfully retire MSA

saraswatpuneet: 👍

dopatraman (commented):
Good stuff... but I am still in the dark as to why we chose to replace the entire integration test system. I think a lot of these changes could have been made in a piece-meal fashion without completely rewriting the original work.

integration-tests/msa/createMsa.test.ts
JoeCap08055: See here: https://github.com/LibertyDSNP/frequency/pull/850#discussion_r1061842811

integration-tests/scaffolding/extrinsicHelpers.ts
rlaferla: I'd like to see the actual name of the account rather than an index into an array.   While it doesn't matter which account is used, it does matter that the developer knows which account it is.

rlaferla (commented):
L👀ks good.  See my comment on `devAccounts` vs. name.

saraswatpuneet (approved):
Since @dopatraman  lead this project, I would definitely wait for his approval
But limited I can understand , it looks good to me

integration-tests/scaffolding/extrinsicHelpers.ts
JoeCap08055: Sounds good. I'll include it in my next PR.

integration-tests/scenarios/grantDelegation.test.ts
```diff
@@ -1,30 +1,63 @@
...
-
-    let api: ApiRx;
     let keys: KeyringPair;
-    let schemaId: Codec;
+    let otherMsaKeys: KeyringPair;
+    let noMsaKeys: KeyringPair;
+    let providerKeys: KeyringPair;
+    let otherProviderKeys: KeyringPair;
+    let schemaId: u16;
+    let providerId: u64;
```
shannonwells, line 15: If these are MessageSourceIds, the naming should match `msaId` and `otherMsaId` because that's what they are, and what people will be searching on when looking at changes/debugging.

shannonwells (dismissed):
This needs to be broken up into two PRs, one for the refactor and one for the added test coverage.  My eyes are glazing over trying to review this between understanding new helpers and refactors versus reading through a bunch of new tests that use the new helpers etc.

The "tl;dr" effect may explain why it's taken so long to get approved.

integration-tests/scenarios/grantDelegation.test.ts
JoeCap08055: I don't see the concept of `provider ID` referenced in the spec anywhere. You must create an MSA first, then create a Provider. So presumably `providerId`, though the same format (`u64`) as an `msaId`, is not the same thing.

integration-tests/messages/addIPFSMessage.test.ts
```diff
@@ -29,45 +22,52 @@ describe("Add Offchain Message", function () {
...
+        const accountWithNoMsa = devAccounts[0].keys;
+        await assert.rejects(ExtrinsicHelper.addIPFSMessage(accountWithNoMsa, schemaId, ipfs_cid, ipfs_payload_len + 1).signAndSend(), {
             name: 'InvalidMessageSourceAccount',
             section: 'messages',
         });
     });
 
     it('should fail if schema does not exist', async function () {
-        await assert.rejects(addIPFSMessage(api, keys, 2, ipfs_cid, ipfs_payload_len + 1), {
+        // If we ever create more than 999 schemas in a test suite/single Frequency instance, this test will fail.
```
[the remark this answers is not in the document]
JoeCap08055: It's simply an arbitrary limit on the number of schemas that might have been created across all integration tests, in order to make sure that `schemaId` 999 does not map to an actual schema. Assuming that schema IDs increase from 0 monotonically.

JoeCap08055:
> https://github.com/LibertyDSNP/frequency/pull/850#pullrequestreview-1240930191

Seems counterproductive to split this now, especially since it's been approved. But I get where you're coming from.

integration-tests/scenarios/grantDelegation.test.ts
shannonwells: No, it's not referenced in the spec, because it's a [convenience wrapper for a MessageSourceId](https://github.com/LibertyDSNP/frequency/blob/main/common/primitives/src/msa.rs#L95), like DelegatorId.  Only by a technicality is it not an MSA Id, but your argument is fair enough, and it's just a nitpick anyway.

PR #854, opened 2022-12-29
dispatch auto-sign workflow

### LINKED ISSUE #792: Integrate Metadata Portal into release process
**TASK**

Update the trigger for the `autosigner` workflow to kick off when a new release is cut.  The scheduled run can be preserved as a backup trigger but probably should be modified to run once every 1-3 hrs.

**BACKGROUND**

The `autosigner` workflow currently runs on a schedule (at midnight UTC every night.)  When it runs it checks (1) the GitHub release assets and (2) the live nodes.   If there are no changes to the metadata, it does nothing.  If there are changes, it generates new signed QR bar codes and automatically creates a PR for them.  It’s a semi-automatic system which is automated but requires someone to merge the PR when the metadata changes.  This is an important safety check for both security and to prevent runaway automation.

demisx:
# Goal
The goal of this PR is trigger `auto-sign.yml` workflow upon successful posting new release assets.

Closes #792

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -386,6 +386,17 @@ jobs:
           tag: "latest"
           force_push_tag: true
           message: "Latest full release"
+      - name: Trigger Auto-signer Workflow
+        uses: actions/github-script@v6
+        with:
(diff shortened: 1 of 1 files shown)


wilwade (commented):
👍

Issue #847, opened 2022-12-22
Frequency Rococo and Mainnet do not have the correct Genesis


wilwade:
## Bisect results:
- Mainnet: #726, Commit: `8354a41603e9f3bab192e5f2bf0fcb9ea24199cf`
- Rococo: #682, Commit: `950691e2aa381155fe563f3325464b05c956f218` (Oddly enough, not the one ^)

## How To Reproduce

1. Checkout `v0.9.29`
2. `make build-rococo`
3. `./target/release/frequency --tmp`
4. See these log messages:
  ```
  2022-12-22 21:34:20 📋 Chain specification: Frequency Rococo
  ...
  2022-12-22 21:42:33 [Parachain] Bootnode with peer id `12D3KooWArmKDbY8Y6XXHGodosWAjRWWxSw5YxWEjSZTBNjJXVSC` is on a different chain (our genesis: 0x0134…0abb theirs: 0x0c33…8e72)
  ```

## Ideas and Thoughts
- How is this possible if the json spec files are being used? Unknown.
- Tested with release and production. That is not the issue (so release is easier).
- Frequency Rococo has had an update already and is on spec_version 2
- Frequency Rococo is on 0.9.29-rc3. Apparently when we built `resources/frequency-rococo.json` we did it wrong
- Check out the `-rc3` release or subwasm can likely be used to get the original wasm to update `resources/frequency-rococo.json` so that it has the same stuff. (Remember that wasms are only deterministic with `srtool`, but it is possible that the original wasm is not an srtool wasm).
- Did a git bisect and got this when testing mainnet:

```
8354a41603e9f3bab192e5f2bf0fcb9ea24199cf is the first bad commit
commit 8354a41603e9f3bab192e5f2bf0fcb9ea24199cf
Date:   Thu Dec 1 11:28:52 2022 -0600

    Chore/unify runtimes (#726)
```

saraswatpuneet:
And I believe since we unified runtimes we should have re-generated with genesis too, but not sure if that would mean we have to stick to it, however this should only be reflective in rococo, given mainnet was launched with unified runtimes

saraswatpuneet:
Or:
* Before unified runtime we were producing wasms without feature flags turned on
* As a result we saw rococo const such as LaunchPeriod etc had mainnet values
* _With the fix the wasm must have changed?  Probably yes_

saraswatpuneet:
[...]

PR #857, opened 2023-01-03
Test Genesis State Output

wilwade:
# Goal
The goal of this PR is to make sure that #847 cannot happen again.

# Discussion
- Added simple test for mainnet and rococo to always be on the same genesis block as those networks require.
- Also ends up being a good basic test for the binary

# Checklist
- [x] Tested: https://github.com/LibertyDSNP/frequency/actions/runs/3803259957
- [x] Updated CI Diagram
image

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -24,14 +24,17 @@ jobs:
         include:
           - network: local
             spec: frequency-rococo-local
+            genesis: "skip"
             build-profile: release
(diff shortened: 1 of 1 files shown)


shannonwells (approved):

wilwade:
@demisx 

1. Fixed

image


2. I think this is the first place we get the binaries to do the check. We don't have a mainnet or rococo binary on the verify PR side of things right?

PR #872, opened 2023-01-04
move genesis check to verify pr workflow

demisx:
# Goal
The goal of this PR is move Genesis check to the Verify PR workflow.

### DIFF
--- a/.github/workflows/verify-pr-commit.yml
+++ b/.github/workflows/verify-pr-commit.yml
@@ -7,7 +7,7 @@ on:
     branches:
       - main
 env:
-  BUILD_PROFILE: release
+  BIN_DIR: target/release
   RUST_TOOLCHAIN: nightly-2022-09-22 # Update this when updating the Rust toolchain
 jobs:
   changes:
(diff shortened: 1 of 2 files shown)


.github/workflows/verify-pr-commit.yml
```diff
@@ -119,20 +128,21 @@ jobs:
         run: |
           CARGO_INCREMENTAL=0 RUSTFLAGS="-D warnings" cargo build --locked --release \
             --features ${{matrix.spec}}
+      - name: Rename Reference Binary
+        if: steps.cache-binary.outputs.cache-hit != 'true'
+        working-directory: ${{env.BIN_DIR}}
+        run: cp ${{env.BUILT_BIN_FILENAME}} ${{env.FINAL_BIN_FILENAME}}
```
wilwade, line 134: Is it worth doing the rename? Couldn't we just use `env.BUILT_BIN_FILENAME` everywhere we currently use `env.FINAL_BIN_FILENAME`?

.github/workflows/verify-pr-commit.yml
```diff
@@ -430,13 +444,16 @@ jobs:
         with:
           name: artifacts-${{github.run_id}}
           path: ${{env.BIN_DIR}}
+      - name: Rename Binary
+        working-directory: ${{env.BIN_DIR}}
+        run: mv ${{env.BUILT_BIN_FILENAME}} ${{env.TEST_BIN_FILENAME}}
```
wilwade, line 449: Same here. Feels like we can just drop all the renaming.

.github/workflows/verify-pr-commit.yml
```diff
@@ -653,4 +677,50 @@ jobs:
...
+        run: |
+          echo "BIN_DIR=target/$BUILD_PROFILE" >> $GITHUB_ENV
+          echo "EXPECTED_GENESIS_STATE_ROCOCO=0x000000000000000000000000000000000000000000000000000000000000000000e3495742b019f5ad49dff7de4040bc965b75eaf46769c24db1027d4ff86fc92703170a2e7597b7b7e3d84c05391d139a62b157e78786d8c082f29dcf4c11131400" >> $GITHUB_ENV
+          echo "EXPECTED_GENESIS_STATE_MAINNET=0x000000000000000000000000000000000000000000000000000000000000000000393a2a0f7778716d006206c5a4787cbf2ea3b26a67379b7a38ee54519d7fd4be03170a2e7597b7b7e3d84c05391d139a62b157e78786d8c082f29dcf4c11131400" >> $GITHUB_ENV
+          echo "BIN_FILENAME_ROCOCO=frequency.rococo-pr" >> $GITHUB_ENV
+          echo "BIN_FILENAME_MAINNET=frequency.mainnet-pr" >> $GITHUB_ENV
+      - name: Check Out Repo
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
```
wilwade, line 697: Feels odd, but I do not believe we need this step. The binaries are all we need.

.github/workflows/verify-pr-commit.yml
```diff
@@ -653,4 +677,50 @@ jobs:
...
+            (echo "ERROR: The actual genesis state does not match the expected" && exit 10)
+      - name: Test Mainnet Genesis State
+        working-directory: ${{env.BIN_DIR}}
+        run: |
+          expected_genesis_state=${{env.EXPECTED_GENESIS_STATE_MAINNET}}
+          echo "Expected genesis state: $expected_genesis_state"
+          actual_genesis_state=$(./${{env.BIN_FILENAME_MAINNET}} export-genesis-state)
+          echo "Actual genesis state: $actual_genesis_state"
+          [[ $actual_genesis_state == $expected_genesis_state ]] || \
+            (echo "ERROR: The actual genesis state does not match the expected" && exit 10)
```
wilwade, line 726: Curious why `exit 10` instead of `exit 1`. GH Actions treats them the same afaik.

wilwade (commented):
Few suggestions. I do wonder if there are some ways to simplify some of the structures. Be nice if the matrix and env vars could be shared somehow, but likely that abstraction would just be more confusing.

.github/workflows/verify-pr-commit.yml
wilwade: Correct. Same in many of the other places.

wilwade (approved):
No other blocking changes from me

wilwade (approved):
👍

saraswatpuneet (approved):
🙇

PR #873, opened 2023-01-04
Update target subdirectory for upgrading

shannonwells:
# Goal
The goal of this PR is to fix a small bug in the init.sh script - it refers to a directory we don't generate any more.

# Discussion
Found when checking that upgrades work via script+sudo.

Checklist items not applicable.

### DIFF
--- a/scripts/init.sh
+++ b/scripts/init.sh
@@ -164,6 +164,7 @@ offboard-frequency)
 upgrade-frequency)
 
   root_dir=$(git rev-parse --show-toplevel)
+  echo "root_dir is set to $root_dir"
 
   cargo build \
     --locked \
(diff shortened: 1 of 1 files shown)


aramikm:
I remember fixing an issue like that before holidays https://github.com/LibertyDSNP/frequency/blame/main/scripts/init.sh#L172

shannonwells:
Thanks @aramikm, I had this change before the holidays and didn't see that you had updated it, however there's one more, so I changed the description.

scripts/init.sh
```diff
@@ -164,17 +164,18 @@ offboard-frequency)
 upgrade-frequency)
 
   root_dir=$(git rev-parse --show-toplevel)
+  echo "root_dir is set to $root_dir"
```
shannonwells, line 167: I'm not sure if we want to leave this or not but I found it helpful

saraswatpuneet (approved):

wilwade (approved):

PR #874, opened 2023-01-05
Update versioning strategy

### LINKED ISSUE #686: [CI] Block the release if version of frequency tomls do not match tag
To prevent version mismatch between a release and binary, release CI path should make sure the [version](https://github.com/LibertyDSNP/frequency/blob/2cc01c4f3da651e079168d9e97de60f12a18bbf4/runtime/frequency/Cargo.toml#L9) of package in runtime toml and others, matches the tag

`./frequency --version` === `git tag`

### LINKED ISSUE #823: Change versioning strategy
We have run into an issue with our versioning. Using polkadot base version `v0.9.29` and then subsequent with a `-X` ie `v0.9.29-1` is great. I like it still, but it causes an issue for any npm packages reusing that version. (0.9.29 > 0.9.29-X).

## Path Forward

1. Our version base will be our own. We will release v1.0.0 soon
2. We will move the polkadot version into the build information with the polkadot version: aka `+polkadot0.9.29`
3. npm versions should strip that portion of the version. So chain version `v1.0.1+polkadot0.9.29` should generate npm matching package version `v1.0.1` (if needed, it can use build information for other data).

## Todo
- [ ] Alter the logic for updating the version for chain
- [ ] Update the ci logic for testing matching versions
- [ ] Update the npm logic for generating a version (dropping the build information)
- [ ] Update the Release page of the wiki with the new setup
- [ ] Release new version
- [ ] Create a new story to consider implementing semantic commits

## Original Options

So we have a few options (from easiest to hardest more or less):

1. Never use the base version, always use a `-X`. Be it `-0` or `-1`, it would solve the issue for npm however would (as our current setup is) be against semver.
1. Stop using the polkadot version as our base. Likely this means our next release is `1.1.0` (to be able to back date `0.9.29` -> `1.0.0` and `0.9.29-2` -> `1.0.2`). Releases that are not hotfixes are likely always minor releases so patches can be reserved for hotfixes?
1. Same as 2 except we use build version information to contain the polkadot version. Eg: `1.0.0+0.9.29` or perhaps `1.1.0+polkadot0.9.30`
1. Use a different version for npm (such as 1), but don't change anything else.
1. Other (I do like https://calver.org options, but that doesn't quite match our situation like our spec and doc website deployments)

**Side Effects**:
- The new release asset naming must be compliant with https://github.com/LibertyDSNP/metadata-portal. Reach out to @johnmela and @rlaferla.

wilwade:
# Goal
The goal of this PR is to switch to a new versioning strategy (shifting the substrate/polkadot version into the build version).

Closes #686 
Closes #823 

# Discussion

Tested with v1.0.0-rc1
```
$ ./frequency.amd64 --version
frequency.amd64 1.0.0-rc1+polkadot0.9.30-9d0f4bc96f9
```
- CI check failure test: https://github.com/LibertyDSNP/frequency/actions/runs/3842190870/jobs/6543671877
- CI check success test: https://github.com/LibertyDSNP/frequency/actions/runs/3842423288

Release Notes Update Test:
image

# Checklist
- [x] CI Check of binary version
- [x] Wiki Update: https://github.com/LibertyDSNP/frequency/wiki/Release
- [ ] [Update Frequency CI Diagram](https://docs.google.com/drawings/d/1IcAzZbulRcntYQ5ub-zDyIkqJY1NE0iqClhkg2PE5lo/edit) (TODO)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -6,8 +6,7 @@ on:
   push:
     tags:
       - "v[0-9]+.[0-9]+.[0-9]+" # ex. v0.29.9
-      - "v[0-9]+.[0-9]+.[0-9]+-[0-9]+" # ex. v0.29.9-1
-      - "v[0-9]+.[0-9]+.[0-9]+-rc[0-9]+" # ex. v0.29.9-1-rc1
+      - "v[0-9]+.[0-9]+.[0-9]+-rc[0-9]+" # ex. v0.29.9-rc1
       - "vX.X.X" # used for testing only
 env:
   RUST_TOOLCHAIN: nightly-2022-09-22 # Update this when updating the Rust toolchain
@@ -322,7 +321,7 @@ jobs:
           fetch-depth: 0
       - name: Full Release?
         run: |
-          is_full_release=$([[ "$NEW_RELEASE_VERSION" =~ ^v[0-9]+\.[0-9]+\.[0-9]+(\-[0-9]+)?$ ]] && \
+          is_full_release=$([[ "$NEW_RELEASE_VERSION" =~ ^v[0-9]+\.[0-9]+\.[0-9]+$ ]] && \
             echo 'true' || echo 'false')
           echo "is_full_release: $is_full_release"
           echo "IS_FULL_RELEASE=$is_full_release" >> $GITHUB_ENV
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -83,13 +83,6 @@ jobs:
         run: |
           file ${{env.BIN_DIR}}/${{env.BUILT_BIN_FILENAME}} && \
             ${{env.BIN_DIR}}/${{env.BUILT_BIN_FILENAME}} --version
-      - name: Testing Version Match
-        run: |
-          EXPECTED_VERSION="${{env.NEW_RELEASE_VERSION}}+polkadot$(make version-polkadot)"
-          ACTUAL_VERSION="v$(${{env.BIN_DIR}}/${{env.BUILT_BIN_FILENAME}} --version | cut -d " " -f 2)"
-          echo "Expected: ${EXPECTED_VERSION}"
-          echo "  Actual: ${ACTUAL_VERSION%-*}"
-          [[ ${ACTUAL_VERSION%-*} == ${EXPECTED_VERSION} ]]
       - name: Rename Binary
         run: |
           cp -p ${{env.BIN_DIR}}/${{env.BUILT_BIN_FILENAME}} \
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -128,14 +128,17 @@ build-rococo-release:
...
+.PHONY: test
 test:
 	cargo test --workspace --locked --features all-frequency-features
 
 integration-test:
 	./scripts/run_integration_tests.sh
 
-PHONY: version
+# Pull the Polkadot version from the polkadot-cli package and remove the leading v
+POLKADOT_VERSION=$(shell cargo tree --depth 0 --manifest-path ./node/service/Cargo.toml --package polkadot-cli | grep polkadot-cli | cut -d " " -f 2 | tr -d "v")
```
wilwade, line 139: This is a bit strong for something that we could make lighter, but it works and is fairly consistent.

Makefile
```diff
@@ -145,6 +148,19 @@ ifneq (,$(findstring v,  $(v)))
...
 	cargo check
 	@echo "All done. Don't forget to double check that the automated replacement worked."
+
+.PHONY: version-polkadot
+version-polkadot:
+	@echo $(POLKADOT_VERSION)
+
+.PHONY: version-reset
+version-reset:
+	find ./ -type f -name 'Cargo.toml' -exec sed -i '' 's/^version = \".*+polkadot.*\"/version = \"0.0.0\"/g' {} \;
```
wilwade, line 166: Mostly just for me to be lazy, but might be helpful for others.

tools/ci/release-notes/build-release-notes.sh
```diff
@@ -1,6 +1,7 @@
 #!/bin/sh
 # set -ex
 
-changelog=$1
+polkadot_version=$1
+changelog=$2
```
wilwade, line 5: Wanted to keep the multi-line action last

.github/workflows/release.yml
```diff
@@ -6,8 +6,7 @@ on:
   push:
     tags:
       - "v[0-9]+.[0-9]+.[0-9]+" # ex. v0.29.9
-      - "v[0-9]+.[0-9]+.[0-9]+-[0-9]+" # ex. v0.29.9-1
```
wilwade, line 9: Removing -X as an option

.github/workflows/release.yml
```diff
@@ -303,8 +302,52 @@ jobs:
...
+          name: artifacts-${{github.run_id}}
+          path: .
+      - name: Move Binary to Target Dir
+        run: |
+          mkdir -p ${{env.BIN_DIR}}
+          mv ${{env.RELEASE_BIN_FILENAME}} ${{env.BIN_DIR}}/${{env.BUILT_BIN_FILENAME}}
+          chmod +x ${{env.BIN_DIR}}/${{env.BUILT_BIN_FILENAME}}
+      - name: Testing Version Match
+        run: |
+          EXPECTED_VERSION="${{env.NEW_RELEASE_VERSION}}+polkadot$(make version-polkadot)"
```
wilwade, line 343: @demisx `make version-polkadot` runs cargo. While this works, should I add the install rust toolchain step?

.github/workflows/release.yml
```diff
@@ -303,8 +302,52 @@ jobs:
...
+        os: [ubuntu-20.04]
+        network: [mainnet]
+        include:
+          - network: mainnet
+            spec: frequency
+            build-profile: production
+            release-file-name-prefix: frequency
+          - os: ubuntu-20.04
+            arch: amd64
+    runs-on: ${{matrix.os}}
```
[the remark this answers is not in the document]
wilwade: I wanted to match the other parts of the code. The `Set Env Vars` in particular would have to be manually set. Using the matrix means I could just duplicate the same setup. (So changes to the spec name or os or such would be easy to find)

.github/workflows/release.yml
wilwade: Switched to an awk. Much faster and doesn't require the step now: https://github.com/LibertyDSNP/frequency/actions/runs/3847691878/jobs/6555082081

wilwade: https://github.com/LibertyDSNP/frequency/pull/874/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R140

wilwade: Right, but the make script now uses `awk`.

PR #884, opened 2023-01-07
Bug Fix (#882) Use correct feature flags with srtool

### LINKED ISSUE #882: [Bug]  mainnet wasm has wrong version
Frequency: Frequency release mainnet wasm has wrong version "frequency-rococo" that should be "frequency"

wilwade:
# Goal
The goal of this PR is to make sure that the correct feature flags are being used with srtool for wasm generation

Closes #882 

# Discussion
- srtool default-features flag does not work as intended
- Instead using build-opts
- Enabled debug logging in srtool for easier checking
- Added subwasm check to test the output wasm

Test Runs:
- Mainnet: https://github.com/LibertyDSNP/frequency/actions/runs/3859784794/jobs/6579633360
    image

- Rococo: https://github.com/LibertyDSNP/frequency/actions/runs/3859784794/jobs/6579633325
     image

# Checklist
- [x] Tests added

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -124,6 +124,7 @@ jobs:
             built-wasm-file-name-prefix: frequency_runtime
             release-wasm-file-name-prefix: frequency-rococo_runtime
             features: frequency-rococo-testnet
+            wasm-core-version: frequency-rococo
           - network: mainnet
             build-profile: production
             package: frequency-runtime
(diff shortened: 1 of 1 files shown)


saraswatpuneet (approved):
🎉🎤 will figure out issue with default features

wilwade:
@demisx 

> I presume doing the "Test WASM this in "Verify PR" workflow is not possible? Just trying to see if there is a way to fail early, not during release.

This really is just a check on srtool not our code, so it is release oriented only. None of our code /should/ impact this.

Also keeping an eye on https://github.com/chevdor/srtool-cli/issues/7 to see if we can use something simpler when it is out.

PR #900, opened 2023-01-10
designdocs: added new state storage proposal

### LINKED ISSUE #898: [Private graph] Schema based generic storage

aramikm:
# Goal
The goal of this PR is provide a proposal for a generic state storage pallet using schemas to allow storage of private graph and keys in a DSNP compatible implementation

Closes #898

# Changes
- Added new design doc
- Renamed uppercased older design docs to lowercase to have consistent naming convention

# Checklist
- [x] Design doc(s) updated


designdocs/state_storage.md
```diff
@@ -0,0 +1,101 @@
...
+
+## Proposal
+Creating a double map of (`SchemaId` and `PageNumber`) as keys and `vec<u8>` as stored values.
+
+Using mentioned double-map will allow storing data using a variety of schemas using pagination.To
+be able to achieve goal number 2 we will need to create a child-tree for each `MSA` and store this
+double map under these child trees for each `MSA`.
+
+### Modes of Storage
+- **Batch**: In this mode we are storing each entity of that schemaId as an item inside an array.
```
wilwade, line 47: Interestingly, batch mode could potentially ignore the pagination and just index the items. However that assumes that we can know the number of items that a page can contain in advance.

designdocs/state_storage.md
```diff
@@ -0,0 +1,101 @@
...
+This is recommended for items that are relatively small and there are more than a few of them.
+- **Single**: In this mode we are storing each entity individually, and it is recommended for items
```
wilwade, line 49: Double checking that I understand (and clear to others). Single refers to a single page. We are not trying to ever have one schema that is combining pages. Is that a correct statement?

designdocs/state_storage.md
```diff
@@ -0,0 +1,101 @@
...
+
+## Non-goals
+- Defining an economic incentive to restrict infinite growth of on-chain database.
+
+## Glossary
+* `DSNP`: Decentralized Social Networking Protocol
+* `MSA`: Message Source Account. A registered identifier with the MSA pallet.
+
+## Proposal
+Creating a double map of (`SchemaId` and `PageNumber`) as keys and `vec<u8>` as stored values.
```
wilwade, line 40: In batch mode, how do we determine the page size given an unknown item size? Does it work to just set it at say 100 and leave it be?

designdocs/state_storage.md
```diff
@@ -0,0 +1,101 @@
...
+    - _Purpose_: Removes an existing item from array
+1. **Replace Item**
+    - _Modes_: Batch
+    - _input_: index of item to replace and `vec<u8>` of new data
+    - _Purpose_: Updates an item in the array
+1. **Upsert**
+    - _Modes_: Single
+    - _input_: Serialized data `vec<u8>`
+    - _Purpose_: Creates or updates an item
+1. **Remove**
```
wilwade, line 77: Do we need to worry about packing?

Say I have 1-10 pages full. If I remove page 3 and gap is that an issue?

designdocs/state_storage.md
```diff
@@ -0,0 +1,101 @@
...
+done with only 1 DB_WRITE.
+- _Composability_: Batch mode allows better composability due to on-chain read requirement and smaller
+size of each item.
+
+### Actions
+1. **Append Item**
+    - _Modes_: Batch
+    - _input_: Serialized data `vec<u8>`
+    - _Purpose_: Adds a new item to existing array
+1. **Remove Item**
```
wilwade, line 65: Do we need to worry about gaps?

If I have pages (page size=100) 1-10 full and remove item from page 2, index 47, how does the next append work? does it fill the hole? Do we need to defrag from time to time?

wilwade (commented):
A few questions and thoughts, but mostly just dialing in the details.

designdocs/state_storage.md
```diff
@@ -0,0 +1,101 @@
...
+![image](https://user-images.githubusercontent.com/9152501/211664660-1d4aa9ac-ed11-44eb-a8f4-66c972750ac3.jpg)
+
+The rationale behind separating these modes is based on two reasons:
+- _Performance_: Batch modes always need 1 on-chain DB_READ and 1 DB_WRITE but Single mode might be
+done with only 1 DB_WRITE.
+- _Composability_: Batch mode allows better composability due to on-chain read requirement and smaller
+size of each item.
+
+### Actions
+1. **Append Item**
```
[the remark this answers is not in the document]
aramikm: I like it due to reduced number of read and writes but we have to think about what happens if one of them are invalid ? - Should we fail the whole operation
- or just apply as many as possible

 I prefer first one but if the providers and clients are not malicious this should happen rarely.

designdocs/state_storage.md
aramikm: That is correct, each item is in it's own page.

designdocs/state_storage.md
aramikm: We will have a constant of MAX_PAGE_SIZE and try to fit as many as items that we can. As long as Items are relatively smaller than a page which they should be it should work.

designdocs/state_storage.md
aramikm: I thought about this and I think one thing might be able do is to replace last page with removed page so pages stay continues.

designdocs/state_storage.md
aramikm: I think since we are doing all this page calculation and management at client side. We can find about a page which is not full and try to pack it, so if our clients behave properly we should have packed pages as much as possible but we might need to have a defrag process once in a while in case some client misbehave.

designdocs/state_storage.md
saraswatpuneet: @aramikm  do you think what we learnt from queue pallet about pov dispatch issue could possibly apply here too?

aramikm: It definitely would.

designdocs/state_storage.md
aramikm: To extend this approach we can do it in 2 different ways if there are more than 1 option(page) to append to

1- Deterministic: Always choosing the page with smallest (or maybe largest) index
  - _tradeoffs_: since the result would always be deterministic across different clients. Instead of having possibility of  duplicate entities inside different pages coming from different clients we will have collisions and one of them will fail but It would cause more collisions.

2-Random: Randomly choose the page from valid options
  - _tradeoffs_: Less collision since the probability of two clients writing to the same page would be lower but more probability of having the same identical item being duplicated in more than one page.

designdocs/state_storage.md
```diff
@@ -0,0 +1,116 @@
...
+- **Paginated**: In this mode we are storing each entity individually and inside a separate page.
+This mode is recommended for entities that are relatively large and there are a few of them.
+
+![image](https://user-images.githubusercontent.com/9152501/213291600-98229ee4-6358-4e0e-abe2-d6da9abe179e.png)
+
+The rationale behind separating these modes is based on two reasons:
+- _Performance_: **Itemized** mode always needs 1 on-chain DB_READ and 1 DB_WRITE but **Paginated**
+mode might be done with only 1 DB_WRITE.
+- _Composability_: **Itemized** mode allows better composability due to on-chain read requirement
+and smaller size of each.
```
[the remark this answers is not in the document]
aramikm: I think in terms of graph that is correct but since this is a generic storage there might be usecases that you might not need to read all pages.
One other thing to remember is that since reading is an off-chain process it does not have any impact on performance of the node.

JoeCap08055 (approved):

saraswatpuneet (approved):
:100:

PR #904, opened 2023-01-11
Feat/withdraw unstaked 819

### LINKED ISSUE #819: [Capacity] Withdraw unstaked amount
### Description
As token user who has scheduled a withdrawal of token - unstaked tokens. I want to be able to to withdraw my tokens.

### Acceptance Criteria
1. Dispatched origin is Signed by Staker.
2. Sums all chunks that are less than or equal to the current Epoch and removes the lock by amount from the account balance.
3. Updates `StakingAccountLedger` total with new locking amount.
4. If an account has nothing at stake, clean up storage by removing StakingLedger and TargetLedger entries.
5. Emits event Withdrawn to notify that a withdrawal was made from the account and amount withdrawn.

shannonwells:
# Goal
The goal of this PR is to implement the `withdraw_unstaked` extrinsic for the capacity pallet.

Closes #819

# Discussion

# Checklist
- [x] Dispatched origin is Signed by Staker.
- [x] Removes all chunks that are less than or equal to the current Epoch and removes the lock by amount from the account balance.
- [x] Updates StakingAccountLedger totals with remaining amounts.
- [x] If an account has no remaining stake, clean up storage by removing StakingLedger
- [x] Emits event `StakeWithdrawn` to notify that a withdrawal was made from the account and amount withdrawn.
- [x] Returns `Error::NotAStakingAccount` if origin does not have anything staked
- [x] Returns `Error::NoUnstakedTokensAvailable` if origin has a staking account but there is nothing available to withdraw
- [x] Tests added
- [x] Benchmarks added
- [x] Weights updated
- N/A Chain spec updated
- N/A Design doc(s) updated
- N/A Custom RPC OR Runtime API added/changed? Updated js/api-augment.


pallets/capacity/src/lib.rs
```diff
@@ -203,6 +218,41 @@ pub mod pallet {
...
+		/// in the caller's token account.
+		///
+		/// ### Errors
+		///   - Returns `Error::NotAStakingAccount` if no StakingAccountDetails are found for `origin`.
+		///
+		pub fn withdraw_unstaked(origin: OriginFor<T>) -> DispatchResult {
+			let staker = ensure_signed(origin)?;
+			let mut amount_withdrawn: BalanceOf<T> = 0u32.into();
+			let current_block = frame_system::Pallet::<T>::block_number();
+			StakingAccountLedger::<T>::mutate_exists(
```
shannonwells, line 234: `mutate_exists` deletes the record if None is returned.

pallets/capacity/src/lib.rs
```diff
@@ -211,6 +226,36 @@ pub mod pallet {
...
 			Ok(())
 		}
+
+		// TODO: real weights
+		#[pallet::weight(T::WeightInfo::stake())]
+		/// removes all thawed UnlockChunks from caller's StakingAccount and unlocks the sum of the thawed values
+		/// in the caller's token account.
+		///
+		/// ### Errors
+		///   - Returns `Error::NotAStakingAccount` if no StakingAccountDetails are found for `origin`.
```
enddynayn, line 236: It looks like the `NoUnstakedTokensAvailable` error was accidentally left out.

pallets/capacity/src/lib.rs
```diff
@@ -211,6 +226,36 @@ pub mod pallet {
...
+			let staker = ensure_signed(origin)?;
+
+			let mut staking_account =
+				Self::get_staking_account_for(&staker).ok_or(Error::<T>::NotAStakingAccount)?;
+			let current_block = frame_system::Pallet::<T>::block_number();
+
+			let amount_withdrawn = staking_account.reap_thawed(current_block);
+			ensure!(!amount_withdrawn.is_zero(), Error::<T>::NoUnstakedTokensAvailable);
+
+			if staking_account.total.is_zero() {
```
enddynayn, line 248: How do you feel about having a function `staking_account.can_delete_account` this would make this line read a little nicer.

pallets/capacity/src/lib.rs
```diff
@@ -278,6 +323,11 @@ impl<T: Config> Pallet<T> {
 		StakingAccountLedger::<T>::insert(staker, staking_account);
 	}
 
+	fn remove_staking_account(staker: &T::AccountId) {
```
enddynayn, line 326: Nice! how about calling it" delete_" instead of "remove" to align with CRUD naming?

pallets/capacity/src/lib.rs
```diff
@@ -78,6 +78,9 @@ mod mock;
 #[cfg(test)]
 mod tests;
 
+#[cfg(test)]
+mod withdraw_unstaked_tests;
```
johnmela, line 82: If we want to follow a pattern of separate test files containing unit tests related to each extrinsic and its helper functions, staking / unstaking should probably also follow this pattern

johnmela: I am a fan of more modular pieces, personally. I also like consistency.

pallets/capacity/src/lib.rs
```diff
@@ -211,6 +226,36 @@ pub mod pallet {
 
 			Ok(())
 		}
+
+		// TODO: real weights
```
johnmela, line 230: Can this TODO line be removed?

johnmela (changes requested):
Do we want to include non capacity related generated weights?

enddynayn:
Looks superb, with some minor comments!

pallets/capacity/src/lib.rs
enddynayn: How about for now we put these in tests and create a ticket to break these up?

pallets/capacity/src/lib.rs
shannonwells: Yes I was planning to remove it once weights are finalized.

pallets/capacity/src/lib.rs
shannonwells: I was really hoping to start a trend with this one.  The tests file is getting pretty long.

pallets/capacity/src/lib.rs
shannonwells: legit

pallets/capacity/src/lib.rs
shannonwells: To be honest I dislike making single-line, single-use functions since it just adds another stack frame, and really, what are we getting out of it that a simple comment cannot straighten out, if needed?  If there is ever another place where we check if the staking account has a zero total balance, or if we add more checks before deleting the staking account, as well, then sure let's abstract it,

pallets/capacity/src/lib.rs
shannonwells: So we decided to break these up slightly differently

enddynayn (approved):
Fabolous! 💯

johnmela (approved):

pallets/capacity/src/types.rs
```diff
@@ -46,6 +51,46 @@ impl<T: Config> StakingAccountDetails<T> {
...
+	/// balance on the token account.
+	/// Returns: the total amount reaped from `unlocking`
+	pub fn reap_thawed(&mut self, current_block: <T>::BlockNumber) -> BalanceOf<T> {
+		let mut total_reaped: BalanceOf<T> = 0u32.into();
+		self.unlocking.retain(|chunk| {
+			if current_block.ge(&chunk.thaw_at) {
+				total_reaped = total_reaped + chunk.value;
+				match self.total.checked_sub(&chunk.value) {
+					Some(new_total) => self.total = new_total,
+					None => {
```
enddynayn, line 80: Not requesting a change, but I just wanted to run a thought. Should we do a log or throw an underflow error instead of printing? Since errors are included in a block, we can look for those errors.

ArithmeticError::Underflow

shannonwells: I'm not sure what you mean by "do a log...instead of printing" since the warn macro is a log output.  When we discussed this before, I mentioned how you can't return an error from `retain` iterator. We decided that it was an extremely unlikely edge case, that it's more or less harmless to just delete the chunk in the case of an underflow, and to log it in case it happens so it can be debugged.  So I chose warn over debug or info.

enddynayn: I was thinking in general printing vs. throwing an error. I did not mean to confuse you. I used to log and print interchangeably. I am still figuring out if there is a rule of thumb to when to print instead of throwing an error. 
Again, I'm not asking for a change.

PR #916, opened 2023-01-17
feat(capacity): add ability to unstake 

### LINKED ISSUE #818: [Capacity] Unstake token amount
### Description
As a token user who has staked to the network, I want to be able to schedule a withdrawal of all or a portion of my staked tokens.

### Acceptance Criteria
1. Dispatched origin is Signed by Staker.
2. Schedules a portion of the stake to be unlocked and ready for transfer after the `confg::UnstakingThawPeriod` ends.
3. The amount unstaked must be greater than 0.
4. Issued Capacity to the target is reduced by the same amount originally issued
5. The amount unstaked cannot exceed the amount staked.
6. If the result of the unstaking would be to leave a balance below `config::MinimumStakingAmount`, the entire amount will be unstaked to avoid leaving dust.
7. when an account has never been a staking account and an attempt to call unstake an error message of NotAStakingAccount should be returned. 
8. If you have a staking account and your active balance is zero, then an error message of AmountToUnstakeExceedsAmountStaked should be returned (the test should include unlocking).
9. Emits Unstake event with unstaker account, target, and amount.

johnmela:
# Goal
The goal of this PR is to allow the unstaking of token for the Capacity feature

Closes #818

### DIFF
--- a/pallets/capacity/src/lib.rs
+++ b/pallets/capacity/src/lib.rs
@@ -51,9 +51,8 @@ use frame_support::{
 };
 
 use sp_runtime::{
-	Perbill,
 	traits::{CheckedAdd, Saturating, Zero},
-	ArithmeticError, DispatchError,
+	ArithmeticError, DispatchError, Perbill,
 };
 
 pub use common_primitives::{
@@ -125,7 +124,7 @@ pub mod pallet {
 		type BenchmarkHelper: RegisterProviderBenchmarkHelper;
 
 		/// The number of Epochs before you can unlock tokens after unstaking.
-    	#[pallet::constant]
+		#[pallet::constant]
 		type UnstakingThawPeriod: Get<u16>;
 	}
 
@@ -183,6 +182,7 @@ pub mod pallet {
 			account: T::AccountId,
 			/// the total amount withdrawn, i.e. put back into free balance.
 			amount: BalanceOf<T>,
+		},
 		/// A token account has unstaked the Frequency network.
 		UnStaked {
 			/// The token account that staked tokens to the network.
@@ -212,17 +212,17 @@ pub mod pallet {
 		/// or nothing has passed the thaw period.
 		NoUnstakedTokensAvailable,
 		/// Unstaking amount should be greater than zero.
-  		UnstakedAmountIsZero,
+		UnstakedAmountIsZero,
 		/// Amount to unstake is greater than the amount staked.
-  		AmountToUnstakeExceedsAmountStaked,
+		AmountToUnstakeExceedsAmountStaked,
 		/// Attempting to unstake from a target that has not been staked to.
- 		StakingAccountNotFound,
+		StakingAccountNotFound,
 		/// Attempting to get a staker / target relationship that does not exist.
 		StakerTargetRelationshipNotFound,
 		/// Attempting to get the target's capacity that does not exist.
 		TargetCapacityNotFound,
 		/// Staker reached the limit number for the allowed amount of unlocking chunks.
-  		MaxUnlockingChunksExceeded,
+		MaxUnlockingChunksExceeded,
 	}
 
 	#[pallet::call]
@@ -291,7 +291,11 @@ pub mod pallet {
 		/// - Returns `Error::InvalidTarget` if `target` is not a valid staking target
 		/// - Returns `Error:: NotAStakingAccount` if `origin` has nothing staked
 		#[pallet::weight(T::WeightInfo::unstake())]
-		pub fn unstake(origin: OriginFor<T>, target: MessageSourceId, amount: BalanceOf<T>) -> DispatchResult {
+		pub fn unstake(
+			origin: OriginFor<T>,
+			target: MessageSourceId,
+			amount: BalanceOf<T>,
+		) -> DispatchResult {
 			let unstaker = ensure_signed(origin)?;
 
 			ensure!(T::TargetValidator::validate(target), Error::<T>::InvalidTarget);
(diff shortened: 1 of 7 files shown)


pallets/capacity/src/lib.rs
```diff
@@ -299,4 +346,41 @@ impl<T: Config> Pallet<T> {
...
+
+		staking_account_details.decrease_by(amount, thaw_at)?;
+		Self::set_staking_account(&unstaker, &staking_account_details);
+
+		Ok(())
+	}
+
+	/// Reduce available capacity of target and return the amount of capacity reduction.
+	fn reduce_capacity(unstaker: &T::AccountId, target: MessageSourceId, amount: BalanceOf<T>) -> Result<BalanceOf<T>, DispatchError>  {
+		let mut staking_target_details = Self::get_target_for(&unstaker, &target).ok_or(Error::<T>::NotStakingAccount)?;
```
enddynayn, line 366: I think we want a different error here: NotStakingAccount
maybe we can use `InvalidTarget`.

enddynayn: In this case, `InvalidTarget` might not be a good error. Maybe we reserve the `InvalidTarget` error when a target does not exist. In this case we are throwing an error because the Staker did not target the provider to receive Capacity. Maybe we need another one like `StakerTargetInvalid.`

enddynayn: @shannonwells thoughts?

pallets/capacity/src/lib.rs
```diff
@@ -299,4 +346,41 @@ impl<T: Config> Pallet<T> {
...
+		let mut capacity_details = Self::get_capacity_for(target).ok_or(Error::<T>::NotStakingAccount)?;
```
enddynayn, line 367: Let's update the error to something like `NotTargetCapacityAccount` since it's the target entry rather than the staking account entry.

pallets/capacity/src/lib.rs
johnmela: This is one of those cases where the storage is in an invalid state. We _shouldn't_ ever reach a state where we have a staking account storage but not a staking target. In addition to a separate error for this case, we probably want to know if it ever happens and trigger a log in our logging infrastructure

enddynayn: I don't think we need to log it since the error will be included in the block.

pallets/capacity/src/lib.rs
```diff
@@ -299,4 +346,41 @@ impl<T: Config> Pallet<T> {
...
+		capacity_details.decrease_by(capacity_reduction, amount);
+
+		Self::set_capacity_for(target, capacity_details);
+		Self::set_target_details_for(unstaker, target, staking_target_details);
+
+		Ok(capacity_reduction)
+	}
+
+	/// Determine the capacity reduction when given total_capacity, unstaking_amount, and total_amount_staked.
+	fn calculate_capacity_reduction(unstaking_amount: BalanceOf<T>, total_amount_staked: BalanceOf<T>, total_capacity: BalanceOf<T>) -> Result<BalanceOf<T>, DispatchError> {
```
enddynayn, line 379: I think it would be a good idea to add some tests to this one.

pallets/capacity/src/tests.rs
```diff
@@ -508,9 +508,237 @@ fn impl_staking_capacity_details_increase_by() {
...
+
+		assert_noop!(
+			Capacity::unstake(RuntimeOrigin::signed(token_account), target, unstaking_amount),
+			Error::<Test>::NotStakingAccount
+		);
+	});
+}
+
+#[test]
+fn impl_staking_account_details_decrease_by() {
```
enddynayn, line 675: nice

enddynayn (approved):
👍 looks amazing!

enddynayn:
Looks great! Don't forget to add that extra test.

pallets/capacity/src/lib.rs
```diff
@@ -178,6 +194,18 @@ pub mod pallet {
 		InsufficientStakingAmount,
 		/// Staker is attempting to stake a zero amount.
 		ZeroAmountNotAllowed,
+		/// Unstaking amount should be greater than zero.
+  		UnstakedAmountIsZero,
+		/// Amount to unstake is greater than the amount staked.
+  		AmountToUnstakeExceedsAmountStaked,
+		/// Attempting to unstake from a target that has not been staked to.
+ 		StakingAccountNotFound,
```
enddynayn, line 202: Nice!

shannonwells (changes requested):
A couple of minor requests and stuff to consider.

pallets/capacity/src/lib.rs
johnmela: Good catch, thanks @shannonwells 
We removed the checked_div function in favor of determining the rate as perbill

pallets/capacity/src/lib.rs
```diff
@@ -299,4 +350,41 @@ impl<T: Config> Pallet<T> {
...
+
+	/// Decrease a staking account's active token and create an unlocking chunk to be thawed at some future block.
+	fn decrease_active_staking_balance(unstaker: &T::AccountId, amount: BalanceOf<T>) -> DispatchResult {
+		let mut staking_account_details = Self::get_staking_account_for(unstaker).ok_or(Error::<T>::StakingAccountNotFound)?;
+		ensure!(amount <= staking_account_details.active, Error::<T>::AmountToUnstakeExceedsAmountStaked);
+
+		let current_block: T::BlockNumber = frame_system::Pallet::<T>::block_number();
+		let thaw_at = current_block.saturating_add(T::BlockNumber::from(T::UnstakingThawPeriod::get()));
+
+		staking_account_details.decrease_by(amount, thaw_at)?;
```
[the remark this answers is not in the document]
johnmela: There should be a test for this error case when the decrease_by fails, `unstake_errors_max_unlocking_chunks_exceeded()`. It checks if the `MaxUnlockingChunksExceeded` error is propagated back to the unstake extrinsic.

I am not opposed to adding unit tests for the pallet helper functions if that's what is decided. I did spend some time thinking about how much redundancy we want (overlap in tests), and it certainly wouldn't hurt to unit test both the extrinsic and its helpers.

shannonwells (approved):

Issue #859, opened 2023-01-03
Remove SUDO


p5150j:
Target date (1/10)

- [x] Rollout v1.0.1 to mainnet
- [x] Merge Polkadot update to 0.9.32
- [x] Cut release v1.1.0
- [x] Roll out v1.1.0 to Rococo
- [x] Roll out v1.1.0 to mainnet (Might be blocked)
- [x] Code and merge removing Sudo from mainnet (while rolling out v1.1.0)
- [x] Cut release v1.2.0
- [x] Roll out v1.2.0 to Rococo
- [x] Roll out v1.2.0 to mainnet
- [x] Sudo Removed!

## Notes
- try-runtime might not be in a working state. @wilwade has some of the fix for this, but it was on an old branch

PR #935, opened 2023-01-19
Remove sudo from mainnet

wilwade:
# Goal
The goal of this PR is to remove the sudo pallet from mainnet

Part of #859

# Discussion
-  It took a bit to get the features working correctly, but got it to the minimal state
- The unified runtime doesn't like running sudo and not sudo, so there is a default config for `all-frequency-features` for checks and tests
- I setup a compile blocker so that it is impossible for a frequency mainnet build to have `all-frequency-features` enabled

# Checklist
- [x] Checks run
- [x] Benchmarks tested locally


node/cli/src/command.rs
```diff
@@ -65,7 +65,7 @@ impl<T: sc_service::ChainSpec + 'static> IdentifyChain for T {
 fn load_spec(id: &str) -> std::result::Result<Box<dyn ChainSpec>, String> {
 	match id {
 		#[cfg(feature = "frequency")]
-		"frequency-bench" => return Ok(Box::new(chain_spec::frequency::frequency())),
+		"frequency-bench" => return Ok(Box::new(chain_spec::frequency::benchmark_mainnet_config())),
```
wilwade, line 68: This is just a rename to remove the confusion as to why this function was still around.

.github/workflows/release.yml
```diff
@@ -236,7 +236,7 @@ jobs:
           shared-key: ${{env.RUST_TOOLCHAIN}}
       - name: Build Docs
         run: |
-          RUSTDOCFLAGS="--enable-index-page -Zunstable-options" cargo doc --no-deps
+          RUSTDOCFLAGS="--enable-index-page -Zunstable-options" cargo doc --no-deps --features frequency
```
wilwade, line 239: Due to the feature issues, docs should only run using the `frequency` feature

.github/workflows/verify-pr-commit.yml
```diff
@@ -271,7 +271,7 @@ jobs:
           profile: minimal
           toolchain: stable
       - name: Check
-        run: SKIP_WASM_BUILD= cargo check --all-features
+        run: SKIP_WASM_BUILD= cargo check --features runtime-benchmarks,all-frequency-features,std
```
wilwade, line 274: `--all-features` doesn't work now due to the mutually exclusive nature of the features.

saraswatpuneet (approved):

shannonwells (changes requested):
After I rebased, `make lint` fails locally for me, and I think there must be some conflict in the feature flags. I tried cargo clippy with no flags, without `runtime-benchmarks` and these fail in the same way. `cargo clippy --features runtime-benchmarks,std,frequency` succeeds though.  Maybe this isn't that big of a deal, but I think we probably want to figure out what is going on.

```
error[E0432]: unresolved import `frequency_runtime`
  --> node/service/src/service.rs:49:10
   |
49 |     pub use frequency_runtime;
   |             ^^^^^^^^^^^^^^^^^ no external crate `frequency_runtime`

error[E0282]: type annotations needed
   --> node/service/src/service.rs:456:2
    |
456 |     start_node_impl(parachain_config, polkadot_config, collator_options, id, hwbench).await
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ cannot infer type

error[E0599]: no method named `clone` found for struct `sc_service::Arc<_>` in the current scope
   --> node/service/src/service.rs:133:10
    |
133 |         client.clone(),
    |                ^^^^^ method not found in `sc_service::Arc<_>`
```

shannonwells (approved):

Issue #474, opened 2022-09-26
[Spike] on offchain storage and query


saraswatpuneet:
Time bound research on:
* Offchain workers
* Offchain storage , query and types of storage
* Propose for story #418 

https://docs.substrate.io/fundamentals/offchain-operations/#:~:text=Offchain%20indexing%20is%20an%20optional,complements%20the%20on%2Dchain%20state.

rlaferla:
### PROPOSAL

Create an off-chain worker in the **MSA** pallet.  This worker will be invoked automatically in a separate WASM environment to process each block in the chain to create maps for each MSA id to a list of account ids.  This process occurs as each block is finalized, it will keep track of the last block number that completed processing and store that block number in the off-chain storage.  This ensures that the maps are valid up to and including the last block processed.

```
fn offchain_worker(block_number: T::BlockNumber) {
}
```

In the worker, it is important to check if the node is a validator using `sp_io::offchain::is_validator()`, otherwise any changes to storage will not be stored.

Determine whether **local** or **persistent** storage will be used.  **Persistent** survives node restarts whereas **local** does not.  Are there security/performance tradeoffs between the two modes?  Initially, we will use **persistent** mode to prevent reprocessing.

Create a **Rust** module for handling off-chain storage that the `offchain_worker()` function will use to store and retrieve the data.

1. `last_processed_block_number`

The value is of type `BlockNumber`

2.  The maps of MSA ids to account ids.  As off-chain storage only has `StorageValueRef`, there needs to be a way to store maps.  This can be accomplished by generating a unique key that is the concatenation of a prefix plus the MSA id as follows:

```
	fn key_for_msa_id(msa_id: MessageSourceId) -> String {
		// e.g.  msa::m2a-1, msa::m2a-2, msa::m2a-5000, etc...
		return format!("msa::m2a-{}", msa_id)
	}
```

The value for the map should be a unique set of account ids.

`BTreeSet<AccountId> accountIds;`

As the **WASM** environment is deterministic and cannot use random data, traditional structures like `HashMap` cannot be used.  Instead, the implementation can use `BTreeMap` and `BTreeSet`.

`use sp_std::collections::{btree_map::BTreeMap,btree_set::BTreeSet};
`

`pub fn add_account_to_msa_id(&mut self, msa_id: MessageSourceId, acct_id: AccountId) {}
`

Lastly, a RPC endpoint needs to be created:

[...]

PR #943, opened 2023-01-19
Spike/offchain indexing: using offchain_index and offchain_worker to persist reverse map(?)

saraswatpuneet:
# Goal
The goal of this PR is to propose an offchain indexing of MSA vs Public Keys by processing block events

Part of #474 

Notes:
* TODO: unit tests, benchmarks (need see impact of whitelisted storage)
* Defined a generic primitive to handle offchain storage in common/primitives/offchain.rs
* Within MSA pallet we have offchain_storage.rs that define msa specific storage items
* In general: we set events (when raised) against a deterministic "key" in offchain_index using block number and a whitelisted counter, which when an offchain worker invokes uses it to find indexed events for a block number.
* May be will drop a diagram

# Checklist
- [ ] ~~Chain spec updated~~
- [ ] ~~Design doc(s) updated~~
- [x] Tests added
- [ ] ~~Benchmarks added~~
- [ ] ~~Weights updated~~


scripts/init.sh
```diff
@@ -83,6 +83,7 @@ start-frequency-instant)
 
   ./target/release/frequency \
     --dev \
+    --enable-offchain-indexing true \
```
saraswatpuneet, line 86: Required by collator node to enable indexing

runtime/frequency/src/lib.rs
```diff
@@ -946,6 +946,10 @@ impl_runtime_apis! {
 				Err(_) => None,
 			}
 		}
+
+		fn get_msa_keys_offchain(msa_id: MessageSourceId) -> (BlockNumber, Vec<AccountId>) {
+			Msa::get_msa_keys_offchain(msa_id)
+		}
```
saraswatpuneet, line 952: Called by RPC , return type is a tuple (BlockNumber, Vec<AccontId>)
BlockNumber states the latest snapshot of when offchain index was updated

pallets/msa/src/rpc/src/lib.rs
```diff
@@ -53,6 +53,10 @@ pub trait MsaApi<BlockHash, AccountId> {
 		delegator_msa_id: DelegatorId,
 		provider_msa_id: ProviderId,
 	) -> RpcResult<Option<Vec<SchemaId>>>;
+
+	/// Retrieve a list of MSA keys for a given MSA ID
+	#[method(name = "msa_getMsaKeys")]
+	fn get_msa_keys(&self, msa_id: MessageSourceId) -> RpcResult<(BlockNumber, Vec<AccountId>)>;
```
saraswatpuneet, line 59: return type for RPC is a tuple (BlockNumber, Vec<AccountId>) is ok? or needs beautification

pallets/msa/src/offchain_storage.rs
```diff
@@ -0,0 +1,181 @@
...
+	B: Encode + Clone + Decode + Eq + Debug + Default,
+{
+	/// msa_id is the key for the index
+	pub msa_id: K,
+	/// public_keys is the value for the index
+	pub public_keys: Vec<V>,
+	/// block_number is the block number of the last update
+	pub block_number: B,
+}
+
```
saraswatpuneet, line 43: Generic type stored on offchain db for each msa

common/primitives/src/offchain.rs
```diff
@@ -0,0 +1,77 @@
...
+		Duration::from_millis(LOCK_DEADLINE),
+	);
+	let lock_status = if let Ok(_guard) = lock.try_lock() {
+		f();
+		LockStatus::Released
+	} else {
+		LockStatus::Locked
+	};
+	lock_status
+}
```
saraswatpuneet, line 41: To acquire lock on an enclosure and do your thing

common/primitives/src/offchain.rs
```diff
@@ -0,0 +1,77 @@
...
+/// Wrapper for offchain_index remove operations
+pub fn remove_offchain_index_value(key: &[u8]) {
+	offchain_index::clear(key);
+}
+
+/// Wrapper for offchain_index set operations
+pub fn set_offchain_index_value(key: &[u8], value: &[u8]) {
+	offchain_index::clear(key);
+	offchain_index::set(key, value);
+}
```
saraswatpuneet, line 58: setting offchain_index when called from RuntimeCode (extrinsics) , acts as channel between Runtime and Offchain workers

common/primitives/src/offchain.rs
```diff
@@ -0,0 +1,77 @@
...
+
+/// Sets a value by the key to offchain index
+pub fn set_index_value<V>(key: &[u8], value: V)
+where
+	V: Encode + Debug,
+{
+	let oci_mem = StorageValueRef::persistent(key);
+	oci_mem.set(&value);
+}
```
saraswatpuneet, line 67: Call it as is or in a locked enclosure to prevent data races
Need to check if lock really works :skull:

common/primitives/src/offchain.rs
```diff
@@ -0,0 +1,77 @@
...
+
+/// Gets a value by the key from persistent storage
+fn get_impl<V: Decode + Debug>(key: &[u8]) -> Result<V, StorageRetrievalError> {
+	let oci_mem = StorageValueRef::persistent(key);
+	if let Ok(Some(data)) = oci_mem.get::<V>() {
+		return Ok(data)
+	} else {
+		return Err(StorageRetrievalError::Undecodable)
+	}
+}
```
saraswatpuneet, line 77: get anything via a key and decode it back into defined `V`

saraswatpuneet: A lot of the code here is kept generic for extensibility

pallets/msa/src/lib.rs
```diff
@@ -104,18 +105,24 @@ mod replay_tests;
...
 
 	use super::*;
 
 	#[pallet::config]
 	pub trait Config: frame_system::Config {
 		/// The overarching event type.
-		type RuntimeEvent: From<Event<Self>> + IsType<<Self as frame_system::Config>::RuntimeEvent>;
+		type RuntimeEvent: From<Event<Self>>
+			+ IsType<<Self as frame_system::Config>::RuntimeEvent>
+			+ TryInto<Event<Self>>;
```
saraswatpuneet, line 125: need runtime event to expose into's generic Event type

pallets/msa/src/lib.rs
```diff
@@ -397,8 +413,55 @@ pub mod pallet {
 	#[pallet::hooks]
 	impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {
 		fn on_initialize(current: T::BlockNumber) -> Weight {
+			<MSAEventCount<T>>::set(0u16);
+			// allocates 1 read and 1 write for any access of `MSAEventCount` in every block
+			T::DbWeight::get().reads(1u64).saturating_add(T::DbWeight::get().writes(1u64));
```
saraswatpuneet, line 418: quickly referencing from messages pallet to use this counter type  to bridge number of unique events indexed in a given block.
* Assuming offchain_workers will see their respective MSAEventCount, as the block_number they are processing?   
* If not do we really save more than just a Counter but also which block_number that counter is for?
@aramikm

pallets/msa/src/lib.rs
```diff
@@ -397,8 +413,55 @@ pub mod pallet {
...
+			let mut event_to_process: Vec<Event<T>> = Vec::new();
+
+			// read the event count for the block
+			let block_event_count = <MSAEventCount<T>>::get();
+			let current_bucket_num: u32 = block_number.try_into().unwrap_or_default();
+
+			// if there are events, read them and process them
+			if block_event_count > 0 {
+				event_to_process = Self::read_events(current_bucket_num, block_event_count);
+			}
```
saraswatpuneet, line 432: if the count of events is >0 we know we sure have indexed some events , read -> process ->(MSAPublicKeys mutation) -> store -> repeat

pallets/msa/src/lib.rs
```diff
@@ -419,8 +482,9 @@ pub mod pallet {
 			let (new_msa_id, new_public_key) =
 				Self::create_account(public_key, |_| -> DispatchResult { Ok(()) })?;
 
-			Self::deposit_event(Event::MsaCreated { msa_id: new_msa_id, key: new_public_key });
-
+			let event = Event::MsaCreated { msa_id: new_msa_id, key: new_public_key };
+			Self::index_event(event.clone());
```
saraswatpuneet, line 486: indexing an event

pallets/msa/src/lib.rs
```diff
@@ -1292,6 +1360,87 @@ impl<T: Config> Pallet<T> {
...
+				[BLOCK_EVENT_KEY, block_number.encode().as_slice(), i.encode().as_slice()].concat();
+			let optional_decoded_event = get_offchain_index::<Event<T>>(key.encode().as_slice());
+			if let Some(decoded_event) = optional_decoded_event {
+				events.push(decoded_event);
+			}
+			// read and clear the event from offchain storage
+			remove_offchain_index(key.encode().as_slice());
+		}
+		events
+	}
```
saraswatpuneet, line 1443: read_events from offchain_index -> decode -> remove from index -> return

pallets/msa/src/lib.rs
```diff
@@ -1292,6 +1360,87 @@ impl<T: Config> Pallet<T> {
...
+		<MSAEventCount<T>>::put(current_event_count);
+		let key = [
+			BLOCK_EVENT_KEY,
+			block_number.encode().as_slice(),
+			current_event_count.encode().as_slice(),
+		]
+		.concat();
+		// set the event in offchain storage
+		set_offchain_index(key.encode().as_slice(), event);
+	}
```
saraswatpuneet, line 1428: Send an event to be indexed, the following key becomes deterministic from  offchain_worker via the counter per block 

`	let key = [
			BLOCK_EVENT_KEY,
			block_number.encode().as_slice(),
			current_event_count.encode().as_slice(),
		]`

pallets/msa/src/lib.rs
```diff
@@ -1292,6 +1360,87 @@ impl<T: Config> Pallet<T> {
...
+
+	fn read_events(block_number: u32, event_count: u16) -> Vec<Event<T>> {
+		let mut events = vec![];
+		for i in 1..event_count + 1 {
+			let key =
+				[BLOCK_EVENT_KEY, block_number.encode().as_slice(), i.encode().as_slice()].concat();
```
saraswatpuneet, line 1434: Same key we set while indexing

pallets/msa/src/offchain_storage.rs
```diff
@@ -0,0 +1,181 @@
...
+	fmt::{Debug, Formatter},
+	vec::Vec,
+};
+
+/// Block event storage prefix
+pub const BLOCK_EVENT_KEY: &[u8] = b"frequency::block_event::msa::";
+
+/// Generic prefix for MSA index storage
+pub const MSA_INDEX_KEY: &[u8] = b"frequency::msa::";
+
```
saraswatpuneet, line 16: Prefixes used in this implementation one to index `BLOCK_EVENT_KEY` to offchain_index and `MSA_INDEX_KEY` to store and retrieve public keys for a given msa_id

common/primitives/src/offchain.rs
```diff
@@ -0,0 +1,77 @@
...
+	storage::{StorageRetrievalError, StorageValueRef},
+	storage_lock::{StorageLock, Time},
+	Duration,
+};
+use sp_std::fmt::Debug;
+
+const DB_LOCK: &[u8] = b"lock";
+
+const LOCK_DEADLINE: u64 = 1000;
+
```
saraswatpuneet, line 13: 1 sec lock , too pessimistic?

pallets/msa/src/lib.rs
```diff
@@ -481,17 +545,14 @@ pub mod pallet {
...
 				})?;
-
-			Self::deposit_event(Event::MsaCreated {
-				msa_id: new_delegator_msa_id,
-				key: new_delegator_public_key,
-			});
-
+			let event =
+				Event::MsaCreated { msa_id: new_delegator_msa_id, key: new_delegator_public_key };
+			Self::index_event(event.clone());
```
rlaferla, line 550: You may want to index events after they are actually deposited.  It shouldn't make a difference but if there was some sort of failure, it could lead to incorrect data.

pallets/msa/src/lib.rs
```diff
@@ -1292,6 +1360,87 @@ impl<T: Config> Pallet<T> {
...
 	}
+
+	/// Get list of account ids that are associated with a given msa
+	pub fn get_msa_keys_offchain(msa_id: MessageSourceId) -> (T::BlockNumber, Vec<T::AccountId>) {
+		let msa_keys_res = get_msa_keys::<MessageSourceId, T::AccountId, T::BlockNumber>(msa_id);
+		if let Err(e) = msa_keys_res {
+			log_err!("Error getting msa keys from offchain storage: {:?}", e);
+			return (T::BlockNumber::zero(), vec![])
+		}
+		msa_keys_res.unwrap()
```
rlaferla, line 1371: Use of `unwrap()` without a test is unsafe.

common/primitives/src/offchain.rs
rlaferla: You could try setting it to 500ms.  Lowering the timeout value may increase the chances of lock contention, but it will also decrease the time that the node needs to wait for a lock, which will lead to a higher transaction throughput.

rlaferla: I can't find the source but somewhere I read the default timeout was 100ms.

common/primitives/src/offchain.rs
rlaferla: Is the `clear()` necessary before `set()`?

pallets/msa/src/lib.rs
rlaferla: Minor:  Rename to `events_to_process` to indicate that it's not a single event but multiple events.

rlaferla (commented):
Looks good for an initial implementation:   As events are created (`MsaCreated`, `PublicKeyAdded`, and `PublicKeyRemoved`) in the MSA pallet, they are stored using `index_event()`,  then the off-chain worker reads and processes all these events by putting them in a `BTreeMap`, locks the storage and writes them out. 

I'd like to see a test that reads the resulting data and generates a report.

rlaferla (approved):

common/primitives/src/offchain.rs
saraswatpuneet: I agree shorter the better, will see the outcome of discussion, there three types of locks

* Time : duration based
* Block: lock acquired on block
* BlockTime: hybrid of both 
https://paritytech.github.io/substrate/master/sp_runtime/offchain/storage_lock/index.html

For our case i believe timed lock is enough

pallets/msa/src/lib.rs
saraswatpuneet: i think its irrelevant, we should not be expecting this part to fail in any situation

pallets/msa/src/lib.rs
saraswatpuneet: I intend to add tests but yes 1367 will ensure error is caught and since get_msa_keys<> does decoding we are sure it will be either pass or fail

pallets/msa/src/lib.rs
aramikm: in theory, you don't need to have the whole key, you should be able to lookup all the keys that have the same prefix, in our case `[BLOCK_EVENT_KEY, block_number.encode().as_slice()]`. So you might not even need the  event_count.

NVM I see that they didn't expose getting values by prefix from offchain storage api which sucks, so it seems like it would be difficult to do that.

aramikm (commented):
Overall looks great, added some comments and concerns.

pallets/msa/src/rpc/src/lib.rs
aramikm: it's fine for a spike. We don't need to make it perfect yet

aramikm:
Uploading to link
[Substrate Node (1).pdf](https://github.com/LibertyDSNP/frequency/files/10471191/Substrate.Node.1.pdf)

pallets/msa/src/lib.rs
```diff
@@ -397,8 +413,55 @@ pub mod pallet {
...
+			// allocates 1 read and 1 write for any access of `MSAEventCount` in every block
+			T::DbWeight::get().reads(1u64).saturating_add(T::DbWeight::get().writes(1u64));
 			Self::reset_virtual_bucket_if_needed(current)
 		}
+
+		fn offchain_worker(block_number: T::BlockNumber) {
+			let mut event_to_process: Vec<Event<T>> = Vec::new();
+
+			// read the event count for the block
+			let block_event_count = <MSAEventCount<T>>::get();
```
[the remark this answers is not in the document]
saraswatpuneet: Yep exactly my major concern about this approach, I think I am getting similar logic sorted out but agreed

pallets/msa/src/lib.rs
```diff
@@ -1292,6 +1360,87 @@ impl<T: Config> Pallet<T> {
...
+			return (T::BlockNumber::zero(), vec![])
+		}
+		msa_keys_res.unwrap()
+	}
+
+	fn process_events(
+		msa_id: MessageSourceId,
+		events: Vec<(T::AccountId, EventType, T::BlockNumber)>,
+	) {
+		let lock_status = offchain_common::lock(msa_id.encode().as_slice(), || {
```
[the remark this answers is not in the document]
saraswatpuneet: Good point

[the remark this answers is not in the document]
saraswatpuneet: Enforced a key pattern

pallets/msa/src/lib.rs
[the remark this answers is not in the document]
saraswatpuneet: So instead of using per block counter, instead during indexing of event we can simly continue to update event count for that block in offchain index

The MSAEventCount will clear up on initialize
The MSAEvent count is a placeholder to count and index using block_number as key and a deterministic prefix

offchain_worker instead of getting the count from MSAEventCount would rather use block_number to get he count and process the events accordingly

common/primitives/src/offchain.rs
saraswatpuneet: It was left over from some a test. removed it

common/primitives/src/offchain.rs
saraswatpuneet: Found that default is even much larger, I think based on how typically of offchain_worker will take to write to db? I think we can go with much lower value
https://github.com/paritytech/substrate/blob/2eb8ad273eeabea380fce00c10893ac788b1cbde/primitives/runtime/src/offchain/storage_lock.rs#L74

pallets/msa/src/lib.rs
saraswatpuneet: This function continues to index events sent to it and also index the count of events it gets per block for offchain reference

pallets/msa/src/lib.rs
```diff
@@ -1292,6 +1353,99 @@ impl<T: Config> Pallet<T> {
...
+	}
+
+	fn read_events(block_number: T::BlockNumber) -> Vec<Event<T>> {
+		let current_block: u32 = block_number.try_into().unwrap_or_default();
+		let count_key = [BLOCK_EVENT_COUNT_KEY, current_block.encode().as_slice()].concat();
+		let optional_event_count = get_offchain_index::<u16>(count_key.encode().as_slice());
+		let mut events = vec![];
+		let event_count =
+			if let Some(event_count) = optional_event_count { event_count } else { return events };
+
```
saraswatpuneet, line 1436: checks if any event count information is indexed if so get the count and find all MSA events index by counter

pallets/msa/src/lib.rs
```diff
@@ -1292,6 +1353,99 @@ impl<T: Config> Pallet<T> {
...
+		for i in 1..event_count + 1 {
+			let key =
+				[BLOCK_EVENT_KEY, block_number.encode().as_slice(), i.encode().as_slice()].concat();
+			let optional_decoded_event = get_offchain_index::<Event<T>>(key.encode().as_slice());
+			if let Some(decoded_event) = optional_decoded_event {
+				events.push(decoded_event);
+			}
+			// read and clear the event from offchain storage
+			remove_offchain_index(key.encode().as_slice());
```
saraswatpuneet, line 1445: Once done clear out the counter so no other offchain worker process the same block again

common/primitives/src/offchain.rs
saraswatpuneet: We can keep in mind that lock is acquired per msa_id so i believe its safe to use even default value, 1 sec sounds reasonable thinking about many many msa_ids being indexed over time

pallets/msa/src/lib.rs
aramikm: Looks great!

pallets/msa/src/lib.rs
aramikm: Now that I think about it maybe it might make sense to have a block_number level offchain lock instead of msa level lock for 3 reasons

1. From user perspective we only want to check to see if this reverse map is updated up to which block number for all inputs but the way we are storing the block number of storage per msa it's hard to tell that if there are more events after this block number that didn't get indexed yet.
2. since we are removing this storage before processing all the events, there is always that possibility that node might die and there is no trace in offchain DB to continue the process.
3. Having point number 2 in mind we will need to be able to process unprocessed events from previous block numbers also. Since we are not sure that the offchain worker will run on every block number and we might need to resume process for a failed one.

I don't think we might need these at this point but it's some nice to haves.

aramikm (approved):
Looks great! thanks for updating.
Added some future improvement notes in case we would like to make it more bulletproof.

pallets/msa/src/lib.rs
saraswatpuneet: 1. Getting the indexed data for a given msa, we get the block_number which is the last updated. Not sure about the overhead
2. Agree a little chance that node crashed before offchain_worker was called, in low transaction world it is instantaneous but i am sure there will be lags. However yes if we can reproduce that else would be worth keeping eye on though we never expect mainnet to stall 
3. Yep we can think of some manual trigger to process missing blocks, but will keep it for laters

Awesome thanks for the amazing review, I will add some unit tests since benchmarks are fairly straightforward

PR #967, opened 2023-01-24
Manual Sealing

dopatraman:
# Goal
The goal of this PR is to add manual sealing mode to the Frequency chain, along with a commandline argument for running it.

# Discussion
Run `make start-manual` to start in manual sealing mode. 

# Checklist
- [ ] Chain spec updated
- [ ] Design doc(s) updated
- [ ] Tests added
- [ ] Benchmarks added
- [ ] Weights updated


saraswatpuneet (approved):

node/service/src/service.rs
```diff
@@ -522,14 +525,16 @@ fn frequency_dev_instant(config: Configuration) -> Result<TaskManager, sc_servic
 		let (command_sink, commands_stream) = futures::channel::mpsc::channel(1024);
 
 		let pool = transaction_pool.pool().clone();
-		let import_stream = pool.validated_pool().import_notification_stream().map(|_| {
-			sc_consensus_manual_seal::rpc::EngineCommand::SealNewBlock {
+		let import_stream = pool
+			.validated_pool()
+			.import_notification_stream()
+			.filter(move |_| futures::future::ready(is_instant))
```
wilwade, line 531: This feels like an odd way to disable the stream, but just filtering everything. A more explicit if statement feels a bit more clear as to what it is doing.

```
let commands_stream = if is_instant {
  let pool = transaction_pool.pool().clone();
  let import_stream = pool.validated_pool()
  // ...
  futures::stream_select!(commands_stream, import_stream)
} else {
  commands_stream
}

wilwade (changes requested):
Few cleanup notes, but it checks out as working!

node/service/src/service.rs
dopatraman: I originally tried that, but the rust compiler would not allow it since it would mean the type of `commands_stream` would be different depending on the value of `is_instant`. 

The `stream_select!` macro returns a different type than `mpsc::channel` -- While both conform to the `Stream<_>` interface, Rust does not seem like it plays well with traits being passed around.

wilwade: Ah and `commands_stream` type is not an easy one to just manually set. I guess that works. Feels like there is an easier way we are missing, but this is simple.

wilwade (approved):

johnmela (approved):

PR #982, opened 2023-01-26
[For reference only, DO NOT MERGE] Spike/governance based registrations

shannonwells:
options for having council-approved schema creation and provider creation/registration

### DIFF
--- a/pallets/msa/src/lib.rs
+++ b/pallets/msa/src/lib.rs
@@ -165,7 +165,7 @@ pub mod pallet {
 		type MaxSignaturesStored: Get<Option<u32>>;
 
 		/// The origin that is allowed to create providers
-		type CreateProviderOrigin: EnsureOrigin<Self::RuntimeOrigin, Success = Self::AccountId>;
+		type CreateProviderOrigin: EnsureOrigin<Self::RuntimeOrigin>;
 	}
 
 	#[pallet::pallet]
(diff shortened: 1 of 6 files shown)


pallets/msa/src/lib.rs
```diff
@@ -853,6 +858,51 @@ pub mod pallet {
 			}
 			Ok(Some(T::WeightInfo::retire_msa(num_deletions)).into())
 		}
+
+		/// create a provider by means of governance approval
+		#[pallet::call_index(11)]
+		#[pallet::weight(T::WeightInfo::create_provider(100_000))]
+		pub fn create_provider_via_governance(
```
rlaferla, line 865: Since this is a public function for end-users, I think we should name it `request_to_be_a_provider()` or similar.

pallets/msa/src/lib.rs
```diff
@@ -853,6 +858,51 @@ pub mod pallet {
...
+			.map_err(|_| Error::<T>::NewKeyOwnershipInvalidSignature)?;
+
+			let provider_msa_id = Self::ensure_valid_msa_key(&create_provider_key)?;
+
+			let bounded_name: BoundedVec<u8, T::MaxProviderNameSize> = create_provider_payload
+				.clone()
+				.provider_name
+				.try_into()
+				.map_err(|_| Error::<T>::ExceedsMaxProviderNameSize)?;
+
```
rlaferla, line 889: Is there a way we could share the same code for this function and `create_provider()` so that changes only need to happen in one place?

PR #983, opened 2023-01-26
feat(event-tuple-to-struct): Use structs instead of tuples for extrinsic events

### LINKED ISSUE #961: Make chain event types use consistent style
So I had been curious as to why, in our TS integration tests, we can do something like:
`msaId = msaCreatedEvent.data.msaId;`
...with all events except for SchemaCreated and SchemaMaxSizeChanged, which require you to do something like:
`schemaId = schemaCreatedEvent.data.1;`

Turns out it's because the Events in Rust are defined using C-like structures for all events in all pallets, except for the schema pallet. Compare:
```
	#[pallet::event]
	#[pallet::generate_deposit(pub(super) fn deposit_event)]
	pub enum Event<T: Config> {
		/// Messages are stored for a specified schema id and block number
		MessagesStored {
			/// The schema for these messages
			schema_id: SchemaId,
			/// The block number for these messages
			block_number: T::BlockNumber,
			/// `DEPRECATED` - Number of messages in this block for this schema
			/// Value does not reflect the actual count and it will be removed in June 2023
			count: u16,
		},
	}
```

to

```
	#[pallet::event]
	#[pallet::generate_deposit(pub (super) fn deposit_event)]
	pub enum Event<T: Config> {
		/// Emitted when a schema is registered. [who, schemas id]
		SchemaCreated(T::AccountId, SchemaId),

		/// Emitted when maximum size for schema model is changed.
		SchemaMaxSizeChanged(u32),
	}
```

Goal:
Make Event definitions consistently use struct-like format in all pallets, like so:

```
pub enum Event<T: Config> {
   SchemaCreated {
      key: T::AccountId,
      schema_id: SchemaId
   },

   SchemaMaxSizeChanged {
      max_size: u32
   }
}
```

JoeCap08055:
Use structs instead of tuples for extrinsic events #961

# Goal
The goal of this PR is to use C-style structs instead of tuples for extrinsic event types.

Closes #961


pallets/schemas/src/lib.rs
```diff
@@ -110,10 +110,19 @@ pub mod pallet {
...
+
+			/// Schema ID of newly-created schema
+			schema_id: SchemaId,
+		},
 
 		/// Emitted when maximum size for schema model is changed.
-		SchemaMaxSizeChanged(u32),
+		SchemaMaxSizeChanged {
+			/// Max size of schema document
+			max_size: u32,
```
johnmela, line 124: 👍 on the struct field comments

rlaferla (approved):
🥇

saraswatpuneet (approved):

PR #985, opened 2023-01-26
Added benchmarking for pallet-collator-selection.

### LINKED ISSUE #608: Run benchmarks and update weights for pallet_collator_selection once benchmarks are fixed for external pallets
Update benchmarks to use the proper local Benchmarks when this is working
- [x] pallet_collator_selection - this should be fixed in [cumulus branch polkadot-v0.9.31](https://github.com/paritytech/cumulus/pull/1714).

## Todo
- [x] Add it to the list of benchmarks to run `scripts/run_all_benchmarks.sh`
- [x] Add the generated file to `runtime/common/src/weights/mod.rs`
- [x] Run the new benchmarks on the reference hardware

rlaferla:
# Goal
The goal of this PR is to add benchmarking for pallet-collator-selection.

Closes #608


saraswatpuneet (approved):
:weight_lifting:

JoeCap08055 (approved):

shannonwells (dismissed):
Please remove all the weights generated that aren't for pallet-collator-selection.  No need to churn those since they'll be generated before a release anyway.

runtime/common/Cargo.toml
```diff
@@ -41,6 +41,7 @@ pallet-treasury = { git = "https://github.com/paritytech/substrate", default-fea
 pallet-utility = { git = "https://github.com/paritytech/substrate", default-features = false, branch = "polkadot-v0.9.36" }
 # cumulus
 cumulus-primitives-core = { git = "https://github.com/paritytech/cumulus", default-features = false, branch = "polkadot-v0.9.36" }
+pallet-collator-selection = { git = "https://github.com/paritytech/cumulus", default-features = false, branch = "polkadot-v0.9.36" }
```
wilwade, line 44: I think adding this should add to `std`, `try-runtime`, and maybe `runtime-benchmarks`

wilwade (changes requested):
One update and I think we need to remove the comment in runtime/frequency/src/lib.rs

// TODO: this uses default but we don't have weights yet. Issue: #608

runtime/frequency/src/lib.rs
```diff
@@ -457,7 +457,6 @@ impl pallet_collective::Config<TechnicalCommitteeCollective> for Runtime {
 	type MaxProposals = TCMaxProposals;
 	type MaxMembers = TCMaxMembers;
 	type DefaultVote = pallet_collective::PrimeDefaultVote;
-	// TODO: this uses default but we don't have weights yet. Issue: #608
 	type WeightInfo = pallet_collective::weights::SubstrateWeight<Runtime>;
```
wilwade, line 460: And Line 448 as well:

```suggestion
	type WeightInfo = weights::pallet_collective::SubstrateWeight<Runtime>;
```

wilwade (changes requested):
One more fix.

runtime/frequency/src/lib.rs
wilwade: This flips it over to use the new one instead of the built in one in the substrate crate.

rlaferla: I applied your change suggestion.  Note that this issue was for the collective pallet (previous PR) while this PR is for the collator selection pallet.

wilwade: @rlaferla Ah so it is

- [ ] We need to fix the collective on line 448 as well.
- [ ] We currently have `type WeightInfo = ();` for the collator selection pallet which should reference the new weights.

rlaferla: Fixed.

wilwade (approved):
👍

PR #1003, opened 2023-01-31
add reset runner action

### LINKED ISSUE #992: Create CI state cleanup Job
Since, the self-hosted runners don't start off with a clean state each time, we need to manually do the cleanup.  Doing it in different jobs makes it hard to maintain. Consider implementing all cleanup in a dedicated CI job that runs before every other job.

demisx:
# Goal
The goal of this PR is to create an reusable action that can be used to reset runner to its original state (as close as possible). Also includes minor cleanup to make jobs consistent with each other.

Closes #992

### DIFF
--- a/.github/workflows/common/reset-self-hosted-runner/action.yml
+++ b/.github/workflows/common/reset-self-hosted-runner/action.yml
@@ -1,0 +1,19 @@
+name: Reset Self-hosted Runner
+description: Shared action that allows to reset a runner to its original state
+runs:
+  using: "composite"
+  steps:
+    - name: Remove Cargo Generated Artifacts
+      shell: bash
+      run: |
+        echo "Removing generated cargo artifacts..."
+        rm -fr ./target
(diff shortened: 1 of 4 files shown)


.github/workflows/common/reset-self-hosted-runner/action.yml
```diff
@@ -0,0 +1,22 @@
+name: Reset Self-hosted Runner
+description: Shared action that allows to reset a runner to its original state
+runs:
+  using: "composite"
+  steps:
+    - name: Remove Cargo Generated Artifacts
+      shell: bash
+      run: |
+        echo "Removing generated cargo artifacts..."
+        rm -fr ./target
```
wilwade, line 10: Do we want this to just use `cargo clean`? I guess we don't want to have to have the cargo dep at this point however.

wilwade (approved):
Need to clear out `runtime/common/992-delete-me.rs`, but otherwise good to go!

Issue #988, opened 2023-01-27
Add new payload locations for schemas


aramikm:
We need to add 2 new payload locations for schemas for **itemized** and **Paginated** data and add these checks in messages and stateful pallets.

PR #1007, opened 2023-02-01
Define stateful storage Payload location types

saraswatpuneet:
# Goal
The goal of this PR is to add ```Itemized``` and ```Paginated```

* Add to PayloadLocation Enum
* Make sure the build passes

Part of : #988 

# Checklist
- [ ] ~~Chain spec updated~~
- [ ] ~~Custom RPC OR Runtime API added/changed? Updated js/api-augment.~~
- [ ] ~~Design doc(s) updated~~
- [ ] ~~Tests added~~
- [ ] ~~Benchmarks added~~
- [ ] ~~Weights updated~~


aramikm (approved):
Looks good. Thanks for rebasing from main!

pallets/messages/src/types.rs
```diff
@@ -62,6 +62,15 @@ where
 					msa_id: None,
 					payload: None,
 				}
+			}, // TODO: Add support for itemized and paginated payloads
```
aramikm, line 65: One question is that do we really want to expose these stateful storages as messages. I'm not totally convinced about that since we have a different definition of a message and stateful storage and even the access is from different pallets. maybe we want to have a more generic return type that will include all types but then that should have a different name than `MessageResponse`

JoeCap08055: Hmm. Well I think the thing is, we're not storing *messages* in stateful storage; we're storing a mutable final state. The *message* will (may) be generated when stateful storage is modified, but the `message payload <> stateful storage payload`. Put simply, I think we should generate an error if we try to store a `Message` in stateful storage. The `Message` would be something like `StatefulStorageUpdated` (or, more likely something more specific like `PrivateGraphUpdated`, and *that* message's storage location would be `OnChain`.

saraswatpuneet: this is just an RPC call to get messages , I have added a comment that for Itemized and Paginated its a different access point but returning default MessageResponse so we dont have to update messages pallets 
What you think

JoeCap08055: I guess the question is, should we return a `MessageResponse` with `None` values, or should we return an error, since there should be no "messages" in Itemized or Paginated storage?

For instance, maybe `get_message_by_schema_and_block()` should just return an empty vector if the storage location specified in the call is Itemized or Paginated?

saraswatpuneet: aah I see what you mean, I will add those checks and bypass

pallets/messages/src/lib.rs
```diff
@@ -393,11 +393,15 @@ impl<T: Config> Pallet<T> {
...
+		match schema_payload_location {
+			PayloadLocation::Itemized | PayloadLocation::Paginated => return Vec::new(),
+			_ =>
+				return <Messages<T>>::get(block_number, schema_id)
+					.into_inner()
+					.iter()
+					.map(|msg| msg.map_to_response(block_number_value, schema_payload_location))
+					.collect(),
+		}
 	}
```
saraswatpuneet, line 405: @JoeCap08055  @aramikm  this would prevent unnecessary read from db.

JoeCap08055 (approved):

PR #1011, opened 2023-02-01
Paginated page types

JoeCap08055:
# Goal
The goal of this PR is to add the types for Paginated pages

# Discussion
NOTE: I did a little bit of refactoring here to take advantage of commonalities between Paginated and Itemized pages.
Also, refactored a bit to try and get more proficient with Rust traits. If it's overkill, let me know & I can de-complicate it.

# Checklist
- [ ] Chain spec updated
- [ ] Design doc(s) updated
- [x] Tests added
- [ ] Benchmarks added
- [ ] Weights updated


saraswatpuneet:
@JoeCap08055  this PR is set to merge in main, can you change it to ```private_graph_main``` branch

pallets/stateful-storage/src/types.rs
```diff
@@ -19,41 +19,108 @@ pub struct ItemHeader {
...
+
+impl PageType for Itemized {}
+impl PageType for Paginated {}
+
+/// A page of data
 #[derive(Encode, Decode, TypeInfo, MaxEncodedLen, PartialEq, Debug)]
 #[scale_info(skip_type_params(PageDataSize))]
 #[codec(mel_bound(PageDataSize: MaxEncodedLen))]
-pub struct ItemPage<PageDataSize: Get<u32>> {
+pub struct Page<P: PageType, PageDataSize: Get<u32>> {
```
aramikm, line 58: Pages can not be under the same Type for a simple reason. `PageDataSize` for itemized pages are different from Paginated ones. I think we will need to have two different Types and we can implement the same Trait on both if we like to treat them as general pages.

pallets/stateful-storage/src/types.rs
```diff
@@ -19,41 +19,108 @@ pub struct ItemHeader {
...
+	PageSizeOverflow,
 }
 
-/// A page of items
+// NOTE: Would prefer to simply have an enum:
+// ```no_compile
+// enum PageType { Itemized, Paginated }
+// ```
+// However, currently can't use enum variants as const generic parameters.
+pub trait PageType {}
```
aramikm, line 43: I think this is a overkill for now. Let's implement a trait when there is a common functionality we need on both

pallets/stateful-storage/src/types.rs
```diff
@@ -19,41 +19,108 @@ pub struct ItemHeader {
 	payload_len: u16,
 }
 
+/// This header is used to specify how much data is in a page and is inserted into the buffer
+/// before every page in a Paginated model.
+#[derive(Clone, Encode, Decode, PartialEq, MaxEncodedLen, Debug)]
+pub struct PageHeader {
```
aramikm, line 25: I don't think we need this for Paginated ones. If we really do we can have something like the following
```
struct Page {
     payload_len:16,
     data: Vec<u8>
}
```
The reason we are doing something like that in `Itemized` for each item is because there are multiple items and if we don't append them like that, each of them might take `PageDataSize` length in PoV (proof of validity). But for Paginated ones we are dealing with only one item and the size of that item in the worst case will be the same size as `PageDataSize`

pallets/stateful-storage/src/types.rs
```diff
@@ -19,41 +19,108 @@ pub struct ItemHeader {
...
+{
+	fn from(bounded: BoundedVec<u8, PageDataSize>) -> Self {
+		Self { data: bounded, page_type: PhantomData }
+	}
+}
+
+impl<PageDataSize: Get<u32>, P: PageType> TryFrom<Vec<u8>> for Page<P, PageDataSize> {
+	type Error = ();
+
+	fn try_from(data: Vec<u8>) -> Result<Self, Self::Error> {
```
aramikm, line 100: I like all these utility Trait implementations.

aramikm (commented):
Added some comments and suggestions. please let me know if you have any questions.

pallets/stateful-storage/src/types.rs
JoeCap08055: But since `Page` is a generic, we can have 2 concrete types:
```
type PaginatedPage = Page<Paginated, MaxPaginatedPageSizeBytes>;
type ItemizedPage = Page<Itemized, MaxItemizedPageSizeBytes>;
```

pallets/stateful-storage/src/types.rs
JoeCap08055: There is common functionality already:
* `From<BoundedVec>`
* `TryFrom<Vec>`

There is also functionality that differentiates on `PageType`:
* `parse()` (though I think this might do well as a `TryFrom` train on the `Parsed*Page` struct)

pallets/stateful-storage/src/types.rs
JoeCap08055: I like having the header as a separate struct for several reasons:
* Can decode the header only
* More manageable if we decide to add fields to the header

Actually, I was thinking of the following (but wanted to discuss first):

* Have only 1 "raw" `Page` type, that basically encapsulates what the `Page<Paginated, PageDataSize>` does now (just handle header/size + payload.
* Move parsing + action logic to `ParsedItemizedPage`. It only handles higher-level logic. Then it marshals back to a raw `Page` which is the object that gets serialized to the tree. We actually wouldn't then even need a `ParsedPaginatedPage` type, then, unless we wanted to add specific paginated-related things to its header other than just a size.

Thoughts? The more I think about it, the more I like that approach.

JoeCap08055: Something like the following:

```
struct Page<PageDataSize: Get<u32>> {
   payload_len: usize,
   data: BoundedVec<u8, PageDataSize>,
}

struct ParsedPage<'a> {
	/// page current size
	page_size: usize,
	/// a map of item index to a slice of blob (header included)
	items: BTreeMap<u16, &'a [u8]>,
}


impl<PageDataSize: Get<u32>> TryFrom<Vec<u8>> for Page<PageDataSize> {}
impl<PageDataSize: Get<u32>> From<BoundedVec<u8, PageDataSize>> for Page<PageDataSize> {}
impl<PageDataSize: Get<u32>> TryFrom<ParsedPage> for Page<PageDataSize> {}

impl<'a, PageDataSize: Get<u32>> TryFrom<Page<PageDataSize>> for ParsedPage<'a> {} // replaces parse() function
```

That leaves a small outstanding question of where `apply_actions` belongs, but otherwise I like this... comments?

JoeCap08055: Okay, I think I've got the `apply_actions` pattern figured out to fit into the ☝🏻 model while still keeping the efficiency of the current approach:
```
struct ParsedPage<'a> {
	/// page current size
	page_size: usize,
	/// a map of item index to a slice of blob (header included)
	items: BTreeMap<u16, &'a [u8]>,
	/// buffer of added items
	add_buffer: Vec<u8>,
}

impl<'a> ParsedPage<'a> {
   /// Apply a single action to the page
   fn apply_action(&mut self, action: &ItemAction) -> &self {
      // 'Remove' deletes from the `items` tree
      // 'Add' appends to `add_buffer`
   }
   
   /// Apply multiple actoins to the page
   fn apply_actions(&mut self, action: &[ItemAction]) -> &self {
      /// iterate over actions and call self.appy_action()
   }
   
   /* The part of the original apply_actions() function that serializes the data back
       to a `Page` is now in the `TryFrom<ParsedPage> for Page` trait implementation. */
}
```

aramikm: I had considered this approach previously but there were two things that pushed me to implement it as is.
1- if you have a mutable ParsedPage then in case of error it can end up in a `dirty state`. And somebody might use that dirty state id they don't know better.
2- in future I might able to not parse on every action and we only need parsing for Remove ones. But based on constant weights calculation complexity I didn't implement this optimization yet.


Overall, I don't have any strong feelings about refactoring this piece and if you think it makes sense go for it. the only issue about refactoring at this stage is that we will need to rewrite all the tests again and it would cause some time delays. I usually like to have the minimum viable implementation and when everything is out we can refactor with peace of mind.

pallets/stateful-storage/src/types.rs
aramikm: I like the different types from common generic, if you can make it work go for it.

I don't think they way you are using the paginated and itemized as first generic to differentiate between them is idiomatic Rust but I have to play with it to see how else we might be able to do it.

pallets/stateful-storage/src/types.rs
aramikm: About having header for paginated ones. We are already using the scale encoding to decode and encode between regular rust types and I'm not sure why we would want to do this in a custom manner when there is no need for it like itemized ones?  We would still need to do migration just like the other ones in case of any change but then we would need to do custom encode/decode every time

pallets/stateful-storage/src/types.rs
aramikm: There are other problem with a generic inner type. The macros we use on each of them might be different like 
`#[codec(mel_bound(T::MaxItemizedPageSizeBytes: MaxEncodedLen))]`

I'm not sure how we would be able to apply this macros based on type

pallets/stateful-storage/src/types.rs
JoeCap08055: Okay, gotcha. I think I like the refactor, but I'll leave out the `PageHeader` since we don't appear to really need it.

JoeCap08055: I'll take care of updating the tests; shouldn't take long. I already did a bunch yesterday.

pallets/stateful-storage/src/types.rs
JoeCap08055: Okay, reworked a bit. Once I eliminated `PageHeader` as unnecessary, there was also no longer a need for `ParsedPaginatedPage`. This simplifies things quite a bit.

saraswatpuneet (approved):
lgtm

aramikm (approved):
Looks good. Thanks

PR #1018, opened 2023-02-02
feat(capacity) Epoch period can be set through governance

### LINKED ISSUE #894: [Capacity] Epoch Period can be set through governance
### Description:
As an stakeholder of the frequency chain I should be able to submit a proposal to update the epoch period length.    

Acceptance Criteria:

- I should be able to submit an extrinsic that updates the period.
- Origin is Root.
- Sets the new Epoch-Period.
- New Epoch-Period begins at the Next Epoch's start.

johnmela:
# Goal
The goal of this PR is to allow the length of the epoch period (in blocks) to be set by a governance action

Closes #894

# Checklist
- [x] Design doc(s) updated
- [x] Tests added
- [x] Benchmarks added
- [x] Weights updated


pallets/capacity/src/lib.rs
```diff
@@ -195,6 +195,15 @@ pub mod pallet {
 	#[pallet::getter(fn get_current_epoch_used_capacity)]
 	pub type CurrentEpochUsedCapacity<T: Config> = StorageValue<_, BalanceOf<T>, ValueQuery>;
 
+	#[pallet::type_value]
+	/// EpochLength defaults to T::MaxEpochLength when not set
```
enddynayn, line 199: Let's consider dropping Max from MaxEpochLengh.

enddynayn: Also, can we remove the configuration if we now have a default value and are setting it through governance?

pallets/capacity/src/lib.rs
```diff
@@ -367,6 +378,26 @@ pub mod pallet {
...
+		/// ### Errors
+		/// - Returns `Error::MaxEpochLengthExceeded` if `length` is greater than T::MaxEpochLength.
+		#[pallet::call_index(3)]
+		#[pallet::weight(T::WeightInfo::unstake())]
+		pub fn set_epoch_length(
+			origin: OriginFor<T>,
+			length: T::BlockNumber,
+		) -> DispatchResult {
+			ensure_root(origin)?;
+			ensure!(length <= T::MaxEpochLength::get(), Error::<T>::MaxEpochLengthExceeded);
```
enddynayn, line 396: What do you think about emitting an event?
Something like, `EpochLengthUpdated`?

enddynayn:
Looks great! Let me know what your thoughts are around emitting and other comments.

pallets/capacity/src/lib.rs
enddynayn: Let's also add a ticket for adding an integration test to update the epoch length.

pallets/capacity/src/lib.rs
```diff
@@ -367,6 +378,26 @@ pub mod pallet {
...
+
+			EpochLength::<T>::set(length);
+			Ok(())
```
johnmela, line 399: Should an event be dispatched here?

johnmela: You beat me to it @enddynayn

pallets/capacity/src/lib.rs
johnmela: I think the intention for `MaxEpochLength` was to set an upper bound for how high we want to allow the epoch length to go, but I suppose if it is up to governance, we can just use a default

johnmela: After discussion internally, we decided to keep `MaxEpochLength` in place for safety

pallets/capacity/src/lib.rs
enddynayn: For a different story. @shannonwells, do we have a way to calculate the start of the next epoch? Also, when does this change take effect?

pallets/capacity/src/lib.rs
```diff
@@ -195,6 +195,18 @@ pub mod pallet {
...
+	/// EpochLength defaults to T::MaxEpochLength when not set
+	pub fn EpochLengthDefault<T: Config>() -> T::BlockNumber {
+		T::MaxEpochLength::get()
+	}
+
+	/// Storage for the epoch length
+	#[pallet::storage]
+	#[pallet::getter(fn get_epoch_length)]
+	pub type EpochLength<T: Config> =
+		StorageValue<_, T::BlockNumber, ValueQuery, EpochLengthDefault<T>>;
```
enddynayn, line 208: whats the default?

pallets/capacity/src/lib.rs
```diff
@@ -195,6 +195,18 @@ pub mod pallet {
 	#[pallet::getter(fn get_current_epoch_used_capacity)]
 	pub type CurrentEpochUsedCapacity<T: Config> = StorageValue<_, BalanceOf<T>, ValueQuery>;
 
+	#[pallet::type_value]
+	/// EpochLength defaults to T::MaxEpochLength when not set
+	pub fn EpochLengthDefault<T: Config>() -> T::BlockNumber {
+		T::MaxEpochLength::get()
```
enddynayn, line 201: Since the Max is used for validation purposes, and I imagine it would be a big number, should we change the default to something else?

pallets/capacity/src/lib.rs
johnmela: It defaults to `MaxEpochLength` set in the config (if there is no value in storage)

pallets/capacity/src/lib.rs
johnmela: We certainly can, currently it is set to 100 blocks, which is not a particularly large number

johnmela: Are you envisioning something like a configurable pallet constant `DefaultEpochLength` also set to something like 100? That would decouple it from `MaxEpochLength` while allowing us to scale up the max if we need to without messing with the default.

shannonwells:
awaiting rebase

enddynayn (approved):
💯 Belleza! LGTM!

shannonwells (approved):
I had the one comment about the mocks/test impact.  Please verify that it doesn't impact test run time much before merging - (say less than a minute), and if it does maybe we should think about a workaround.  I hate to spend a bunch of time dealing with a "simple" solution to a thing that's only ever going to run once.

.github/workflows/merge-pr.yml
```diff
@@ -6,6 +6,7 @@ on:
   push:
     branches:
       - main
+      - capacity-development
```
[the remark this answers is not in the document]
johnmela: This change is in `capacity-development`, odd that you were seeing it. Maybe the comment went in before your browser refreshed and registered the rebase.

pallets/capacity/src/extrinsics_tests.rs
```diff
@@ -55,7 +56,7 @@ fn withdraw_unstaked_correctly_sets_new_lock_state() {
 		assert_eq!(10u64, Balances::locks(&staker)[0].amount);
 
 		// Epoch length = 10, we want to run to epoch 3
-		run_to_block(31);
+		run_to_block(301);
```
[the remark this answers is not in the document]
johnmela: Ah thanks for the heads up. I didn't realize it was producing each block sequentially, yes the reason for this is the hard coded default value vs a configurable pallet constant. The tests (on my development machine) run extremely quickly. However, for the sake of good performance practices moving forward, I have reduced the epoch_length in tests where we call `run_to_block` to 10 blocks

PR #1028, opened 2023-02-02
321 propose to be a provider

### LINKED ISSUE #321: [Provider Registration] Provider Creation limits
Frequency needs an easy way to figure out who is a provider and what the gates are to become one. 
In the short term, we want to make this a governance action, but still be frictionless.

Related: #578

## User Journey (Spike: `spike/governance-based-registrations`)
- User creates a MSA
- User wants to be a provider
- User uses an MSA control key to sign the `create_provider_via_governance` payload
- User asks off-chain (discord, email, etc...) one Frequency Council member executes the extrinsic on their behalf.
- The end.

## User Journey (Spike + Small changes)
- User creates a MSA
- User wants to be a provider
- User asks off-chain (discord, email, etc...) one Frequency Council member votes to create a motion to grant them provider status
- The Frequency Council member creates and votes for for the proposal
- Anyone (likely the one voter) closes the ^ proposal
- Proposal executes and creates provider for original user's MSA
- ALT: User does full referendum process.

## User Journey (User initiated Option) (@rlaferla to look at time boxed to 1-2 days)
- User creates a MSA
- User wants to be a provider
- *User calls extrinsic* that generates a proposal in the Frequency Council
- At least One Frequency Council member votes for the proposal
- Anyone (likely the one voter) closes the ^ proposal
- Proposal executes and creates provider for original user's MSA
- ALT: User does full referendum process.

## Long term: User Journey
- User creates a MSA
- User wants to be a provider
- User stakes/pays fee/etc... and executes extrinsic and becomes provider. The end.

## Acceptance Criteria

- Only a frequency council member can approve an MSA to become a provider
- Local and rococo also allow anyone to become an provider
- Any associated with the MSA must have requested provider creation 
- Documentation updated to reflect

### LINKED ISSUE #1008: DESIGN: Request to be a provider
@rlaferla

rlaferla:
# Goal
The goal of this PR is to handle provider creation by governance

Closes #321


enddynayn:
Looks great. Why do we need a collective pallet for providers? Also, maybe we should extra provider registration into its pallet?

pallets/msa/src/lib.rs
```diff
@@ -161,6 +183,12 @@ pub mod pallet {
 		/// calculated value.
 		#[pallet::constant]
 		type MaxSignaturesStored: Get<Option<u32>>;
+
+		/// The origin that is allowed to create providers
+		type CreateProviderOrigin: EnsureOrigin<Self::RuntimeOrigin, Success = AccountId32>;
+
+		// The origin that is allowed to create providers via governance
+		type CreateProviderViaGovernanceOrigin = EnsureOrigin<Self::RuntimeOrigin>;
```
enddynayn, line 191: why is this one needed if you have CreateProviderOrigin?

pallets/msa/src/lib.rs
```diff
@@ -848,11 +867,48 @@ pub mod pallet {
...
+					);
 				},
 			}
 			Ok(Some(T::WeightInfo::retire_msa(num_deletions)).into())
 		}
+
+		/// Request to be a provider
+		#[pallet::call_index(11)]
+		#[pallet::weight(1000)]
+		pub fn request_to_be_provider(
```
enddynayn, line 882: Should the request be the one that is being called through governance? Once the request is accepted then the provider fills in their info?

enddynayn:
Maybe use of feature flags for extrinsic?

rlaferla:
> Maybe use of feature flags for extrinsic?

What do you propose?

pallets/msa/src/weights.rs
```diff
@@ -18,7 +18,7 @@
 //! Autogenerated weights for pallet_msa
 //!
 //! THIS FILE WAS AUTO-GENERATED USING THE SUBSTRATE BENCHMARK CLI VERSION 4.0.0-dev
-//! DATE: 2023-01-27, STEPS: `20`, REPEAT: 10, LOW RANGE: `[]`, HIGH RANGE: `[]`
+//! DATE: 2023-02-09, STEPS: `20`, REPEAT: 10, LOW RANGE: `[]`, HIGH RANGE: `[]`
 //! EXECUTION: Some(Wasm), WASM-EXECUTION: Compiled, CHAIN: Some("frequency-bench"), DB CACHE: 1024
```
enddynayn, line 22: Can you please just checking the weights for MSA?

pallets/msa/src/lib.rs
```diff
@@ -1068,6 +1146,40 @@ impl<T: Config> Pallet<T> {
...
+	///
+	/// # Events
+	/// * [`Event::ProviderCreated`]
+	///
+	/// # Errors
+	/// * [`Error::NoKeyExists`] - account does not have an MSA
+	/// * [`Error::ExceedsMaxProviderNameSize`] - Too long of a provider name
+	/// * [`Error::DuplicateProviderRegistryEntry`] - a ProviderRegistryEntry associated with the given MSA id already exists.
+	///
+	pub fn do_create_provider(
```
enddynayn, line 1160: Maybe we can simplify this function to something like this and use this in the extrinsic.
```
pub fn create_registered_provider(
		provider_id: ProviderId,
		name: BoundedVec<u8, T::MaxProviderNameSize>,
	) -> DispatchResult {
		ProviderToRegistryEntry::<T>::try_mutate(provider_id, |maybe_metadata| -> DispatchResult {
			ensure!(maybe_metadata.take().is_none(), Error::<T>::DuplicateProviderRegistryEntry);
			*maybe_metadata = Some(ProviderRegistryEntry { provider_name: name });
			Ok(())
		})
	}
```

pallets/msa/src/lib.rs
```diff
@@ -1068,6 +1146,40 @@ impl<T: Config> Pallet<T> {
...
+			|maybe_metadata| -> DispatchResult {
+				ensure!(
+					maybe_metadata.take().is_none(),
+					Error::<T>::DuplicateProviderRegistryEntry
+				);
+				*maybe_metadata = Some(ProviderRegistryEntry { provider_name: bounded_name });
+				Ok(())
+			},
+		)?;
+		Self::deposit_event(Event::ProviderCreated { provider_id: ProviderId(provider_msa_id) });
```
enddynayn, line 1179: I think we should move this out of this function. We have a pattern of logging at the end of extrinsics.

pallets/msa/src/lib.rs
```diff
@@ -511,26 +548,8 @@ pub mod pallet {
 		#[pallet::call_index(2)]
 		#[pallet::weight(T::WeightInfo::create_provider(provider_name.len() as u32))]
 		pub fn create_provider(origin: OriginFor<T>, provider_name: Vec<u8>) -> DispatchResult {
```
enddynayn, line 550: Maybe I am not understading. It looks like I can still register to be a provider without governance?

enddynayn: We can call this in local an rococo.

pallets/msa/src/lib.rs
```diff
@@ -511,26 +548,8 @@ pub mod pallet {
...
-					);
-					*maybe_metadata = Some(ProviderRegistryEntry { provider_name: bounded_name });
-					Ok(())
-				},
-			)?;
-			Self::deposit_event(Event::ProviderCreated {
-				provider_id: ProviderId(provider_msa_id),
-			});
-			Ok(())
+			let provider_key = T::CreateProviderOrigin::ensure_origin(origin)?;
```
enddynayn, line 551: if we can only call this function in rococo and local only why use ensure_signed here?

pallets/msa/src/lib.rs
enddynayn: ahh I see the feature flag is on the origin. Why not put the feature flag on the function itself?

pallets/msa/src/lib.rs
```diff
@@ -848,11 +876,61 @@ pub mod pallet {
...
 				},
 			}
 			Ok(Some(T::WeightInfo::retire_msa(num_deletions)).into())
 		}
+
+		/// Request to be a provider.  Creates a proposal for council approval to create a provider from a MSA
+		///
+		/// # Errors
+		/// - [`NoKeyExists`](Error::NoKeyExists) - If there is not MSA for `origin`.
+
```
enddynayn, line 892: here

pallets/msa/src/lib.rs
```diff
@@ -848,11 +876,61 @@ pub mod pallet {
...
+		#[pallet::call_index(11)]
+		#[pallet::weight(T::WeightInfo::request_to_be_provider(provider_name.len() as u32))]
+		pub fn request_to_be_provider(
```
enddynayn, line 895: maybe propose_provider since it submits a proposal?

pallets/msa/src/lib.rs
```diff
@@ -848,11 +876,61 @@ pub mod pallet {
...
+		/// # Events
+		/// * [`Event::ProviderCreated`]
+		///
+		/// # Errors
+		/// * [`Error::NoKeyExists`] - account does not have an MSA
+		/// * [`Error::ExceedsMaxProviderNameSize`] - Too long of a provider name
+		/// * [`Error::DuplicateProviderRegistryEntry`] - a ProviderRegistryEntry associated with the given MSA id already exists.
+		#[pallet::call_index(12)]
+		#[pallet::weight(T::WeightInfo::create_provider_via_governance(provider_name.len() as u32))]
+		pub fn create_provider_via_governance(
```
enddynayn, line 925: I think we can remove this function.

pallets/msa/src/lib.rs
enddynayn: I think we can have:

```
pallet_something::Config for Runtime {
....
blah 
blahe
#[feature....]
type CreateProviderOrigin = MsaCreateProviderOrigin;
// The origin that is allowed to create providers via governance
#[feature....]
type CreateProviderViaGovernanceOrigin = MsaCreateProviderViaGovernanceOrigin;
````

File: pallets/msa/src/lib.rs [hunk 11]
```diff
@@ -117,12 +129,26 @@ pub mod pallet {
 		/// The overarching event type.
 		type RuntimeEvent: From<Event<Self>> + IsType<<Self as frame_system::Config>::RuntimeEvent>;
 
+		/// The runtime call dispatch type.
+		type Proposal: Parameter
+			+ Dispatchable<RuntimeOrigin = Self::RuntimeOrigin, PostInfo = PostDispatchInfo>
+			+ From<Call<Self>>
+			+ IsSubType<Call<Self>>
```
enddynayn, line 136: I do not think you are using this trait.

pallets/msa/src/lib.rs
```diff
@@ -117,12 +129,26 @@ pub mod pallet {
...
+			+ IsType<<Self as frame_system::Config>::RuntimeCall>;
```
enddynayn, line 137: or this one.

pallets/msa/src/lib.rs
```diff
@@ -305,6 +337,11 @@ pub mod pallet {
 			/// The Delegator MSA Id
 			delegator_id: DelegatorId,
 		},
+		/// A proposal motion has been created
+		CreateProviderProposalCreated {
```
enddynayn, line 341: I not think this is being used.

pallets/msa/src/lib.rs
```diff
@@ -848,11 +876,61 @@ pub mod pallet {
...
+		/// # Errors
+		/// - [`NoKeyExists`](Error::NoKeyExists) - If there is not MSA for `origin`.
+
+		#[pallet::call_index(11)]
+		#[pallet::weight(T::WeightInfo::request_to_be_provider(provider_name.len() as u32))]
+		pub fn request_to_be_provider(
+			origin: OriginFor<T>,
+			provider_name: Vec<u8>,
+		) -> DispatchResult {
+			let proposer = ensure_signed(origin)?;
```
enddynayn, line 899: We probably dont want to create a proposal if we dont have an msa.

pallets/msa/src/lib.rs
shannonwells: @rlaferla @wilwade ^^ pls to check my explanation in case I missed something

pallets/msa/src/lib.rs
shannonwells: We can't.

pallets/msa/src/lib.rs
shannonwells: This has to stay here.

shannonwells (commented):
I'll review more once we've hashed out some other details.

pallets/msa/src/benchmarking.rs
```diff
@@ -195,6 +195,29 @@ benchmarks! {
 		assert!(Msa::<T>::get_provider_registry_entry(ProviderId(provider_msa_id)).is_some());
 	}
 
+	create_provider_via_governance {
+		let s in 0 .. T::MaxProviderNameSize::get();
+
+		let provider_name = (1 .. s as u8).collect::<Vec<_>>();
```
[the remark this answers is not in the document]
rlaferla: `create_provider()` also uses varies it. If we change it, I recommend changing it in both places.

pallets/msa/src/lib.rs
rlaferla: Which trait?  Do you mean the associated type Proposal (which is being used)?

pallets/msa/src/lib.rs
rlaferla: removed

pallets/msa/src/lib.rs
rlaferla: It's now `propose_to_be_provider() `

pallets/msa/src/lib.rs
rlaferla: It's now on the two extrinsics.

pallets/msa/src/weights.rs
rlaferla: I only checked in the MSA weights because two new benchmarks were added to it.  Jenkins created the rest.  I will work with @shannonwells to remove them.

pallets/msa/src/lib.rs
rlaferla: Fixed

pallets/msa/src/lib.rs
rlaferla: fixed

pallets/msa/src/lib.rs
rlaferla: which one?

pallets/msa/src/lib.rs
rlaferla: removed

pallets/msa/src/lib.rs
rlaferla: removed

pallets/msa/src/lib.rs
enddynayn: can we please rename this to create_provider

pallets/msa/src/lib.rs
```diff
@@ -161,6 +180,12 @@ pub mod pallet {
 		/// calculated value.
 		#[pallet::constant]
 		type MaxSignaturesStored: Get<Option<u32>>;
+
+		/// The origin that is allowed to create providers
+		type CreateProviderOrigin: EnsureOrigin<Self::RuntimeOrigin, Success = Self::AccountId>;
+
+		/// The origin that is allowed to create providers via governance
+		type CreateProviderViaGovernanceOrigin: EnsureOrigin<Self::RuntimeOrigin>;
```
enddynayn, line 188: I think we should consider adding feature flags here and on the extrinsic so that they do not appear in the Frequency production if not needed.

rlaferla: If we are able to conditionally compile the extrinsic based on the `frequency` feature flag, we can remove the `CreateProviderOrigin` type.  However, we'd have to also verify that the removal of the extrinsic doesn't break any other integrations, etc..

wilwade: Hmm... We have a rococo only pallet (the sudo pallet), but I don't think we want to have an extrinsic skipped in mainnet. The `pallet::call_index` makes it at least a possibility to do however.

That said, if we conditionally compiled it, we would need to do separate benchmarks or modify the generated benchmarks which is a bad idea.

Finally, this is a short-term fix (see Long Term: User Journey in the issue: https://github.com/LibertyDSNP/frequency/issues/321)

rlaferla: This was not possible but more importantly, we are planning to keep `create_provider() `in the future.

pallets/msa/src/lib.rs
```diff
@@ -161,6 +178,12 @@ pub mod pallet {
...
```
enddynayn, line 186: These can be named the same. CreateProviderOrigin since they run on different runtimes.

rlaferla: What do you suggest?

rlaferla: We cannot because governance origins cannot use `AccountId`.

enddynayn: Maybe we can pair.

shannonwells (changes requested):
Events do need to be deposited only inside of extrinsics and the event needs to be different for each. I know it will duplicate some code, but that's where those calls belong. We would especially want to know if somehow the `create_provider` extrinsic "escaped" to mainnet.

shannonwells (approved):
LGTM

PR #1036, opened 2023-02-06
Implement paginated extrinsics

### LINKED ISSUE #926: [Private Graph] Create extrinsic single page generic data
Create an extrinsic to Create_or_update a paginated storage.

### Inputs:
- msa_id: MessageSourceId
- schema_id: SchemaId
- page: StoragePage
```

struct StoragePage {
  number: PageNumber,
  data: vec<u8>,
  prev_hash: Option<[u8; 32]> // or T::Hash
}
```

The data should be store under a child tree of the msa and map between `(schemaId,pageNumber`)  to `vec<u8>`

### Acceptance Criteria:
- a user should be able to create a new or update an existing page for a Paginated Storage
- unit tests and integration tests
- benchmarks

More details in [here](https://github.com/LibertyDSNP/frequency/pull/900/files)

Noes:
ping @aramikm for child tree branch

JoeCap08055:
# Goal
The goal of this PR is to implement extrinsics for upserting/removing pages in Paginated Stateful Storage.

Closes #926

# Checklist
- [ ] Chain spec updated
- [ ] Design doc(s) updated
- [x] Tests added
- [x] Benchmarks added
- [x] Weights updated


common/primitives/src/stateful_storage.rs
```diff
@@ -0,0 +1,2 @@
+/// PageId is the unique identifier for a Page in Stateful Storage
+pub type PageId = u8;
```
aramikm, line 2: It's less likely that we might have pages more than 256 but I think to future proof this for groups let's change it to `u16`

pallets/stateful-storage/Cargo.toml
```diff
@@ -32,26 +32,11 @@ sp-std = { git = "https://github.com/paritytech/substrate", default-features = f
...
-  'sp-std/std',
-  'sp-core/std',
-  'sp-io/std',
-  'sp-runtime/std',
-  'frame-support/std',
-  'frame-system/std',
-  'frame-benchmarking/std',
-  'common-primitives/std',
-]
+runtime-benchmarks = ["frame-benchmarking/runtime-benchmarks", "frame-support/runtime-benchmarks", "frame-system/runtime-benchmarks", "common-primitives/runtime-benchmarks"]
```
aramikm, line 40: We should use the same 1 item per line styling, It's easier to see the diffs

pallets/stateful-storage/src/lib.rs
```diff
@@ -260,17 +263,94 @@ pub mod pallet {
...
+				DelegatorId(state_owner_msa_id),
+				schema_id,
+				current_block,
+			)
+			.map_err(|_| Error::<T>::UnAuthorizedDelegate)?;
+
+			let schema_key: StatefulPageKeyPart = schema_id.encode();
+			let page_key: StatefulPageKeyPart = page_id.encode();
+
+			StatefulChildTree::write(&state_owner_msa_id, &[schema_key, page_key], payload);
```
aramikm, line 304: I think you need to wrap your payload inside `Page<T::MaxPaginatedPageSizeBytes)::from(payload.into())` before storing

pallets/stateful-storage/src/types.rs
```diff
@@ -73,6 +72,7 @@ impl<PageDataSize: Get<u32>> Page<PageDataSize> {
 
 		let mut updated_page_buffer = Vec::with_capacity(parsed.page_size);
 		let mut add_buffer = Vec::new();
+		log::info!("parsed page len: {:?}", parsed.page_size);
```
aramikm, line 75: should be `debug`?

common/primitives/src/stateful_storage.rs
JoeCap08055: Done.

pallets/stateful-storage/src/types.rs
JoeCap08055: Meant to remove that.

pallets/stateful-storage/Cargo.toml
JoeCap08055: Weird; must have been some auto-formatter hook. Fixed.

pallets/stateful-storage/src/lib.rs
```diff
@@ -260,17 +263,94 @@ pub mod pallet {
...
+		) -> DispatchResult {
 			ensure!(
 				payload.len() as u32 <= T::MaxPaginatedPageSizeBytes::get(),
 				Error::<T>::PageExceedsMaxPageSizeBytes
 			);
+			ensure!(
+				page_id as u32 <= T::MaxPaginatedPageId::get(),
+				Error::<T>::PageIdExceedsMaxAllowed
+			);
+			let provider_key = ensure_signed(origin)?;
```
enddynayn, line 281: let provider_key = ensure_signed(origin)?;
I would move this check to the top since it checks that this transaction's origin is signed.

pallets/stateful-storage/src/lib.rs
```diff
@@ -260,17 +263,94 @@ pub mod pallet {
...
+			let provider_msa_id = T::MsaInfoProvider::ensure_valid_msa_key(&provider_key)
+				.map_err(|_| Error::<T>::InvalidMessageSourceAccount)?;
+
+			let schema = T::SchemaProvider::get_schema_by_id(schema_id);
```
enddynayn, line 285: Code golf:
	```
	let schema = T::SchemaProvider::get_schema_by_id(schema_id).amp_err (|_|  Error::<T>::InvalidSchemaId)?;
	```

Doing this will allow you to get rid of unwrap() below and do:
```
ensure!(
				schema.payload_location == PayloadLocation::Paginated,
				Error::<T>::SchemaPayloadLocationMismatch
			);
```

pallets/stateful-storage/src/lib.rs
```diff
@@ -260,17 +263,94 @@ pub mod pallet {
...
+		) -> DispatchResult {
+			ensure!(
+				page_id as u32 <= T::MaxPaginatedPageId::get(),
+				Error::<T>::PageIdExceedsMaxAllowed
+			);
+			let provider_key = ensure_signed(origin)?;
+			let provider_msa_id = T::MsaInfoProvider::ensure_valid_msa_key(&provider_key)
+				.map_err(|_| Error::<T>::InvalidMessageSourceAccount)?;
+
+			let schema = T::SchemaProvider::get_schema_by_id(schema_id);
```
enddynayn, line 329: Seems like this block of code can be refactored into a helper method.  Both extrinsic seem to be doing these checks. By moving into a helper method it allows for it to be unit tested.


```
fn ensure_valid_schema_something() {
		let schema = T::SchemaProvider::get_schema_by_id(schema_id);
			ensure!(schema.is_some(), Error::<T>::InvalidSchemaId);
			ensure!(
				schema.unwrap().payload_location == PayloadLocation::Paginated,
				Error::<T>::SchemaPayloadLocationMismatch
			);

			let current_block = frame_system::Pallet::<T>::block_number();
			T::SchemaGrantValidator::ensure_valid_schema_grant(
				ProviderId(provider_msa_id),
				DelegatorId(state_owner_msa_id),
				schema_id,
				current_block,
			)
			.map_err(|_| Error::<T>::UnAuthorizedDelegate)?;
			....
			}
```

enddynayn:
Looking great!

pallets/stateful-storage/src/lib.rs
```diff
@@ -260,17 +263,94 @@ pub mod pallet {
...
+				schema_id,
+				current_block,
+			)
+			.map_err(|_| Error::<T>::UnAuthorizedDelegate)?;
+
+			let schema_key: StatefulPageKeyPart = schema_id.encode();
+			let page_key: StatefulPageKeyPart = page_id.encode();
+
+			StatefulChildTree::write(&state_owner_msa_id, &[schema_key, page_key], payload);
+			Self::deposit_event(Event::PaginatedPageUpdated {
```
enddynayn, line 305: This extrinsic name contains `upsert` in the name. So, perhaps, we should consider calling this event PaginatedPageUpsert?

pallets/stateful-storage/src/lib.rs
```diff
@@ -260,17 +263,94 @@ pub mod pallet {
...
+				schema_id,
+				current_block,
+			)
+			.map_err(|_| Error::<T>::UnAuthorizedDelegate)?;
+
+			let schema_key = schema_id.encode();
+			let page_key = page_id.encode();
+
+			StatefulChildTree::kill(&state_owner_msa_id, &[schema_key, page_key]);
+			Self::deposit_event(Event::PaginatedPageRemoved {
```
enddynayn, line 349: To follow the CRUD naming conventions, perhaps we call this event Event::PaginatedPageDeleted?

pallets/stateful-storage/src/benchmarking.rs
```diff
@@ -32,13 +35,13 @@ benchmarks! {
 		let s in 1 .. T::MaxItemizedBlobSizeBytes::get()- 1;
 		let provider_msa_id = 1u64;
 		let delegator_msa_id = 2u64;
-		let schema_id = 1u16;
+		let schema_id = ITEMIZED_SCHEMA;
 		let caller: T::AccountId = whitelisted_caller();
 		let payload = vec![0u8; s as usize];
 
-		assert_ok!(create_schema::<T>(PayloadLocation::Itemized)); // TODO change this to itemized
+		// assert_ok!(create_schema::<T>(PayloadLocation::Itemized)); // TODO change this to itemized
```
saraswatpuneet, line 42: we can delete this and also remove the create_schema helper function looks like its not used?

saraswatpuneet (commented):
lgtm [including comments from reviewers]

pallets/stateful-storage/src/lib.rs
JoeCap08055: But that method doesn't return a `Result`, it returns an `Option`. But you got me thinking--a little checking, and I can use `.ok_or()` instead of `.map_err()`. Thanks.

pallets/stateful-storage/src/lib.rs
JoeCap08055: Nice idea; I like it. Actually I think the same treatment can be applied to most of the checks in the extrinsics for this pallet.

pallets/stateful-storage/src/lib.rs
saraswatpuneet: We can either create a chore for this if we want to handle this once all pieces are in ? But since there are no downstream consumers as of yet, I am ok with either or

pallets/stateful-storage/src/lib.rs
JoeCap08055: I've refactored this into a `check_origin` method that also gets the `MessageSourceId`. But I'm not calling it first thing because that requires a DB read; I figured better to check the payload first that doesn't require any DB reads.

pallets/stateful-storage/src/lib.rs
```diff
@@ -290,4 +336,33 @@ pub mod pallet {
 	}
 }
 
-impl<T: Config> Pallet<T> {}
+impl<T: Config> Pallet<T> {
+	fn check_origin(origin: OriginFor<T>) -> Result<MessageSourceId, DispatchError> {
```
enddynayn, line 340: I see what you did here. Though it seems that ensure_valid_msa_key contributes to checking the origin, they are two different things.

When one checks the origin, it simply checks that the origin enum is a `Signed` origin. It does not do any DB reads. So it would be best not to wrap these together or change the name of this function. I do not have any recommendations for naming at the moment.

pallets/stateful-storage/src/lib.rs
JoeCap08055: Take a look at the latest commit; refactored function `check_origin` does both signature verification AND looks up MSA (incurs DB read). If we really want signature verification to come first ALWAYS, I can rework it a little.

pallets/stateful-storage/src/lib.rs
```diff
@@ -290,4 +336,33 @@ pub mod pallet {
...
+		payload_location: PayloadLocation,
+	) -> DispatchResult {
+		let schema =
+			T::SchemaProvider::get_schema_by_id(schema_id).ok_or(Error::<T>::InvalidSchemaId)?;
+		ensure!(
+			schema.payload_location == payload_location,
+			Error::<T>::SchemaPayloadLocationMismatch
+		);
+
+		let current_block = frame_system::Pallet::<T>::block_number();
```
enddynayn, line 359: I wonder if removing this call is ok:
```
let schema =
			T::SchemaProvider::get_schema_by_id(schema_id).ok_or(Error::<T>::InvalidSchemaId)?;
		ensure!(
			schema.payload_location == payload_location,
			Error::<T>::SchemaPayloadLocationMismatch
		);
```

It might be that we do not need it if you can grant that schema implies that the grant exists?

```
T::SchemaGrantValidator::ensure_valid_schema_grant(
			ProviderId(provider_msa_id),
			DelegatorId(state_owner_msa_id),
			schema_id,
			current_block,
		)
		```

File: pallets/stateful-storage/src/lib.rs (hunk 11 above)
Comment (JoeCap08055) [reply pallets/stateful-storage/src/lib.rs:340]: Gotcha.

File: pallets/stateful-storage/src/lib.rs (hunk 12 above)
Comment (JoeCap08055) [reply pallets/stateful-storage/src/lib.rs:359]: Problem is we need to verify the payload location matches. That information only exists on the schema itself, we can't get or infer it from the grant.

File: pallets/stateful-storage/src/benchmarking.rs [hunk 13]
```diff
@@ -47,6 +46,26 @@ benchmarks! {
 		assert!(page_result.data.len() > 0);
 	}
 
+	upsert_page {
```
enddynayn, line 49: Nice. I am guessing your updating the weights on Jenkins and weighting for them to complete to wire this up?

pallets/stateful-storage/src/lib.rs
enddynayn: ahhh, yeah your right.

pallets/stateful-storage/src/benchmarking.rs
JoeCap08055: I'm running benchmarks locally to check, but I think we're holding off on Jenkins and weighting until just before we merge the whole `private-graph-main` branch to `main`

pallets/stateful-storage/src/benchmarking.rs
JoeCap08055: Hmmm... okay, I guess `create_schema` is necessary when actually running benchmarks (since it just failed on me locally). Updating...

pallets/stateful-storage/src/benchmarking.rs
enddynayn: It might not be a bad idea to generate this and check them since you already created the benchmark.

pallets/stateful-storage/src/mock.rs
```diff
@@ -160,45 +170,52 @@ impl<BlockNumber> SchemaGrantValidator<BlockNumber> for SchemaGrantValidationHan
...
-		// 	})
-		// }
-
-		Some(SchemaResponse {
-			schema_id,
-			model: r#"schema"#.to_string().as_bytes().to_vec(),
-			model_type: ModelType::AvroBinary,
-			payload_location: PayloadLocation::OnChain,
-		})
+		match schema_id {
```
enddynayn, line 175: I would avoid mocking these out.

saraswatpuneet: Agree with enddy, ideally  that would give us a complete roundtrip similar to main runtime

pallets/stateful-storage/src/lib.rs
JoeCap08055: Done.

pallets/stateful-storage/src/lib.rs
JoeCap08055: OK, refactored once more. Should be good now.

pallets/stateful-storage/src/benchmarking.rs
JoeCap08055: Fixed.

pallets/stateful-storage/src/lib.rs
JoeCap08055: Refactored.

pallets/stateful-storage/src/mock.rs
JoeCap08055: Hmmm... isn't the point of unit testing to only execute the unit under test and mock all external dependencies? (btw, this is how `pallet_messages` is doing it; I just copied the approach). Though personally I think we could benefit from test-specific implementations of `SchemaBenchmarkHelper` and `SchemaProvider` that kept track of created schemas using a BTreeMap or HashMap, so then we wouldn't have to do this hokey stuff. But I don't like the idea of taking a dependency on the Schemas pallet for unit testing.

saraswatpuneet: Right, then we have to agree all such tests should live in integration tests ? I think we can definitely investigate this as a spike, since many other pallets would benefit from it like messages pallet which depends on schema as well as msa etc.

aramikm (approved):
Approved! I'll take care of outstanding issues and questions in my PR https://github.com/LibertyDSNP/frequency/pull/1041 to keep this moving.

Issue #815, opened 2022-12-19
[Private graph] exposing custom rpc to generic storage based on schema 


p5150j:
Expose a custom rpc to return all generic items stored for a schema. Same RPC exposes paginated and itemized storages.

### input
- msa_id: MessageSourceId
- schema_id: Schema_id

## Output
- output: vec<vec<u8>>

## Acceptance Criteria:
- returns all items stored under msa and schema_id for both storage types
- unit tests and integration tests

p5150j:
- new rpc endpoint for passing in the msa pris to get a list of the pris

p5150j:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Blockchain-Team-638f7004a2c37722fe2850e8/issues/LibertyDSNP/frequency/815?planning-poker) with Zenhub @aramikm @JoeCap08055 @saraswatpuneet

PR #1041, opened 2023-02-07
Stateful-pallet: added rpc support to get pages

aramikm:
# Goal
The goal of this PR is to implement a unified RPC endpoint to get Itemized and paginated stored items.

Closes issue #815 

# Discussion
- Same RPC exposes all Itemized and Paginated Items (we might want to filter them in future but no required usecase currently)
- Modifies Child tree to use TwoX hash instead of identity
- Clean ups

# Checklist
- [x] Tests added


pallets/stateful-storage/src/lib.rs
```diff
@@ -274,20 +277,84 @@ pub mod pallet {
...
+							},
+						}
+					}
+				},
+				PayloadLocation::Paginated => {
+					let keys: Vec<StatefulPageKeyPart> = vec![schema_id.encode().to_vec()];
+					let mut pages: Vec<PageResponse> = StatefulChildTree::prefix_iterator::<
+						Page<T::MaxPaginatedPageSizeBytes>,
+					>(&msa_id, &keys[..])
+					.filter_map(|(key, v)| match <u16>::decode(&mut &key[..]) {
```
aramikm, line 326: u16 should be replaced with `PageId` type after merge

pallets/stateful-storage/src/lib.rs
```diff
@@ -274,20 +277,84 @@ pub mod pallet {
...
-			page_response
+impl<T: Config> Pallet<T> {
+	/// This function returns all the pages associated with `msa_id` and `schema_id` params
+	/// and support both types of storage types
+	/// Itemized - For this storage it parses all the items and returns each item with its index
+	/// Paginated - For this storage it finds all the pages and return each page with its pageId
+	///
+	/// [Warning] since this function iterates over all the potential keys it should never called
+	/// from runtime.
+	pub fn get_pages(msa_id: MessageSourceId, schema_id: SchemaId) -> Vec<PageResponse> {
```
aramikm, line 290: This does not return any error and just ignores any if happened and return only valid items.
Let me know if you think it makes sense to expose errors also.

pallets/stateful-storage/src/lib.rs
```diff
@@ -234,10 +240,7 @@ pub mod pallet {
 
 			match updated_page.is_empty() {
 				true => {
-					StatefulChildTree::kill::<Page<T::MaxItemizedPageSizeBytes>>(
-						&state_owner_msa_id,
-						&keys,
-					);
+					StatefulChildTree::kill(&state_owner_msa_id, &keys);
 					Self::deposit_event(Event::PageRemoved {
```
saraswatpuneet, line 244: :+1:

pallets/stateful-storage/src/lib.rs
saraswatpuneet: Imo: its ok like this, if request is legit! then they would expect results else empty , we follow this pattern at quite a few places, but not opposed to errors if they will actually be meaningful

saraswatpuneet (approved):
:+1:  can we put a TODO comment for changes needed to be done for post merge like

 <u16> to <PageId>

pallets/stateful-storage/src/lib.rs
JoeCap08055: I agree. No need to return an error unless there's something meaningful or critical.

saraswatpuneet (approved):
👍

PR #1043, opened 2023-02-08
Event Custom RPC 

### LINKED ISSUE #994: Custom RPC to return Filtered Events
To be able to reduce the scale decoding issues for consumers of events we need to reduce the events to only the ones that a consumer cares about to reduce the possibility of decoding issues for unwanted events.

wilwade:
# Goal
The goal of this PR is to offer an RPC endpoint with easier options for getting events and handling scale decoding.

It reduces the scale decoding issue to one of just the event enum/struct instead of needing to worry about the substrate wrapping of the events.

Closes #994

## Testing

1. Start instant sealing
2. Create schema, msa, or other things that trigger events
3. Snag the block hash and put in into this curl command (feel free to skip the `jq` call if you want non-formatted response):
```
curl --location --request POST 'localhost:9933' --header 'Content-Type: application/json' --data-raw '{
    "id":1,
    "jsonrpc":"2.0",
    "method": "frequency_getEvents",
    "params": ["'$HASH'"]
}' | jq
```

4. You should see the events and can scale decode them using the event structure and https://www.shawntabrizi.com/substrate-js-utilities/codec/

MSA Create Event SCALE
```
{
  "Msa.MsaCreated": {
    "msa_id": "u64",
    "key": "[u8; 32]"
  }
}
```

# To Do
- [x] Test js/api-augment updates

# Checklist
- [x] Updated js/api-augment.
- [x] Tests added


node/service/src/rpc/tests/mod.rs
```diff
@@ -0,0 +1,55 @@
...
+#[derive(Debug, TypeInfo, Clone, Encode, Decode, Eq, PartialEq)]
+enum FakeRuntime {
+	#[codec(index = 60)]
+	FakePallet(FakePalletEvents),
+}
+
+#[derive(Debug, TypeInfo, Clone, Encode, Decode, Eq, PartialEq)]
+enum FakePalletEvents {
+	#[codec(index = 7)]
+	FakeEvent { msa_id: u64 },
```
shannonwells, line 24: nitpick - it's generally better to test N>1 things, and 3 is more N-like than 2.   I would want to test that N>1 events and N>1 pallets output worked.  Though it does work fine now, so maybe for later.

shannonwells (approved):
Does the SCALE decoding information fall out of the RPC setup? How do you get it?

wilwade:
@shannonwells 
> Does the SCALE decoding information fall out of the RPC setup? How do you get it?

You still have to get that information from the Metadata or from hard coding it. The primary benefit is that you know what event it is before you decode it. This means that you could even ignore decoding the event if all you want to know is that the event happened. (or only decoding the events you care about the data from)

node/service/src/rpc/tests/mod.rs
wilwade: Good point, but agreed that it is a maybe later action. I want to get some feedback on it first.

aramikm (approved):
Ship it!

PR #1049, opened 2023-02-09
feat(generic-child-tree): Implement generic key decoding and generic …

### LINKED ISSUE #1042: [private graph] More generic child tree utility
### Description
Currently we have a child tree utility that allow us to interact with a child tree but it lacks some generic items such as
- Generic hash function
- Generic Key type

Not having a generic key type forces us to decode all keys to `Vec<u8>` first and then we can decode to the desired key type.

### AC
- Support Generic Reversible hash function
- Support Generic Key Type

### Additional info
for an example of how this can be done we can look at https://github.com/paritytech/substrate/blob/38c3e5ee10be760115a77a115ef0ac6ba381542b/frame/support/src/storage/generator/double_map.rs#L76

JoeCap08055:
# Goal
The goal of this PR is to implement generics for child tree key types and hash method.

Closes #1042 

# Checklist
- [ ] Chain spec updated
- [ ] Design doc(s) updated
- [ ] Tests added
- [ ] Benchmarks added
- [ ] Weights updated


pallets/stateful-storage/src/lib.rs
```diff
@@ -73,17 +73,16 @@ pub use weights::*;
...
+	use crate::{stateful_child_tree::StatefulChildTree, types::*};
 	use common_primitives::{
 		msa::{MessageSourceId, MsaLookup, MsaValidator, SchemaGrantValidator},
 		schema::{SchemaId, SchemaProvider},
 		stateful_storage::PageId,
 	};
 	use frame_support::pallet_prelude::*;
 
+	type Tree = StatefulChildTree<Identity>;
+
```
saraswatpuneet, line 85: :1st_place_medal:

aramikm: There might be a miscommunication on my part. We want `TwoX` for this pallet but for tests and maybe other pallets the `Identity` default makes sense

pallets/stateful-storage/src/stateful_child_tree.rs
```diff
@@ -1,37 +1,54 @@
...
 	/// data doesn't store under the given `key` `None` is returned.
-	pub fn try_read<V: Decode + Sized>(
+	pub fn try_read<K: IsStatefulKey, V: Decode + Sized>(
 		msa_id: &MessageSourceId,
-		keys: &[StatefulPageKeyPart],
+		keys: &K,
 	) -> Result<Option<V>, ()> {
 		let child = Self::get_child_tree(*msa_id);
-		let value = child::get_raw(&child, &Identity::hash(Self::concat_keys(keys).as_bytes_ref()));
+		let value = child::get_raw(&child, &H::hash(&keys.encode()).as_ref());
```
aramikm, line 51: I don't think this is going to work. We need to hash each key separately and concat those hashes. To be able to look them up by prefix

pallets/stateful-storage/src/lib.rs
JoeCap08055: Right; I was waiting for a decision before updating this. I'll change to `TwoX`.

pallets/stateful-storage/src/stateful_child_tree.rs
JoeCap08055: Ah, right. Duh. `hash(x) + hash(y) <> hash(x + y)`

pallets/stateful-storage/src/types.rs
```diff
@@ -1,9 +1,13 @@
 use codec::{Decode, Encode, MaxEncodedLen};
+use common_primitives::{schema::SchemaId, stateful_storage::PageId};
 use frame_support::pallet_prelude::*;
 use scale_info::TypeInfo;
 use sp_core::bounded::BoundedVec;
 use sp_std::{cmp::*, collections::btree_map::BTreeMap, fmt::Debug, prelude::*};
 
+pub type ItemizedKey = (SchemaId,);
```
aramikm, line 8: It seems like you have duplicated these types in lib.rs also. I would suggest to define all these types inside here and remove the duplicates defined in `lib.rs`

aramikm (approved):
Looks great!. Just need to rebase from `private-graph-main` since my changes got in.

pallets/stateful-storage/src/lib.rs
```diff
@@ -178,6 +178,10 @@ pub mod pallet {
 		PaginatedPageRemoved { msa_id: MessageSourceId, schema_id: SchemaId, page_id: PageId },
 	}
 
+	type ItemizedFullKey = (SchemaId,);
```
enddynayn, line 181: Can we please add some docs for this type,

pallets/stateful-storage/src/stateful_child_tree.rs
```diff
@@ -1,37 +1,238 @@
...
 
-/// Paginated Stateful data access utility
-pub struct StatefulChildTree;
-impl StatefulChildTree {
-	pub fn concat_keys(keys: &[StatefulPageKeyPart]) -> Vec<u8> {
-		let mut key = Vec::<u8>::new();
-		for k in keys {
-			let mut k2 = k.to_owned();
-			key.append(&mut k2);
+pub trait IsStatefulKey<H: MultipartKeyStorageHasher>:
```
enddynayn, line 40: This trait does more than check if something is a stateful key. I don't have any suggestions for naming at the moment.

pallets/stateful-storage/src/lib.rs
```diff
@@ -325,11 +320,12 @@ pub mod pallet {
 			msa_id: MessageSourceId,
 			schema_id: SchemaId,
 		) -> Page<T::MaxItemizedPageSizeBytes> {
-			let storage_key = &schema_id.encode()[..];
-			let keys = vec![storage_key.to_vec()];
-			let page_response =
-				StatefulChildTree::try_read::<Page<T::MaxItemizedPageSizeBytes>>(&msa_id, &keys)
-					.map_or_else(|_| Page::default(), |page| page.unwrap_or_default());
+			let key: ItemizedFullKey = (schema_id,);
+			let page_response = StatefulChildTree::<T::Hasher>::try_read::<
```
enddynayn, line 324: Can we call it like this?
StatefulChildTree::<T::Hasher>::try_read::<_, ...>(&msa_id, &key). unwrap_or_default()

pallets/stateful-storage/src/stateful_child_tree.rs
```diff
@@ -42,37 +243,50 @@ impl StatefulChildTree {
 	///
 	/// Allows getting all the keys having the same prefix
 	/// Warning: This should not be used from any on-chain transaction!
-	pub fn prefix_iterator<V: Decode + Sized>(
+	pub fn prefix_iterator<
+		V: Decode + Sized,
+		K: IsStatefulKey<H> + Sized,
```
enddynayn, line 248: Do you expect that different hashers are going to be used on the same type? I guess why not set this as a generic one level up?

pallets/stateful-storage/src/stateful_child_tree.rs
```diff
@@ -42,37 +243,50 @@ impl StatefulChildTree {
...
+		prefix_keys: &PrefixKey,
+	) -> Box<impl Iterator<Item = (K, V)>> {
 		let child = Self::get_child_tree(*msa_id);
-		ChildTriePrefixIterator::<_>::with_prefix(&child, Self::concat_keys(keys).as_bytes_ref())
+		let result = ChildTriePrefixIterator::<(Vec<u8>, V)>::with_prefix(
+			&child,
+			&prefix_keys.hash_prefix_only(),
+		)
+		.filter_map(|(k, v)| {
+			if let Ok(key) =
```
enddynayn, line 260: the match thing seems more of a rust way than an if statement. No strong opinion.

pallets/stateful-storage/src/stateful_child_tree.rs
```diff
@@ -1,37 +1,238 @@
...
+		let hash_len: usize = (Self::HashSize::get() * num_key_parts).into();
+		if x.len() < hash_len {
+			log::error!("Invalid reverse: hash length too short");
+			return &[]
+		}
+		&x[hash_len..]
+	}
+}
+/// 128-bit Blake2 hash.
+impl MultipartKeyStorageHasher for Blake2_128 {
```
enddynayn, line 24: nice

pallets/stateful-storage/src/lib.rs
JoeCap08055: Removed; after merge from pg-main, got `ItemizedKey` from `common_primitives::stateful_storage`

pallets/stateful-storage/src/types.rs
JoeCap08055: Done.

pallets/stateful-storage/src/stateful_child_tree.rs
JoeCap08055: Hmm, good point. It kind of grew. Maybe `MultipartKey` is a bettername?

JoeCap08055: Renamed to `MultipartKey`

pallets/stateful-storage/src/stateful_child_tree.rs
JoeCap08055: I'm not understanding what you're getting at... the hasher is a type parameter to `StatefulChildTree`, but it needs to make sure that the MultipartKey uses the same hasher. What am I missing?

pallets/stateful-storage/src/stateful_child_tree.rs
JoeCap08055: It's more than an `if`... it's an `if let`. Which is just syntactic sugar for a basic `match`.

pallets/stateful-storage/src/lib.rs
JoeCap08055: No... the return value is an `Option` wrapped in a `Result`. `unwrap_or_default()` on the Result may still yield a `None` value.

PR #1066, opened 2023-02-10
Chore/remove sudo storage #859

shannonwells:
# Goal
The goal of this PR is to create a runtime migration that:
* On **frequency-rococo/local**, _does nothing_.
* On **frequency mainnet**:
    * Deletes all storage associated with Sudo pallet:  The Key and the PalletVersion.
    * transfers all token from the Sudo account to Treasury and causes it to be reaped.
* Provides a best-practices template for future storage migrations. 

References #859 

# Discussion
I know this is long but please read carefully and critically since this is our first storage migration.

* Note that it's standard practice to remove these migrations in a future PR after the upgrade has been run on mainnet.
* I am not sure if it's necessary to use generics in the RemoveSudo struct.
* Please pay particular attention to anything that may affect testnet/local
* Please verify that my assumptions about what features are enabled for frequency-rococo in particular is correct

#### Local
You will need to have a local chain running from `main`, then check out this branch and run the commands shown. You may wish to do this by cloning frequency in a different folder from where you are running `try-runtime`.

Try both lines shown below for mainnet and testnet but with `--uri ws://localhost:9944` instead of pointing at a mainnet or testnet node.

#### Mainnet 
```shell
RUST_LOG=debug cargo run --features=frequency,try-runtime try-runtime --runtime target/debug/wbuild/frequency-runtime/frequency_runtime.compact.compressed.wasm --chain=frequency on-runtime-upgrade live --uri wss://1.rpc.frequency.xyz:443
```

The output for mainnet, which has `frequency` feature enabled, should be:
```
2023-02-10 13:25:53.510  INFO                 main frequency_runtime::remove_sudo: 🟢        Current Sudo account balance: free: 1000000000000, reserved: 0, fee_frozen: 0, misc_frozen: 0    
2023-02-10 13:25:53.511  INFO                 main frequency_runtime::remove_sudo: ❓       Sudo Key storage exists ===>  true    
2023-02-10 13:25:53.511  INFO                 main frequency_runtime::remove_sudo: ❓       Sudo PalletVersion storage exists ===>  true    
2023-02-10 13:25:53.515  INFO                 main frequency_runtime::remove_sudo: ❓       Sudo Key storage exists ===>  false    
2023-02-10 13:25:53.515  INFO                 main frequency_runtime::remove_sudo: ❓       Sudo PalletVersion storage exists ===>  false    
[...]

PR #1079, opened 2023-02-13
[WIP] Schema Grants

saraswatpuneet:
# Goal
The goal of this PR is to introduce concept of schema grants which are kind of acl per schema for fine grained controls especially needed in stateful storage operations

With @aramikm 

Closes 

# Discussion

# Checklist
- [ ] Chain spec updated
- [ ] Design doc(s) updated
- [ ] Tests added
- [ ] Benchmarks added
- [ ] Weights updated


runtime/common/src/constants.rs
```diff
@@ -132,6 +132,8 @@ parameter_types! {
 pub type SchemasMinModelSizeBytes = ConstU32<8>;
 /// The maximum length of a schema model (in bytes)
 pub type SchemasMaxBytesBoundedVecLimit = ConstU32<65_500>;
+/// The maximum number of grants allowed per schema
+pub type MaxSchemaGrantsPerSchema = ConstU16<1>;
```
saraswatpuneet, line 136: just one for now? either AppendOnly or SignatureRequired grants

common/primitives/src/schema.rs
```diff
@@ -35,6 +38,28 @@ pub enum PayloadLocation {
...
+	AppendOnly,
+	/// The metadata of this collection can be modified.
+	SignatureRequired,
+}
+
+impl Default for Grant {
+	fn default() -> Self {
+		Self::AppendOnly
+	}
+}
```
saraswatpuneet, line 57: Mostly for mocks/tests gets easier to read

runtime/common/src/constants.rs
saraswatpuneet: Or we allow more ?

common/primitives/src/macros.rs
```diff
@@ -0,0 +1,76 @@
...
+					.type_params(vec![TypeParameter::new("T", Some(meta_type::<$bitflag_enum>()))])
+					.composite(
+						Fields::unnamed()
+							.field(|f| f.ty::<$size>().type_name(stringify!($bitflag_enum))),
+					)
+			}
+		}
+	};
+}
+// pub(crate) use impl_codec_bitflags;
```
saraswatpuneet, line 76: $wrapper wraps a type $size that is used to store a bit flags enum $bitflag_enum and is used to generated derived types with basic features like encoding, decoding etc.

Used for bitflags (grants)

common/primitives/src/schema.rs
```diff
@@ -35,6 +38,28 @@ pub enum PayloadLocation {
...
+/// Support for up to 16 user-enabled features on a collection.
+#[bitflags]
+#[repr(u16)]
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+#[derive(Copy, Clone, RuntimeDebug, PartialEq, Eq, Encode, Decode, MaxEncodedLen, TypeInfo)]
+pub enum Grant {
+	/// Items in this collection are transferable.
+	AppendOnly,
+	/// The metadata of this collection can be modified.
+	SignatureRequired,
```
JoeCap08055, line 50: How is this going to work:
1. Schema with this property will ALWAYS require a signed encapsulated payload, or
2. Schema with this property may be called normally by owner or delegate, but ALSO may be used with a "one-time delegation" via a signed payload

If (2), then perhaps a different name, like `AllowSignedDelegation`, or something?

JoeCap08055: Also, the documentation comments need to be updated.

common/primitives/src/schema.rs
JoeCap08055: This seems dangerous... maybe qualify inside `#[cfg="test"]` or something?

pallets/schemas/src/lib.rs
```diff
@@ -278,9 +285,14 @@ pub mod pallet {
 			model: BoundedVec<u8, T::SchemaModelMaxBytesBoundedVecLimit>,
 			model_type: ModelType,
 			payload_location: PayloadLocation,
+			grants: BoundedVec<Grant, T::SchemaModelMaxBytesBoundedVecLimit>,
```
saraswatpuneet, line 288: Question: Do we really need a new extrinsic, since grants are optional, this Option should prevent any breaking changes if anyone is actually consuming this ?

common/primitives/src/schema.rs
saraswatpuneet: That will be our next story, the focus of this PR is just to update schemas pallet

saraswatpuneet:
@JoeCap08055  when we initially worked on schemas, we said permission are nothing but schemas delegation (publishing rights) ,  Grants we discussed to be called to schema level controls.
We can bring it up in parking lot tomorrow :)

common/primitives/src/schema.rs
```diff
@@ -35,6 +38,28 @@ pub enum PayloadLocation {
 	Paginated,
 }
 
+/// Support for up to 16 user-enabled features on a collection.
+#[bitflags]
+#[repr(u16)]
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+#[derive(Copy, Clone, RuntimeDebug, PartialEq, Eq, Encode, Decode, MaxEncodedLen, TypeInfo)]
+pub enum Grant {
```
aramikm: I agree with Joe to Not call these grants since these are directly associated with Schemas and they are more Properties or Attributes of that schema.

common/primitives/src/schema.rs
```diff
@@ -35,6 +38,28 @@ pub enum PayloadLocation {
...
+	/// Items in this collection are transferable.
+	AppendOnly,
+	/// The metadata of this collection can be modified.
+	SignatureRequired,
+}
+
+impl Default for Grant {
+	fn default() -> Self {
+		Self::AppendOnly
```
aramikm: I think AppendOnly as the default is a bit odd. Ideally the default should be the default of underlying type `u16` which is 0

pallets/schemas/src/lib.rs
aramikm: I think for backwards compatibility we will need a new extrinsic since the signature of the extrinsic in metadata will change

pallets/schemas/src/types.rs
```diff
@@ -18,4 +17,6 @@ where
 	pub model: BoundedVec<u8, MaxModelSize>,
 	/// The payload location
 	pub payload_location: PayloadLocation,
+	/// grants for the schema
+	pub grants: Grants,
```
aramikm: I think we will need to create a new Version of Schema to be able to migrate between old version to new.

aramikm (commented):
The main logic seems solid just a couple of comments and concerns:

- The Backwards compatibility issue (We probably need new extrinsic) 
- Migration of old Schemas to new ones (we will need this due to rococo and it will allow us to  experience a data migration for the first time)
- `Grants` name doesn't seem very suitable here.

pallets/schemas/src/types.rs
saraswatpuneet: Understood! yeah did not cross my mind we are deep in it now

saraswatpuneet:
I renamed Grant->SchemaSetting and Grants-> SchemaSettings to keep it simple

common/primitives/src/schema.rs
saraswatpuneet: For naming

Should we be explicit about who and what is being signed?

pallets/schemas/src/types.rs
aramikm: @JoeCap08055 we have to look into what is the preferred way in substrate 
here is an example
https://github.com/paritytech/substrate/blob/master/frame/membership/src/migrations/v4.rs

integration-tests/scaffolding/extrinsicHelpers.ts
```diff
@@ -179,11 +179,16 @@ export class ExtrinsicHelper {
...
     }
 
-    /** Schema Extrinsics */
+    /** Generic Schema Extrinsics */
     public static createSchema(keys: KeyringPair, model: any, modelType: "AvroBinary" | "Parquet", payloadLocation: "OnChain" | "IPFS"| "Itemized" | "Paginated"): Extrinsic {
         return new Extrinsic(() => ExtrinsicHelper.api.tx.schemas.createSchema(JSON.stringify(model), modelType, payloadLocation), keys, ExtrinsicHelper.api.events.schemas.SchemaCreated);
     }
 
+    /** Generic Schema Extrinsics */
+    public static createSchemaWithGrants(keys: KeyringPair, model: any, modelType: "AvroBinary" | "Parquet", payloadLocation: "OnChain" | "IPFS"| "Itemized" | "Paginated", grant: "AppendOnly"| "SignatureRequired"): Extrinsic {
```
JoeCap08055, line 188: nit: change name to `createSchemaWithSettings`

pallets/schemas/src/migrations.rs
```diff
@@ -0,0 +1,33 @@
...
+//! - Rename SchemaV2 to Schema and respective storage once migration is complete
+use super::*;
+use crate::types::{Schema, SchemaV2};
+use codec::{Decode, Encode};
+use common_primitives::schema::{SchemaId, SchemaSettings};
+use frame_support::pallet_prelude::Weight;
+
+pub fn migrate_schema_to_schema_v2<T: Config>() -> Weight {
+	let mut weight: Weight = Weight::zero();
+	let schema_count = <Schemas<T>>::iter().count();
```
JoeCap08055, line 15: Maybe more efficient to use `iter().enumerate()` here instead of `count()`? Would save having to re-fetch the schema later on.

pallets/schemas/src/lib.rs
```diff
@@ -174,6 +181,19 @@ pub mod pallet {
 		OptionQuery,
 	>;
 
+	/// Storage for SchemaV2
+	/// - Key: Schema Id
+	/// - Value: [`SchemaV2`](SchemaV2)
+	#[pallet::storage]
+	#[pallet::getter(fn get_schema_v2)]
+	pub(super) type SchemasV2<T: Config> = StorageMap<
```
JoeCap08055, line 189: Just a thought: If we rename the current/original `Schemas` storage to `SchemasV1` and call the new one `Schemas`, it will ensure that any existing references will use the new storage (and eliminate the need for post-migration renaming/code modification). (Will still need to delete the V1 storage post-migration, but that's all).

pallets/schemas/src/types.rs
saraswatpuneet: There are few different kinds of migrations, like runtime etc. since we are simply translating our schema to new type it will just be a simple hook 

Example for such https://github.com/paritytech/polkadot/pull/4545

pallets/schemas/src/lib.rs
```diff
@@ -263,6 +283,41 @@ pub mod pallet {
...
+		/// * [`Error::InvalidSchema`] - Schema is malformed in some way
+		/// * [`Error::SchemaCountOverflow`] - The schema count has exceeded its bounds
+		#[pallet::call_index(2)]
+		#[pallet::weight(T::WeightInfo::create_schema(model.len() as u32 + settings.len() as u32))]
+		pub fn create_schema_with_settings(
+			origin: OriginFor<T>,
+			model: BoundedVec<u8, T::SchemaModelMaxBytesBoundedVecLimit>,
+			model_type: ModelType,
+			payload_location: PayloadLocation,
+			settings: BoundedVec<SchemaSetting, T::MaxSchemaSettingsPerSchema>,
```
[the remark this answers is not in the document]
saraswatpuneet: That would break the metadata so we would rather create a new extrinsic then forcing downstream to update metadata if they arent using stateful related stuff when we upgrade the chain and also we have a clear visibility.

open to suggestion

pallets/schemas/src/migrations.rs
```diff
@@ -0,0 +1,23 @@
...
+		weight = weight.saturating_add(T::DbWeight::get().reads_writes(1, 1));
+		Some(Schema {
+			model_type: old_schema.model_type,
+			model: old_schema.model,
+			payload_location: old_schema.payload_location,
+			settings: SchemaSettings::all_disabled(),
+		})
+	});
+	weight
+}
```
saraswatpuneet, line 23: @aramikm  so since we are simply translating values, this is some places where parity simply uses on_runtime_upgrade hook to run migrations

pallets/schemas/src/lib.rs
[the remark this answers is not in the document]
saraswatpuneet: And keeping option then it just behaves like original one, keeping it simple

pallets/schemas/src/lib.rs
```diff
@@ -193,6 +200,13 @@ pub mod pallet {
 		}
 	}
 
+	#[pallet::hooks]
+	impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {
+		fn on_runtime_upgrade() -> Weight {
+			migrations::migrate_schemas_with_additional_settings::<T>()
+		}
+	}
+
```
saraswatpuneet, line 209: This i havent not tested yet, but thats the plan, to do some upgrades and check if migration works, then we can add some try-runtime flags to have a test case

pallets/schemas/src/migrations.rs
```diff
@@ -0,0 +1,23 @@
...
+//! Note: Post migration, this file should be deleted.
+use crate::{types::Schema, *};
+use common_primitives::schema::SchemaSettings;
+use frame_support::pallet_prelude::Weight;
+
+pub fn migrate_schemas_with_additional_settings<T: Config>() -> Weight {
+	let mut weight: Weight = Weight::zero();
+	//translate_values of Schemas storage add settings: SchemaSettings::all_disabled()
+	//to all schemas
+	<Schemas<T>>::translate_values(|old_schema: Schema<T::SchemaModelMaxBytesBoundedVecLimit>| {
```
aramikm, line 13: I don't see us using different Schema types for old and new ones. How does substrate know how to decode old values? I'm not sure if this will work

aramikm: I also like enwrapping the migration inside a mod version 

```rust
mod v1{
    // migration logic
}
```
example https://github.com/paritytech/substrate/blob/master/frame/contracts/src/migration.rs#L87

saraswatpuneet: In general through any language or db i expect decoders to just populate fields that exists, hoping its the same here as I expect this to working similar to going through 
https://github.com/paritytech/polkadot/pull/4545

saraswatpuneet: I also like the ides of modulating it, will do after these arduous upgrade testing

aramikm: but even in that example there is 2 versions for **SessionInfo**
```

<Pallet<T> as Store>::Sessions::translate_values(|old: primitives::**v1**::SessionInfo| {
		vs += 1;
		Some(primitives::**v2**::SessionInfo::from(old))
```

saraswatpuneet:
A full upgrade test to make sure we have successful migration

https://docs.google.com/document/d/1_CrTeVKh_sq_twsB60WNMd1SKnyWxOL_AW6I9a-5XKI/edit?usp=sharing

JoeCap08055:
> A full upgrade test to make sure we have successful migration
> 
> https://docs.google.com/document/d/1_CrTeVKh_sq_twsB60WNMd1SKnyWxOL_AW6I9a-5XKI/edit?usp=sharing

As part of the validation, should verify that all schemas were upgraded... the document only shows 1 schema queried after upgrade (maybe just not shown in the doc)

common/primitives/src/schema.rs
```diff
@@ -48,6 +67,8 @@ pub struct SchemaResponse {
 	pub model_type: ModelType,
 	/// The payload location
 	pub payload_location: PayloadLocation,
+	/// grants for the schema
+	pub settings: Vec<SchemaSetting>,
```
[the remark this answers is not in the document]
saraswatpuneet: @aramikm  i am a bit divided here, to support SchemaSetting directly we would definitely need a custom serializer and de for Bitflags<SchemaSetting, u16> given that what essentially SchemaSettings is

I was under the impression to return actual types as vec to prevent doing all above 

What you suggest, or i can set a ticket out to fix this?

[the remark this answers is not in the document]
saraswatpuneet: Another thing is in our users mind they are setting some values and exactly what they get back I think that will keep consistency than serializing flags what you think?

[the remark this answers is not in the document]
saraswatpuneet: Looks like this never translates back I get nil so this needs fix before merge 

Will do that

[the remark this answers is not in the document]
aramikm: I think this is fine. It might be confusing if the input and output is not consistent from user perspective as you pointed out.

pallets/schemas/src/lib.rs
```diff
@@ -193,6 +200,13 @@ pub mod pallet {
 		}
 	}
 
+	#[pallet::hooks]
+	impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {
+		fn on_runtime_upgrade() -> Weight {
+			migrations::v1::migrate_schemas_with_additional_settings::<T>()
```
aramikm, line 206: it does seem like we will need a cleanup ticket to remove all these runtime upgrade logic

pallets/schemas/src/migrations.rs
aramikm: Is there any possibly that a runtime migration hook might get triggered more than 1 time?

aramikm (approved):
Looks great to me. Let's make sure somebody else with migration experience also checks this before merge.

pallets/schemas/src/lib.rs
saraswatpuneet: yep https://github.com/LibertyDSNP/frequency/issues/1106

JoeCap08055 (approved):

pallets/schemas/src/migrations.rs
shannonwells: For safety reasons, a migration hook should be designed to be idempotent even if you expect to remove it before the next runtime upgrade.  There should be some check like for if it's OldSchema or not before calling `translate_values`.

pallets/schemas/src/lib.rs
shannonwells: try-runtime is meant to be used for testing, esp. storage migrations.  If you have a look at the [remove sudo storage PR](https://github.com/LibertyDSNP/frequency/pull/1066) it has a bunch of logging in the try-runtime, but logging is a good idea anyway in case something goes wrong.

pallets/schemas/src/tests.rs
```diff
@@ -350,3 +350,27 @@ fn dsnp_broadcast() {
 	);
 	assert_ok!(result);
 }
+
+#[test]
+fn create_schema_with_settings_should_work() {
```
shannonwells, line 355: Error tests?

pallets/schemas/src/lib.rs
```diff
@@ -335,6 +386,23 @@ pub mod pallet {
 
 			Ok(next)
 		}
+
+		fn validate_schema_model(
```
shannonwells, line 390: 🏅

shannonwells (commented):
Please see comments

pallets/schemas/src/migrations.rs
saraswatpuneet: @shannonwells I believe this `|old_schema: OldSchema<T::SchemaModelMaxBytesBoundedVecLimit>|` cast will prevent translation of any schemas that dont fit the OldSchema model since any new schemas wont decode to OldSchema type

Or you mean its safe to` iter -> check -> update?`

pallets/schemas/src/tests.rs
saraswatpuneet: Since schema dont directly check settings but consumers of schema does, I have added more tests and negative tests for SchemaSetting in stateful pallet

pallets/stateful-storage/src/lib.rs
```diff
@@ -430,55 +428,96 @@ impl<T: Config> Pallet<T> {
...
+		// Ensure that the schema allows signed payloads.
+		// If so, calling extrinsic must be of signature type.
+		if schema.settings.contains(&SchemaSetting::SignatureRequired) {
+			ensure!(is_payload_signed, Error::<T>::SchemaNotSupported);
+		}
+
+		// Ensure that the schema does not allow deletion for AppendOnly SchemaSetting.
+		if schema.settings.contains(&SchemaSetting::AppendOnly) {
+			ensure!(!is_deleting, Error::<T>::SchemaNotSupported);
+		}
```
saraswatpuneet, line 508: @aramikm  what you think, wanted to keep it simple, given we will circle back at validation

runtime/frequency/src/lib.rs
```diff
@@ -1100,6 +1115,38 @@ cumulus_pallet_parachain_system::register_validate_block! {
...
+	fn post_upgrade(_state: Vec<u8>) -> Result<(), &'static str> {
+		let weight = pallet_schemas::migrations::v1::post_migrate_schemas_to_v1::<Runtime>();
+		log::info!("post_upgrade weight: {:?}", weight);
+		Ok(())
+	}
+}
+// ==============================================
+//       END RUNTIME STORAGE MIGRATION: Schemas
+// ==============================================
+
```
saraswatpuneet, line 1149: @shannonwells  tested with this and works fine, i see schemas migrated

pallets/schemas/src/migrations.rs
aramikm: I think she means checking even before iteration. One thing I've seen some pallets do it that they have a separate storage to keep track of current version of migration and update it after the migration is complete. So if the process runs again it would check with that version and not run if it's the latest version. There might be other ways also to achieve this.

pallets/stateful-storage/src/benchmarking.rs
```diff
@@ -11,7 +11,7 @@ use frame_system::RawOrigin;
 use sp_core::bounded::BoundedVec;
 use stateful_child_tree::StatefulChildTree;
 
-pub const ITEMIZED_SCHEMA: SchemaId = 100; // keep in sync with mock.rs. TODO: refactor
+pub const ITEMIZED_APPEND_ONLY_SCHEMA: SchemaId = 100; // keep in sync with mock.rs. TODO: refactor
```
aramikm, line 14: these are refactored. Please make sure to use the `test_common.rs` in the pallet.

pallets/schemas/src/migrations.rs
saraswatpuneet: I see, yes I remember seeing that around, will redo this part

pallets/schemas/src/lib.rs
```diff
@@ -143,6 +150,7 @@ pub mod pallet {
 
 	#[pallet::pallet]
 	#[pallet::generate_store(pub(super) trait Store)]
+	#[pallet::storage_version(SCHEMA_STORAGE_VERSION)]
```
saraswatpuneet, line 153: Storage version will be checked at the time of migration and anything older will be translated while newer ignored

aramikm: I'm curious how this works under the hood. Just want to make sure there are no surprises

aramikm (approved):
Looks good! Ship it ✅

PR #1090, opened 2023-02-14
feat(race-prevention-hash): Add page content hash to prevent data rac…

### LINKED ISSUE #1023: [Private graph] race condition prevention
### Definition
To remove race condition of stateful storage changes we are going to send the hash of current state in itemized and paginated requests.

### AC
- Modify Itemized extrinsic to get a hash value and compare it with the hash of current page and fail if the don't match
- Modify Paginated extrinsic to get the hash value (ignore comparing for now since we will need to read the page on chain) but this will allow us to add that later.
- Return hash with the storage rpc and add to page update event (validate against current hash)

### Question
- What hash function we should use and what should be the input param type be ? It can be a multihash type to support multiple hashes but ideally ti should only be 32 bytes hash algorithms. Please refer to @wesbiggs for that decision.

JoeCap08055:
# Goal
The goal of this PR is to add content hash checking to Itemized and Paginated extrinsics and RPC responses, in order to prevent data race conditions from allowing invalid stateful data overwrites.

Closes #1023 

# Checklist
- [ ] Chain spec updated
- [ ] Design doc(s) updated
- [ ] Tests added
- [ ] Benchmarks added
- [ ] Weights updated


common/primitives/src/stateful_storage.rs
```diff
@@ -11,6 +11,8 @@ use utils::*;
 
 /// PageId is the unique identifier for a Page in Stateful Storage
 pub type PageId = u16;
+/// PageHash is the type/size of hash of the page content.
+pub type PageHash = u32;
```
aramikm, line 15: should we use `u32` or `[u8; 4]`?  Which one seems more compatible with other language implementations?

pallets/stateful-storage/Cargo.toml
```diff
@@ -30,29 +30,14 @@ sp-runtime = { git = "https://github.com/paritytech/substrate", default-features
 sp-std = { git = "https://github.com/paritytech/substrate", default-features = false, branch = "polkadot-v0.9.36" }
 # Frequency related dependencies
 common-primitives = { default-features = false, path = "../../common/primitives" }
+twox-hash = { version = "1.6.3", default-features = false, features = ["digest_0_10"] }
```
aramikm, line 33: nit: This should be above `# Frequency related dependencies` since it's not a frequency crate

pallets/stateful-storage/Cargo.toml
```diff
@@ -30,29 +30,14 @@ sp-runtime = { git = "https://github.com/paritytech/substrate", default-features
...
-  'sp-std/std',
-  'sp-core/std',
-  'sp-io/std',
-  'sp-runtime/std',
-  'frame-support/std',
-  'frame-system/std',
-  'frame-benchmarking/std',
-  'common-primitives/std'
-]
+runtime-benchmarks = ["frame-benchmarking/runtime-benchmarks", "frame-support/runtime-benchmarks", "frame-system/runtime-benchmarks", "common-primitives/runtime-benchmarks"]
```
aramikm, line 41: another nit

pallets/stateful-storage/src/lib.rs
```diff
@@ -170,15 +170,37 @@ pub mod pallet {
...
+			schema_id: SchemaId,
+			page_id: PageId,
+			prev_content_hash: PageHash,
+			curr_content_hash: PageHash,
+		},
+		PaginatedPageDeleted {
+			msa_id: MessageSourceId,
+			schema_id: SchemaId,
+			page_id: PageId,
+			prev_content_hash: PageHash,
```
aramikm, line 202: nit: To be consistent we should either include `prev_content_hash` for both `ItemizedPageDeleted` and `PaginatedPageDeleted` or for none of them.

common/primitives/src/stateful_storage.rs
saraswatpuneet: > should we use `u32` or `[u8; 4]`? Which one seems more compatible with other language implementations?

👍

common/primitives/src/stateful_storage.rs
```diff
@@ -35,6 +39,8 @@ pub struct ItemizedStoragePageResponse {
 	pub msa_id: MessageSourceId,
 	///  Schema id of the
 	pub schema_id: SchemaId,
+	/// Hash of the page content
+	pub content_hash: PageHash,
 	/// Items in a page
```
saraswatpuneet, line 44: Do we want to call it content_hash ?

saraswatpuneet (commented):
Good for me since comments covered more

Nice work 👍

common/primitives/src/stateful_storage.rs
JoeCap08055: Rather than block this PR, going to file a 1-point spike to answer this & also determine if the `TwoX32` hash is a valid selection given support in other languages.

pallets/stateful-storage/Cargo.toml
JoeCap08055: Fixed.

pallets/stateful-storage/Cargo.toml
JoeCap08055: `cargo add` rewrites this as a single line, FYI.
Fixed.

pallets/stateful-storage/src/lib.rs
JoeCap08055: Fair point. Done.

common/primitives/src/stateful_storage.rs
JoeCap08055: I don't have strong feelings... suggestions?

saraswatpuneet (approved):
Right before the demo 🔥

aramikm (approved):
Looks great! Ship it!

Issue #171, opened 2022-06-13
Capacity unstaking


enddynayn:
**Description** 
As a service provider, I want to be able to unstake token, so that I can use my tokens.

**Acceptance Criteria**:
- a thaw period should be [insert here the period]
- Account should reflect reserved amount to be reduced by unstaking amount
- Account free balance should increase by unstaking amount

p5150j:
@enddynayn Thoughts on closing this one out as its invalidated by #817 #818 #819 and #820 now?

wilwade:
Duplicate

PR #1105, opened 2023-02-16
Zero-out staking account if unstake to below minimum

shannonwells:
# Goal
The goal of this PR is to implement clearing the staking account if the unstake amount would put the account below the configured minimum.

In the process had to fix a bug when updating capacity and staking target accounts.

Related to #171 

# Discussion
Because it's possible that the actual unstaked amount could be more than what was requested, `StakinAccountDetails::decrease_by` must now return a value of the actual amount that was unstaked in order to correctly adjust the target staking account and the capacity.  The needed updates revealed the bug in `calculate_capacity_reduction` where the return value was the remaining amount, not the deducted amount. Dependent tests were updated.

# Checklist
- [ ] Chain spec updated
- [x] Tests added

## Not applicable
- Design doc(s) updated
- Benchmarks added
- Weights updated

### DIFF
--- a/pallets/capacity/src/lib.rs
+++ b/pallets/capacity/src/lib.rs
@@ -524,7 +524,7 @@ impl<T: Config> Pallet<T> {
 		let thaw_at =
 			current_epoch.saturating_add(T::EpochNumber::from(T::UnstakingThawPeriod::get()));
 
-		let unstake_result = staking_account.deduct(amount, thaw_at)?;
+		let unstake_result = staking_account.withdraw(amount, thaw_at)?;
 
 		Self::set_staking_account(&unstaker, &staking_account);
 
@@ -546,8 +546,8 @@ impl<T: Config> Pallet<T> {
 			capacity_details.total_tokens_staked,
 			capacity_details.total_available,
 		);
-		staking_target_details.deduct(amount, capacity_reduction);
-		capacity_details.deduct(capacity_reduction, amount);
+		staking_target_details.withdraw(amount, capacity_reduction);
+		capacity_details.withdraw(capacity_reduction, amount);
 
 		Self::set_capacity_for(target, capacity_details);
 		Self::set_target_details_for(unstaker, target, staking_target_details);
@@ -591,7 +591,7 @@ impl<T: Config> Nontransferable for Pallet<T> {
 		}
 	}
 
-	fn deduct(msa_id: MessageSourceId, amount: Self::Balance) -> Result<(), DispatchError> {
+	fn withdraw(msa_id: MessageSourceId, amount: Self::Balance) -> Result<(), DispatchError> {
 		let mut capacity_details =
 			Self::get_capacity_for(msa_id).ok_or(Error::<T>::TargetCapacityNotFound)?;
 
--- a/pallets/capacity/src/lib.rs
+++ b/pallets/capacity/src/lib.rs
@@ -369,21 +369,19 @@ pub mod pallet {
 		pub fn unstake(
 			origin: OriginFor<T>,
 			target: MessageSourceId,
-			amount: BalanceOf<T>,
+			requested_amount: BalanceOf<T>,
 		) -> DispatchResult {
 			let unstaker = ensure_signed(origin)?;
 
-			ensure!(amount > Zero::zero(), Error::<T>::UnstakedAmountIsZero);
+			ensure!(requested_amount > Zero::zero(), Error::<T>::UnstakedAmountIsZero);
 
-			// 1. return actual amount unstaked here
-			Self::decrease_active_staking_balance(&unstaker, amount)?;
-			// 2. pass in the actual amount here
-			let capacity_reduction = Self::reduce_capacity(&unstaker, target, amount)?;
+			let actual_amount = Self::decrease_active_staking_balance(&unstaker, requested_amount)?;
+			let capacity_reduction = Self::reduce_capacity(&unstaker, target, actual_amount)?;
 
 			Self::deposit_event(Event::UnStaked {
 				account: unstaker,
 				target,
-				amount,
+				amount: actual_amount,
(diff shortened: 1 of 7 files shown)


pallets/capacity/src/types.rs
```diff
@@ -93,17 +93,32 @@ impl<T: Config> StakingAccountDetails<T> {
...
+		}
+		let unlock_chunk = UnlockChunk { value: actual_unstaked, thaw_at };
+
+		// we've already done the check but it's fine, we need to handle possible errors.
 		self.unlocking
 			.try_push(unlock_chunk)
 			.map_err(|_| Error::<T>::MaxUnlockingChunksExceeded)?;
 
-		Ok(())
+		self.active = active;
```
shannonwells, line 120: don't update `self` if we were not able to add a new unlocking chunk.

pallets/capacity/src/types.rs
```diff
@@ -93,17 +93,32 @@ impl<T: Config> StakingAccountDetails<T> {
 	}
 
 	/// Decrease the amount of active stake by an amount and createa an UnlockChunk.
-	pub fn decrease_by(&mut self, amount: BalanceOf<T>, thaw_at: T::EpochNumber) -> DispatchResult {
-		let new_active = self.active.saturating_sub(amount);
-
-		let unlock_chunk = UnlockChunk { value: amount, thaw_at };
-
-		self.active = new_active;
+	pub fn decrease_by(
```
shannonwells, line 96: should we actually rename `decrease_by` in StakingAccountDetails and StakingTargetDetails to `deduct` to match the capacity ledger stuff?

Likewise `increase_by` to `deposit`?

enddynayn (approved):
Phenomenal! 💯

pallets/capacity/src/types.rs
enddynayn: I like that idea.

pallets/capacity/src/types.rs
```diff
@@ -93,17 +93,32 @@ impl<T: Config> StakingAccountDetails<T> {
 	}
 
 	/// Decrease the amount of active stake by an amount and createa an UnlockChunk.
-	pub fn decrease_by(&mut self, amount: BalanceOf<T>, thaw_at: T::EpochNumber) -> DispatchResult {
-		let new_active = self.active.saturating_sub(amount);
-
-		let unlock_chunk = UnlockChunk { value: amount, thaw_at };
-
-		self.active = new_active;
+	pub fn deduct(
```
wilwade, line 96: Suggestion only: In my head `withdraw` would be the antonym to `deposit`.
```suggestion
	pub fn withdraw(
```

pallets/capacity/src/types.rs
```diff
@@ -93,17 +93,32 @@ impl<T: Config> StakingAccountDetails<T> {
...
+		&mut self,
+		amount: BalanceOf<T>,
+		thaw_at: T::EpochNumber,
+	) -> Result<BalanceOf<T>, DispatchError> {
+		// let's check for an early exit before doing all these calcs
```
wilwade, line 101: These are very simple calculations right? I guess there is one constant pull, but I wouldn't think the savings could be very much for the math.

pallets/capacity/src/types.rs
```diff
@@ -93,17 +93,32 @@ impl<T: Config> StakingAccountDetails<T> {
...
+			Error::<T>::MaxUnlockingChunksExceeded
+		);
+		let mut active = self.active.saturating_sub(amount);
+		let mut actual_unstaked: BalanceOf<T> = amount;
+
+		if active.le(&T::MinimumStakingAmount::get()) {
+			actual_unstaked = amount.saturating_add(active);
+			active = Zero::zero();
+		}
+		let unlock_chunk = UnlockChunk { value: actual_unstaked, thaw_at };
```
wilwade, line 113: This is just code golf and making sure I understand the logic. Feel free to ignore.

This has no mutation, but it is the same amount of math.

Original: sub, le, add
Below: sub, le, sub

```suggestion
		let actual_unstaked = if self.active.saturating_sub(amount).le(&T::MinimumStakingAmount::get()) {
			actual_unstaked = self.active;
		} else {
    		amount
		}
		let unlock_chunk = UnlockChunk { value: actual_unstaked, thaw_at };
		
		// self.unlocking...
		
		self.active = self.active.saturating_sub(actual_unstaked);
```

wilwade (commented):
Looks good to me. A few comments, but feel free to ignore all of them.

pallets/capacity/src/types.rs
shannonwells: My thinking was, not individually no, but for a network with millions of txns it can add up.

pallets/capacity/src/types.rs
shannonwells: I like @wilwade 's suggestion of withdraw, partly because it's a much different word.

pallets/capacity/src/lib.rs
```diff
@@ -369,19 +369,19 @@ pub mod pallet {
 		pub fn unstake(
 			origin: OriginFor<T>,
 			target: MessageSourceId,
-			amount: BalanceOf<T>,
+			requested_amount: BalanceOf<T>,
```
johnmela, line 372: I like the more precise parameter naming

pallets/capacity/src/types.rs
johnmela: I agree with the use of `withdraw` instead of `deduct`

pallets/capacity/src/types.rs
shannonwells: This particular code doesn't compile, because it references itself inside the if statement. Also it does the saturating sub twice. The existing way is pretty clear and easy to read; it does the sub once, and initializes the two variables to the most likely values.

johnmela (approved):

PR #1108, opened 2023-02-17
Update the instant sealing docker to use the docker volume

### LINKED ISSUE #1107: Enable Docker Volume and Manual Sealing Options for Instant Sealing Docker
The Instant Sealing Docker forces the use of the tmp directory when it should use the docker volume and docker handle the use or removal of the chain data.

Secondly the docker image should allow for use of manual sealing in addition to the instant sealing mode.

wilwade:
# Goal
The goal of this PR is to update the instant sealing docker.

No breaking changes for most people:
- if you had it setup to use the volume, it will correctly start loading up the chain data from prior runs when it wouldn't before.

Closes #1107

# Discussion
- Better Docker documentation
- Allow instant-sealing or manual-sealing in the docker
- Use the docker volume

### DIFF
--- a/docker/instant-seal-node.dockerfile
+++ b/docker/instant-seal-node.dockerfile
@@ -36,9 +36,6 @@ ENTRYPOINT ["/frequency/frequency", \
 	# Required params for starting the chain
 	"--dev", \
 	"-lruntime=debug", \
-	"--instant-sealing", \
-	"--wasm-execution=compiled", \
-	"--execution=wasm", \
 	"--no-telemetry", \
 	"--no-prometheus", \
 	"--port=30333", \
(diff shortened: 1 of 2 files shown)


saraswatpuneet (approved):

docker/instant-seal-node.dockerfile
```diff
@@ -36,9 +36,6 @@ ENTRYPOINT ["/frequency/frequency", \
 	# Required params for starting the chain
 	"--dev", \
 	"-lruntime=debug", \
-	"--instant-sealing", \
```
pfrank13, line 39: Question, do we have to remove this? I ask because of the latency of when I will be able to see a potential issue versus when it "happens" e.g. instant-sealing keeps things nice and neat and doesn't create a crap load of blocks, Matter of being able to come in in the morning and just being able to look at the latest block MeWe had an issue with (something like this)

pfrank13 (commented):
Just a question, otherwise I think we are good.

docker/instant-seal-node.dockerfile
wilwade: It is removed from the entrypoint, but added back into the `CMD`. So it still works that way by default, but now it can be overridden if you want the docker image to do something different!

pfrank13 (approved):

PR #1130, opened 2023-02-23
set home dir to /root on self-hosted runners in remaining workflows

### LINKED ISSUE #1129: Set home dir to root on self-hosted runners
Currently, all self-hosted runners (all 8 of them) execute code as root. We need to set `HOME: root` env var in all workflows as long as it stays this way. The Verify PR workflow has been already adjusted and ran without issues for a couple of days. Time to sync up.

demisx:
# Goal
The goal of this PR is to sync up remaining workflows and set home dir to `/root` for all jobs executed on the current self-hosted build runners. 

Closes #1129

### DIFF
--- a/.github/workflows/common/reset-self-hosted-runner/action.yml
+++ b/.github/workflows/common/reset-self-hosted-runner/action.yml
@@ -6,8 +6,13 @@ runs:
     - name: Remove /home/runner
       shell: bash
       run: |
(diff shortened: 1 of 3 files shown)


wilwade (approved):

saraswatpuneet (approved):
:100:

PR #1131, opened 2023-02-23
Added upsert_schema_permissions()

### LINKED ISSUE #876: Schema permissions can only be inserted/updated with mutual agreement between provider and delegator.
Schema permissions can only be inserted/updated (upsert) with mutual agreement between provider and delegator.  Granting, revoking/removing of schemas will be removed.

For example, a user signs up with a provider.  At a later time, provider deploys a new feature.  The user is prompted whether they want to use this new feature.  When they agree, the schema permissions are upserted to reflect it.

## Acceptance Criteria
- Permissions should not be able to be added without permission of both the delegator and the provider.
- Permissions changes are *not* required to be additive only
- Provider should resubmit to `grant_delegation` the new delegation (_aka put_) to completely replace the prior delegation
- grant_schema_permissions should be removed (_security issue_)

rlaferla:
# Goal
The goal of this PR is to add upsert schema permissions functionality for `grant_delegation()`

Closes #876

### DIFF
--- a/integration-tests/scenarios/grantDelegation.test.ts
+++ b/integration-tests/scenarios/grantDelegation.test.ts
@@ -159,16 +159,6 @@ describe("Delegation Scenario Tests", function () {
             }
         });
 
-        it("should fail to grant a duplicate delegation to a provider (DuplicateProvider)", async function () {
-            const payload = await generateDelegationPayload({
-                authorizedMsaId: providerId,
-                schemaIds: [schemaId],
-            });
-            const addProviderData = ExtrinsicHelper.api.registry.createType("PalletMsaAddProvider", payload);
-
-            const grantDelegationOp = ExtrinsicHelper.grantDelegation(keys, providerKeys, signPayloadSr25519(keys, addProviderData), payload);
-            await assert.rejects(grantDelegationOp.fundAndSend(), { name: 'DuplicateProvider' });
-        });
     });
 
     describe("schema permission grants", function () {
--- a/pallets/msa/src/lib.rs
+++ b/pallets/msa/src/lib.rs
@@ -572,6 +572,7 @@ pub mod pallet {
 				add_provider_payload.authorized_msa_id == provider_id.0,
 				Error::<T>::UnauthorizedDelegator
 			);
+
 			Self::upsert_schema_permissions(
 				provider_id,
 				delegator_id,
@@ -1167,7 +1168,7 @@ impl<T: Config> Pallet<T> {
 
 			// Revoke any that are not in the new list
 			let current_block = frame_system::Pallet::<T>::block_number();
-			PermittedDelegationSchemas::<T>::try_revoke_schemas_not_in_list(
+			PermittedDelegationSchemas::<T>::try_get_mut_schemas(
 				delegation,
 				revoke_ids,
 				current_block,
--- a/pallets/msa/src/lib.rs
+++ b/pallets/msa/src/lib.rs
@@ -1152,9 +1152,16 @@ impl<T: Config> Pallet<T> {
 
 			let existing_keys: Vec<SchemaId> =
 				delegation.schema_permissions.keys().cloned().collect();
+
 			for existing_schema_id in &existing_keys {
 				if !schema_ids.contains(&existing_schema_id) {
-					revoke_ids.push(*existing_schema_id);
+					match delegation.schema_permissions.get(&existing_schema_id) {
+						Some(block) =>
+							if *block == T::BlockNumber::zero() {
+								revoke_ids.push(*existing_schema_id);
+							},
+						None => {},
+					}
 				}
 			}
 			for schema_id in &schema_ids {
@@ -1163,8 +1170,9 @@ impl<T: Config> Pallet<T> {
 				}
 			}
 
-			// Revoke any that are not in the new list
 			let current_block = frame_system::Pallet::<T>::block_number();
+
+			// Revoke any that are not in the new list that are not already revoked
 			PermittedDelegationSchemas::<T>::try_get_mut_schemas(
 				delegation,
 				revoke_ids,
--- a/pallets/msa/src/lib.rs
+++ b/pallets/msa/src/lib.rs
@@ -572,7 +572,11 @@ pub mod pallet {
 				add_provider_payload.authorized_msa_id == provider_id.0,
 				Error::<T>::UnauthorizedDelegator
 			);
-			Self::add_provider(provider_id, delegator_id, add_provider_payload.schema_ids)?;
+			Self::upsert_schema_permissions(
+				provider_id,
+				delegator_id,
+				add_provider_payload.schema_ids,
+			)?;
 			Self::deposit_event(Event::DelegationGranted { delegator_id, provider_id });
 
 			Ok(())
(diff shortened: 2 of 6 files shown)


shannonwells (approved):
🚢 it
👁️s 👍 2️⃣ Me

pallets/msa/src/tests.rs
```diff
@@ -645,6 +645,114 @@ pub fn add_provider_to_msa_is_success() {
...
+		assert_ok!(sp.try_insert(1u16, 1u64)); // schema id 1 revoked at block 1
+		assert_ok!(sp.try_insert(2u16, 1u64)); // schema id 2 revoked at block 1
+		assert_ok!(sp.try_insert(3u16, 0u64));
+		assert_ok!(sp.try_insert(4u16, 0u64));
+		assert_ok!(sp.try_insert(5u16, 0u64));
+		assert_ok!(sp.try_insert(6u16, 0u64));
+
+		let expected = Delegation { revoked_at: 0, schema_permissions: sp };
+
+		assert_eq!(Msa::get_delegation(delegator, provider), Some(expected));
```
shannonwells, line 752: I just added the following to the end of this test, which makes it fail.  So I'm not sure if this is the correct expectation. I would think that when sending a new set of permissions, schemas that are already revoked should _not_ have the expiration block updated. Secondly I expected to be able to re-permit/renew a previously expired schema:

```

		System::set_block_number(5);
		let (delegator_signature, add_provider_payload) =
			create_and_sign_add_provider_payload_with_schemas(
				delegator_pair.clone(),
				provider_msa,
				Some(vec![1, 3, 6]),
			);

		assert_ok!(Msa::grant_delegation(
			RuntimeOrigin::signed(provider_account.into()),
			delegator_account.into(),
			delegator_signature,
			add_provider_payload
		));

		let mut sp = BoundedBTreeMap::<SchemaId, u64, MaxSchemaGrantsPerDelegation>::new();
		assert_ok!(sp.try_insert(1u16, 0u64)); // schema id 1 should be renewed
		assert_ok!(sp.try_insert(2u16, 1u64)); // schema id 2 should stay revoked at block 1
		assert_ok!(sp.try_insert(3u16, 0u64)); // leave alone
		assert_ok!(sp.try_insert(4u16, 5u64)); // revoke
		assert_ok!(sp.try_insert(5u16, 5u64)); // revoke
		assert_ok!(sp.try_insert(6u16, 0u64)); // leave alone

		let exp = Delegation { revoked_at: 5, schema_permissions: sp };
		assert_eq!(Some(exp), Msa::get_delegation(delegator, provider), );
```

In this portion of the test, schema 2 has its expiration block updated to block 5.  Schema 1 is not re-permitted/renewed.

shannonwells (changes requested):
I just added to your test and this does not seem like the right response:

At the end of

pallets/msa/src/tests.rs
rlaferla: My test was set up for block 0.  I can change it to use the current block and I think it will pass.

integration-tests/scenarios/grantDelegation.test.ts
```diff
@@ -159,16 +159,6 @@ describe("Delegation Scenario Tests", function () {
             }
         });
 
-        it("should fail to grant a duplicate delegation to a provider (DuplicateProvider)", async function () {
```
wilwade, line 162: Could we switch this to a positive test now? Aka an integration test of upsert with just a simple alter of the assert?

pallets/msa/src/tests.rs
rlaferla: I think there is a bug but I have to change your test because you reversed the assert from what I used in the prior two calls.  There's also a bug in your test.  The `revoked_at` is for revoking all permissions and is only mutated/changed by `revoke_provider() `and not `grant_delegation()`.

rlaferla: If the schema is not in the new grant delegation list but is in the current list, it is revoked at the current block.  It needs to do a further test to not include blocks that were already revoked at a previous block so that they stay revoked at the original block and not the current one.

pallets/msa/src/lib.rs
```diff
@@ -1129,6 +1134,49 @@ impl<T: Config> Pallet<T> {
...
+	/// Modify delegation's schema permissions
+	///
+	/// # Errors
+	/// * [`Error::ExceedsMaxSchemaGrantsPerDelegation`]
+	pub fn upsert_schema_permissions(
+		provider_id: ProviderId,
+		delegator_id: DelegatorId,
+		schema_ids: Vec<SchemaId>,
+	) -> DispatchResult {
+		Self::try_mutate_delegation(delegator_id, provider_id, |delegation, _is_new_delegation| {
```
wilwade, line 1146: I don't see a mutation of `delegation` happening in here. Am I missing something?

pallets/msa/src/lib.rs
```diff
@@ -1129,6 +1134,49 @@ impl<T: Config> Pallet<T> {
...
+			Self::ensure_all_schema_ids_are_valid(&schema_ids)?;
+
+			// Create revoke and insert lists
+			let mut revoke_ids: Vec<SchemaId> = Vec::new();
+			let mut insert_ids: Vec<SchemaId> = Vec::new();
+
+			let existing_keys: Vec<SchemaId> =
+				delegation.schema_permissions.keys().cloned().collect();
+			for existing_schema_id in &existing_keys {
```
wilwade, line 1155: I wonder if using a hashset would be worth the cost here. If we assume that most users have 10 schemas (in and out), this is ~`O(10*5*2) = 100` between these two nested contains checks. With the worse case being `O(30*30*2) = 1_800`  (with MaxSchemas set to 30 and users swapping 30 for 30).

Assuming a set is possible (with normal insert at `O(1)` and check at `O(1)`) then 10 worst case would be `O(10+10+10+10) = 40` and 30 worst case would be `O(30+30+30+30) = 120`

Something like tinyset: https://crates.io/crates/tinyset would work with default features disabled.
hashbrown is already used in substrate, but not in a no_std env as far as I can tell, but might work: https://github.com/rust-lang/hashbrown#features

pallets/msa/src/tests.rs
rlaferla: Fixed in next update.

pallets/msa/src/lib.rs
rlaferla: I think we should consider it in a future story.

integration-tests/scenarios/grantDelegation.test.ts
rlaferla: What would be the benefit of this over the unit test that does that?

wilwade: Mostly just that extra end to end test that shows how it works. Not a blocker

pallets/msa/src/lib.rs
wilwade: Agreed. Don't want it to block the next release. I'll add it to the top of the tech debt so it gets done soon.

shannonwells (approved):
W00t!

pallets/msa/src/lib.rs
wilwade: Added an updated version of the suggestion into a new story: https://github.com/LibertyDSNP/frequency/issues/1155

wilwade (approved):
👍

PR #1138, opened 2023-02-25
feat(tx-payment): add ability to use Capacity

### LINKED ISSUE #1026: [Capacity][SubTask-2-of-4] Fee withdrawal from token or capacity account
This is a subtask of #1010 as the parent see for details

Update ChargeTransaction signed extension validate && pre-dispatch for calculating payment and withdrawing from either Capacity or Token.

enddynayn:
# Goal
Add the ability to use Capacity for transacting.

This implementation wraps the substrate transaction-payment
pallet, and it is used to toggle between token transactions
and capacity transactions.

When Capacity is used to transact, tipping is
set to zero regardless if a user attempts to tip
with Capacity. Token transaction tipping is
still supported.

Closes #1026 

https://www.loom.com/share/b0d3a0df837547dfa7b0195db51cd0e7

### DIFF
--- a/pallets/frequency-tx-payment/src/lib.rs
+++ b/pallets/frequency-tx-payment/src/lib.rs
@@ -1,33 +1,114 @@
+//! # Frequency-Transactions Pallet
+//! The Frequency-transaction pallet allows users to transact using either capacity or token.
+//!
+//! - [Configuration: `Config`](Config)
+//! - [Extrinsics: `Call`](Call)
+//! - [Event Enum: `Event`](Event)
+//! - [Error Enum: `Error`](Error)
+//!
+//! ## Overview
+//! Capacity is a refillable resource that can be used to make transactions on the network.
+//! This pallets allows users to toggle between submitting transactions with capacity or tokens.
+//! Note that tipping for capacity transactions is not allowed and is set to zero. Users who
+//! use tokens to transact can continue to use a tip.
+//!
+//! The Frequency-transaction pallet provides functions for:
+//!
+//! - pay_with_capacity
+//!
+
 #![cfg_attr(not(feature = "std"), no_std)]
 
 use frame_support::{
-	dispatch::{DispatchInfo, GetDispatchInfo, PostDispatchInfo},
+	dispatch::{DispatchInfo, Dispatchable, GetDispatchInfo, PostDispatchInfo},
 	pallet_prelude::*,
-	traits::IsType,
+	traits::{IsSubType, IsType},
+	DefaultNoBound,
 };
+use frame_system::pallet_prelude::*;
 
 use codec::{Decode, Encode};
 use scale_info::TypeInfo;
 use sp_std::prelude::*;
 
 use sp_runtime::{
-	traits::{DispatchInfoOf, Dispatchable, PostDispatchInfoOf, SignedExtension},
+	traits::{DispatchInfoOf, PostDispatchInfoOf, SignedExtension, Zero},
 	transaction_validity::{TransactionValidity, TransactionValidityError},
 	FixedPointOperand,
 };
 
+use common_primitives::capacity::Nontransferable;
+
 use pallet_transaction_payment::OnChargeTransaction;
 
-// Type aliases used for interaction with `OnChargeTransaction`.
+/// Type aliases used for interaction with `OnChargeTransaction`.
 pub(crate) type OnChargeTransactionOf<T> =
 	<T as pallet_transaction_payment::Config>::OnChargeTransaction;
 
-// Balance type alias.
+/// Balance type alias.
 pub(crate) type BalanceOf<T> = <OnChargeTransactionOf<T> as OnChargeTransaction<T>>::Balance;
 
-// Liquidity info type alias (imbalances).
+/// Liquidity info type alias (imbalances).
 pub(crate) type LiquidityInfoOf<T> =
 	<OnChargeTransactionOf<T> as OnChargeTransaction<T>>::LiquidityInfo;
+
+/// Capacity Balance type
+pub(crate) type CapacityOf<T> = <T as Config>::Capacity;
+
+/// Capacity Balance alias
+pub(crate) type CapacityBalanceOf<T> = <CapacityOf<T> as Nontransferable>::Balance;
+
+/// Used to pass the initial payment info from pre- to post-dispatch.
+#[derive(Encode, Decode, DefaultNoBound, TypeInfo)]
+pub enum InitialPayment<T: Config> {
+	/// No initial fee was paid.
+	#[default]
+	Free,
+	/// The initial fee was payed in the native currency.
+	Token(LiquidityInfoOf<T>),
+	/// The initial fee was paid in an asset.
+	Capacity,
+}
+
+#[cfg(feature = "std")]
+impl<T: Config> InitialPayment<T> {
+	pub fn is_free(&self) -> bool {
+		match *self {
+			InitialPayment::Free => true,
+			_ => false,
+		}
+	}
+
+	pub fn is_capacity(&self) -> bool {
+		match *self {
+			InitialPayment::Capacity => true,
+			_ => false,
+		}
+	}
+
+	pub fn is_token(&self) -> bool {
+		match *self {
+			InitialPayment::Token(_) => true,
+			_ => false,
+		}
+	}
+}
+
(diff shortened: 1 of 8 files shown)


shannonwells (approved):
Looks good, just had 2 questions and a non-blocking comment about EMPTY_FUNCTION.

mattheworris (commented):
Looks good to me.

pallets/frequency-tx-payment/src/lib.rs
```diff
@@ -102,16 +166,62 @@ pub struct ChargeFrqTransactionPayment<T: Config>(#[codec(compact)] BalanceOf<T>
 
 impl<T: Config> ChargeFrqTransactionPayment<T>
 where
-	BalanceOf<T>: Send + Sync + FixedPointOperand,
+	BalanceOf<T>: Send + Sync + FixedPointOperand + IsType<CapacityBalanceOf<T>> + From<u64>,
```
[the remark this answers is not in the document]
enddynayn: Both CapacityBalanceOf and BalanceOf are associate types that come from the Currency trait. Although they look the same because they are defined in different places to rust, they are two different types. If I understand correctly, this trait allows us to say that these are technically the same thing.

pallets/frequency-tx-payment/src/lib.rs
```diff
@@ -102,16 +166,62 @@ pub struct ChargeFrqTransactionPayment<T: Config>(#[codec(compact)] BalanceOf<T>
...
+		let fee =
+			pallet_transaction_payment::Pallet::<T>::compute_fee(len as u32, info, self.tip(call));
+
+		match call.is_sub_type() {
+			Some(Call::pay_with_capacity { .. }) => {
+				let msa_id = pallet_msa::Pallet::<T>::ensure_valid_msa_key(who).map_err(
+					|_| -> TransactionValidityError { InvalidTransaction::Payment.into() },
+				)?;
+
+				T::Capacity::withdraw(msa_id, fee.into()).map_err(
```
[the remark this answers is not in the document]
enddynayn: I think we can talk about this more tomorrow. It is worth exploring.

pallets/frequency-tx-payment/src/tests.rs
```diff
@@ -181,3 +238,153 @@ fn pay_with_capacity_returns_weight_of_child_call() {
...
+				weight: Weight::from_ref_time(5),
+				pays_fee: Pays::No,
+				..Default::default()
+			};
+			let len = 10;
+			let result = charge_tx_payment.withdraw_fee(&who, call, &info, len).unwrap();
+
+			// fee = base_weight(5)
+			//   + extrinsic_weight(5) * WeightToFee(1)
+			//   + TransactionByteFee(1)* len(10) = 20
```
johnmela, line 381: Do we still need these lines?

johnmela (changes requested):

pallets/frequency-tx-payment/src/tests.rs
enddynayn: It's up to you. I added them for you.

johnmela: Ah, let's keep them

pallets/frequency-tx-payment/src/lib.rs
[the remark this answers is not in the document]
johnmela: I think this may be worth exploring also, but I will approve since I don't think this is blocking

johnmela (approved):

pallets/frequency-tx-payment/src/tests.rs
```diff
@@ -112,6 +116,7 @@ fn transaction_payment_with_token_and_post_dispatch_refund_is_succesful() {
 		.build()
 		.execute_with(|| {
 			let account_id = 1u64;
+			// update to run test on all capacity transactions.
```
enddynayn, line 119: remove

pallets/frequency-tx-payment/src/lib.rs
```diff
@@ -1,34 +1,96 @@
 #![cfg_attr(not(feature = "std"), no_std)]
```
wilwade, line 1: Not this PR, but we should make sure the pallet has docs and lints that follows the pattern of the other ones.

pallets/frequency-tx-payment/src/lib.rs
```diff
@@ -1,34 +1,96 @@
...
+	Capacity,
+}
+
+#[cfg(feature = "std")]
+impl<T: Config> InitialPayment<T> {
+	pub fn is_free(&self) -> bool {
+		match *self {
+			InitialPayment::Free => true,
+			_ => false,
+		}
```
wilwade, line 60: I like match as well, but not sure why not to just derive Eq and use `==` for the smaller result
```suggestion
		self == InitialPayment::Free
```

pallets/frequency-tx-payment/src/lib.rs
```diff
@@ -1,34 +1,96 @@
...
+	}
+}
+
+impl<T: Config> sp_std::fmt::Debug for InitialPayment<T> {
+	#[cfg(feature = "std")]
+	fn fmt(&self, f: &mut sp_std::fmt::Formatter) -> sp_std::fmt::Result {
+		match *self {
+			InitialPayment::Free => write!(f, "Nothing"),
+			InitialPayment::Capacity => write!(f, "Token"),
+			InitialPayment::Token(_) => write!(f, "Imbalance"),
```
wilwade, line 84: Free -> Nothing makes sense, but Capacity -> Token and Token -> Imbalance do not to me.

pallets/frequency-tx-payment/src/lib.rs
```diff
@@ -102,16 +166,62 @@ pub struct ChargeFrqTransactionPayment<T: Config>(#[codec(compact)] BalanceOf<T>
...
-	pub fn tip(&self) -> BalanceOf<T> {
-		self.0
+	pub fn tip(&self, call: &<T as frame_system::Config>::RuntimeCall) -> BalanceOf<T> {
+		match call.is_sub_type() {
+			Some(Call::pay_with_capacity { .. }) => Zero::zero(),
+			_ => self.0,
+		}
+	}
+
+	// Note consider setting the tip at the benning
```
wilwade, line 186: sp
```suggestion
	// Note consider setting the tip at the beginning
```

pallets/frequency-tx-payment/src/lib.rs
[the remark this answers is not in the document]
wilwade: (Feel free to disregard as I am not into the weeds and this might be really wrong)

I think all we care is that they have capacity. Currently capacity is restricted to providers, so that is enough without the additional check.

If we were to change the requirements for capacity, it would be on the acquisition of capacity not the use of it. (for faster checks at use time)

pallets/frequency-tx-payment/src/lib.rs
```diff
@@ -157,10 +268,12 @@ where
 		info: &DispatchInfoOf<Self::Call>,
 		len: usize,
 	) -> TransactionValidity {
-		let result = pallet_transaction_payment::ChargeTransactionPayment::<T>::from(self.0)
-			.validate(who, call, info, len)?;
+		use pallet_transaction_payment::ChargeTransactionPayment;
```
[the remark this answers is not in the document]
enddynayn: Good catch. I thought I had undone this.

pallets/frequency-tx-payment/src/lib.rs
enddynayn: Ah yeah, this is used most for debugging in tests. When you withdraw tokens for an account, you create an imbalance.

https://blog.polymath.network/substrate-deep-dive-imbalances-8dfa89cc1d1

pallets/frequency-tx-payment/src/lib.rs
enddynayn: I did this because I could not add the `==.` It was a workaround since `LiquidityInfoOf<T>` does not implement
equals. LiquidityInfoOf<T> is an alias for something from the original transaction-payment pallet.

These are just for testing and are used in std only.

```
pub enum InitialPayment<T: Config> {
	/// No initial fee was paid.
	#[default]
	Free,
	/// The initial fee was paid in the native currency.
	Token(LiquidityInfoOf<T>),
	/// The initial fee was paid in an asset.
	Capacity,
}
```

PR #1143, opened 2023-02-27
Remove on_initialize from MSA Pallet with new Ring Based Signature Registry

### LINKED ISSUE #785: budget on_initialize
Ensure that "on_initialize" function for clearing out signatures
does not consume all resources for a block.

Description:
The benchmarks show that if we were to reach our MSAMaxSignaturesPerBucket limit of 50_000 that it would overfill a block and prevent blocks from being produced:

```
Weight::from_ref_time(0 as u64)
    // Standard Error: 6_000
    .saturating_add(Weight::from_ref_time(1_252_000 as u64).saturating_mul(m as u64))
    .saturating_add(T::DbWeight::get().writes((1 as u64).saturating_mul(m as u64)))
```

Inserting 50K into m results in exceeding over 5*10^12.
We can only consume 2*10^12 of weight in a block, which exceeds that limit by more than twice.

Questions:

- Is there a recommended block space allocation for all on_initialize functions?
- What is the realistic max amount of signatures that could be inserted through extrinsics?
- Given that realistic (_worst case_) max amount, how much block space does it consume?
- What is the prescribed solutions (_eg: [stackexchange](https://substrate.stackexchange.com/questions/5842/on-initialize-best-practice-for-executing-a-large-for-loop) [queue, using on_idle to drain on each block]_) 

(_note* the result of this spike will help define the AC for the implementation ticket_ )

[ref function](https://github.com/LibertyDSNP/frequency/pull/777/files#diff-6843eb02bb7e86a38f71d2e5c22b6bef714af6c8e01024ab4e4bbfefca831f24R156)

also see branch https://github.com/LibertyDSNP/frequency/tree/spike/msa-initializer

wilwade:
# Goal
The goal of this PR is to remove the need to use `on_initialize` from the msa pallet.

Closes #785

# Discussion

Implements a new system to protect against https://github.com/LibertyDSNP/frequency/issues/327

The new system is a forward linked list ring.

Simple test of 50k signatures? `make integration-load-test`

## Adding with a Full Ring

The ring is implemented on a Storage Map that uses the signatures as the key and the value is a tuple with the expire block number and the next step in the ring.

#### Example State (Ring size of 5)
- signature, next step signature
- 1,2
- 2,3
- 3,4
- 4,5
- 5,1

#### Adding new signature `n`
- signature, next step signature
- `n`,2
- 2,3
- 3,4
- 4,5
- 5,`n`

We must update both sides of the edge of the ring to maintain the forward linked list.

We also maintain a pointer state so we know where the head and tail of the ring is, and if it is in the filling state.

## Filling the Ring

To support filling the signature for the first time, we merely skip the removal of the oldest state.

#### State (ring size > 5)
- signature, next step signature
- 1,2
- 2,3
- 3,1

#### Adding new signature `n`
- signature, next step signature
- 1,2
- 2,3
- 3,`n`
- `n`,1

# Checklist
- [x] Chain spec version updated
- [x] Migration
- [x] Tests added
- [x] Benchmarks added (Not needed)
- [x] Weights updated


integration-tests/load-tests/msaInitializerLoad.test.ts
```diff
@@ -0,0 +1,155 @@
...
+
+        // Make 250 MSAs
+        const msaKeys: GeneratedMsa[] = await generateMsas(2300);
+
+        let blockNumber = await getBlockNumber();
+        console.log("Starting Block Number:", blockNumber);
+
+        let signatureCallTests: Array<KeyringPair> = [];
+
+        // Make 49,998 signatures via 24_998 loops
```
rlaferla, line 32: Should this be 49,996 (2 * 24,998)?

pallets/msa/src/lib.rs
```diff
@@ -1319,83 +1333,105 @@ impl<T: Config> Pallet<T> {
...
+				tail: signature.clone(),
+				count: 0,
+			});
+
+		// Default to the current tail in case we are still filling up
+		let mut new_tail: MultiSignature = pointer.tail.clone();
+
+		// We are now wanting to overwrite prior signatures
+		// Do we need this > 0?
+		let looping: bool =
```
wilwade, line 1358:
```suggestion
		let isRingBufferFull: bool =
```

JoeCap08055 (commented):
Would it be possible to implement the RB as simply a blob containing a fixed-size array (`BoundedVec`) plus a current index?

Pros:
* 1 read, 1 write TOTAL
* no pointer management

Cons:
* 1 read, but it's of the entire RB
* 1 write, but ☝🏻 
* No keyed lookup of signatures, must search the array

(Basically, once the RB is deserialized into memory, it is managed as a typical in-memory ringbuffer, then serialized back)

wilwade:
@JoeCap08055 
> Would it be possible to implement the RB as simply a blob containing a fixed-size array (BoundedVec) plus a current index?

The size of it would be 1.6 Mb (50k sigs at ~32 bytes per sig), so I don't think we want to do that that much in memory work. Second the size of the proof for Weights 2.0 would grow as well (although I do not know by how much).

So I think it is worse. We could still do a map and pointer and serialize it into one storage, but I think for 50k+, that is going to be too much.

JoeCap08055:
> @JoeCap08055
> 
> > Would it be possible to implement the RB as simply a blob containing a fixed-size array (BoundedVec) plus a current index?
> 
> The size of it would be 1.6 Mb (50k sigs at ~32 bytes per sig), so I don't think we want to do that that much in memory work. Second the size of the proof for Weights 2.0 would grow as well (although I do not know by how much).
> 
> So I think it is worse. We could still do a map and pointer and serialize it into one storage, but I think for 50k+, that is going to be too much.

As I suspected. Just thought I'd ask 😅

Makefile
```diff
@@ -141,6 +141,18 @@ test:
 integration-test:
 	./scripts/run_integration_tests.sh
 
+.PHONY: try-runtime
```
wilwade, line 144: Initial work on making `try-runtime` and `on-runtime-upgrade` easier to run.

integration-tests/load-tests/signatureRegistry.test.ts
```diff
@@ -0,0 +1,160 @@
+import "@frequency-chain/api-augment";
```
wilwade, line 1: This is the actual load test. Other changes to integration test files are just to support this.

integration-tests/load-tests/signatureRegistry.test.ts
```diff
@@ -0,0 +1,160 @@
...
+interface GeneratedMsa {
+    id: u64;
+    nonce: number;
+    controlKey: KeyringPair;
+}
+
+describe("MSA Initializer Load Tests", function () {
+
+    beforeEach(async function () {
+        await createBlock();
```
wilwade, line 19: There are a lot of createBlocks in here to support the manual sealing.

This also makes a lot of the code complex as we have to create a promise (or lots of them) then await them to be resolved after we create a block.

integration-tests/load-tests/signatureRegistry.test.ts
```diff
@@ -0,0 +1,160 @@
...
+    // Create a second block in case there were more than could fit in ^ blocks
+    await createBlock();
+    const msaIds = await Promise.all(msaPromises)
+    console.log("Generated Msas: ", count);
+    return msaIds;
+}
+
+async function createBlock(wait: number = 300) {
+    // Wait ms before creating the block to give the chain time to process the transaction pool
+    await new Promise(r => setTimeout(r, wait));
```
wilwade, line 121: Validating transactions takes real time, so we need to wait for them to happen. With the larger blocks, it can take longer.

integration-tests/load-tests/signatureRegistry.test.ts
```diff
@@ -0,0 +1,160 @@
...
+    const nonce = await firstValueFrom(ExtrinsicHelper.api.call.accountNonceApi.accountNonce(keys.address));
+    return nonce.toNumber();
+}
+
+async function createMsa(keys: KeyringPair): Promise<u64> {
+    const result = await ExtrinsicHelper.createMsa(keys).signAndSend();
+    const msaRecord = result[1]["msa.MsaCreated"];
+    if (msaRecord) return msaRecord.data[0] as u64;
+
+    // This doesn't always work I think due to the load. So doing a backup call
```
wilwade, line 139: Overloading the WebSocket limit specifically. So those createMsa subscriptions are getting closed before they can respond.

integration-tests/scaffolding/extrinsicHelpers.ts
```diff
@@ -91,34 +92,34 @@ export class Extrinsic<T extends ISubmittableResult = ISubmittableResult, C exte
...
-        return firstValueFrom(this.extrinsic().signAndSend(this.keys).pipe(
+    public signAndSend(nonce?: number): Promise<ParsedEventResult> {
+        return firstValueFrom(this.extrinsic().signAndSend(this.keys, {nonce: nonce}).pipe(
             filter(({ status }) => status.isInBlock || status.isFinalized),
             this.parseResult(this.event),
         ))
     }
 
-    public async sudoSignAndSend(): Promise<[ParsedEvent<C, N> | undefined, EventMap]> {
+    public sudoSignAndSend(): Promise<[ParsedEvent<C, N> | undefined, EventMap]> {
```
wilwade, line 102: Cleaned up a few of extra asyncs hanging around.

integration-tests/scaffolding/extrinsicHelpers.ts
```diff
@@ -91,34 +92,34 @@ export class Extrinsic<T extends ISubmittableResult = ISubmittableResult, C exte
         this.api = ExtrinsicHelper.api;
     }
 
-    public async signAndSend(): Promise<[ParsedEvent<C, N> | undefined, EventMap]> {
-        return firstValueFrom(this.extrinsic().signAndSend(this.keys).pipe(
+    public signAndSend(nonce?: number): Promise<ParsedEventResult> {
```
wilwade, line 95: Lots of optional nonce adding to support lots of transactions in one block.

pallets/msa/src/lib.rs
```diff
@@ -1319,83 +1316,105 @@ impl<T: Config> Pallet<T> {
...
+				count: 0,
+			});
+
+		// Default to the current tail in case we are still filling up
+		let mut new_tail: MultiSignature = pointer.tail.clone();
+
+		// We are now wanting to overwrite prior signatures
+		// Do we need this > 0?
+		let is_ring_buffer_full: bool =
+			pointer.count > 0 && pointer.count == T::MaxSignaturesStored::get().unwrap_or(0);
```
enddynayn, line 1342: I think we can use Zero::zero() here.

Why do we call unwrap_or(0) here?
> T::MaxSignaturesStored::get().unwrap_or(0);

pallets/msa/src/migration.rs
```diff
@@ -0,0 +1,158 @@
...
+
+	if version < 1 {
+		weight = weight.saturating_add(v1::migrate::<T>());
+		// Updated version inside so we can drain the prefix if needed
+	}
+
+	weight
+}
+
+/// Migrating to remove old storage
```
wilwade, line 45: This does not try to deal with the situation of moving the old signatures into the new system. This is due to the system not being used on mainnet yet.

pallets/msa/src/migration.rs
```diff
@@ -0,0 +1,158 @@
...
+				v0::PayloadSignatureRegistry::<T>::set(0, MultiSignature::Sr25519(msg), Some(i));
+			}
+
+			let iter = v0::PayloadSignatureRegistry::<T>::iter();
+			assert_eq!(iter.count(), 102);
+
+			// Migrate and Check
+
+			v1::migrate::<T>();
+			assert_eq!(StorageVersion::get::<Pallet<T>>(), 1);
```
wilwade, line 153: This shouldn't work, but I am looking into if it does due to the `storage_alias` macro working differently than the storage macro.

Not a blocker for release as shouldn't be the case on rococo or mainnet.

pallets/msa/src/weights.rs
```diff
@@ -76,89 +75,89 @@ impl<T: frame_system::Config> WeightInfo for SubstrateWeight<T> {
...
 	fn create() -> Weight {
-		Weight::from_ref_time(19_842_000 as u64)
+		Weight::from_ref_time(20_514_000 as u64)
 			.saturating_add(T::DbWeight::get().reads(3 as u64))
 			.saturating_add(T::DbWeight::get().writes(3 as u64))
 	}
-	// Storage: Msa PayloadSignatureBucketCount (r:1 w:1)
-	// Storage: Msa PayloadSignatureRegistry (r:1 w:1)
+	// Storage: Msa PayloadSignatureRegistryRing (r:1 w:1)
+	// Storage: Msa PayloadSignatureRegistryRingPointer (r:1 w:1)
```
wilwade, line 83: No changes in number of reads and writes! 🎉

pallets/msa/src/weights.rs
```diff
@@ -76,89 +75,89 @@ impl<T: frame_system::Config> WeightInfo for SubstrateWeight<T> {
...
 	}
-	// Storage: Msa PayloadSignatureBucketCount (r:1 w:1)
-	// Storage: Msa PayloadSignatureRegistry (r:2 w:2)
+	// Storage: Msa PayloadSignatureRegistryRing (r:2 w:2)
+	// Storage: Msa PayloadSignatureRegistryRingPointer (r:1 w:1)
 	// Storage: Msa PublicKeyToMsaId (r:2 w:1)
 	// Storage: Msa PublicKeyCountForMsaId (r:1 w:1)
 	fn add_public_key_to_msa() -> Weight {
-		Weight::from_ref_time(147_786_000 as u64)
+		Weight::from_ref_time(155_708_000 as u64)
```
wilwade, line 109: Does appear there is a slight increase in base weight, but this is also within the range of variability of weight generation.

runtime/common/src/constants.rs
```diff
@@ -114,13 +114,9 @@ pub type MsaMaxPublicKeysPerMsa = ConstU8<25>;
...
 pub type MSAMortalityWindowSize = ConstU32<100>;
-/// The maximum number of signatures per virtual bucket
-pub type MSAMaxSignaturesPerBucket = ConstU32<50_000>;
-/// The total number of virtual buckets
-pub type MSANumberOfBuckets = ConstU32<2>;
 /// The upper limit on total stored signatures.
-/// This MUST be MaxSignaturesPerBucket * NumberOfBuckets.
-pub type MSAMaxSignaturesStored = ConstU32<100_000>;
+/// Set to an average of 50 signatures per block
+pub type MSAMaxSignaturesStored = ConstU32<50_000>;
```
wilwade, line 119: Value per 100 blocks has not changed.

scripts/run_integration_tests.sh
```diff
@@ -1,5 +1,13 @@
 #!/usr/bin/env bash
```
wilwade, line 2: Extended to support integration load tests

pallets/msa/src/lib.rs
```diff
@@ -173,10 +159,20 @@ pub mod pallet {
...
+	#[pallet::storage_version(STORAGE_VERSION)]
 	pub struct Pallet<T>(_);
 
+	#[pallet::hooks]
+	impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {
+		fn on_runtime_upgrade() -> frame_support::weights::Weight {
+			migration::migrate::<T>()
+		}
+	}
+
```
saraswatpuneet, line 175: you might not want to do this in pallet hooks,  instead use `OnRuntimeUpgrade` in runtime
https://github.com/LibertyDSNP/frequency/commit/f05867c0cafb0f0bbc6f1dd4a6c5eceab8db536c

aramikm (approved):
Implementation looks correct but for some reason the load test fails when running with
`make start-manual` with following error

Screenshot 2023-02-28 at 5 56 33 PM

Another way to do load test would be to just do it in rust and disable the test an example would be https://github.com/paritytech/substrate/blob/master/frame/message-queue/src/integration_test.rs

pallets/msa/src/lib.rs
wilwade: It was hard to tell which was better, but this way appeared to be the latest pattern? Thoughts as to why one way instead of another?

wilwade:
> Implementation looks correct but for some reason the load test fails when running with `make start-manual` with following error

@aramikm The load tests can be iffy as it isn't always waiting for the correct amount of time. As long as it is connecting to a separate node, the same setup will occur (unless you were to slow it down more).

Could you try it again and see if it errors out at the same point?

pallets/msa/src/lib.rs
saraswatpuneet: OnRuntimeUpgrade is preferred

Parity states a warning for using hook

https://paritytech.github.io/substrate/master/frame_support/traits/trait.Hooks.html#method.on_runtime_upgrade

> While this function can be freely implemented, using on_runtime_upgrade from inside the pallet is discouraged and might get deprecated in the future. Alternatively, export the same logic as a free-function from your pallet, and pass it to type Executive from the top-level runtime.
> 
> [source](https://paritytech.github.io/substrate/master/src/frame_support/traits/hooks.rs.html#275-277)
> 
>

wilwade: Updated... I think

runtime/frequency/src/lib.rs
```diff
@@ -179,14 +179,17 @@ pub type UncheckedExtrinsic =
 /// Extrinsic type that has already been checked.
 pub type CheckedExtrinsic = generic::CheckedExtrinsic<AccountId, RuntimeCall, SignedExtra>;
 
+/// Migrations for Frequency
+pub type Migrations = (remove_sudo::RemoveSudo, pallet_msa::migration::Migration<Runtime>);
+
```
saraswatpuneet, line 184: cool, schemas one can be added here too when merged

runtime/frequency/src/lib.rs
```diff
@@ -179,14 +179,17 @@ pub type UncheckedExtrinsic =
...
 /// Executive: handles dispatch to the various modules.
 pub type Executive = frame_executive::Executive<
 	Runtime,
 	Block,
 	frame_system::ChainContext<Runtime>,
 	Runtime,
 	AllPalletsWithSystem,
-	remove_sudo::RemoveSudo,
+	Migrations,
```
saraswatpuneet, line 192: I like how cleaner it is now

:+1:

pallets/msa/src/benchmarking.rs
```diff
@@ -213,27 +213,6 @@ benchmarks! {
 		assert_eq!(frame_system::Pallet::<T>::events().len(), 1);
 	}
 
-	on_initialize {
```
enddynayn, line 216: woot woot! 💯

pallets/msa/src/lib.rs
```diff
@@ -228,33 +217,41 @@ pub mod pallet {
...
+	///     - `BlockNumber` when the keyed signature can be ejected from the registry
+	///     - [`MultiSignature`] the linked list pointer. This pointer is written as the oldest value, but updated to be the "next" value when there is one
 	#[pallet::storage]
 	#[pallet::getter(fn get_payload_signature_registry)]
-	pub(super) type PayloadSignatureRegistry<T: Config> = StorageDoubleMap<
-		_,                      // prefix
-		Twox64Concat,           // hasher for key1
-		T::BlockNumber, // Bucket number. Stored as BlockNumber because I'm done arguing with rust about it.
-		Twox64Concat,   // hasher for key2
+	pub(super) type PayloadSignatureRegistryRing<T: Config> = StorageMap<
```
enddynayn, line 238: Nice you got rid of the double map

pallets/msa/src/lib.rs
```diff
@@ -1360,83 +1351,105 @@ impl<T: Config> Pallet<T> {
...
+
+		// We are now wanting to overwrite prior signatures
+		// Do we need this > 0?
+		let is_ring_buffer_full: bool =
+			pointer.count > 0 && pointer.count == T::MaxSignaturesStored::get().unwrap_or(0);
+
+		// Overwrite oldest signature, update tail
+		if is_ring_buffer_full {
+			<PayloadSignatureRegistryRing<T>>::try_mutate(
+				pointer.tail,
```
enddynayn, line 1382: Here we are dropping the tail, but I would expect to drop the head. The head being the oldest or next in line to be dropped. Like when you wait in line, the person in the front of the line who is next is the head.

wilwade: I think of the tail as the oldest entry in the ring and the head as the newest entry.

`1,2 (tail) -> 2,3 -> 3,4 -> 4,1 (head)`

I'll add some notes to the documentation to show that.

pallets/msa/src/lib.rs
```diff
@@ -1360,83 +1351,105 @@ impl<T: Config> Pallet<T> {
...
+				count: 0,
+			});
+
+		// Default to the current tail in case we are still filling up
+		let mut new_tail: MultiSignature = pointer.tail.clone();
+
+		// We are now wanting to overwrite prior signatures
+		// Do we need this > 0?
+		let is_ring_buffer_full: bool =
+			pointer.count > 0 && pointer.count == T::MaxSignaturesStored::get().unwrap_or(0);
```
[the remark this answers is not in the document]
wilwade: You are right. It is not optional. I'll remove and that removes any possible need for the pointer.count > 0.

pallets/msa/src/lib.rs
```diff
@@ -1360,83 +1351,105 @@ impl<T: Config> Pallet<T> {
...
+
+		// Overwrite oldest signature, update tail
+		if is_ring_buffer_full {
+			<PayloadSignatureRegistryRing<T>>::try_mutate(
+				pointer.tail,
+				|maybe_block_and_next_oldest_pointer| -> DispatchResult {
+					match maybe_block_and_next_oldest_pointer {
+						// Loop Complete, remove and remember the new tail value
+						Some((test_block, tail)) => {
```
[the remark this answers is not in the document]
wilwade: Updated to `expire_block_number` which reads well:

`ensure!(current_block.gt(expire_block_number), ...)`

pallets/msa/src/migration.rs
wilwade: It is now working correctly. It was only working with the overlay storage before.

pallets/msa/src/lib.rs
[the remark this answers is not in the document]
wilwade: So it turns out that it is an optional because it is used as the max for the StorageMap max and that requires it to implement `Get<Option<u32>>`.

I still removed the pointer.count > 0 as we should not be adding signatures if the MaxSignaturesStored is unset or set to 0.

pallets/msa/src/lib.rs
```diff
@@ -1360,83 +1354,103 @@ impl<T: Config> Pallet<T> {
...
+							// Update the old head with a pointer to the new head
+							*maybe_old_head = Some((old_head.0, signature.clone()));
 
-		// If we did not cross a bucket boundary block, stop
-		if prior_bucket_num == current_bucket_num {
-			return Weight::zero()
+							Ok(())
+						},
+						// If we don't have an old head, that is very strange
+						None => fail!("Missing data from the Signature Registry!"),
```
enddynayn, line 1425: same to this one.

pallets/msa/src/lib.rs
wilwade: Updating to `newest` and `oldest` for a much clearer naming.

enddynayn (approved):
💯 You make it look easy.  Superb!🥇

PR #1147, opened 2023-02-27
Initial implementation of revised retire_msa() design

### LINKED ISSUE #678: Handle Deleting Delegations over Expected Max in `retire_msa`
## Force Removal of Delegations before Retirement

**Implementation Details**

- [x] `ensure_msa_can_retire` checks for zero delegations
- [x] Add error to `ensure_msa_can_retire` for existing delegations
- [x] Update `retire_msa` to not remove delegations
- [x] Remove unused code
- [x] Update benchmarks
- [x] Update tests

- Con: Takes time to retire
- Pro: Generally not a very used function
- Pro: Simple solution (aka force the user/client to deal with it)
- Pro: No need to track delegation count
- Pro: Generally understandable error
- Pro: We already require all but one key to be removed first
- Future: Could also be a pattern for requiring the user remove user storage items first as well.

### Rejected Options

**Require the user call `retire_msa` more than once**

- Just remove n delegations each time it is called until no delegations remain
- Con: Confusing errors from `retire_msa`

**Use on_initialize / on_idle to clean up**
- Con: Complexity of tracking what needs to be cleaned up
- User Pro: Chain does the work instead of the user
- Consensus Con: Chain does the work

1. Add the resource to delete in the end of a queue from extrinsic
2. `on_initialize` will get the first item in queue and try to delete SOME_DEFINED_MAX_NUMBER from it and check if there still exists more
2.1. If not then remove the top from queue
2.2. If there are more just leave it in the queue
3. Repeat step 2 on next block

rlaferla:
# Goal
The goal of this PR is to revise MSA retirement to require that the user/delegator remove all provider delegations before retiring their MSA account.

Closes #678


pallets/msa/src/lib.rs
```diff
@@ -78,7 +78,6 @@ use common_primitives::{
 	msa::{
 		Delegation, DelegationValidator, DelegatorId, MsaLookup, MsaValidator, ProviderId,
 		ProviderLookup, ProviderRegistryEntry, SchemaGrantValidator,
-		EXPECTED_MAX_NUMBER_OF_PROVIDERS_PER_DELEGATOR,
```
wilwade, line 81: We should delete this in common primitives as well.

pallets/msa/src/lib.rs
```diff
@@ -862,7 +860,7 @@ pub mod pallet {
 					);
 				},
 			}
-			Ok(Some(T::WeightInfo::retire_msa(num_deletions)).into())
+			Ok(Some(T::WeightInfo::retire_msa()).into())
```
wilwade, line 863: As the weight is static, I think we can now just return `Ok`. (and update `DispatchResultWithPostInfo` to just `DispatchResult`)
```suggestion
			Ok()
```

pallets/msa/src/lib.rs
```diff
@@ -1655,6 +1643,14 @@ impl<T: Config + Send + Sync> CheckFreeExtrinsicUse<T> {
 			key_count == 1,
 			InvalidTransaction::Custom(ValidityError::InvalidMoreThanOneKeyExists as u8)
 		);
+
+		let num_delegations =
+			DelegatorAndProviderToDelegation::<T>::iter_key_prefix(DelegatorId(msa_id)).count();
+		ensure!(
+			num_delegations == 0,
+			InvalidTransaction::Custom(ValidityError::InvalidNonZeroProviderDelegations as u8)
+		);
```
wilwade, line 1652: I don't think we need the whole count here, just that any exist.

```suggestion
		let has_delegations: bool =
			DelegatorAndProviderToDelegation::<T>::iter_key_prefix(DelegatorId(msa_id)).any(|_| true);

		ensure!(
			!has_delegations,
			InvalidTransaction::Custom(ValidityError::InvalidNonZeroProviderDelegations as u8)
		);
```

wilwade (changes requested):
Two small things and one optimization.

pallets/msa/src/lib.rs
rlaferla: Fixed in next commit.

pallets/msa/src/lib.rs
rlaferla: I think this needs to be `Ok(())`

wilwade (approved):
🌟

pallets/msa/src/lib.rs
```diff
@@ -862,7 +860,7 @@ pub mod pallet {
 					);
 				},
 			}
-			Ok(Some(T::WeightInfo::retire_msa(num_deletions)).into())
```
enddynayn, line 865: 🙌

pallets/msa/src/weights.rs
```diff
@@ -107,26 +107,23 @@ impl<T: frame_system::Config> WeightInfo for SubstrateWeight<T> {
...
 	// Storage: Msa PublicKeyCountForMsaId (r:1 w:1)
-	// Storage: Msa DelegatorAndProviderToDelegation (r:0 w:3)
-	fn retire_msa(s: u32, ) -> Weight {
-		Weight::from_ref_time(26_276_224 as u64)
-			// Standard Error: 8_763
-			.saturating_add(Weight::from_ref_time(969_676 as u64).saturating_mul(s as u64))
+	fn retire_msa() -> Weight {
+		Weight::from_ref_time(26_981_000 as u64)
 			.saturating_add(T::DbWeight::get().reads(2 as u64))
-			.saturating_add(T::DbWeight::get().writes((1 as u64).saturating_mul(s as u64)))
```
enddynayn, line 129: 🙌

common/primitives/src/msa.rs
```diff
@@ -16,9 +16,6 @@ pub use crate::schema::SchemaId;
 /// Message Source Id or msaId is the unique identifier for Message Source Accounts
 pub type MessageSourceId = u64;
 
-/// Maximum # of providers per delegator for deletion/benchmarking use
-pub const EXPECTED_MAX_NUMBER_OF_PROVIDERS_PER_DELEGATOR: u32 = 128;
```
enddynayn, line 20: 🙌

pallets/msa/src/lib.rs
```diff
@@ -1655,6 +1643,16 @@ impl<T: Config + Send + Sync> CheckFreeExtrinsicUse<T> {
 			key_count == 1,
 			InvalidTransaction::Custom(ValidityError::InvalidMoreThanOneKeyExists as u8)
 		);
+
+		let has_delegations: bool =
+			DelegatorAndProviderToDelegation::<T>::iter_key_prefix(DelegatorId(msa_id))
```
enddynayn, line 1648: ohh are you sure this does not invoke a read per item, which is probably not being caught by the benchmarks?

enddynayn, line 1648: ohh are you sure this does not invoke a read per item, which is probably not being caught by the benchmarks?

rlaferla: This was @wilwade's addition.  What do you think @wilwade ?

enddynayn: It might not be a bad idea to update this storage to a map count thing that Wil found.

rlaferla: Agreed but that is dependent on upgrading Polkadot.  I think it's in 0.9.38.

enddynayn (approved):
Nice! 💯

pallets/msa/src/lib.rs
rlaferla: After reviewing `count()` vs `any()` with @enddynayn,` any()` does one additional read while count calls `next()` repeatedly, counting the number of iterations.

PR #1156, opened 2023-02-28
Stateful Strorage

### LINKED ISSUE #1153: private graph branch final clean up
### Problem
We will need a final clean up before merging this branch into main

### Acceptance criteria

- [x]  Check and Add detailed documentations and cost of execution like big O notation if makes sense
- [x] Verify benchmarks and weights
- [x] Verify Spec version
- [x] Verify and update design docs
- [x] Verify and clean up tests if possible
- [x] Check all scripts

aramikm:
# Goal
The goal of this PR is to get feedback for private graph branch before merging into main.

Closes #1153 

# Main parts
- Schema pallet updates to support new properties
- Schema pallet storage migrations
- Stateful pallet implementation
- Child Tree utility
- Overhead benchmark support

# Checklist
- [x] Chain spec updated
- [x] Design doc(s) updated
- [x] Tests added
- [x] Benchmarks added
- [x] Weights updated


.github/workflows/.polkadot-latest-full-release.txt
```diff
@@ -0,0 +1 @@
+v0.9.37
```
wilwade, line 1: File can be removed.

.github/workflows/misc/polkadot-latest-full-release.txt
```diff
@@ -0,0 +1 @@
+v0.9.37
```
wilwade, line 1: File can be removed.

.github/workflows/track-polkadot-releases.yml
```diff
@@ -0,0 +1,42 @@
+name: Track Polkadot Releases
```
wilwade, line 1: File can be removed.

common/primitives/src/macros.rs
```diff
@@ -0,0 +1,76 @@
+// This file is part of Substrate.
```
wilwade, line 1: Is this actually copied over from Substrate?

pallets/stateful-storage/src/rpc/src/lib.rs
```diff
@@ -0,0 +1,109 @@
...
+use sp_runtime::{generic::BlockId, traits::Block as BlockT, DispatchError};
+use sp_std::vec::Vec;
+use std::sync::Arc;
+
+#[cfg(test)]
+mod tests;
+
+/// Frequency Stateful Storage Custom RPC API
+#[rpc(client, server)]
+pub trait StatefulStorageApi<BlockHash> {
```
wilwade, line 33: I doubt it, but would it be possible to fold this into the messages rpc? Not just add the endpoints, but add the responses to the existing `messages.getBySchemaId` endpoint?

pallets/stateful-storage/src/lib.rs
```diff
@@ -0,0 +1,776 @@
...
+//!
+//! - Appending items in an **itemized** model
+//! - Removing items in an **itemized** model
+//! - Upserting items in a **paginated** model
+//! - Removing pages in a **paginated**  model
+//!
+//! ## Terminology
+//! * **Item:** Data payload mapping to a defined schema
+//! * **Page:** Block of on-chain data of a fixed size, containing one (Paginated model) or more (Itemized model) **items**.
+//!
```
wilwade, line 33: Be nice to have the quick links like we do on the other pallets.

pallets/schemas/src/migrations.rs
```diff
@@ -0,0 +1,106 @@
...
+	BoundedVec,
+};
+use scale_info::TypeInfo;
+use sp_std::marker::PhantomData;
+
+/// Migrations for the schemas pallet.
+/// Following migrations are required:
+/// - Adding settings to the Schema struct
+/// Note: Post migration, this file should be deleted.
+pub mod v1 {
```
wilwade, line 20:
```suggestion
pub mod v0 {
```

pallets/msa/Cargo.toml
```diff
@@ -54,6 +54,6 @@ std = [
   "sp-core/std",
   "pallet-schemas/std",
   "common-primitives/std",
-  "common-runtime/std",
+	"common-runtime/std",
```
wilwade, line 57: Feels like a typo or merge typo

runtime/common/src/weights/mod.rs
```diff
@@ -32,5 +32,4 @@ pub mod pallet_timestamp;
 pub mod pallet_treasury;
 pub mod pallet_utility;
 
-pub use block_weights::constants::BlockExecutionWeight;
-pub use extrinsic_weights::constants::ExtrinsicBaseWeight;
+pub use frame_support::weights::constants::{BlockExecutionWeight, ExtrinsicBaseWeight};
```
wilwade, line 35: Don't we want this to be our new weights?
- runtime/common/src/weights/block_weights.rs
- runtime/common/src/weights/extrinsic_weights.rs

```suggestion
pub mod block_weights;
pub mod extrinsic_weights;
```

runtime/frequency/src/lib.rs
```diff
@@ -170,7 +172,11 @@ pub type UncheckedExtrinsic =
 pub type CheckedExtrinsic = generic::CheckedExtrinsic<AccountId, RuntimeCall, SignedExtra>;
 
 /// Migrations for Frequency
-pub type Migrations = (remove_sudo::RemoveSudo, pallet_msa::migration::Migration<Runtime>);
+pub type Migrations = (
+	remove_sudo::RemoveSudo,
+	pallet_msa::migration::Migration<Runtime>,
+	pallet_schemas::migrations::SchemaMigrationToV1<Runtime>,
```
wilwade, line 178: Might want to follow the pallet_msa pattern for single migration block that calls the versioned ones.

scripts/kill_freq.sh
```diff
@@ -1,11 +1,11 @@
-#!/bin/zsh
+#!/bin/bash
```
wilwade, line 1: 🎉

scripts/run_integration_tests.sh
```diff
@@ -1,5 +1,27 @@
 #!/usr/bin/env bash
 
+function get_frequency_pid () {
```
wilwade, line 3: @wilwade Create PR to update to support load testing again.

wilwade (commented):
Few things looking for merge issues more than anything else

pallets/stateful-storage/src/lib.rs
```diff
@@ -0,0 +1,776 @@
...
+				schema_id,
+				page_id,
+				prev_content_hash,
+			});
+		}
+
+		Ok(())
+	}
+
+	#[cfg(feature = "runtime-benchmarks")]
```
wilwade, line 746: These two functions should be able to be moved into the benchmarks.rs file.

pallets/stateful-storage/src/tests.rs
```diff
@@ -0,0 +1,2533 @@
...
+				owner_signature,
+				payload
+			),
+			Error::<Test>::ProofHasExpired
+		)
+	})
+}
+
+#[test]
+fn upsert_page_with_signature_having_out_of_window_payload_should_fail() {
```
wilwade, line 1720: I don't see a specific test for a replay. That might be a good one to have.

pallets/stateful-storage/src/tests.rs
```diff
@@ -0,0 +1,2533 @@
...
+
+		// act
+		assert_err!(
+			StatefulStoragePallet::upsert_page_with_signature(
+				RuntimeOrigin::signed(caller_1),
+				delegator_key.into(),
+				owner_signature,
+				payload
+			),
+			Error::<Test>::InvalidSignature
```
wilwade, line 1783: Is this situation possible?

1. Signature 0 to approve change to `state Null -> A`
2. Signature 1 to approve change to `state A -> B`
3. Signature 2 to approve change to back to A: `state B -> A`
4. Reuse signature 1 (`state A->B`) to force the user to move back again to `state B`

scripts/run_integration_tests.sh
wilwade: @JoeCap08055 https://github.com/LibertyDSNP/frequency/pull/1197

pallets/stateful-storage/src/tests.rs
JoeCap08055: I have a draft PR up with the SignedExtrinsic stuff, that will need to be merged before we do this final merge to `main` (unless we just want to merge this & I'll rebase my PR off of `main`). I need to do a bit of test cleanup/rework in there anyway, I can add one for replay.

pallets/stateful-storage/src/tests.rs
JoeCap08055: That's why we need the ringbuffer. I think we need the same replay prevention logic that's in the msa pallet. Or, probably, given the way things will be structured with Capacity, maybe we need to refactor the signature replay so it can be easily shared.

pallets/stateful-storage/src/tests.rs
aramikm: 1 step replay is kinda tested with different hash values. If it has the same hash value it means the page didn't change.

pallets/stateful-storage/src/tests.rs
aramikm: This is a possibility if two signatures 1 and 2 are signed close to each other. We can reduce the possibility by choosing a narrower signature validation window. 
I think this should be a rare case since you would need all of following conditions to happen:

1. two signatures signed inside defined mortality window.
2. second signature should happen after first signature since it should know the hash. (or can predict it)
3. second signature should reverse the state to exactly a previous version inside the mortality window.

I think due to the nature of data in graph we can get away with it. This might be more important for itemized keys but since we are applying append only for those it can not go back to a previous state.

Anyway I don't think adding additional DB reads and writes to handle this is feasible.

pallets/stateful-storage/src/types.rs
```diff
@@ -0,0 +1,273 @@
...
+impl<PageDataSize: Get<u32>> Page<PageDataSize> {
+	pub fn is_empty(&self) -> bool {
+		self.data.is_empty()
+	}
+
+	pub fn get_hash(&self) -> PageHash {
+		if self.is_empty() {
+			return PageHash::default()
+		}
+		let mut hasher = XxHash32::with_seed(0);
```
wilwade, line 164: Is it worth switching to XXH3 and then just taking the first 32 bits of it? It is a lot faster. https://crates.io/crates/twox-hash and https://cyan4973.github.io/xxHash/

pallets/stateful-storage/src/tests.rs
wilwade: I agree that this doesn't need as much of a protection as account control does like in the msa pallet. Also anything that would be problematic to reuse a signature for should use the Append Only permission.

I think the options are:
1. Leave as is. Some schemas might not have this concern or can use append only if they do.
2. Skip the requirement of ability to calculate the state hash off-chain and add the last edited block number to the data. (once per page and once for the entire itemized list). (Allows add page->delete page->add page replay still, but removes all others)
3. Almost the same as 2, but use a nonce instead of the hash and don't use a hash at all. Just use the nonce. (Allows add page->delete page->add page replay still, but removes all others)
4. (I agree with Aramik, not an option from performance requirements) Use a separate piece of state to handle it.
5. Require signature would also require append only.
6. (Already rejected. Causes any state change to be dependent on others and doesn't work with the user-provider model) Add a nonce to the msa or such.

I wonder if 3 gives us any other improvements in that we don't have to hash the data, but that's a tiny amount of savings.

I think I am currently leaning toward 1, but maybe 5 or 3? Curious if there are other ideas.

pallets/stateful-storage/src/rpc/src/lib.rs
aramikm: there are a number of challenges with trying to use the same rpc for getting messages and storages.
- Messages are schema oriented, which means we only need schema but storages are msa_id and schema oriented.
- Messages response types are very different from storage since we have pagination on them but storages right not are not paginated since the assumption is that they can not grow that big (might be a wrong assumption)
- Rust would not allow us to return different types unless we change the rpc to return an enum and define those different response types in there.

We can always unify them after merge since that would need to be a separate new endpoint and existing message rpc endpoint would not support it without changes which would break backwards compatibility.

runtime/common/src/weights/mod.rs
aramikm: done

pallets/stateful-storage/src/types.rs
aramikm: Modified to use xxh3 first 4 bytes

runtime/frequency/src/lib.rs
aramikm: @saraswatpuneet you might know better about this one

pallets/stateful-storage/src/tests.rs
aramikm: there was a conversation with @wesbiggs about using block number or hash and hash was superior based on some scenarios. One of them was the following:

> Wes Biggs:
> The only problem I see with block number from the DSNP perspective is that an application that is making a series of 
> writes over the same session (but in different blocks) will not know what block its previous change was part of (without always querying for state before submitting any updates). 
> OK yeah, we might be talking about different problems. Here's my scenario:
> Alice uses DSNP through two different providers' apps. Both apps allow her to manipulate her social graph, and she uses them interchangeably to do so. We want to ensure that if the cached graph in either app gets stale, this condition will be detected at write time and not allow changes made in App 1 to overwrite changes that were made, unbeknownst to App 1, in App 2.
> 1. [Block 0] Suppose Alice's starting graph contains { Bob, Carol }
> 2. [Block 1] Alice opens App 1. Provider 1 fetches Alice's current graph, and caches { Bob, Carol }
> 3. [Block 2] Alice opens App 2. Provider 2 fetches Alice's current graph, and caches { Bob, Carol }
> 4. [Block 3] Alice adds a connection to Dan in App 2. App 2's cache now has { Bob, Carol, Dan }. Provider 2 writes { Bob, Carol, Dan } to the chain.
> 5. [Block 4] Alice switches back to App 1. She adds a connection to Erin. App 1's cache now has { Bob, Carol, Erin }. App 2 attempts to write { Bob, Carol, Erin } to the chain.

> We want the chain to reject 5 because its preimage, { Bob, Carol } is out of sync.

aramikm: > Require signature would also require append only.

Having append only for paginated schemas it a bit weird. For current use cases it seems fine since we need both of them for adding keys but I can see that in future we might be able to use signature-required ones even for paginated for cost shifting or any other reason.

runtime/frequency/src/lib.rs
saraswatpuneet: It is very similar, just the naming is such, I will just name it to `SchemaMigration`

pallets/stateful-storage/src/tests.rs
JoeCap08055: @wilwade I agree--the current interpretation of `AppendOnly` for paginated is really `UpsertOnly`. Which doesn't really seem to provide a benefit (we can't delete a page, but we can overwrite it as empty...). Maybe we should enforce in the `schemas` pallet that `AppendOnly` can only be applied to `PayloadLocation::Itemized` and update the attribute documentation accordingly.

aramikm: Decided to spike using a nonce inside the page so that the overall state can not back to a previous one without page removal.

common/primitives/src/schema.rs
```diff
@@ -61,3 +88,27 @@ pub trait SchemaValidator<SchemaId> {
 	#[cfg(any(feature = "std", feature = "runtime-benchmarks", test))]
 	fn set_schema_count(n: SchemaId);
 }
+
+impl SchemaSettings {
+	/// Set all settings to disabled
```
shannonwells, line 93:
```suggestion
	/// Returns new SchemaSettings with all settings disabled
```

pallets/schemas/src/lib.rs
```diff
@@ -306,20 +324,54 @@ pub mod pallet {
...
+		#[pallet::call_index(4)]
+		#[pallet::weight(T::WeightInfo::create_schema(model.len() as u32 + settings.len() as u32))]
+		pub fn create_schema_with_settings(
+			origin: OriginFor<T>,
+			model: BoundedVec<u8, T::SchemaModelMaxBytesBoundedVecLimit>,
+			model_type: ModelType,
+			payload_location: PayloadLocation,
+			settings: BoundedVec<SchemaSetting, T::MaxSchemaSettingsPerSchema>,
+		) -> DispatchResult {
+			let sender = ensure_signed(origin)?;
```
shannonwells, line 368: Why is there a separate extrinsic for this, that allows anyone with enough token to create a schema?  It's the same as the governance-based one except for the origin requirement.  Was it left over from before all the governance-based schema creation was done?

pallets/schemas/src/tests.rs
```diff
@@ -516,3 +518,70 @@ fn dsnp_broadcast() {
 	);
 	assert_ok!(result);
 }
+
+#[test]
+fn create_schema_with_settings_should_work() {
```
shannonwells, line 523: Assuming the extrinsic isnt' supposed to be there any more, these tests can use `create_schema_via_governance`.

pallets/stateful-storage/src/rpc/src/tests/mod.rs
```diff
@@ -0,0 +1,136 @@
...
+					0,
+					msa_id,
+					schema_id,
+					SUCCESSFUL_PAYLOAD.to_vec(),
+					DUMMY_STATE_HASH)]),
+				_ => Err(DispatchError::Other("some error")),
+			}
+		}
+
+		fn get_itemized_storages(msa_id: MessageSourceId, schema_id: SchemaId) -> Result<ItemizedStoragePageResponse, DispatchError> {
```
shannonwells, line 32:
```suggestion
		fn get_itemized_storage(msa_id: MessageSourceId, schema_id: SchemaId) -> Result<ItemizedStoragePageResponse, DispatchError> {
```

shannonwells (changes requested):
I've only covered 55 of 71 files, but I wanted to put in my current comments before proceeding. 

I had a question about whether you should be doing saturated_adds in the #[pallet::weight(..)] instead of a + and was going to look it up myself before mentioning but I wanted to get some comments posted so here it is.

The main problem is the `create_schema_with_settings` extrinsic. I'm guessing it was in there just for development and hasn't been removed yet.

pallets/schemas/src/lib.rs
saraswatpuneet: We added this to not break downstream consumers like saas team.

To have a separate extrinsic to create schema with settings

And thank you bringing this, governance based schema creation was done while this work was ongoing.

Do we need to do anything else with this extrinsic? as this will also be called via governance.

pallets/schemas/src/tests.rs
saraswatpuneet: If we can get away without creating a new extrinsic and only updating create_schema function then all good, though that's a breaking change 👀

integration-tests/scaffolding/helpers.ts
```diff
@@ -49,6 +57,42 @@ export async function generateAddKeyPayload(payloadInputs: AddKeyData, expiratio
...
+
+  return {
+    expiration,
+    ...payload,
+  }
+}
+
+export async function generatePaginatedUpsertSignaturePayload(payloadInputs: PaginatedUpsertSignaturePayload, expirationOffset?: number): Promise<PaginatedUpsertSignaturePayload> {
+  let { expiration, ...payload } = payloadInputs;
+  if (!expiration) {
```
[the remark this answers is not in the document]
aramikm: Followed the existing pattern that we have for Msa payloads. I think the main idea behind it was to simplify getting the expiration block in payload creation but open to suggestions how we can simplify these

pallets/schemas/src/lib.rs
aramikm: As Puneet mentioned we are not that involved with the governance related changes but the main reason to add this was to not break backwards compatibility of existing extrinsic. Would like to know how this would work with governance

saraswatpuneet: Looking at code again, looks like create_schema_by_goverance and this new one, does the same thing, so we can get rid of this one

Will push a commit

integration-tests/scaffolding/extrinsicHelpers.ts
```diff
@@ -236,4 +245,37 @@ export class ExtrinsicHelper {
...
+
+    public static removePageWithSignature(delegatorKeys: KeyringPair, providerKeys: KeyringPair, signature: Sr25519Signature, payload: PaginatedDeleteSignaturePayload): Extrinsic {
+      return new Extrinsic(() => ExtrinsicHelper.api.tx.statefulStorage.deletePageWithSignature(delegatorKeys.publicKey, signature, payload), providerKeys, ExtrinsicHelper.api.events.statefulStorage.PaginatedPageDeleted);
+    }
+
+    public static upsertPageWithSignature(delegatorKeys: KeyringPair, providerKeys: KeyringPair, signature: Sr25519Signature, payload: PaginatedUpsertSignaturePayload): Extrinsic {
+      return new Extrinsic(() => ExtrinsicHelper.api.tx.statefulStorage.upsertPageWithSignature(delegatorKeys.publicKey, signature, payload), providerKeys, ExtrinsicHelper.api.events.statefulStorage.PaginatedPageUpdated);
+    }
+
+    public static getItemizedStorages(msa_id: MessageSourceId, schemaId: any): Promise<ItemizedStoragePageResponse> {
```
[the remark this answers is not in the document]
shannonwells: Actually I take it back. Apparently "storages" is all over the substrate repo, it's become a thing now so...what do I know

pallets/schemas/src/lib.rs
```diff
@@ -306,20 +324,54 @@ pub mod pallet {
...
 		#[pallet::call_index(3)]
-		#[pallet::weight(T::WeightInfo::create_schema_via_governance(model.len() as u32))]
+		#[pallet::weight(T::WeightInfo::create_schema_via_governance(model.len() as u32+ settings.len() as u32))]
 		pub fn create_schema_via_governance(
 			origin: OriginFor<T>,
 			creator_key: T::AccountId,
 			model: BoundedVec<u8, T::SchemaModelMaxBytesBoundedVecLimit>,
 			model_type: ModelType,
 			payload_location: PayloadLocation,
+			settings: BoundedVec<SchemaSetting, T::MaxSchemaSettingsPerSchema>,
```
saraswatpuneet, line 334: @shannonwells  this is also updated, if this is not a breaking change, we can drop the extrinsic below and keep this

pallets/stateful-storage/src/rpc/src/tests/mod.rs
shannonwells: Ignore me, apparently all the cool Substrate kids are using "storages" these days 🤷‍♀️

pallets/schemas/src/lib.rs
```diff
@@ -236,7 +247,12 @@ pub mod pallet {
 		) -> DispatchResult {
 			let sender = ensure_signed(origin)?;
```
shannonwells, line 248: `create_shema` is still here, it just creates a schema with empty settings, so I'm not sure what you mean backward compatibility. It seems like you have it already.

pallets/stateful-storage/src/lib.rs
```diff
@@ -0,0 +1,746 @@
...
+	/// # Errors
+	/// * [`Error::ItemExceedsMaxBlobSizeBytes`]
+	///
+	fn check_actions(
+		actions: &BoundedVec<ItemAction, T::MaxItemizedActionsCount>,
+	) -> DispatchResult {
+		ensure!(
+			actions.iter().all(|a| match a {
+				ItemAction::Add { data } =>
+					data.len() <= T::MaxItemizedBlobSizeBytes::get() as usize,
```
shannonwells, line 599: is this the size check I was looking for?

integration-tests/stateful-pallet-storage/handleAppendOnly.test.ts
```diff
@@ -0,0 +1,76 @@
...
+        }
+        const target_hash = await getCurrentItemizedHash(msa_id, itemizedSchemaId);
+
+        let add_actions = [add_action, update_action, delete_action];
+
+        let itemized_add_result_1 = ExtrinsicHelper.applyItemActions(providerKeys, itemizedSchemaId, msa_id, add_actions, target_hash);
+        await assert.rejects(async () => {
+          await itemized_add_result_1.fundAndSend();
+        }, {
+          name: 'SchemaNotSupported',
```
shannonwells, line 71: This returned error doesn't make sense to me in this case. I would've expected this to return `InvalidItemAction` instead. I see in the pallet code that's what it returns, it's just surprising.

pallets/schemas/src/lib.rs
aramikm: I think since this is only get triggered by governance, it should not not be considered breaking.

shannonwells (commented):
Moar comments & Qs

pallets/stateful-storage/src/lib.rs
```diff
@@ -0,0 +1,746 @@
...
+			_ => 0,
+			})
+		))]
+		pub fn apply_item_actions_with_signature(
+			origin: OriginFor<T>,
+			delegator_key: T::AccountId,
+			proof: MultiSignature,
+			payload: ItemizedSignaturePayload<T>,
+		) -> DispatchResult {
+			ensure_signed(origin)?;
```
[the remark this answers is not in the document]
JoeCap08055: We don't need to call `check_grants`, since we verify that the signer of the payload is the owner of the msa_id being acted upon.

Further question, though: do we want to additionally support a use case where a _delegate_ can issue a signed payload? Right now it looks like we don't.

integration-tests/stateful-pallet-storage/handleAppendOnly.test.ts
JoeCap08055: We're sending an action array that contains non-`Add` actions to an `AppendOnly` schema. The `check_schema` function returns `SchemaNotSupported`.

pallets/stateful-storage/src/types.rs
```diff
@@ -0,0 +1,275 @@
...
+			updated_page_buffer.extend_from_slice(slice);
+		}
+		updated_page_buffer.append(&mut add_buffer);
+
+		Page::<PageDataSize>::try_from(updated_page_buffer).map_err(|_| PageError::PageSizeOverflow)
+	}
+
+	/// Parses all the items inside an ItemPage
+	/// This has O(n) complexity when n is the number of all the bytes in that itemized storage
+	pub fn parse_as_itemized(&self, include_header: bool) -> Result<ParsedItemPage, PageError> {
```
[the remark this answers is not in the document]
saraswatpuneet: I think we missed out and yes it should be BoundedVec with limit set to the constant, good catch :+1:

pallets/stateful-storage/src/lib.rs
saraswatpuneet: Correct but I believe we should also reflect that in our ds using BoundedVec over page data size limit

pallets/stateful-storage/src/types.rs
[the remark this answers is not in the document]
aramikm: We do check all of them and we have tests for it. I tried adding BoundedVec on ItemAction but Rust didn't make it easy at the time so decided to go with a simpler Vec. But if we can make it work with BoundedVec that is preferable. We do check the item size in `check_actions`

pallets/stateful-storage/src/lib.rs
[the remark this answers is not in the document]
aramikm: One of the use-cases of these extrinisics is to allow **on-time delegation**. 

> Further question, though: do we want to additionally support a use case where a delegate can issue a signed payload? Right now it looks like we don't.

I don't think we need that at the moment but we can always add later if that scenario became important.

pallets/stateful-storage/src/lib.rs
```diff
@@ -0,0 +1,746 @@
...
+		/// since the signature of delegator is checked there is no need for delegation validation
+		#[pallet::call_index(4)]
+		#[pallet::weight(T::WeightInfo::upsert_page_with_signature(payload.payload.len() as u32))]
+		pub fn upsert_page_with_signature(
+			origin: OriginFor<T>,
+			delegator_key: T::AccountId,
+			proof: MultiSignature,
+			payload: PaginatedUpsertSignaturePayload<T>,
+		) -> DispatchResult {
+			ensure_signed(origin)?;
```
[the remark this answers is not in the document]
aramikm: Same explanation as the ther one

pallets/stateful-storage/src/lib.rs
```diff
@@ -0,0 +1,746 @@
...
+			)?;
+			Ok(())
+		}
+
+		/// Deletes a Paginated storage that requires signature
+		/// since the signature of delegator is checked there is no need for delegation validation
+		#[pallet::call_index(5)]
+		#[pallet::weight(T::WeightInfo::delete_page_with_signature())]
+		pub fn delete_page_with_signature(
+			origin: OriginFor<T>,
```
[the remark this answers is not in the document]
aramikm: Same explanation as the other one

pallets/schemas/src/lib.rs
aramikm: I'll schedule a huddle time to talk about your concerns here.

pallets/stateful-storage/src/lib.rs
aramikm: @saraswatpuneet  I'm not sure if I understand what you mean but we check the page size whenever we are applying those actions

pallets/stateful-storage/src/types.rs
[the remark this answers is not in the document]
saraswatpuneet: @aramikm  this is what I meant below , I am ok with going for BoundedVec but we check the size already.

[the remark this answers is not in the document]
shannonwells: Yes, I specifically meant the data size.

pallets/stateful-storage/src/lib.rs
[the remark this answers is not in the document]
shannonwells: More specifically I was asking if it's intended to allow these types of actions to be delegated to an MSA that is not a registered Provider.  It was clear that the signature itself and the payload are validated.

integration-tests/stateful-pallet-storage/handleAppendOnly.test.ts
shannonwells: Thanks @aramikm, it makes sense, I was coming at it from the standpoint of someone looking at their error, and there are a couple of places where that error crops up so I was thinking it could be good to have something more distinct in this case.

aramikm: Changed it to **UnsupportedOperationForSchema** if you think it still sucks please let me know what would you name it

pallets/stateful-storage/src/lib.rs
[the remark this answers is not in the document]
shannonwells: Okay, just making sure that was intentional.

shannonwells (approved):
Removing the "chagnes requested" given convo/comments and upcoming changes, and look at the new commits when they come in

pallets/stateful-storage/src/lib.rs
```diff
@@ -0,0 +1,797 @@
...
+		    0
+		))]
+		pub fn apply_item_actions(
+			origin: OriginFor<T>,
+			#[pallet::compact] state_owner_msa_id: MessageSourceId,
+			#[pallet::compact] schema_id: SchemaId,
+			#[pallet::compact] target_hash: PageHash,
+			actions: BoundedVec<ItemAction, T::MaxItemizedActionsCount>,
+		) -> DispatchResult {
+			let provider_key = ensure_signed(origin)?;
```
enddynayn, line 280: Maybe we do not call this provider_key and instead just key. This makes me think that the signer needs to be a provider.

pallets/stateful-storage/src/lib.rs
```diff
@@ -0,0 +1,797 @@
...
+			let is_pruning = actions.iter().any(|a| matches!(a, ItemAction::Delete { .. }));
+			Self::check_actions(&actions)?;
+			Self::check_schema(schema_id, PayloadLocation::Itemized, false, is_pruning)?;
+			Self::check_grants(provider_key, state_owner_msa_id, schema_id)?;
```
enddynayn, line 284: I think we should probably move this above check schema since you need an MSA to do anything. I might even move it up before check_actions but those do not require a db read.

pallets/stateful-storage/src/types.rs
```diff
@@ -0,0 +1,299 @@
...
+
+/// Defines the actions that can be applied to an Itemized storage
+#[derive(Clone, Encode, Decode, Debug, TypeInfo, MaxEncodedLen, PartialEq)]
+#[scale_info(skip_type_params(DataSize))]
+#[codec(mel_bound(DataSize: MaxEncodedLen))]
+pub enum ItemAction<DataSize: Get<u32> + Clone + sp_std::fmt::Debug + PartialEq> {
+	/// Adding new Item into page
+	Add {
+		/// The data to add
+		data: BoundedVec<u8, DataSize>,
```
enddynayn, line 47: Woot woot! 👍

pallets/stateful-storage/src/lib.rs
```diff
@@ -0,0 +1,777 @@
...
+		///
+		#[pallet::call_index(0)]
+		#[pallet::weight(T::WeightInfo::apply_item_actions( actions.len() as u32 ,
+			actions.iter().fold(0, |acc, a| acc.saturating_add(match a {
+				ItemAction::Add { data } => data.len() as u32,
+				_ => 0,
+			})),
+		    0
+		))]
+		pub fn apply_item_actions(
```
enddynayn, line 270: In the spirit of sql transactions were commit is called to save a transaction, perhaps, we use commit instead of apply.

pallets/stateful-storage/src/lib.rs
```diff
@@ -0,0 +1,777 @@
...
+	/// * [`Event::ItemizedPageDeleted`]
+	///
+	fn modify_itemized(
+		state_owner_msa_id: MessageSourceId,
+		schema_id: SchemaId,
+		target_hash: PageHash,
+		actions: BoundedVec<ItemAction<T::MaxItemizedBlobSizeBytes>, T::MaxItemizedActionsCount>,
+	) -> DispatchResult {
+		let key: ItemizedKey = (schema_id,);
+		let existing_page = StatefulChildTree::<T::KeyHasher>::try_read::<_, ItemizedPage<T>>(
```
enddynayn, line 637: it looks kinda nicer if we wrote like
```
let existing_page: ItemizedPage<T> = StatefulChildTree::<T::KeyHasher>::try_read(....)
```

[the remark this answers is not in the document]
enddynayn: same thing for getting a paginated page.

pallets/stateful-storage/src/types.rs
```diff
@@ -0,0 +1,299 @@
...
+	fn try_from(data: Vec<u8>) -> Result<Self, Self::Error> {
+		let bounded: BoundedVec<u8, PageDataSize> = BoundedVec::try_from(data).map_err(|_| ())?;
+		Ok(Page::from(bounded))
+	}
+}
+
+impl<PageDataSize: Get<u32>> Page<PageDataSize> {
+	/// Applies all actions to specified page and returns the updated page
+	/// This has O(n) complexity when n is the number of all the bytes in that itemized storage
+	pub fn apply_item_actions<T: Config>(
```
enddynayn, line 225: This is interesting. I do not see


pub type ItemizedPage<T> = Page<Config<Self = T>::MaxItemizedPageSizeBytes>

pub type PaginatedPage<T> = Page<Config<Self = T>::MaxPaginatedPageSizeBytes>


It seems that apply_item_actions is only used for Page<Config<Self = T>::MaxItemizedPageSizeBytes> however, it is only needed for ItemizedPage.

I think it might be easier to follow if we create a trait called something like
```
pub trait CommitActions {
 ...
}
```
And then we implemented on ItemizedPage.

By doing this we target only the type we want to implement that behavior.

pallets/stateful-storage/src/types.rs
```diff
@@ -0,0 +1,299 @@
...
+	}
+}
+
+impl<PageDataSize: Get<u32>> From<BoundedVec<u8, PageDataSize>> for Page<PageDataSize> {
+	fn from(bounded: BoundedVec<u8, PageDataSize>) -> Self {
+		Self { data: bounded }
+	}
+}
+
+impl<PageDataSize: Get<u32>> TryFrom<Vec<u8>> for Page<PageDataSize> {
```
enddynayn, line 213: This seems helpful do we want to feature flag since we do use it outside of test?

pallets/stateful-storage/src/types.rs
```diff
@@ -0,0 +1,299 @@
...
+		Ok(Page::from(bounded))
+	}
+}
+
+impl<PageDataSize: Get<u32>> Page<PageDataSize> {
+	/// Applies all actions to specified page and returns the updated page
+	/// This has O(n) complexity when n is the number of all the bytes in that itemized storage
+	pub fn apply_item_actions<T: Config>(
+		&self,
+		actions: &[ItemAction<T::MaxItemizedBlobSizeBytes>],
```
enddynayn, line 227: 🔥

pallets/stateful-storage/src/lib.rs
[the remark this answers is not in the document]
enddynayn: looks like you already have a function called `get_itemized_storage` you are just not using it.

pallets/stateful-storage/src/types.rs
enddynayn, line 227: should this be a bounded vector?
```
BoundedVec<
		ItemAction<T::MaxItemizedBlobSizeBytes>,
		T::MaxItemizedActionsCount,
	>
```

pallets/stateful-storage/src/types.rs
```diff
@@ -0,0 +1,299 @@
...
+	) -> Result<Self, PageError> {
+		let mut parsed = self.parse_as_itemized(true)?;
```
enddynayn, line 229: seems like this could go into the CommitActions or something else that is specific to ItemizedPage?

pallets/stateful-storage/src/types.rs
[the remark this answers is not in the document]
enddynayn: It would be nice to have some unit test for this. It would help me understand this a little better.

pallets/stateful-storage/src/types.rs
```diff
@@ -0,0 +1,299 @@
...
+
+/// Errors dedicated to parsing or modifying pages
+#[derive(Debug, PartialEq)]
+pub enum PageError {
+	/// Unable to decode the data in the item
+	ErrorParsing(&'static str),
+	/// Add or Delete Operation was not possible
+	InvalidAction(&'static str),
+	/// ItemPage count overflow catch
+	ArithmeticOverflow,
```
enddynayn, line 72: sp_runtime has ArithmeticError we use it in the MSA pallet.

https://github.com/LibertyDSNP/frequency/blob/db81e4fa142087d8b5cd37b181852f3bb8407803/pallets/msa/src/lib.rs#L985

pallets/stateful-storage/src/types.rs
```diff
@@ -0,0 +1,299 @@
...
+	/// Parses all the items inside an ItemPage
+	/// This has O(n) complexity when n is the number of all the bytes in that itemized storage
+	pub fn parse_as_itemized(&self, include_header: bool) -> Result<ParsedItemPage, PageError> {
+		let mut count = 0u16;
+		let mut items = BTreeMap::new();
+		let mut offset = 0;
+		while offset < self.data.len() {
+			ensure!(
+				offset + ItemHeader::max_encoded_len() <= self.data.len(),
+				PageError::ErrorParsing("wrong header size")
```
enddynayn, line 274: maybe we can remove these strings  and create another enum like
```
pub enum <Something>Error {
   InvalidHeaderSize,
   InvalidPayloadSize,
   HeaderDecodeFail,
} 
```

enddynayn:
Looks Great! I am 90% done going through it. I have no significant concerns! This is some fantastic work! You guys did it!

pallets/stateful-storage/src/types.rs
aramikm: Created a trait to handle itemized ones

pallets/stateful-storage/src/lib.rs
```diff
@@ -0,0 +1,778 @@
...
+	///
+	fn delete_paginated(
+		state_owner_msa_id: MessageSourceId,
+		schema_id: SchemaId,
+		page_id: PageId,
+		target_hash: PageHash,
+	) -> DispatchResult {
+		let keys: PaginatedKey = (schema_id, page_id);
+		if let Some(existing_page) =
+			StatefulChildTree::<T::KeyHasher>::try_read::<_, PaginatedPage<T>>(
```
enddynayn, line 752: We use this in a couple places. I think it would be a good idea to make helper methods.

pallets/stateful-storage/src/lib.rs
```diff
@@ -0,0 +1,778 @@
...
+		};
+		Ok(())
+	}
+
+	/// Modifies a page from paginated storage by provided new page
+	///
+	/// # Events
+	/// * [`Event::PaginatedPageUpdated`]
+	///
+	fn modify_paginated(
```
enddynayn, line 702: this this an update or create? What does modify mean?

pallets/stateful-storage/src/stateful_child_tree.rs
```diff
@@ -0,0 +1,395 @@
...
+use common_primitives::msa::MessageSourceId;
+use frame_support::{
+	storage::{child, child::ChildInfo, ChildTriePrefixIterator},
+	Blake2_128, Blake2_256, Hashable, StorageHasher, Twox128, Twox256,
+};
+use sp_core::{ConstU8, Get};
+use sp_io::hashing::twox_64;
+use sp_std::{fmt::Debug, prelude::*};
+
+pub trait MultipartKeyStorageHasher: StorageHasher {
```
enddynayn, line 13: pretty cool how you create this trait with a default and did implementations below.

pallets/stateful-storage/src/stateful_child_tree.rs
```diff
@@ -0,0 +1,395 @@
...
+
+	fn arity(&self) -> u8 {
+		Self::Arity::get()
+	}
+
+	fn hash(&self) -> Vec<u8>;
+	fn hash_prefix_only(&self) -> Vec<u8>;
+
+	fn decode(hash: &[u8]) -> Result<Self, codec::Error> {
+		let mut key_material = H::reverse(hash, Self::Arity::get());
```
enddynayn, line 61: Did you mean to use the function defined above in used here?
```
fn arity(&self) -> u8 {
		Self::Arity::get()
	}
```

pallets/stateful-storage/src/stateful_child_tree.rs
```diff
@@ -0,0 +1,395 @@
...
+}
+impl<T> MultipartStorageKeyPart for T where
+	T: Clone + Debug + Default + Decode + Encode + Eq + PartialEq + Hashable
+{
+}
+
+pub trait MultipartKey<H: MultipartKeyStorageHasher>: MultipartStorageKeyPart {
+	type Arity: Get<u8>;
+
+	fn arity(&self) -> u8 {
```
enddynayn, line 53: I do not think this is being used?

pallets/stateful-storage/src/stateful_child_tree.rs
```diff
@@ -0,0 +1,395 @@
...
+			.iter()
+			.chain(H::hash(&encoded_2).as_ref().iter())
+			.chain(H::hash(&encoded_3).as_ref().iter())
+			.chain(H::hash(&encoded_4).as_ref().iter())
+			.cloned()
+			.collect::<Vec<_>>()
+	}
+}
+
+pub trait IsTuplePrefix<H: MultipartKeyStorageHasher, T: MultipartStorageKeyPart>:
```
enddynayn, line 222: do we need to add docs for these?

pallets/stateful-storage/src/stateful_child_tree.rs
```diff
@@ -0,0 +1,395 @@
...
+pub trait MultipartStorageKeyPart:
+	Clone + Debug + Default + Decode + Encode + Eq + PartialEq + Hashable
+{
+}
+impl<T> MultipartStorageKeyPart for T where
+	T: Clone + Debug + Default + Decode + Encode + Eq + PartialEq + Hashable
+{
+}
+
+pub trait MultipartKey<H: MultipartKeyStorageHasher>: MultipartStorageKeyPart {
```
enddynayn, line 50: do we need to add docs for these?

pallets/stateful-storage/src/stateful_child_tree.rs
```diff
@@ -0,0 +1,395 @@
...
+	type Arity: Get<u8>;
```
enddynayn, line 51: In Cs arity means:
> in logic, mathematics, and computer science, arity is the number of arguments or operands taken by a function, operation or relation. 

I am trying to draw the connection between the definition and use. Can you please help understand?

pallets/stateful-storage/src/tests.rs
aramikm: Fixed via https://github.com/LibertyDSNP/frequency/pull/1220

pallets/stateful-storage/src/stateful_child_tree.rs
enddynayn: maybe we can remove this from here and use the `arity` function to implement what it needs.

aramikm: Arity is the length of the tuple as you mentioned. It's the number of keys. Rust allows different kinds of implementations for doing the same thing. What is the advantage of changing this to a function?

pallets/stateful-storage/src/stateful_child_tree.rs
aramikm: nah, it wouldn't work with self. Removed the function

pallets/stateful-storage/src/types.rs
aramikm: It does but we can't use it here, We want to return All errors under the same Enum

pallets/stateful-storage/src/types.rs
aramikm: Nope. It's a slice because it's a best practice to use slices if possible.

pallets/stateful-storage/src/types.rs
aramikm: Thought about that but then I'll have to create another enum for actions also and include both in PageError enum. Not sure if it worths the changes. I also print these in case of failure since it will allow us to know what is a wrong with storage later.

scripts/run_all_benchmarks.sh
```diff
@@ -41,5 +226,12 @@ for pallet_name in "${CUSTOM_PALLETS[@]}"; do
...
-  run_benchmark pallet_${pallet_name} ${steps} ${repeat} ${output} ${template} || exit_err
+  run_benchmark ${pallet_name} ${steps} ${repeat} ${output} ${template} || exit_err
 done
+
+if [[ -n "${OVERHEAD}" ]]
+then
+  echo "Running extrinsic and block overhead benchmark"
+  echo " "
+  ${BENCHMARK} overhead --execution=wasm --wasm-execution=compiled --weight-path=runtime/common/src/weights --chain=dev --warmup=10 --repeat=100 || exit_err
+fi
```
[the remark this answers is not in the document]
aramikm: I think we can do this change in another PR since it is not directly related to private graph and it's over improvement for benchmarking. 
Created an issue to track this https://github.com/LibertyDSNP/frequency/issues/1231

pallets/stateful-storage/src/lib.rs
enddynayn: This is like a transaction where we are doing multiple operations. In SQL commit is what is used to save transactions.

pallets/stateful-storage/src/types.rs
enddynayn: I am not sure why we are deviating from what we do everywhere else?

pallets/stateful-storage/src/stateful_child_tree.rs
enddynayn: It's minor; however, notice that when the trait is implemented on a tuple, you also provide the tuple length. The tuple you implemented this trait knows its own length.

pallets/stateful-storage/src/lib.rs
aramikm: All extrinsic calls are transactions anyway if I remember correctly. I don't see any value in indicating the obvious.

pallets/stateful-storage/src/types.rs
aramikm: Added a trait for this

pallets/stateful-storage/src/types.rs
aramikm: Because we are trying to create self contained structs that can be tested in isolation. That's why the parsing logic is implemented here which is inside the types.

Makefile
```diff
@@ -86,6 +86,12 @@ benchmarks-messages:
 benchmarks-schemas:
 	./scripts/run_benchmark.sh -p schemas
 
+benchmarks-stateful-storage:
+	./scripts/run_benchmark.sh -p stateful-storage
+
+benchmarks-overhead:
+	./scripts/run_benchmark.sh -p overhead
+
```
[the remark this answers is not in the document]
aramikm: tracking the issue in https://github.com/LibertyDSNP/frequency/issues/1231

pallets/stateful-storage/src/lib.rs
enddynayn: The function name might be misleading or doing too much. Because it also checks to see if the key is associated with a valid MSA id. This is the root of being able to do anything.

```
let provider_msa_id = T::MsaInfoProvider::ensure_valid_msa_key(&provider_key)
			.map_err(|_| Error::<T>::InvalidMessageSourceAccount)?;
```

pallets/stateful-storage/src/stateful_child_tree.rs
aramikm: This is Joe's implementation so I'm not sure why he has chosen this approach but from my understanding you can not get the length of a tuple

pallets/stateful-storage/src/lib.rs
enddynayn: What could be more obvious is that you have a batch of operations users wants to perform. We can write this in a manner in which we discard the failed ones and continue with the rest of the operations. So it could be more evident that these actions are transactional. Recall that I had to ask you this question myself? I was not sure what actions would be applied. For this reason, it's not apparent. 

However, it's not a big deal; I only recommended this to align more with db operations. It's good as is.

aramikm: That is an issue that we seems like we should handle by adding more documentation.

pallets/stateful-storage/src/types.rs
enddynayn: I think thats good thinking. Something I just noticed is that maybe these errors are not reaching their full potential since they all get swallowed into corrupted state CorruptedState. Maybe we can put this error messages to work. Even if we had an arithemtic error we would not know or a parsing error or anything bad that happens here.

Perhaps it would help us is we had something along the lines of 
```
 pub enum Error<T> {
      CorruptedState(PageError | WhataverError)
     }
```

pallets/stateful-storage/src/types.rs
enddynayn: Yea your right I guess its not a big deal here since it does not need to care about constraints.

pallets/stateful-storage/src/types.rs
aramikm: We are printing all the parsing related errors. I think we can print the actions. also but that's less crucial since it will indicate an erroneous input. The parsing problem is way more important that's why we are printing.
```
			ItemizedOperations::<T>::apply_item_actions(&existing_page, &actions[..]).map_err(
				|e| match e {
					PageError::ErrorParsing(err) => {
						log::warn!(
							"failed parsing Itemized msa={:?} schema_id={:?} {:?}",
							state_owner_msa_id,
							schema_id,
							err
						);
						Error::<T>::CorruptedState
					},
					_ => Error::<T>::InvalidItemAction,
				},
			)?;
```

PR #1160, opened 2023-03-01
Overhead benchmarking

JoeCap08055:
# Goal
The goal of this PR is to add overhead benchmarking capability to Frequency.

# Checklist
- [ ] Chain spec updated
- [ ] Design doc(s) updated
- [ ] Tests added
- [x] Benchmarks added
- [x] Weights updated


saraswatpuneet (approved):
:1st_place_medal:

node/cli/src/benchmarking.rs
```diff
@@ -0,0 +1,186 @@
...
+
+/// Generates inherent data for the `benchmark overhead` command.
+///
+/// Note: Should only be used for benchmarking.
+pub fn inherent_benchmark_data() -> Result<InherentData> {
+	let mut inherent_data = InherentData::new();
+	let d = Duration::from_millis(0);
+	let timestamp = sp_timestamp::InherentDataProvider::new(d.into());
+	let mock_para_inherent_provider =
+		cumulus_primitives_parachain_inherent::MockValidationDataInherentDataProvider {
```
aramikm, line 167: Inherents are part of the block header overhead. Just want to make sure mocking the inherent validator would not have any real effect on weights.

aramikm (approved):
Looks great! Awesome work. Added one question.

node/cli/src/benchmarking.rs
JoeCap08055: Good point. There's more research to be done; this is currently a first pass for a working solution. I'm going to sync from pg-main & research while cleaning up. Worst case, we'll merge this and open another issue to track using real inherent data instead of mocking.

JoeCap08055: @aramikm -- looking at this, I don't think it will impact the overhead extrinsic benchmark. We're not mocking the validator, only the inherent data from the relay chain that gets validated (but not the validation code). The main goal of the overhead benchmark is to get the overhead involved in executing all of the signed extensions involved in an extrinsic call, which is still being done. I think it's OK, but we can ask Parity for confirmation.

Issue #1163, opened 2023-03-02
[Bug] Should allow re-delegation


wilwade:
This scenario should work:

1. User A delegates to Provider X with Schema Ids 1, 2, and 3
2. User A revokes all delegation to Provider X at Block Number 100
3. User A delegates to Provider X again with Schema Ids 2, 3, and 4

Expected Final State:
- User A has a Delegation with Provider X
- Schema 2, 3, and 4 are granted
- Schema 1 is marked as revoked at Block Number 100

Actual Final State:
- ✅ User A has a Delegation with Provider X
-  ✅ Schema 4 is granted
-  ✅ Schema 1 is marked as revoked at Block Number 100
- ❌ Schema 2 and 3 are still marked as revoked as of block number 100

_Originally posted by @wilwade in https://github.com/LibertyDSNP/frequency/issues/876#issuecomment-1448981869_

rlaferla:
The issue here is conceptual and the complexity stems from the fact that delegation/schema permissions are a contract between the provider and delegator.  This contract not only governs the relationship going forward but also historically because the blockchain is a timeline based on block numbers.  

Therefore by allowing re-delegation, it implies that the provider is not only granted permission to use a particular schema going forward from that block but also historically from block 1.

I can change this behavior but we should consider an alternative (clearer and more flexible) design which allows for a history of contracts where a contract is enacted at a specific block number.  This means that the latest contract is enforced from that block forward.

rlaferla:
From my tests, it appears that the current implementation works as intended.

PR #1164, opened 2023-03-02
Fix for re-delegation of schema permissions

rlaferla:
- Fixed re-delegation of schema permissions (updated grant vs inserted grant)
- Expanded create_sign_add_provider_payload_with_schemas to include an expiration block
- Updated test to use example in issue #1163

# Goal
The goal of this PR is to fix re-delegation of schema permissions (existing but not newly inserted)

Closes #1160


shannonwells (approved):
Do it to it

wilwade (approved):

PR #1166, opened 2023-03-03
Chore/renames and fix bug

shannonwells:
# Goal
The goals of this PR:
1. Fix a DoS vulnerability/bug
2. Push some logic into CapacityDetails
3. Rename some functions and members to be clearer and self-consistent.

Closes 

# Discussion
The `deposit` function was really a "replenish_by_amount" function when I would have thought it would call `CapacityDetails::deposit`.  It was only ever being used by `replenish_by_amount` anyway. 

Additionally, a lot of work was being done outside of `CapacityDetails` struct that it ought to do. I added some functions that have the same name as the functions we want in Replenishable.

Now the Capacity impl of `Replenishable`, all each function does is try to fetch the `CapacityDetails` for the MSA, then calls the function of the same name on the retrieved `CapacityDetails`, and sets the new value in storage.

In the process of updating these things, I found a bug in `deposit` which created a DoS vulnerability, which is that we should not be updating "`last_replenished_at`" on a deposit.  A deposit is done when someone stakes for a target.  By touching "`last_replenished_at`" on a deposit, it behaves the same as "`replenish_all`" in that sense.  This created a DoS vulnerability:
1. Epoch N, Provider has used up all their capacity. 
2. User stakes minimum amount and targets Provider MSA 1.
3. Provider is effectively replenished by minimum amount.
4. Epoch N+1 begins
5. Immediately afterward, User stakes just a tiny bit more - but less than needed for a message, updating "`last_replenished_at`" to the current epoch.
6. Provider tries to send a message and cannot because they don't have enough Capacity, and they will not be replenished because the current epoch is the same as "`last_replenished_at`."
7. Repeat steps 4-6 ad infinitum

# Checklist
- [x] Design doc(s) updated
- [x] Tests added

# N/A
- Chain spec updated
- Benchmarks added
- Weights updated

### DIFF
--- a/common/primitives/src/capacity.rs
+++ b/common/primitives/src/capacity.rs
@@ -24,7 +24,7 @@ pub trait Nontransferable {
 	fn balance(msa_id: MessageSourceId) -> Self::Balance;
 
 	/// Reduce Capacity of an MSA account by amount.
-	fn withdraw(msa_id: MessageSourceId, amount: Self::Balance) -> Result<(), DispatchError>;
+	fn deduct(msa_id: MessageSourceId, amount: Self::Balance) -> Result<(), DispatchError>;
 
 	/// Increase Capacity of an MSA account by an amount.
 	fn deposit(msa_id: MessageSourceId, amount: Self::Balance) -> Result<(), DispatchError>;
--- a/pallets/capacity/src/helpers_tests.rs
+++ b/pallets/capacity/src/helpers_tests.rs
@@ -169,8 +169,8 @@ fn increase_stake_and_issue_capacity_is_successful() {
 
 		let capacity_details = Capacity::get_capacity_for(&target).unwrap();
 
-		assert_eq!(capacity_details.remaining, 55);
-		assert_eq!(capacity_details.total_available, 55);
+		assert_eq!(capacity_details.remaining_capacity, 55);
+		assert_eq!(capacity_details.total_capacity_issued, 55);
 		assert_eq!(capacity_details.last_replenished_epoch, 0);
 
 		let target_details = Capacity::get_target_for(&staker, &target).unwrap();
@@ -223,9 +223,9 @@ fn set_capacity_details_is_successful() {
 
 		let capacity_details: CapacityDetails<BalanceOf<Test>, <Test as Config>::EpochNumber> =
 			CapacityDetails {
-				remaining: 10u64,
+				remaining_capacity: 10u64,
 				total_tokens_staked: 10u64,
-				total_available: 10u64,
+				total_capacity_issued: 10u64,
 				last_replenished_epoch: 1u32,
 			};
 
@@ -233,8 +233,8 @@ fn set_capacity_details_is_successful() {
 
 		let stored_capacity_details = Capacity::get_capacity_for(target).unwrap();
 
-		assert_eq!(stored_capacity_details.remaining, 10);
-		assert_eq!(stored_capacity_details.total_available, 10);
+		assert_eq!(stored_capacity_details.remaining_capacity, 10);
+		assert_eq!(stored_capacity_details.total_capacity_issued, 10);
 		assert_eq!(stored_capacity_details.last_replenished_epoch, 1);
 	});
 }
@@ -270,12 +270,12 @@ fn calculate_capacity_reduction_determines_the_correct_capacity_reduction_amount
 fn impl_balance_is_successful() {
 	new_test_ext().execute_with(|| {
 		let target_msa_id = 1;
-		let remainging_amount = 10u32;
+		let remaining_amount = 10u32;
 		let total_available_amount = 10u32;
 
 		let _ = create_capacity_account_and_fund(
 			target_msa_id,
-			remainging_amount.into(),
+			remaining_amount.into(),
 			total_available_amount.into(),
 			1u32,
 		);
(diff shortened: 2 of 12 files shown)


pallets/capacity/src/helpers_tests.rs
```diff
@@ -395,37 +395,6 @@ fn impl_deposit_errors_target_capacity_not_found() {
 	});
 }
 
-#[test]
-fn impl_deposit_errors_increase_exceeds_available_when_trying_deposit_a_larger_amount_than_the_available_capacity(
```
shannonwells, line 399: this is no longer applicable

pallets/msa/src/weights.rs
```diff
@@ -64,7 +64,7 @@ pub trait WeightInfo {
 	fn create_provider() -> Weight;
 	fn create_provider_via_governance() -> Weight;
 	fn propose_to_be_provider() -> Weight;
-	fn on_initialize(m: u32, ) -> Weight;
+	fn  on_initialize(m: u32, ) -> Weight;
```
johnmela, line 67: Was this extra space intended?

pallets/capacity/src/types.rs
```diff
@@ -158,40 +158,73 @@ impl<Balance: Saturating + Copy + CheckedAdd> StakingTargetDetails<Balance> {
 #[derive(PartialEq, Eq, Clone, Default, Encode, Decode, RuntimeDebug, TypeInfo, MaxEncodedLen)]
 pub struct CapacityDetails<Balance, EpochNumber> {
 	/// The Capacity remaining for the `last_replenished_epoch`.
-	pub remaining: Balance,
+	pub remaining_capacity: Balance,
 	/// The amount of tokens staked to an MSA.
 	pub total_tokens_staked: Balance,
 	/// The total Capacity issued to an MSA.
-	pub total_available: Balance,
+	pub total_capacity_issued: Balance,
```
johnmela, line 165: 🙌

johnmela (changes requested):
That's a great catch! 
My inline comments are non-blocking however, requesting changes for the failing test.

pallets/msa/src/weights.rs
shannonwells: nope, it's fixed in my local.

shannonwells:
@johnmela ah yeah, that's what I get for running tests only on capacity-pallet.

common/primitives/src/capacity.rs
```diff
@@ -24,7 +24,7 @@ pub trait Nontransferable {
 	fn balance(msa_id: MessageSourceId) -> Self::Balance;
 
 	/// Reduce Capacity of an MSA account by amount.
-	fn withdraw(msa_id: MessageSourceId, amount: Self::Balance) -> Result<(), DispatchError>;
+	fn deduct(msa_id: MessageSourceId, amount: Self::Balance) -> Result<(), DispatchError>;
```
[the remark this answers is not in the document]
shannonwells: done

johnmela (approved):
Fantastic!

common/primitives/src/capacity.rs
enddynayn, line 27: FYI, in the fungible trait that parity is swapping to, they call it decrease_balance. I am good with this thought.

https://github.com/paritytech/substrate/blob/master/frame/support/src/traits/tokens/fungible.rs#L349

enddynayn (approved):
🚀 ship it! 💯

pallets/capacity/src/helpers_tests.rs
```diff
@@ -439,14 +408,16 @@ fn impl_replenish_all_for_account_is_successful() {
...
 		assert_ok!(Capacity::replenish_all_for(target_msa_id));
 
 		let mut capacity_details =
 			CapacityDetails::<BalanceOf<Test>, <Test as Config>::EpochNumber>::default();
 
-		capacity_details.remaining = 10u32.into();
+		capacity_details.remaining_capacity = 10u32.into();
 		capacity_details.total_tokens_staked = 10u32.into();
-		capacity_details.total_available = 10u32.into();
+		capacity_details.total_capacity_issued = 10u32.into();
```
enddynayn, line 420: I like these new names! It's much more clear.

pallets/capacity/src/lib.rs
```diff
@@ -642,26 +629,29 @@ impl<T: Config> Replenishable for Pallet<T> {
...
+		let mut capacity_details =
+			Self::get_capacity_for(msa_id).ok_or(Error::<T>::TargetCapacityNotFound)?;
+		capacity_details.replenish_by_amount(amount, &Self::get_current_epoch());
 		Ok(())
 	}
 
 	fn can_replenish(msa_id: MessageSourceId) -> bool {
 		if let Some(capacity_details) = Self::get_capacity_for(msa_id) {
-			return capacity_details.last_replenished_epoch.lt(&Self::get_current_epoch())
+			return capacity_details.can_replenish(Self::get_current_epoch())
```
enddynayn, line 653: nice!

pallets/capacity/src/types.rs
```diff
@@ -158,40 +158,73 @@ impl<Balance: Saturating + Copy + CheckedAdd> StakingTargetDetails<Balance> {
...
+{
+	/// Increase a targets total Tokens staked and Capacity total issuance by an amount.
+	pub fn deposit(&mut self, amount: &Balance) -> Option<()> {
+		self.remaining_capacity = amount.checked_add(&self.remaining_capacity)?;
 		self.total_tokens_staked = amount.checked_add(&self.total_tokens_staked)?;
-		self.total_available = amount.checked_add(&self.total_available)?;
-		self.last_replenished_epoch = replenish_at;
+		self.total_capacity_issued = amount.checked_add(&self.total_capacity_issued)?;
 
+		// We do not touch last_replenished epoch here, because it would create a DoS vulnerability.
```
enddynayn, line 181: this is a good comment

PR #1184, opened 2023-03-06
Optimize registry and fix weights

wilwade:
# Goal
The goal of this PR is to fix the weights for msa pallet and optimize the signature registry

Amends: #785 and #1143

# Discussion
- Weights were wrong because we were not testing with a full registry, which is eventually the only case.
- By switching from a ring to a list + keeping the most recent value in the pointer allowed a reduction of one r/w, although it does not show up because weights were wrong.
- A few other minor cleanups

# Checklist
- [x] Benchmarks added
- [x] Weights updated


pallets/msa/src/benchmarking.rs
```diff
@@ -67,6 +67,20 @@ fn create_msa_account_and_keys<T: Config>() -> (T::AccountId, SignerId, MessageS
 	(account_id, key_pair, msa_id)
 }
 
+fn generate_fake_signature(i: u8) -> MultiSignature {
```
enddynayn, line 70: nice!

pallets/msa/src/benchmarking.rs
```diff
@@ -80,16 +94,18 @@ benchmarks! {
...
 		assert_ok!(Msa::<T>::create(RawOrigin::Signed(caller.clone()).into()));
 		assert_ok!(Msa::<T>::create_provider(RawOrigin::Signed(caller.clone()).into(),Vec::from("Foo")));
 
 		let schemas: Vec<SchemaId> = (0 .. s as u16).collect();
 		T::SchemaValidator::set_schema_count(schemas.len().try_into().unwrap());
 		let (payload, signature, key) = create_payload_and_signature::<T>(schemas, 1u64.into());
-	}: _ (RawOrigin::Signed(caller), key, signature, payload)
+	}: _ (RawOrigin::Signed(caller), key.clone(), signature, payload)
 	verify {
-		assert_eq!(frame_system::Pallet::<T>::events().len(), 2);
```
enddynayn, line 92: It does not hurt to keep this one too.

pallets/msa/src/lib.rs
```diff
@@ -217,25 +217,25 @@ pub mod pallet {
...
 	/// ### Storage
 	/// - Key: Signature
 	/// - Value: Tuple
 	///     - `BlockNumber` when the keyed signature can be ejected from the registry
-	///     - [`MultiSignature`] the linked list pointer. This pointer is written as the oldest value, but updated to be the "next" value when there is one
+	///     - [`MultiSignature`] the forward linked list pointer. This pointer is the next "newest" value.
 	#[pallet::storage]
 	#[pallet::getter(fn get_payload_signature_registry)]
-	pub(super) type PayloadSignatureRegistryRing<T: Config> = StorageMap<
+	pub(super) type PayloadSignatureRegistryList<T: Config> = StorageMap<
```
enddynayn, line 238: I think queue makes sense as a postfix Queue.

pallets/msa/src/lib.rs
```diff
@@ -1345,35 +1348,42 @@ impl<T: Config> Pallet<T> {
...
-			PayloadSignatureRegistryRingPointer::<T>::get().unwrap_or(SignatureRegistryPointer {
+			PayloadSignatureRegistryPointer::<T>::get().unwrap_or(SignatureRegistryPointer {
 				newest: signature.clone(),
+				newest_expires_at: signature_expires_at,
 				oldest: signature.clone(),
 				count: 0,
 			});
 
+		// Make sure it is not sitting as the `newest` in the pointer
+		ensure!(
```
enddynayn, line 1374: 💯

pallets/msa/src/lib.rs
```diff
@@ -1383,39 +1393,59 @@ impl<T: Config> Pallet<T> {
...
 			newest: signature.clone(),
+			newest_expires_at: signature_expires_at,
 			oldest,
 		});
 
 		Ok(())
 	}
 
+	/// Used to fill up the registry for benchmarks
+	#[cfg(feature = "runtime-benchmarks")]
```
enddynayn, line 1423: I think we can move this too the benchmark file.

wilwade: Yes. Yes we can. I didn't realize I could call the storage directly.

Moving...

enddynayn (approved):
💯 exquisite 👍

shannonwells (changes requested):
The audit replay test isn't valid any more because we've entirely changed the system.  I think the entire file should be removed because leaving it implies it's testing something real.

pallets/msa/src/weights.rs
```diff
@@ -185,86 +185,86 @@ impl WeightInfo for () {
...
 	// Storage: Msa DelegatorAndProviderToDelegation (r:1 w:1)
 	fn revoke_delegation_by_provider() -> Weight {
-		Weight::from_ref_time(22_948_000 as u64)
+		Weight::from_ref_time(22_287_000 as u64)
 			.saturating_add(RocksDbWeight::get().reads(2 as u64))
 			.saturating_add(RocksDbWeight::get().writes(1 as u64))
 	}
-	// Storage: Msa PayloadSignatureRegistryRing (r:2 w:2)
-	// Storage: Msa PayloadSignatureRegistryRingPointer (r:1 w:1)
+	// Storage: Msa PayloadSignatureRegistryList (r:4 w:4)
```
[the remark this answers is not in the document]
wilwade: It is because this has two signatures that are added to the registry.

shannonwells (approved):

PR #1189, opened 2023-03-07
chore(Capacity) add createMsaAndProvider helper function for use with capacity integration tests

### LINKED ISSUE #889: [Capacity] integration tests for transacting
AC
- [x] When a user does not have enough capacity to pay for the transaction fee and is NOT eligible to replenish Capacity, it should error and be dropped from the transaction pool.
- [x] When a user attempts to pay for a non-capacity transaction with Capacity, it should error and drop the transaction from the transaction-pool.
- [x] When Capacity has withdrawn an event CapacityDeducted is emitted.
- [x] When a user is not a registered provider and attempts to pay with Capacity, it should error with InvalidTransaction::Payment.
- [x] Integration test and manual test. A registered provider with Capacity but no tokens associated with the key should still be able to use polkadot UI to submit a capacity transaction.
- [x] Token transactions should not be affected by supporting Capacity transactions.

johnmela:
# Goal
The goal of this PR is to add a helper function `createMsaAndProvider` to consolidate some repeated code used for capacity integration test setup.


enddynayn (approved):
💯 fabulous!

mattheworris (approved):
Looks good.

PR #1192, opened 2023-03-07
final changes

aramikm:
# Goal
The goal of this PR is to address https://github.com/LibertyDSNP/frequency/issues/1153

### DIFF
--- a/pallets/stateful-storage/src/stateful_child_tree.rs
+++ b/pallets/stateful-storage/src/stateful_child_tree.rs
@@ -377,9 +377,9 @@ impl<H: MultipartKeyStorageHasher> StatefulChildTree<H> {
 		storage_name: &[u8],
 	) -> ChildInfo {
 		let trie_root = Self::get_tree_prefix(msa_id);
-		// child tree root should be hashed by Twox128 to avoid probability of conflict
+		// child tree root should be hashed by Blake128 to avoid probability of conflict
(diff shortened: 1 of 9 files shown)


pallets/stateful-storage/src/stateful_child_tree.rs
```diff
@@ -377,9 +377,9 @@ impl<H: MultipartKeyStorageHasher> StatefulChildTree<H> {
 		storage_name: &[u8],
 	) -> ChildInfo {
 		let trie_root = Self::get_tree_prefix(msa_id);
-		// child tree root should be hashed by Twox128 to avoid probability of conflict
+		// child tree root should be hashed by Blake128 to avoid probability of conflict
```
aramikm, line 380: changing to Blake to use cryptographic hash

saraswatpuneet (approved):
lgtm

pallets/stateful-storage/src/stateful_child_tree.rs
JoeCap08055: Do we really need it? Is it worth the expense?

aramikm: From my perspective not but Parity recommended to use a Cryptographic hash function for child tree root. I don't think the expense is that bad compared to overall weights.

PR #1194, opened 2023-03-07
feat(stateful-signed-extension): add SignedExtension for stateful-sto…

### LINKED ISSUE #1095: [Private graph] research extrinsic overhead for using signed extensions
[Paused]
Research extrinsic overhead for using signed extensions.

We would like to use some of the checks in signed extensions to remove invalid requests to go inside the block too early

- race condition
- schema related checks   
- calculate the wight from signed extension 

AC:

- calculate extrinsic overhead without changes 
- new signed extension with PG and pull checks in 
- recalculate extrinsic overhead with changes

JoeCap08055:
# Goal
The goal of this PR is to implement a `SignedExtension` for the `stateful-storage` pallet. Modifies benchmarks to include SignedExtension logic in individual extrinsic benchmarks.

Closes #1095 

# Checklist
- [ ] Chain spec updated
- [ ] Design doc(s) updated
- [ ] Tests added
- [ ] Benchmarks added
- [ ] Weights updated


pallets/stateful-storage/src/stateful_signed_extension.rs
```diff
@@ -0,0 +1,284 @@
...
+	/// * target_hash: the content hash representing the known state of the Paginated storage.
+	///
+	/// # Errors (as u8 wrapped by `InvalidTransaction::Custom`)
+	/// * [`Error::InvalidSchemaId`]
+	/// * [`Error::SchemaPayloadLocationMismatch`]
+	/// * [`Error::SchemaNotSupported`]
+	/// * [`Error::CorruptedState`]
+	/// * [`Error::StalePageState`]
+	///
+	fn validate_itemized(
```
aramikm, line 79: Now that I'm thinking about itemized this would generate 2 READ in weights for reading itemized and that's not so great. This works for paginated since we don't need to read the item for modification but for itemized it is problematic.

aramikm: maybe we just check the hash for itemized in extrinsic? It would be different from paginated but not sure how else we can keep the weights optimized.

pallets/stateful-storage/src/stateful_signed_extension.rs
```diff
@@ -0,0 +1,284 @@
...
+		const TAG_PREFIX: &str = "StatefulStorageItemized";
+		let is_pruning = actions.iter().any(|a| matches!(a, ItemAction::Delete { .. }));
+
+		Self::check_schema(*schema_id, PayloadLocation::Itemized, is_payload_signed, is_pruning)
+			.map_err(|e| map_dispatch_error(e))?;
+
+		Self::get_checked_itemized_page(msa_id, schema_id, target_hash)
+			.map_err(|e| map_dispatch_error(e))?;
+
+		return ValidTransaction::with_tag_prefix(TAG_PREFIX).and_provides(account_id).build()
```
aramikm, line 96: I think provides should probably be msa_id and schema_id for itemized and msa_id and schema_id and page_id for paginated

aramikm (commented):
Awesome! Added some questions and concerns. I think we might want to run benchmarks to see the generated weights.

pallets/stateful-storage/src/stateful_signed_extension.rs
JoeCap08055: As far as I can tell, the number of reads hasn't changed; only the order they're done in. Yes, we read Itemized twice, but the second read is from Overlay cache.

aramikm: It is but weights don't know about overlay so it would appear as 2 reads in weights and block space would be wasted.

JoeCap08055: Here are the reads the benchmark sees for `Itemized`:
```
	// Storage: Schemas Schemas (r:1 w:0)
	// Storage: Msa PublicKeyToMsaId (r:1 w:0)
	// Storage: Msa DelegatorAndProviderToDelegation (r:1 w:0)
	// Storage: unknown [0xbd1557c8db6bd8599a811a7175fbc2fc6400] (r:1 w:1)
```

pallets/stateful-storage/src/stateful_signed_extension.rs
```diff
@@ -0,0 +1,281 @@
...
+		let is_pruning = actions.iter().any(|a| matches!(a, ItemAction::Delete { .. }));
+
+		Self::check_schema(*schema_id, PayloadLocation::Itemized, is_payload_signed, is_pruning)
+			.map_err(|e| map_dispatch_error(e))?;
+
+		Self::get_checked_itemized_page(msa_id, schema_id, target_hash)
+			.map_err(|e| map_dispatch_error(e))?;
+
+		return ValidTransaction::with_tag_prefix(TAG_PREFIX)
+			.and_provides(msa_id)
```
aramikm, line 94: I think all of these should be combined (maybe in a tuple) and only use one `provides` call. Looking at `and_provides` implementation is seems like it treats each as a prefix.

aramikm: Looking at `ValidTransaction` documents it seems like we should include msa_id and schema_id in our prefix for itemized and msa_id, schema_id and page_id in our paginated and get rid of provides.
Provides and requires are there to allow creating a dependency graph between transactions.

JoeCap08055: Updated.

pallets/stateful-storage/src/stateful_signed_extension.rs
aramikm: Hmm, it is weird that it just showed up as 1 read instead of 2 reads. Maybe they fixed that overlay part for weights? I'm a bit skeptical 🤔

pallets/stateful-storage/src/stateful_signed_extension.rs
JoeCap08055: Looking at the code, the prefix is not part of the `ValidTransaction`; it only gets used in the builder when you add a tag with either `provides` or `requires`.

pallets/stateful-storage/src/stateful_signed_extension.rs
saraswatpuneet: I hope there is no mock call that is hiding one read but otherwise looks like cache works

pallets/stateful-storage/src/stateful_signed_extension.rs
```diff
@@ -0,0 +1,278 @@
...
+
+		return ValidTransaction::with_tag_prefix(TAG_PREFIX)
+			.and_provides((msa_id, schema_id, page_id))
+			.build()
+	}
+}
+
+pub fn map_dispatch_error(err: DispatchError) -> InvalidTransaction {
+	InvalidTransaction::Custom(match err {
+		DispatchError::Module(module_err) => module_err.index,
```
shannonwells, line 149: nice

shannonwells (commented):
👍

PR #1207, opened 2023-03-09
change hash and restrict append only

aramikm:
# Goal
The goal of this PR is added 2 of following
- Change hash to XxHash64 
- Restrict AppendOnly to only be applicable to Itemized Location


saraswatpuneet (approved):
Lgtm!

PR #1225, opened 2023-03-13
[Handles] Design Doc

saraswatpuneet:
# Goal
The goal of this PR is  to layout design document for user handles requirements on frequency chain

# Checklist
- [ ] Chain spec updated
- [x] Design doc(s) updated
- [ ] Tests added
- [ ] Benchmarks added
- [ ] Weights updated

### DIFF
--- a/designdocs/user-handles.md
+++ b/designdocs/user-handles.md
@@ -3,7 +3,7 @@
 ## Context and Scope
 
 The Frequency blockchain aims to provide a mechanism to register and retrieve user handles on chain to make it easier to use MSA Ids.
-To achieve this, we propose the creation of a user handle ```registry``` system on Frequency chain, coupled with ```MessageSourceAccount```, that allows users to choose a handle and select a suffix from a set of options allowed by the chain that together are unique.
+To achieve this, we propose the creation of a user handle ```registry``` system on Frequency chain, coupled with ```MessageSourceAccount```, that allows users to choose a handle for their MSA Ids.Handle system would then append an available numeric suffix to the chosen user handle, separated by a delimiter[.], to make it unique and store the mapping of the handle to the MSA Id in the registry.
 The user handle system will also enforce guidelines to ensure that handles are user-friendly and easy to remember, as well as prevent misuse and abuse.
 
 The goal of this proposal is to enable users to create unique handles on the Frequency blockchain using a registry, making it more accessible for users to engage with the network.
@@ -117,13 +117,12 @@ sequenceDiagram
 
 MsaHandlePayload {
     handle: &[u8],
-    suffix: Option<u32>,
 }
 ```
 
-### Create user handle with chosen handle and optional suffix
+### Create user handle with chosen handle
 
- As a network, Frequency should allow users to choose their own handle and suffix. This extrinsic will allow users to create a handle with a chosen suffix. If the suffix is not provided, the chain will generate a random suffix within the range of suffixes allowed by the chain.
+ As a network, Frequency should allow users to choose their own handle, while chain will generate a random numeric suffix within the range of suffixes allowed. The full handle will be the handle with the suffix.
 
 ``` rust
 Input
@@ -143,10 +142,10 @@ Output
 
 Validation requirements
 * handle must follow handle guidelines.
-* suffix must be in the range of the handle.
 * handle must be unique.
 * handle must be available.
 * suffix must be available.
+* suffix must be in the range of the handle.
 * MSA ID must exist. Signing keys must resolve to MSA ID.
 
 Signature requirements
--- a/designdocs/user-handles.md
+++ b/designdocs/user-handles.md
@@ -115,10 +115,13 @@ sequenceDiagram
 ``` rust
 
 MsaHandlePayload {
-    base_handle: &[u8],
+    display_handle: &[u8],
 }
 ```
 
+display_handle is the handle that the user wants to use. This is the handle that will be displayed to other users.
+
+``` rust
 ### Create user handle with chosen handle
 
  As a network, Frequency should allow users to choose their own handle, while chain will generate a random numeric suffix within the range of suffixes allowed. The full handle will be the handle with the suffix.
--- a/designdocs/user-handles.md
+++ b/designdocs/user-handles.md
@@ -27,14 +27,16 @@ User handle ```registry``` on frequency chain.
 * Generate the next 10/20 ```u32``` values from the PRNG these are the possible suffixes within the range of suffixes allowed by the chain.
 * Query the chain to check which values are available, check if the handle with the suffix is available.
 * Choose an available value and attempt to claim it by submitting the full handle with the suffix to the chain.
-* Chain creates a available msa and map to valid handle and suffix chosen by user.
+* Chain check runs some validation and map the handle to the msa.
 * Chain maintains a reverse mapping from handle to msa for offchain use.
 
 ### Chain Steps
 
 * Every 100 blocks (e.g., when block_number % 100 == 1), update the current seed by taking the Merkle root of the previous block. Keep the current and one previous seed so that two seeds, s0 and s1, are part of the chain state at any given time.
 * Provide an RPC to get the current seed.
+* Provide an RPC to get the suffix options for a given handle.
 * When a user submits a handle, check the numeric suffix for availability and that it honors range defined for suffix window.
+* Store ```msa_id``` to ```handle``` mapping and ```handle``` to ```msa_id``` mapping.
 
 ### Handle Guidelines
 
--- a/designdocs/user-handles.md
+++ b/designdocs/user-handles.md
@@ -269,7 +269,7 @@ The crate will use a PRNG (pseudo-random number generator) to generate suffix va
 
 ## frequency-handles crate sequence diagram
 
-Frequency-handles is a standalone rust crate that provide basic functionality for generating and validating handles. It is intended to be used by the frequency crate to provide handle generation and validation functionality.
+Frequency-handles is a standalone rust crate that provide basic functionality for generating and validating handles. It is intended to be used by the frequency chain to provide handle generation and validation functionality.
 
 ```mermaid
 sequenceDiagram
--- a/designdocs/user-handles.md
+++ b/designdocs/user-handles.md
@@ -2,7 +2,7 @@
 
 ## Context and Scope
 
-The Frequency blockchain aims to provide a mechanism to register and retrieve user handles on chain.
+The Frequency blockchain aims to provide a mechanism to register and retrieve user handles on chain to make it easier to use MSA Ids.
 To achieve this, we propose the creation of a user handle ```registry``` system on frequency chain, coupled with ```MessageSourceAccount```, that allows users to choose a unique handle and select a suffix from a set of options allowed by the chain. The user handle system will also enforce guidelines to ensure that handles are user-friendly and easy to remember, as well as prevent misuse and abuse.
 
 The goal of this proposal is to enable users to create unique handles on the Frequency blockchain using a registry, making it more accessible for users to engage with the network.
--- a/designdocs/user-handles.md
+++ b/designdocs/user-handles.md
@@ -15,7 +15,7 @@ The goal of this proposal is to enable users to create unique handles on the Fre
 * Handles are unique and have a 1:1 mapping with ```MessageSourceAccount```.
 * Governance will define the range of suffixes allowed for a given handle.
 * Governance will define the time period a handle must be retired before it can be reused.
-* The handle create/change/(and retire, if not ```Pays::No```) supports capacity based transactions.
+* The handle create/change/(and retire, if its not feeless) supports capacity based transactions.
 
 ## Requirements
 
--- a/designdocs/user-handles.md
+++ b/designdocs/user-handles.md
@@ -267,7 +267,7 @@ The crate will provide a verification function that takes a user handle and a su
 
 The crate will use a PRNG (pseudo-random number generator) to generate suffix values for user handles. The PRNG will take a seed value derived from the current block's Merkle root and the user's desired handle base, and generate a sequence of 20 values that will be used as suffixes. These suffixes will then be checked for availability on the chain before attempting to create a new MSA.
 
-## Crate sequence diagram
+## frequency-handles crate sequence diagram
 
 Frequency-handles is a standalone rust crate that provide basic functionality for generating and validating handles. It is intended to be used by the frequency crate to provide handle generation and validation functionality.
 
--- a/designdocs/user-handles.md
+++ b/designdocs/user-handles.md
@@ -269,7 +269,7 @@ The crate will provide a verification function that takes a user handle and a su
 * Verify that the handle is valid, i.e. it follows the handle guidelines.
 * Check for homoglpyhs.
 
-## PRNG
+### PRNG
 
 The crate will use a PRNG (pseudo-random number generator) to generate suffix values for user handles. The PRNG will take a seed value derived from the current block's Merkle root and the user's desired handle base, and generate a sequence of 10/20 values that will be used as suffixes. These suffixes will then be checked for availability on the chain before attempting to create a new handle.
 
--- a/designdocs/user-handles.md
+++ b/designdocs/user-handles.md
@@ -24,9 +24,9 @@ Make the system easy to use and integrate with existing UI and wallet systems.
 * Apply the resulting value as the seed to the specified PRNG.
 * Generate the next 10 uint16 values from the PRNG.
 * For each value, calculate modulo 9000, then add 1000, to get a number between 1000 and 9999.
-* Generate MSAs by concatenating the suffix to the requested handle base and hashing mod 232.
-* Query the chain to check which values are available.
-* Choose an available value and attempt to claim it by submitting the full handle with the MSA creation RPC.
+* Query the chain to check which values are available. Via lookup, check if the handle with the suffix is available.
+* Create a available msa and map it to the handle with the suffix.
+* Choose an available value and attempt to claim it by submitting the full handle with the suffix to the chain.
 
 ### Chain Steps
 
--- a/designdocs/user-handles.md
+++ b/designdocs/user-handles.md
@@ -71,36 +71,27 @@ sequenceDiagram
     participant Wallet
     actor User
     participant App
-    participant RPC as RPC Node
-    participant Chain as Frequency
-    User->>App: Enter desired handle
-    App->>RPC: get_suffix_options(handle)
-    RPC->>Chain: Query current seed
-    Chain-->>RPC: Return seed
-    RPC->>RPC: Compute suffix options
-    RPC->>Chain: check_handle_availability(options)
-    Chain-->>RPC: Return subset of options
-    RPC-->>App: Return subset of options
-    App->>User: Present suffix options
-    User->>App: Choose desired suffix
-    App->>Wallet: Send signup data
-    Wallet->>Wallet: Perform 2FA if first attempt
-    Wallet-->>App: Return signed payload
-    App->>RPC: create_msa_handle(..., handle, suffix)*
-    RPC->>Chain: Submit transaction
-    Chain->>Chain: Check validity and availability
-    Chain-->>App: MsaHandleCreated event with handle
-    App->>User: Proceed with setup
-    User->>App: Request to get MSA ID for handle
-    App->>RPC: get_msa_id_for_handle(handle)
-    RPC->>Chain: Query handle mapping state (State Query MSA<->Handle)
-    Chain-->>RPC: Return handle mapping state (State Query MSA<->Handle)
-    RPC-->>App: Return MSA ID for handle
-    User->>App: Request to get handle for MSA ID
+    participant RPC
+    participant Frequency
+    User->>App: enter desired handle
+    App->>Wallet: send signup data
+    Wallet->>Wallet: perform 2FA if first attempt
+    Wallet-->>App: return signed payload
+    App->>Frequency: create_msa_handle(.., signed_payload)
+    Frequency-->>Frequency: query current seed
+    Frequency-->>Frequency: compute suffix options
+    Frequency-->>Frequency: check_handle_availability(...handle, suffix)
+    Frequency-->>Frequency: submit transaction
+    Frequency-->>App: MsaHandleCreated event with (msa_id, handle+suffix) 
+    App->>User: proceed with setup
+    User->>App: request to get msa_id
+    App->>Frequency: query state for msa_id (handle<->msa_id)
+    Frequency-->>App: return msa_id (State Query handle<->msa_id)
+    User->>App: request to get handle for msa_id
     App->>RPC: get_handle_for_msa_id(msa_id)
-    RPC->>Chain: Query MSA ID mapping state (Reverse Index Handle<->MSA)
-    Chain-->>RPC: Return MSA ID mapping state (Reverse Index Handle<->MSA)
-    RPC-->>App: Return handle for MSA ID
+    RPC->>Frequency: query for handle given msa_id (index handle<->msa_id)
+    Frequency-->>RPC: return handle for msa_id
+    RPC-->>App: return handle for msa_id
 ```
 
 ## Storage
@@ -124,13 +115,13 @@ sequenceDiagram
 
 MsaHandlePayload {
     handle: &[u8],
-    suffix: u32,
+    suffix: Option<u32>,
 }
 ```
 
-### Create user handle with chosen suffix
+### Create user handle with chosen handle and optional suffix
 
- As a network frequency should allow users to choose their own handle and suffix. This extrinsic will allow users to create a handle with a chosen suffix.
+ As a network frequency should allow users to choose their own handle and suffix. This extrinsic will allow users to create a handle with a chosen suffix. If the suffix is not provided, the chain will generate a random suffix within the range of suffixes allowed by the chain.
 
 ``` rust
 Input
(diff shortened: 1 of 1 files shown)


designdocs/user-handles.md
```diff
@@ -0,0 +1,198 @@
+# Frequency User Handles
```
saraswatpuneet, line 1: Focus on this doc

designdocs/user-handles.md
```diff
@@ -0,0 +1,308 @@
+# Frequency User Handles Registry
+
+## Context and Scope
+
+The Frequency blockchain aims to provide a mechanism to register and retrieve user handles on chain.
```
wilwade, line 5:
```suggestion
The Frequency blockchain aims to provide a mechanism to register and retrieve user handles on chain to make it easier to use MSA Ids.
```

designdocs/user-handles.md
```diff
@@ -0,0 +1,308 @@
...
+To achieve this, we propose the creation of a user handle ```registry``` system on frequency chain, coupled with ```MessageSourceAccount```, that allows users to choose a unique handle and select a suffix from a set of options allowed by the chain. The user handle system will also enforce guidelines to ensure that handles are user-friendly and easy to remember, as well as prevent misuse and abuse.
```
wilwade, line 6:
```suggestion
To achieve this, we propose the creation of a user handle ```registry``` system on frequency chain, coupled with ```MessageSourceAccount```, that allows users to choose a handle and select a suffix from a set of options allowed by the chain that together are unique.
The user handle system will also enforce guidelines to ensure that handles are user-friendly and easy to remember, as well as prevent misuse and abuse.
```

designdocs/user-handles.md
```diff
@@ -0,0 +1,308 @@
...
+* Apply the resulting value as the seed to the specified PRNG.
+* Generate the next 10/20 ```u32``` values from the PRNG these are the possible suffixes within the range of suffixes allowed by the chain.
+* Query the chain to check which values are available, check if the handle with the suffix is available.
+* Choose an available value and attempt to claim it by submitting the full handle with the suffix to the chain.
+* Chain validates the handle and suffix and maps the handle to the ```msa_id```.
+* Chain maintains a reverse mapping from handle to msa for offchain use.
+
+### Chain Steps
+
+* Every 100 blocks (e.g., when block_number % 100 == 1), update the current seed by taking the Merkle root of the previous block. Keep the current and one previous seed so that two seeds, s0 and s1, are part of the chain state at any given time.
```
wilwade, line 55: Curious why mod(100) == 1 instead of 0. Is that because we have a lot of things that already happen on those blocks?

designdocs/user-handles.md
```diff
@@ -0,0 +1,308 @@
...
+    App->>User: Present suffix options
+    User->>App: Choose desired suffix
+    App->>Wallet: Send signup data
+    Wallet->>Wallet: Perform 2FA if first attempt
+    Wallet-->>App: Return signed payload
+    App->>RPC: create_msa_handle(..., handle, suffix)*
+    RPC->>Chain: Submit transaction
+    Chain->>Chain: Check validity and availability
+    Chain-->>App: MsaHandleCreated event with handle
+    App-->>RPC: Publish Profile Announcement/batch with (canonical) DSNP handle
```
wilwade, line 91: I think we will want to skip this for simplicity.
```suggestion
```

designdocs/user-handles.md
```diff
@@ -0,0 +1,308 @@
...
+## Required Extrinsics
+
+### Primitives
+
+``` rust
+
+MsaHandlePayload {
+    handle: &[u8],
+    suffix: u32,
+    msa_id: u64,
```
wilwade, line 125: We can derive the MSA Id from the signing key. (You can only create a handle for your own MSA, and keys can only be attached to one MSA at a time)

designdocs/user-handles.md
```diff
@@ -0,0 +1,308 @@
...
+* **UsedSuffixes**: This storage will keep track of all used suffixes to ensure that no two handles have the same suffix. (May not be necessary)
+* **Seed (current and previous)**: This storage values (```tuple (s0, s1)```) will keep track of the current and previous seed values.
+
+## Required Extrinsics
+
+### Primitives
+
+``` rust
+
+MsaHandlePayload {
```
wilwade, line 122: Do we need to put in an additional mortality even if the suffix has an effective mortality?

designdocs/user-handles.md
```diff
@@ -0,0 +1,308 @@
...
+* handle must follow handle guidelines.
+* suffix must be in the range of the handle.
+* handle must be unique.
+* handle must be available.
+* suffix must be available.
+* MSA ID must exist.
+
+Signature requirements
+
+The extrinsic must be signed by the user/provider private key. The signature must be verified on-chain to ensure that the user is the owner of the private key.
```
wilwade, line 160: Side note, but a reminder that as this does not have a general replay issue, it does NOT need to be added into the signature registry to protect against replay.

designdocs/user-handles.md
```diff
@@ -0,0 +1,308 @@
...
+
+### Create user handle with chosen suffix
+
+ As a network frequency should allow users to choose their own handle and suffix. This extrinsic will allow users to create a handle with a chosen suffix.
+
+``` rust
+Input
+
+* origin - must be a signed origin
+* owner msa id - the MSA ID of the user. MSA ID must be created before calling this extrinsic.
```
wilwade, line 137:
```suggestion
* owner key for the msa id - one of the owner keys for the MSA Id of the user. MSA Id must be created before calling this extrinsic.
```

designdocs/user-handles.md
```diff
@@ -0,0 +1,308 @@
...
+* proof - the proof of ownership of the handle, ```MsaHandlePayload``` signed by the user/provider private key
```
wilwade, line 138:
```suggestion
* proof - the proof of ownership of the handle, ```MsaHandlePayload``` signed by an msa owner's private key
```

designdocs/user-handles.md
```diff
@@ -0,0 +1,308 @@
...
+
+Output
+* Event - `MsaHandleCreated` with the MSA ID and the handle
+
+* Errors - 
+    * `HandleAlreadyExists` if the handle already exists
+    * `InvalidMsaHandle` if the handle is invalid
+    * `Unauthorized` if the signature is invalid
+    * `InvalidSuffix` if the suffix is invalid
+    * `SuffixAlreadyExists` if the suffix already exists
```
wilwade, line 148: I don't think `SuffixAlreadyExists` is a needed error. We can just say handle not available to refer to the handle+suffix.

designdocs/user-handles.md
```diff
@@ -0,0 +1,308 @@
...
+
+## Storage
+
+* **HandleSuffixMin**: This storage value (```u32``) will keep track of the minimum suffix value allowed for a given handle.
+* **HandleSuffixMax**: This storage value (```u32``) will keep track of the maximum suffix value allowed for a given handle.
+* **HandleRetirementPeriod**: This storage value (```u32``) will keep track of the number of blocks a handle must be retired before it can be reused.
+* **MSAIdToUserHandle**: This storage will map MSAIds to their corresponding user handles.
+* **UserHandlesToMSAId**: This storage will map user handles to their corresponding MSAIds.
+* **RetiredHandles**: This storage will keep track of all retired handles and the block number at which they were retired. This probably is a spike to run few ideas on how to implement this.
+* **UsedSuffixes**: This storage will keep track of all used suffixes to ensure that no two handles have the same suffix. (May not be necessary)
```
wilwade, line 113: > May not be necessary

Agreed. We should just consider it part of the handle storage

designdocs/user-handles.md
```diff
@@ -0,0 +1,308 @@
...
+```
+
+## Storage
+
+* **HandleSuffixMin**: This storage value (```u32``) will keep track of the minimum suffix value allowed for a given handle.
+* **HandleSuffixMax**: This storage value (```u32``) will keep track of the maximum suffix value allowed for a given handle.
+* **HandleRetirementPeriod**: This storage value (```u32``) will keep track of the number of blocks a handle must be retired before it can be reused.
+* **MSAIdToUserHandle**: This storage will map MSAIds to their corresponding user handles.
+* **UserHandlesToMSAId**: This storage will map user handles to their corresponding MSAIds.
+* **RetiredHandles**: This storage will keep track of all retired handles and the block number at which they were retired. This probably is a spike to run few ideas on how to implement this.
```
wilwade, line 112: Idea: We could store this in the `UserHandlesToMSAId` data as "retired" with the block number it retired on. (That way it is merely waiting for `current_block > (retire_block + HandleRetirementPeriod)`)

designdocs/user-handles.md
```diff
@@ -0,0 +1,308 @@
...
+    RPC-->>App: Return handle for MSA ID
+```
+
+## Storage
+
+* **HandleSuffixMin**: This storage value (```u32``) will keep track of the minimum suffix value allowed for a given handle.
+* **HandleSuffixMax**: This storage value (```u32``) will keep track of the maximum suffix value allowed for a given handle.
+* **HandleRetirementPeriod**: This storage value (```u32``) will keep track of the number of blocks a handle must be retired before it can be reused.
+* **MSAIdToUserHandle**: This storage will map MSAIds to their corresponding user handles.
+* **UserHandlesToMSAId**: This storage will map user handles to their corresponding MSAIds.
```
wilwade, line 111: Side note. While this is correct, I think we would want to store the key of `UserHandlesToMSAId` as either the reduced handle (aka post homoglyph replacement) or the hash of the reduced handle.

designdocs/user-handles.md
```diff
@@ -0,0 +1,308 @@
...
+    Chain-->>RPC: Return MSA ID mapping state (Reverse Index Handle<->MSA)
+    RPC-->>App: Return handle for MSA ID
+```
+
+## Storage
+
+* **HandleSuffixMin**: This storage value (```u32``) will keep track of the minimum suffix value allowed for a given handle.
+* **HandleSuffixMax**: This storage value (```u32``) will keep track of the maximum suffix value allowed for a given handle.
+* **HandleRetirementPeriod**: This storage value (```u32``) will keep track of the number of blocks a handle must be retired before it can be reused.
+* **MSAIdToUserHandle**: This storage will map MSAIds to their corresponding user handles.
```
wilwade, line 110: Side note: Remember that this one should map to the "actual" user handle. Aka before any homoglyph replacement.

designdocs/user-handles.md
```diff
@@ -0,0 +1,308 @@
...
+### Retire user handle
+
+As a network frequency should allow users to retire their handles. This extrinsic will allow users to retire their handles. Retired handles will be available for reuse after a time period set by governance.
+
+``` rust
+Input
+
+* origin - must be a signed origin
+* owner msa id - the MSA ID of the user. MSA ID must be created before calling this extrinsic.
+* proof_of_ownership - the proof of ownership of the handle to be retired, ```MsaHandlePayload``` signed by the user/provider private key
```
wilwade, line 172: If we were ok with making retiring a `Pays::No` transaction (I'm thinking yes?), the we could skip both `owner msa id` and `proof_of_ownership`.

If not, this should be `owner_key`, `proof_of_ownership`.

designdocs/user-handles.md
```diff
@@ -0,0 +1,308 @@
...
+```
+
+### Change handle
+
+As a network frequency should allow users to change their handles. This extrinsic will allow users to change their handles. Retired handles will be available for reuse after a time period set by governance.
+
+``` rust
+Input
+* origin - must be a signed origin
+* owner msa ID - the MSA ID of the user
```
wilwade, line 194: Swap to owner_key

designdocs/user-handles.md
```diff
@@ -0,0 +1,308 @@
...
+* proof_of_ownership - the proof of ownership of the handle to be changed, ```MsaHandlePayload``` signed by the user/provider private key
```
wilwade, line 195: I don't think we need proof of current ownership as the `proof_of_new_ownership` provides the validation that this is requested by the MSA owner.

designdocs/user-handles.md
```diff
@@ -0,0 +1,308 @@
...
+* Governance will define the range of suffixes allowed for a given handle.
+* Governance will define the time period a handle must be retired before it can be reused.
+
+## Requirements
+
+The high level requirements for user handles are:
+
+* Support user ability to choose any valid handle and select a suffix from a set of options allowed by the chain.
+* Ensure handles and suffixes are unique and non-conflicting.
+* Make the system resistant to namespace exhaustion, homoglyphs and race conditions.
```
wilwade, line 24: One more thing. Reduce the likelihood of a secondary market for handles.

wilwade (changes requested):
A list of mostly small things.

designdocs/user-handles.md
saraswatpuneet: he reason for choosing mod(100) == 1 instead of mod(100) == 0 is to avoid conflicts with any other system events that may be happening on mod(100) == 0. 

Subject to change: this is what i inferred from Wess' proposal

designdocs/user-handles.md
saraswatpuneet: Will ad it to notes in extrinsic to remember during story creation

pallets/msa/src/signature_registry_tests.rs
rlaferla, line 1: I don't think you intended to include this file.

pallets/msa/src/tests.rs
rlaferla, line 1: I don't think you intended to include this file either.  It is obsolete as of yesterday.

tools/ci/scripts/run-benchmark.sh
rlaferla, line 1: Nor this one.

designdocs/user-handles.md
```diff
@@ -0,0 +1,323 @@
...
+    User->>App: Request to get handle for MSA ID
+    App->>RPC: get_handle_for_msa_id(msa_id)
+    RPC->>Chain: Query MSA ID mapping state (Reverse Index Handle<->MSA)
+    Chain-->>RPC: Return MSA ID mapping state (Reverse Index Handle<->MSA)
+    RPC-->>App: Return handle for MSA ID
+```
+
+## Storage
+
+* **HandleSuffixMin**: This storage value (```u32``) will keep track of the minimum suffix value allowed for a given handle.
```
rlaferla, line 108: Do suffixes (used to distinguish accounts with same base handle) really need to be as large as 4294967295 (u32)? or can they be 65535 (u16)?

rlaferla: I see.  There are ~294,282 Zhang Wei and ~44,935 John Smiths.

designdocs/user-handles.md
```diff
@@ -0,0 +1,323 @@
...
+* Apply the resulting value as the seed to the specified PRNG.
+* Generate the next 10/20 ```u32``` values from the PRNG these are the possible suffixes within the range of suffixes allowed by the chain.
+* Query the chain to check which values are available, check if the handle with the suffix is available.
+* Choose an available value and attempt to claim it by submitting the full handle with the suffix to the chain.
+* Chain validates the handle and suffix and maps the handle to the ```msa_id```.
+* Chain maintains a reverse mapping from handle to msa for offchain use.
+
+### Chain Steps
+
+* Every 100 blocks (e.g., when block_number % 100 == 1), update the current seed by taking the Merkle root of the previous block. Keep the current and one previous seed so that two seeds, s0 and s1, are part of the chain state at any given time.
```
rlaferla, line 57: It sounds as if we need to calculate the Merkle root when it's actually just the hash of the previous block.  Correct?

designdocs/user-handles.md
```diff
@@ -0,0 +1,323 @@
...
+Signature requirements
+
+The extrinsic must be signed by the user/provider private key. The signature must be verified on-chain to ensure that the user is the owner of the private key.
+```
+
+**Note:** If retiring a ```Pays::No transaction```? then, we could skip both owner_key and proof_of_ownership.
+
+### Change handle
+
+As a network frequency should allow users to change their handles. This extrinsic will allow users to change their handles. Retired handles will be available for reuse after a time period set by governance.
```
rlaferla, line 197: How often can a user change their handle?  e.g.  total number of times or elapsed time since last change.  It seems that there should be some limits set by governance.

designdocs/user-handles.md
saraswatpuneet: We can decide of types but in general the window of min max suffix range will move around in that space

pallets/msa/src/signature_registry_tests.rs
saraswatpuneet: Will be rebased

designdocs/user-handles.md
```diff
@@ -0,0 +1,326 @@
+# Frequency User Handles Registry
+
+## Context and Scope
+
+The Frequency blockchain aims to provide a mechanism to register and retrieve user handles on chain to make it easier to use MSA Ids.
+To achieve this, we propose the creation of a user handle ```registry``` system on frequency chain, coupled with ```MessageSourceAccount```, that allows users to choose a handle and select a suffix from a set of options allowed by the chain that together are unique.
```
johnmela, line 6:
```suggestion
To achieve this, we propose the creation of a user handle ```registry``` system on the Frequency chain, coupled with ```MessageSourceAccount```, that allows users to choose a handle and select a suffix from a set of options allowed by the chain that together are unique.
```

johnmela: Or.. 
`To achieve this, we propose the creation of a user handle ```registry``` system on chain`

designdocs/user-handles.md
```diff
@@ -0,0 +1,326 @@
...
+The user handle system will also enforce guidelines to ensure that handles are user-friendly and easy to remember, as well as prevent misuse and abuse.
+
+The goal of this proposal is to enable users to create unique handles on the Frequency blockchain using a registry, making it more accessible for users to engage with the network.
+
+## Assumptions
+
+* ```MessageSourceAccount``` exists for a user to create a handle.
+* It is optional for a user to create a handle.
+* Handles are unique and has a 1:1 mapping with ```MessageSourceAccount```.
```
johnmela, line 15:
```suggestion
* Handles are unique and have a 1:1 mapping with ```MessageSourceAccount```.
```

designdocs/user-handles.md
```diff
@@ -0,0 +1,326 @@
...
+
+* Every 100 blocks (e.g., when block_number % 100 == 1), update the current seed by taking the Merkle root of the previous block. Keep the current and one previous seed so that two seeds, s0 and s1, are part of the chain state at any given time.
+* Provide an RPC to get the current seed.
+* Provide an RPC to get the suffix options for a given handle.
+* When a user submits a handle, check the numeric suffix for availability and that it honors range defined for suffix window.
+* Store ```msa_id``` to ```handle``` mapping and ```handle``` to ```msa_id``` mapping.
+
+### Handling Race Conditions
+
+There is an exceedingly small chance that two users may attempt to claim the same handle and suffix combination within the same block. In such cases, the transaction will fail to create the MSA, and the app will receive a failure for the transaction rather than the expected MsaHandleCreated event.
```
enddynayn, line 65: If the transaction fails who pays for the fee?

designdocs/user-handles.md
aramikm: There is already a frame for that https://github.com/paritytech/substrate/tree/master/frame/insecure-randomness-collective-flip this is not a secure random just as they mention it.

aramikm: If seed is not random, the values generated using that seed is also not considered random but it might be fine for our usecase

designdocs/user-handles.md
enddynayn: why?

designdocs/user-handles.md
```diff
@@ -0,0 +1,326 @@
...
+
+Signature requirements
+
+The extrinsic must be signed by the user private key. The signature must be verified on-chain to ensure that the user is the owner of the private key. The signature must also include the old handle and the new handle to prevent unauthorized handle changes.
+```
+
+**Note** :
+
+* Change handle should also retire the old handle and suffix.
+* These are possible extrinsics and inputs required for the handle feature. The exact extrinsics and inputs may change based on the implementation.
```
enddynayn, line 232: What about transferring ownership?

designdocs/user-handles.md
```diff
@@ -0,0 +1,326 @@
...
+
+## Proposal
+
+User handle ```registry``` on frequency chain.
+
+### Handle Guidelines
+
+* User handles must be unique and each msa can only be be mapped to one handle.
+* User handles must be between 2 and 32 characters (32b limit) long.
+* User handles cannot contain the following substrings (or homoglyphs): @, #, :, ., ```
```
enddynayn, line 37: Are we thinking of following http://unicode.org/reports/tr46/ similar to ENS?

https://docs.ens.domains/contract-api-reference/name-processing

designdocs/user-handles.md
wilwade: Style guide note. Just "Frequency", not "the" Frequency.

designdocs/user-handles.md
```diff
@@ -0,0 +1,326 @@
+# Frequency User Handles Registry
+
+## Context and Scope
+
+The Frequency blockchain aims to provide a mechanism to register and retrieve user handles on chain to make it easier to use MSA Ids.
+To achieve this, we propose the creation of a user handle ```registry``` system on frequency chain, coupled with ```MessageSourceAccount```, that allows users to choose a handle and select a suffix from a set of options allowed by the chain that together are unique.
+The user handle system will also enforce guidelines to ensure that handles are user-friendly and easy to remember, as well as prevent misuse and abuse.
+
+The goal of this proposal is to enable users to create unique handles on the Frequency blockchain using a registry, making it more accessible for users to engage with the network.
```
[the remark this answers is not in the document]
wilwade: Correct. The other options outside the chain were not able to meet the needed goals. :(

designdocs/user-handles.md
wilwade: Not sure if it got added to the list, but one of the design goals is that we do not have a secondary market for handles. So a explicit transfer is not possible intentionally.

designdocs/user-handles.md
wilwade: Hmm... If someone can continuously change their handle, they could lock up a large number of the common handles.

However suffixes make that a lot harder to do.

It still costs (Capacity) for each new handle or change.

designdocs/user-handles.md
```diff
@@ -0,0 +1,316 @@
...
+
+``` rust
+
+MsaHandlePayload {
+    handle: &[u8],
+    suffix: Option<u32>,
+}
+```
+
+### Create user handle with chosen handle and optional suffix
```
wilwade, line 122: We should mention somewhere that `create`, and `change` (and `retire` depending on if it is Pays::No) should be Capacity possible transactions.

enddynayn: Making these Capacity transactions might incentive providers to create MSA ids and handles. These could then be resold and create a market for them. However, given that we do not have any punishment system, we put too much trust in providers not to exploit.

designdocs/user-handles.md
enddynayn: Can we prevent this from happening? I can create accounts with handles associated with them and sell the MSA.

designdocs/user-handles.md
```diff
@@ -0,0 +1,326 @@
...
+User handle ```registry``` on frequency chain.
+
+### Handle Guidelines
+
+* User handles must be unique and each msa can only be be mapped to one handle.
+* User handles must be between 2 and 32 characters (32b limit) long.
+* User handles cannot contain the following substrings (or homoglyphs): @, #, :, ., ```
+* User handles cannot be (@admin, @everyone, @all) *blocklist of handles we reserve
+* Suffixes are u32 limited to a range defined by governance
+* Suffix will be randomly constrained.
```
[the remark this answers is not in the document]
saraswatpuneet: basically we will have SuffixMin and SuffixMax range which can be altered by governance, 
now that will give control over which range of suffixes will need to exhaust first before window is stretched, shrinked or moved around the space

We are using going to use seed for PRNG and constraint the returned suffix in defined window range

designdocs/user-handles.md
saraswatpuneet: Cool added both to the notes for reference

designdocs/user-handles.md
saraswatpuneet: Hmm good point, I was expecting handles to have a permanence with msa unless specifically retired or revoke?

designdocs/user-handles.md
saraswatpuneet: Added notes and assumption for the same , thanks

designdocs/user-handles.md
saraswatpuneet: Yep not needed now, however since we are doing optional suffix, what if someone uses extrinsic to claim some handle and suffix without checking if that suffix will be available or not

So not sure if its useful to send an error stating exactly same, but yes user can always check range of allowed suffixes.

designdocs/user-handles.md
saraswatpuneet: For our use case, we can rely on the network effect for the randomness of the seed. As long as we keep track of both seed-prev and seed-current at any given moment, we can ensure that the range of values generated using the seed will be different.

saraswatpuneet: Let me add the details on seed computation from Wess's proposal

designdocs/user-handles.md
saraswatpuneet: Name squatting is hard to prevent but since we have suffixes and a window to control what range of suffixes are being allowed by chain so that takes care of things like TM/Squating etc to some degree.

However, now with updated path, we are not moving the burden of choosing suffixes to user but rather going to put any randomly available suffix to desired user handle

Full Handle can be retired and left to be picked when it becomes available again

designdocs/user-handles.md
saraswatpuneet: Added a spike for the same, either use msa_id or some variant of it as seed or use chain for the random seed

saraswatpuneet: #1250

johnmela (approved):

designdocs/user-handles.md
saraswatpuneet: #1250 , we will probably be spiking on seed and PRNG

designdocs/user-handles.md
saraswatpuneet: :+1:

designdocs/user-handles.md
saraswatpuneet: Totally 👍

PR #1241, opened 2023-03-15
add universal benchmarks target to Makefile

### LINKED ISSUE #1231: Improve benchmark script
1. Rename `run_all_benchmarks.sh` -> `run_benchmarks.sh` as "all" in the name implies that this script runs benchmarks for all pallets and doesn't have selective pallet support. Maybe misleading to some developers. Perhaps the old `run_benchmarks.sh` can be removed or deprecated as this new script seems to do its job already.

1. Instead of creating a new make target for each pallet (just think of 1000s pallets), we can create one universal target that passes pallet name as an argument. Something like this:
   ```make
   .PHONY: benchmarks
   benchmarks:
     ./scripts/run_benchmark.sh -p $(pallet)
   ```
   Then call it with:
   
   `$ make benchmarks pallet="<pallet_name>"`

demisx:
# Goal
The goal of this PR is improve benchmarks target in the Makefile

Closes #1231

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -73,9 +73,15 @@ upgrade-local:
 	./scripts/init.sh upgrade-frequency
 
 
+# Run benchmarks for all pallets
+.PHONY: benchmarks-all
+benchmarks-all:
+	./scripts/run_all_benchmarks.sh
+
+# Run benchmark for a given pallet
 .PHONY: benchmarks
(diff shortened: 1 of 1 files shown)


wilwade:
@demisx is this needed with #1231 merged?

PR #1246, opened 2023-03-15
feat(capacity): allow p.g tx's to pay with capacity

enddynayn:
# Goal
Allow for the following private-graph calls to pay transaction fees with capacity:
- apply_item_actions
- upsert_page
- delete_page
- apply_item_actions_with_signature
- upsert_page_with_signature
- delete_page_with_signature

The fees for these transactions in Capacity are fixed to this point in time by
not allowing new benchmarks to affect the weight of these transactions.

issue-1266

### DIFF
--- a/pallets/frequency-tx-payment/src/capacity_stable_weights.rs
+++ b/pallets/frequency-tx-payment/src/capacity_stable_weights.rs
@@ -61,6 +61,13 @@ pub trait WeightInfo {
 	// Messages
 	fn add_onchain_message(n: u32, ) -> Weight;
 	fn add_ipfs_message() -> Weight;
+	// Stateful-storage
+	fn apply_item_actions(n: u32, ) -> Weight;
+	fn upsert_page(s: u32, ) -> Weight;
+	fn delete_page() -> Weight;
+	fn apply_item_actions_with_signature(s: u32, ) -> Weight;
+	fn upsert_page_with_signature(s: u32, ) -> Weight;
+	fn delete_page_with_signature() -> Weight;
 }
 
 /// Weights for pallet_msa using the Substrate node and recommended hardware.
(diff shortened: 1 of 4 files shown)


pallets/frequency-tx-payment/src/capacity_stable_weights.rs
```diff
@@ -61,6 +61,13 @@ pub trait WeightInfo {
 	// Messages
 	fn add_onchain_message(n: u32, ) -> Weight;
 	fn add_ipfs_message() -> Weight;
+	// Stateful-storage
```
aramikm, line 64: This seems to be a bit tricky, how would you know what stable weights are correct? Weights can not be assumed to be stable and I'm not sure what is the consequences for  relying on something like this

pallets/stateful-storage/src/tests.rs
```diff
@@ -2640,3 +2641,22 @@ fn signature_replay_on_deleted_page_check() {
 		));
 	})
 }
+
+#[test]
+fn test_sum_add_actions_bytes() {
```
aramikm, line 2646: I think you'll need to rebase from main since your test files structure is changed

aramikm (approved):
The implementation looks good but the underlying assumptions might be concerning.

shannonwells (approved):
Looks good to me, I'm still assuming we're intentionally setting capacity fees based on current weight calculations.  I think we should do a last run on Jenkins and update all these before the release though.

pallets/frequency-tx-payment/src/capacity_stable_weights.rs
enddynayn: I'll be working on putting something to address some of your questions tomorrow.

pallets/stateful-storage/src/tests.rs
enddynayn: Sounds good. I will rebase the capacity-development branch and update this.

PR #1259, opened 2023-03-17
Update integration test scripts and options

mattheworris:
# Goal
The goal of this PR is to update the run_integration_tests.sh script to:

1. Handle signals and clean up.
2. Add a command line argument to skip the js build.

Change Explanations:
Makefile:
- Add $(ARGS) as pass through to enable: `make integration-test ARGS="-s 1"` to enable skipping js build.

README.md:
- Add instructions to enable skipping js build.

run_integration_tests.sh:
- Modify `cleanup` not to use the kill_freq.sh script. As signals are asynchronous, `cleanup` does not know how to navigate to the scripts directory using relative paths. It's a small script, so the commands are included directly in `cleanup`.
- Fix syntax errors so that `getopts` will correctly parse the `-s` option.
- Add `if-then` conditional for `SKIP_JS_BUILD`.
- Remove `if-then` conditional which is handled by `cleanup` on receiving the `EXIT` signal when the script finishes.


scripts/run_integration_tests.sh
```diff
@@ -5,23 +5,48 @@ function get_frequency_pid () {
...
-    then
-        ${RUNDIR}/kill_freq.sh
-    fi
+    local signal="$1"
+
+    case "$signal" in
+        TERM|INT) 
+            # Catch TERM and INT signals and exit gracefully
+            exit
+            ;;
```
JoeCap08055, line 14: Seems superflous to bother catching TERM & INT if we're not going to do anything. In that case, might as well just not have a signal handler.

BUT... the point of the signal handler is to make sure that no matter how the script exits, it kills the running frequency node if it's supposed to.

If the main issue for this fix is that the script was sometimes not in the correct directory to find the path to the `kill_freq.sh` script, I think there are 2 main ways to resolve:


1. Get the directory of the running script at the top of the script file, and use that to construct the path to the `kill_freq.sh` script, instead of using relative paths.
-or-
2. Only save the PID for frequency if we start it ourselves; then in the cleanup handler, if PID is non-empty, kill it. No need for the `kill_freq.sh` script.

scripts/run_integration_tests.sh
```diff
@@ -5,23 +5,48 @@ function get_frequency_pid () {
...
+trap 'cleanup TERM' TERM
+trap 'cleanup INT' INT
 
-while getopts "s" OPTNAME
+while getopts "s:" OPTNAME
 do
-    case ${OPTNAME} in
-        "s") SKIP_JS_BUILD=1
+    case "${OPTNAME}" in
+        "s") SKIP_JS_BUILD=$OPTARG
```
JoeCap08055, line 44: Why the change to use `$OPTARG` here? Seems like it just puts more burden on the user to pass a meaningless value on the CLI.

scripts/run_integration_tests.sh
```diff
@@ -61,7 +86,7 @@ fi
 
 declare -i timeout_secs=30
 declare -i i=0
-while (( !PID && i < timeout_secs ))
+while [[ -z "${PID}" && i -lt timeout_secs ]]
```
JoeCap08055, line 89: 1. Why this change? Seems like just another way of doing exactly the same thing.
2. I don't think this works... if you're using normal expression evaluation (`[[ expression ]]`) instead of arithmetic evaluation (`(( _expression_ ))`), then you need to dereference variables with `$var`; otherwise you're just comparing the strings of the variable names.

Makefile
```diff
@@ -145,7 +145,7 @@ test:
 	cargo test --workspace --locked --features runtime-benchmarks,all-frequency-features
 
 integration-test:
-	./scripts/run_integration_tests.sh
+	./scripts/run_integration_tests.sh $(ARGS)
```
JoeCap08055, line 148: This works, but forces the user to really read the docs in order to understand how to run tests. If we create multiple targets in the makefile, we can take advantage of makefile CLI auto-complete in some shells so that the user can hit TAB to see the available commands. Targets might look like this:

```
integration-test:
   
integration-test-only:
   ./scripts/run_integration_tests.sh -s
   
integration-test-load:
   ./scripts/run_integration_tests.sh load
   
integration-test-load-only:
   ./scripts/run_integration_tests.sh -s load
```

Then, if the user types `make int<TAB>`, they'll get:
```
integration-test integration-test-only integration-test-load integration-test-load-only
```

scripts/run_integration_tests.sh
mattheworris: Good point, reverting to the original.

Makefile
mattheworris: Good idea, refactoring it in.

scripts/run_integration_tests.sh
mattheworris: Agreed, this change was not necessary and should have been omitted from the commit: will revert.

saraswatpuneet (approved):

JoeCap08055 (approved):

PR #1268, opened 2023-03-22
1267 benchmarking writes pallet collective twice

### LINKED ISSUE #1267: Benchmarking writes `pallet_collective` twice
When running the benchmarking job it writes the `pallet_collective` file twice.

This might be a problem, but also could be due to the fact that we use the pallet_collective twice. (Council and Technical Committee)

Let's double check that this is not an issue.

https://github.com/LibertyDSNP/frequency/actions/runs/4485490323/jobs/7887093046

> Multiple benchmarks were written to the same file: "./scripts/../runtime/common/src/weights/pallet_collective.rs".

JoeCap08055:
# Goal
The goal of this PR is to:
1. Account for multiple instances of `pallet_collective` when benchmarking & using weights.
2. Cleanup/enhance benchmarking scripts

Closes #1267 

# Checklist
- [ ] Chain spec updated
- [ ] Design doc(s) updated
- [ ] Tests added
- [ ] Benchmarks added
- [x] Weights updated


wilwade (approved):
Let's do it.

PR #1289, opened 2023-03-29
When comparing UTF8 strings, account for multibyte characters

### LINKED ISSUE #1253: [Bug] Byte index not aligned with char boundary in multi-byte strings
The extrinsic add_ipfs_message in [pallets/messages/src/lib.rs](https://github.com/LibertyDSNP/frequency/blob/c6c53f3d8b246da402a8bc97f51b2d98d8341398/pallets/messages/src/lib.rs#L222) invokes validate_cid function in which a conversion of the parameter (an array) is [made into a UTF-8 string](https://github.com/LibertyDSNP/frequency/blob/c6c53f3d8b246da402a8bc97f51b2d98d8341398/pallets/messages/src/lib.rs#L407-L410). An attacker could manipulate that parameter and cause a panic, leading to a DOS attack.

## Issue

The below code converts the &Vec<u8> in_cid into a str with the UTF-8 format.

    pub fn validate_cid(in_cid: &Vec<u8>) -> Result<Vec<u8>, DispatchError> {
        let cstr: &str = sp_std::str::from_utf8(&in_cid[..]).map_err(|_| Error::<T>::InvalidCid)?;
        ...
        ensure!(&cstr[0..2] != "Qm", Error::<T>::UnsupportedCidVersion);
        ...
    }

The below call can be used to reproduce the issue and cause a crash:

    call: RuntimeCall::Messages(Call::add_ipfs_message { schema_id: 12, cid: [55, 197, 136, 0, 0, 0, 0, 0, 0, 0, 0, 0], payload_length: 332 })

Once executed, the above call will throw the following error:

    byte index 2 is not a char boundary; it is inside 'ň' (bytes 1..3) of `7ň`
    thread 'X' panicked at 'byte index 2 is not a char boundary; it is inside 'ň' (bytes 1..3) of `7ň`', pallets/messages/src/lib.rs:413:18
    The index used to access a character in the cstr string is not aligned with a character boundary, meaning that the index 
    selected doesn't correspond with the beginning or the end of a character in the UTF-8 encoding. Here, the error occurred when checking the length of the cstr string and when verifying if the first two characters are equal to "Qm" : the invalid byte at index 2 is inside a multi-byte character, leading to the panic error.

## Risk

Since this can be triggered by a call to an extrinsic an attacker could cause a validator node to crash by submitting a specific crafted extrinsic and cause them missing their block authoring slot. This could lead to a denial-of-service of the whole chain with only very low cost requirements for the attacker.

## Mitigation

We suggest to make sure that the selected byte indexes align with the beginning and the end of complete UTF-8 characters, and to perform the necessary checks.

mattheworris:
# Goal
The goal of this PR is to fix bug #1253 (Byte index not aligned with char boundary in multibyte strings)
The function `validate_cid` currently assumes that the first two characters are not multibyte. Therefore, a multibyte character in the first two positions will cause a panic when checking for CIDv0 == "Qm".

Changes:
Use the `starts_with()` function to safely access the multibyte characters.

Closes #1253

# Checklist
- [ ] Chain spec updated
- [ ] Design doc(s) updated
- [x] Tests added
- [ ] Benchmarks added
- [ ] Weights updated


pallets/messages/src/lib.rs
```diff
@@ -410,7 +410,17 @@ impl<T: Config> Pallet<T> {
 		// Decode SCALE encoded CID into string slice
 		let cstr: &str = sp_std::str::from_utf8(&in_cid[..]).map_err(|_| Error::<T>::InvalidCid)?;
 		ensure!(cstr.len() > 2, Error::<T>::InvalidCid);
-		ensure!(&cstr[0..2] != "Qm", Error::<T>::UnsupportedCidVersion);
+
```
enddynayn, line 413: Are you ok with renaming cstr to cid_str? 

We could also simplify the code using starts_with.

```
  ensure!(!cid_str.starts_with("Qm"), Error::<T>::UnsupportedCidVersion);
```

enddynayn (approved):
👍 Nice!

shannonwells (approved):

pallets/messages/src/lib.rs
```diff
@@ -408,12 +408,14 @@ impl<T: Config> Pallet<T> {
...
 	pub fn validate_cid(in_cid: &Vec<u8>) -> Result<Vec<u8>, DispatchError> {
 		// Decode SCALE encoded CID into string slice
-		let cstr: &str = sp_std::str::from_utf8(&in_cid[..]).map_err(|_| Error::<T>::InvalidCid)?;
-		ensure!(cstr.len() > 2, Error::<T>::InvalidCid);
-		ensure!(&cstr[0..2] != "Qm", Error::<T>::UnsupportedCidVersion);
+		let cid_str: &str =
+			sp_std::str::from_utf8(&in_cid[..]).map_err(|_| Error::<T>::InvalidCid)?;
+		ensure!(cid_str.len() > 2, Error::<T>::InvalidCid);
+		// starts_with handles Unicode multibyte characters safely
+		ensure!(!cid_str.starts_with("Qm"), Error::<T>::UnsupportedCidVersion);
```
aramikm, line 415: So this is the line that fixes the issue.

aramikm (approved):
Looks great!

PR #1290, opened 2023-03-29
fix(cov): ensure test pass

enddynayn:
# Goal
Running the test with llvm-cov fails because the
state for mock BlockNumberProvider is modified
between tests and not reset.

Ensure that the state for the mock is reset
before running a test.

### DIFF
--- a/pallets/time-release/src/benchmarking.rs
+++ b/pallets/time-release/src/benchmarking.rs
@@ -72,8 +72,7 @@ benchmarks! {
 			schedule.start = i.into();
 			TimeReleasePallet::<T>::transfer(RawOrigin::Signed(from.clone()).into(), to_lookup.clone(), schedule.clone())?;
 		}
-
-		frame_system::Pallet::<T>::set_block_number(schedule.end().unwrap() + 1u32.into());
+		T::BlockNumberProvider::set_block_number(schedule.end().unwrap() + 1u32.into());
 	}: _(RawOrigin::Signed(to.clone()))
 	verify {
 		assert_eq!(
--- a/pallets/time-release/src/mock.rs
+++ b/pallets/time-release/src/mock.rs
@@ -125,6 +125,7 @@ pub struct ExtBuilder;
 impl ExtBuilder {
 	pub fn build() -> sp_io::TestExternalities {
 		let mut t = frame_system::GenesisConfig::default().build_storage::<Runtime>().unwrap();
+		MockBlockNumberProvider::set(0);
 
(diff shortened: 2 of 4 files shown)


wilwade (approved):
🥇

pallets/time-release/src/benchmarking.rs
```diff
@@ -72,8 +72,6 @@ benchmarks! {
 			schedule.start = i.into();
 			TimeReleasePallet::<T>::transfer(RawOrigin::Signed(from.clone()).into(), to_lookup.clone(), schedule.clone())?;
 		}
-
-		frame_system::Pallet::<T>::set_block_number(schedule.end().unwrap() + 1u32.into());
```
aramikm, line 76: why are you removing this? Isn't this to alter what is going to happen in the benchmark?

enddynayn: It does not do anything. This pallet does not depend on the system block number.

pallets/time-release/src/mock.rs
```diff
@@ -125,6 +125,7 @@ pub struct ExtBuilder;
 impl ExtBuilder {
 	pub fn build() -> sp_io::TestExternalities {
 		let mut t = frame_system::GenesisConfig::default().build_storage::<Runtime>().unwrap();
+		MockBlockNumberProvider::set(0);
```
aramikm, line 128: Tests can run in parallel so having a static value being set and read from different tests without any proper locking seems to interfere with having isolated tests.

pallets/time-release/src/benchmarking.rs
aramikm: it definitely does, when trying to lazily clean up the storage if everything is unlocked

pallets/time-release/src/mock.rs
enddynayn: This is called before every test is run.

pallets/time-release/src/benchmarking.rs
enddynayn: This benchmark does not is what I mean to say.

pallets/time-release/src/mock.rs
enddynayn: I see. I had not thought about it. Hmmm... I think we do this in other places where we update constants.

enddynayn: ahh ok. I think we are ok because we are using parameter types and we reset storage on every run.

aramikm (approved):
Looks great!

PR #1321, opened 2023-04-04
Update pays with capacity integration test to check for correctly reduced capacity

mattheworris:
# Goal
The goal of this PR is check that the capacity fee is correctly withdrawn from the capacity ledger.

### DIFF
--- a/integration-tests/capacity/transactions.test.ts
+++ b/integration-tests/capacity/transactions.test.ts
@@ -1,6 +1,6 @@
 import "@frequency-chain/api-augment";
 import { KeyringPair } from "@polkadot/keyring/types";
-import { u64, u16 } from "@polkadot/types";
+import { u64, u16, u128 } from "@polkadot/types";
 import assert from "assert";
 import { AddKeyData, EventMap, ExtrinsicHelper } from "../scaffolding/extrinsicHelpers";
 import {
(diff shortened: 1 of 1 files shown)


integration-tests/capacity/transactions.test.ts
```diff
@@ -70,7 +72,22 @@ describe("Capacity Transactions", function () {
...
+                    fee = chainEvents["capacity.CapacityWithdrawn"].data.amount;
+                }
+                else {
+                    assert.fail("should return a CapacityWithdrawn event");
+                }
+                const reducedCapacity = amountStaked - fee.toBigInt();
+
+                // Check for remaining capacity to be reduced by correct amount
+                const capacityStaked = (await firstValueFrom(ExtrinsicHelper.api.query.capacity.capacityLedger(stakeProviderId))).unwrap();
+                assert.equal(capacityStaked.remainingCapacity, reducedCapacity, "should return a capacityLedger with " + reducedCapacity + " remainingCapacity");
```
shannonwells, line 88: it would be more readable to use [template literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) here: 
```
`should return a capacityLedger with ${reducedCapacity} remainingCapacity`
```

Also it would be less confusing to call the variable remainingCapacity.

mattheworris: Good idea, will update.

enddynayn (approved):
👍 awesome!

shannonwells (approved):
Nice!

enddynayn (approved):
👍

PR #1324, opened 2023-04-05
fix(FrqcyTxPayment): ensure pay_with_capacity errors

enddynayn:
# Goal
Ensure that when calling "pay_with_capacity," errors are propagated.
Errors are being swallowed and not propagated to the client when an extrinsic call fails.

Update benchmarks for frequency-tx-payment pallet.


mattheworris (approved):
✅

integration-tests/capacity/transactions.test.ts
```diff
@@ -113,26 +114,21 @@ describe("Capacity Transactions", function () {
 
             // A registered provider with Capacity but no tokens associated with the
             // key should still be able to use polkadot UI to submit a capacity transaction.
-            // *All accounts will have at least an EXISTENTIAL_DEPOSIT = 1M.
-            // This test may still be valid if the txn cost is greater than 1M.
             it("pays for a transaction with available capacity and no free tokens", async function () {
                 let delegatorKeys = createKeys("delegatorKeys");
-                let _delegatorMsaId = await ExtrinsicHelper.createMsa(delegatorKeys);
+                await fundKeypair(devAccounts[0].keys, delegatorKeys, 100n * dollars);
```
shannonwells, line 119: I don't understand the purpose of these changes in this test and above.  What was wrong with the test setup?

pallets/frequency-tx-payment/src/lib.rs
```diff
@@ -177,11 +177,10 @@ pub mod pallet {
 		pub fn pay_with_capacity(
 			origin: OriginFor<T>,
 			call: Box<<T as Config>::RuntimeCall>,
-		) -> DispatchResult {
+		) -> DispatchResultWithPostInfo {
```
shannonwells, line 180: bingo

Makefile
```diff
@@ -154,9 +156,6 @@ benchmarks-multi-local:
 benchmarks-capacity:
 	./scripts/run_benchmark.sh -p capacity
 
-benchmarks-frequency-tx-payment:
```
shannonwells, line 157: why is “benchmarks-frequency-tx-payment” removed from the Makefile?

integration-tests/capacity/transactions.test.ts
enddynayn: The await does not belong there. It is doing nothing. It's broken.

shannonwells (approved):
Thanks, looks good

PR #1376, opened 2023-04-11
1370 SignedExtension for Handles

### LINKED ISSUE #1370: [Handles] Signed Extension for retire_handle
## Goals 
Frequency `Pays::No` extrinsics should implement respective `SignedExtensions` for pre-dispatch checks. In the handles pallet, we would like to implement the same for `retire_handle`

Acceptance Criteria:

- [x] `retire_handle()` being a `Pays::No` type transaction should have some pre-dispatch checks via an implementation of `HandlesSignedExtension` (an example being [here](https://github.com/LibertyDSNP/frequency/blob/5b880d3c9bc2069d767ba0b2744a29419c537e8e/pallets/stateful-storage/src/stateful_signed_extension.rs))
- [x] Implement validation logic for `HandlesSignedExtension`, some possible validations that can be done in extension are listed below
- [x] Unit tests(if needed)

_**Possible validations from retire_handle that can be moved to signed extension(please refine implementation time)**_
- [x] Validate if `msa` exists, an example being [here](https://github.com/LibertyDSNP/frequency/blob/5b880d3c9bc2069d767ba0b2744a29419c537e8e/pallets/msa/src/lib.rs#L1568)
- [x] Check for existence of `Handle` of `msa_id`
- [x] Possible check for encoding? (If we are checking encoding while claiming, this might be a redundant one)

rlaferla:
# Goal
The goal of this PR is to implement a SignedExtension for retire_handle() which is a free extrinsic

Closes #1370


JoeCap08055 (approved):
Looks good. Only 2 comments:
* This SE still needs to be activated by adding to the Frequency runtime as well as the overhead benchmark
* The benchmark for any extrinsic that is in the hot path for this SE needs to be updated to manually call the SE. This is because the overhead benchmark will not include the cost of the internal SE logic, due to the fact that the extrinsic subtype wouldn't match the extrinsic being targeted by this SE. So we need to include the weight in the benchmark for each extrinsic targeted by the SE.

As this is now our 3rd (4th if you count the `stateful-storage` SE that is still pending) SE, it's probably time to have an architectural discussion about our use of SEs and how we might be able to consolidate them into a common framework.

I'm approving this, since it's correct according to our current as-built design, and as it doesn't actually activate the SE yet, there's no harm.

pallets/handles/src/handles_signed_extension.rs
```diff
@@ -0,0 +1,130 @@
...
+	pallet_prelude::ValidTransaction,
+	traits::IsSubType,
+};
+use scale_info::TypeInfo;
+use sp_runtime::{
+	traits::{DispatchInfoOf, SignedExtension},
+	transaction_validity::{InvalidTransaction, TransactionValidity, TransactionValidityError},
+	DispatchError,
+};
+/// The SignedExtension trait is implemented on CheckFreeExtrinsicUse to validate the request. The
```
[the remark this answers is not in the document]
rlaferla: Fixed.

saraswatpuneet (approved):
Looks good few things

* Run `make format`  to resolve linting issues
* Nit: No sure about @JoeCap08055 's feedback on overheads, but if we have covered it then good

let @wilwade  run a pass before merging

rlaferla:
Formatting has been fixed.

aramikm (approved):
Looks good to me. The only concern might be the overhead that we have to address but it doesn't have to be in this PR if it has to go out.

pallets/handles/src/handles_signed_extension.rs
```diff
@@ -0,0 +1,130 @@
...
+/// purpose of this is to ensure that the retire_handle extrinsic cannot be
+/// repeatedly called to flood the network.
+#[derive(Encode, Decode, Clone, Eq, PartialEq, TypeInfo)]
+#[scale_info(skip_type_params(T))]
+pub struct HandlesSignedExtension<T: Config + Send + Sync>(PhantomData<T>);
+
+impl<T: Config + Send + Sync> HandlesSignedExtension<T> {
+	/// Create new `SignedExtension`.
+	pub fn new() -> Self {
```
enddynayn, line 26: Since this is only needed for tests, consider adding a feature flag.

pallets/handles/src/tests/signed_extension_tests.rs
```diff
@@ -0,0 +1,36 @@
+use crate::{handles_signed_extension::HandlesSignedExtension, tests::mock::*};
+use frame_support::{assert_ok, dispatch::DispatchInfo};
+use sp_core::{sr25519, Pair};
+use sp_runtime::traits::SignedExtension;
+
+/// Assert that retiring a handle passes the signed extension HandlesSignedExtension
+#[test]
+fn signed_extension_retire_handle_success() {
```
enddynayn, line 8: can we add test for the not happy path?

pallets/handles/src/handles_signed_extension.rs
```diff
@@ -0,0 +1,130 @@
+//! Substrate Signed Extension for validating requests to the handles pallet
```
enddynayn, line 1: What about just calling the file extension.rs since we are already in the handles directory?

pallets/handles/src/handles_signed_extension.rs
```diff
@@ -0,0 +1,130 @@
...
+	///
+	/// * The delegator must already have a MSA id
+	/// * The MSA must already have a handle associated with it
+	///
+	/// Returns a `ValidTransaction` or wrapped [`pallet::Error`]
+	///
+	/// # Errors (as u8 wrapped by `InvalidTransaction::Custom`)
+	/// * [`Error::InvalidMessageSourceAccount`]
+	/// * [`Error::MSAHandleDoesNotExist`]
+	fn validate_retire_handle(delegator_key: &T::AccountId) -> TransactionValidity {
```
enddynayn, line 59: Why do we implement handles for the pallet?

pallets/handles/src/handles_signed_extension.rs
```diff
@@ -0,0 +1,130 @@
...
+
+/// Map a module DispatchError to an InvalidTransaction::Custom error
+pub fn map_dispatch_error(err: DispatchError) -> InvalidTransaction {
+	InvalidTransaction::Custom(match err {
+		DispatchError::Module(module_err) =>
+			<u32 as Decode>::decode(&mut module_err.error.as_slice())
+				.unwrap_or_default()
+				.try_into()
+				.unwrap_or_default(),
+		_ => 255u8,
```
enddynayn, line 80: can you please explain what 255 is?

enddynayn (commented):
Looking good!

pallets/handles/src/handles_signed_extension.rs
```diff
@@ -0,0 +1,130 @@
...
+	}
+}
+
+/// Validator trait to be used in validating Handles requests from a Signed Extension.
+pub trait HandlesValidate<C: Config> {
+	/// Validates calls to the retire_handle() extrinsic
+	fn validate_retire_handle(delegator_key: &C::AccountId) -> TransactionValidity;
+}
+
+impl<T: Config> HandlesValidate<T> for Pallet<T> {
```
enddynayn, line 48: If we do keep this then maybe we move this into the lib.rs file?

pallets/handles/src/handles_signed_extension.rs
```diff
@@ -0,0 +1,130 @@
...
+/// repeatedly called to flood the network.
+#[derive(Encode, Decode, Clone, Eq, PartialEq, TypeInfo)]
+#[scale_info(skip_type_params(T))]
+pub struct HandlesSignedExtension<T: Config + Send + Sync>(PhantomData<T>);
+
+impl<T: Config + Send + Sync> HandlesSignedExtension<T> {
+	/// Create new `SignedExtension`.
+	pub fn new() -> Self {
+		Self(sp_std::marker::PhantomData)
+	}
```
enddynayn, line 28: you probably move your `fn validate_retire_handle(delegator_key: &T::AccountId) -> TransactionValidity` here if we are not using anywhere else?

pallets/handles/src/handles_signed_extension.rs
```diff
@@ -0,0 +1,130 @@
...
+			.map_err(|_| map_dispatch_error(DispatchError::CannotLookup))?;
+		return ValidTransaction::with_tag_prefix(TAG_PREFIX).build()
+	}
+}
+
+/// Map a module DispatchError to an InvalidTransaction::Custom error
+pub fn map_dispatch_error(err: DispatchError) -> InvalidTransaction {
+	InvalidTransaction::Custom(match err {
+		DispatchError::Module(module_err) =>
+			<u32 as Decode>::decode(&mut module_err.error.as_slice())
```
enddynayn, line 76: can you please explain this?

pallets/handles/src/handles_signed_extension.rs
JoeCap08055: `InvalidTransaction::Custom` takes a `u8`. If the `DispatchError` is not a `DispatchError::Module` from which we can extract a numeric code, then we just send the max u8 value (255) to indicate it's an unknown error. It's a little bit hacky, but there's no really good way to map a DispatchError to an InvalidTransaction error.

pallets/handles/src/handles_signed_extension.rs
rlaferla: For now, just being consistent with stateful storage.  This all likely going to change anyways as we have been accruing signed extensions.

pallets/handles/src/handles_signed_extension.rs
rlaferla: @JoeCap08055 Thoughts as this would likely be the case for stateful storage as well?

pallets/handles/src/handles_signed_extension.rs
rlaferla: This is to map a DispatchError to an InvalidTransaction error as per @JoeCap08055

pallets/handles/src/tests/signed_extension_tests.rs
rlaferla: Adding :-( path.

pallets/handles/src/handles_signed_extension.rs
```diff
@@ -0,0 +1,130 @@
...
+	/// * [`Error::InvalidMessageSourceAccount`]
+	/// * [`Error::MSAHandleDoesNotExist`]
+	fn validate_retire_handle(delegator_key: &T::AccountId) -> TransactionValidity {
+		const TAG_PREFIX: &str = "HandlesRetireHandle";
+
+		// Validation: The delegator must already have a MSA id
+		let delegator_msa_id = T::MsaInfoProvider::ensure_valid_msa_key(&delegator_key)
+			.map_err(|e| map_dispatch_error(e))?;
+		// Validation: The MSA must already have a handle associated with it
+		_ = MSAIdToDisplayName::<T>::try_get(delegator_msa_id)
```
enddynayn, line 66: can we do here:
```
 MSAIdToDisplayName::<T>::try_get(delegator_msa_id).ok_or(UnknownTransaction::CannotLookup)?
 ```
 Also, I think we should use the helper methods for MSAIdToDisplayName?

File: pallets/handles/src/handles_signed_extension.rs (hunk 6 above)
Comment (enddynayn) [reply pallets/handles/src/handles_signed_extension.rs:80]: I noticed that Substrate uses zero. Why did we decide to use 255?

File: pallets/handles/src/handles_signed_extension.rs (hunk 9 above)
Comment (enddynayn) [inline pallets/handles/src/handles_signed_extension.rs:76]: I think you can remove
```

<u32 as Decode>::decode(&mut module_err.error.as_slice())
	.unwrap_or_default()
				.try_into()
				.unwrap_or_default(),
```
In favor of
```
DispatchError::Module(ModuleError { error, .. }) => error[0],
		_ => 0,
	};
```

Module Error looks like it has the index.

https://paritytech.github.io/substrate/master/sp_runtime/struct.ModuleError.html

File: pallets/handles/src/handles_signed_extension.rs [hunk 11]
```diff
@@ -0,0 +1,130 @@
...
+	type Call = T::RuntimeCall;
+	type AdditionalSigned = ();
+	type Pre = ();
+	const IDENTIFIER: &'static str = "HandlesSignedExtension";
+
+	fn additional_signed(&self) -> sp_std::result::Result<(), TransactionValidityError> {
+		Ok(())
+	}
+
+	fn pre_dispatch(
```
enddynayn, line 98: docs

pallets/handles/src/handles_signed_extension.rs
```diff
@@ -0,0 +1,130 @@
...
+where
+	T::RuntimeCall: Dispatchable<Info = DispatchInfo> + IsSubType<Call<T>>,
+{
+	type AccountId = T::AccountId;
+	type Call = T::RuntimeCall;
+	type AdditionalSigned = ();
+	type Pre = ();
+	const IDENTIFIER: &'static str = "HandlesSignedExtension";
+
+	fn additional_signed(&self) -> sp_std::result::Result<(), TransactionValidityError> {
```
enddynayn, line 94: docs

pallets/handles/src/tests/signed_extension_tests.rs
```diff
@@ -0,0 +1,36 @@
...
+	new_test_ext().execute_with(|| {
+		let base_handle_str = "test1";
+
+		// Claim the handle
+		let alice = sr25519::Pair::from_seed(&[0; 32]);
+		let (payload, proof) =
+			get_signed_claims_payload(&alice, base_handle_str.as_bytes().to_vec());
+		assert_ok!(Handles::claim_handle(
+			RuntimeOrigin::signed(alice.public().into()),
```
enddynayn, line 17: Why is it necessary to create keys here? We do not do this in any other mock file.

pallets/handles/src/tests/signed_extension_tests.rs
```diff
@@ -0,0 +1,36 @@
...
+#[test]
+fn signed_extension_retire_handle_success() {
+	new_test_ext().execute_with(|| {
+		let base_handle_str = "test1";
+
+		// Claim the handle
+		let alice = sr25519::Pair::from_seed(&[0; 32]);
+		let (payload, proof) =
+			get_signed_claims_payload(&alice, base_handle_str.as_bytes().to_vec());
+		assert_ok!(Handles::claim_handle(
```
enddynayn, line 16: Can we set this directly in the storage? This would simplify the test and speed it up.

pallets/handles/src/tests/signed_extension_tests.rs
rlaferla: Added!

pallets/handles/src/handles_signed_extension.rs
rlaferla: Yes...

```
pub fn map_dispatch_error(err: DispatchError) -> InvalidTransaction {
	InvalidTransaction::Custom(match err {
		DispatchError::Module(ModuleError { error, .. }) => error[0],
		_ => 0,
	})
}
```

enddynayn (approved):
👍 Nice!

pallets/handles/src/handles_signed_extension.rs
JoeCap08055: zero could be a valid error enum value from the pallet. 255 is highly unlikely to be.

PR #1400, opened 2023-04-14
Fixed default Makefile target

rlaferla:
# Goal
The goal of this PR is to fix the default Makefile target.

Restores the default from before the clean target was added.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,8 @@
 UNAME := $(shell uname)
 
+.PHONY: all
+all: start
+
 .PHONY: clean
 clean:
 	cargo clean


wilwade (approved):

Makefile
```diff
@@ -1,5 +1,8 @@
 UNAME := $(shell uname)
 
+.PHONY: all
+all: start
+
```
JoeCap08055, line 5: This works. Another way to do this would be:
```
.DEFAULT_GOAL := start
```

JoeCap08055 (approved):

PR #1402, opened 2023-04-14
Moved suffix.rs to handles-utils package

rlaferla:
# Goal
The goal of this PR is to move suffix generation into the handles-utils package.


Makefile
wilwade, line 1: Rebase issue?

wilwade (approved):
One rebase issue likely, but otherwise looks good to me

PR #1409, opened 2023-04-14
 Add unit test for multihash panic in multiformats/multihash

### LINKED ISSUE #1233: [Bug] validate_cid can cause a panic
> Unsafe unwrap() in rust-multihash dependency

The extrinsic add_ipfs_message in [pallets/messages/src/lib.rs](https://github.com/LibertyDSNP/frequency/blob/c6c53f3d8b246da402a8bc97f51b2d98d8341398/pallets/messages/src/lib.rs#L222) invokes `validate_cid()` function in which a call to `Cid::read_bytes()` is also made. This function uses the [rust-multihash](https://github.com/multiformats/rust-multihash) dependency where the function [read_u64](https://github.com/multiformats/rust-multihash/blob/master/src/multihash.rs#LL360C14-L360C24) contains [Ok(decode::u64(&b[..=i]).unwrap()](https://github.com/multiformats/rust-multihash/blob/master/src/multihash.rs#L368). If the unwrap() fails, this could potentially cause nodes to panic. Thus, it could provide attackers a low cost opportunity to perform a DoS attack against the whole parachain.
Issue

The below code invokes Cid::read_bytes where, if cid_b is not in a correct format, a panic could happen.
```
pub fn validate_cid(in_cid: &Vec<u8>) -> Result<Vec<u8>, DispatchError> {
		...
		ensure!(Cid::read_bytes(&cid_b[..]).is_ok(), Error::<T>::InvalidCid);
		...
}
```

The below call can be used to reproduce the unwrap() issue and cause a crash:

`RuntimeCall::Messages(Call::add_ipfs_message { schema_id: 0, cid: [0, 1, 0, 1, 203, 155, 0, 0, 0, 5, 67], payload_length: 407371785 })`

## Risk

Since this can be triggered by a call to an extrinsic an attacker could cause a validator node to crash by submitting a specific crafted extrinsic and cause them missing their block authoring slot. This could lead to a denial-of-service of the whole chain with only very low cost requirements for the attacker.

## Mitigation

There are multiple possibilities to mitigate the issue. The first one is forking the dependency, removing the unwrap() and matching the Error from there. A second possible fix would be to check for the read bytes to be formatted correctly (with a correct length for example) so the unwrap issue could never be triggered.

mattheworris:
# Goal
The goal of this PR is to add a unit test in the messages pallet to detect a panic on some types of bad cids.
`multihash` v0.18.1 fixes an issue with calling `unwrap()` on certain types of errors in a `no-std` environment.
`multihash` is included by the `cid` crate used for IPFS messages.
[`Cargo.lock` was updated by another PR before this was merged, so v0.18.1 is already there]

Closes #1233 

Further updates in progress #1417 
https://github.com/multiformats/rust-cid/pull/135

### DIFF
--- a/pallets/messages/src/tests/other_tests.rs
+++ b/pallets/messages/src/tests/other_tests.rs
@@ -643,3 +643,17 @@ fn validate_cid_not_utf8_aligned_errors() {
 		assert_noop!(MessagesPallet::validate_cid(&bad_cid), Error::<Test>::InvalidCid);
 	})
 }
+
+#[test]
+fn validate_cid_not_correct_format_errors() {
+	new_test_ext().execute_with(|| {
+		// This should not panic, but should return an error.
+		let bad_cid = vec![0, 1, 0, 1, 203, 155, 0, 0, 0, 5, 67];
+
+		assert_noop!(MessagesPallet::validate_cid(&bad_cid), Error::<Test>::InvalidCid);
+		// This should not panic, but should return an error.
+		let another_bad_cid = vec![241, 0, 0, 0, 0, 0, 128, 132, 132, 132, 58];
+
+		assert_noop!(MessagesPallet::validate_cid(&another_bad_cid), Error::<Test>::InvalidCid);
+	})
+}


Cargo.lock
```diff
@@ -859,7 +909,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "fd94671561e36e4e7de75f753f577edafb0e7c05d6e4547229fdf7938fbcd2c3"
 dependencies = [
  "core2",
- "multihash 0.18.0",
+ "multihash 0.18.1",
```
mattheworris, line 912: This is the crate version that fixes the panic.

wilwade (approved):

Cargo.lock
JoeCap08055: I suggest bumping the dependency in `Cargo.toml` as well.

JoeCap08055: This will technically work, but it might be nice to submit a PR to https://github.com/multiformats/rust-cid/blob/master/Cargo.toml to bump the actual dependency there.

JoeCap08055: Then we could bump our `cid` dependency in `Cargo.toml`.
Meanwhile, approving--but maybe file a chore to follow up once the `cid` crate bumps the underlying dependency.

JoeCap08055 (approved):

Cargo.lock
JoeCap08055: @saraswatpuneet , it's not a manual update of the lock file; I think Cargo updated the indirect dependency based on the rules, now that multihash has published a version with the fix. But my previous comments stand--we should encourage the `cid` crate to update their dependency, and then update ours explicitly.

saraswatpuneet (approved):

PR #1411, opened 2023-04-16
Sequence index reset when min suffix changes 

saraswatpuneet:
# Goal
The goal of this PR is  to prevent requirement of large migrations when suffix ranges are updated

Please think about it and propose some interesting solution, in this PR solution 1 is proposed

**Solution 1**
* Save suffix index as a tuple of (index, min)
* When check the index for next suffix if min has changes, suffix index we changed to 0

**Solution 2**
?


pallets/handles/src/lib.rs
```diff
@@ -146,7 +146,7 @@ pub mod pallet {
 	#[pallet::storage]
 	#[pallet::getter(fn get_current_suffix_index_for_canonical_handle)]
 	pub type CanonicalBaseHandleToSuffixIndex<T: Config> =
-		StorageMap<_, Blake2_128Concat, Handle, SequenceIndex, OptionQuery>;
+		StorageMap<_, Blake2_128Concat, Handle, (SequenceIndex, HandleSuffix), OptionQuery>;
```
wilwade, line 149: That's an confusing type name

pallets/handles/src/lib.rs
```diff
@@ -146,7 +146,7 @@ pub mod pallet {
 	#[pallet::storage]
```
wilwade, line 146: Update Docs

wilwade (commented):
Nice simple solution. Two small notes

pallets/handles/src/lib.rs
```diff
@@ -227,8 +230,14 @@ pub mod pallet {
 				None => {
 					next = 0;
 				},
-				Some(current) => {
-					next = current.checked_add(1).ok_or(Error::<T>::SuffixesExhausted)?;
+				Some((current, suffix_min)) => {
+					let current_suffix_min = T::HandleSuffixMin::get();
+					// if saved suffix min has changed, reset the suffix index
+					if current_suffix_min != suffix_min {
+						next = 0;
```
rlaferla, line 237: If the sequence is reset to zero, won't this create duplicates?  If not, why?

wilwade: This assumes that we never have an overlapping min-max range.

If it overlaps ever, it doesn't matter what we do, we might duplicate as the shuffle will be different.

Example:

1-5 Shuffled (easy)
- 3
- 4
- 2
- 1
- 5

2-6 Shuffled
- 4
- 6
- 3
- 5
- 2

wilwade (approved):
When we move to Config constants we can add a test for this, but currently I don't think that is possible.

PR #1418, opened 2023-04-17
Handles replay tests cleanups

### LINKED ISSUE #1413: [Handles] Recheck replay tests
- [ ]  Check over replay tests and make sure they are testing the failure case when inside of 100 blocks. test_handle_claim_replay_attack does not. Some of these might be duplicates. See pallets/handles/src/tests/handles_replay_attack_test.rs
- [ ] ?

saraswatpuneet:
# Goal
The goal of this PR is to cleanup replay tests and with better naming 

Closes #1413 

# Checklist
- [ ] Chain spec updated
- [ ] Design doc(s) updated
- [x] Tests added
- [ ] Benchmarks added
- [ ] Weights updated

### DIFF
--- a/pallets/handles/src/tests/handle_retirements_tests.rs
+++ b/pallets/handles/src/tests/handle_retirements_tests.rs
@@ -1,12 +1,13 @@
-use crate::{tests::mock::*, Error, Event};
+use crate::{handles_signed_extension::HandlesSignedExtension, tests::mock::*, Error, Event};
 use codec::Decode;
 use common_primitives::{handles::*, msa::MessageSourceId};
 use frame_support::{
 	assert_noop, assert_ok,
-	dispatch::{GetDispatchInfo, Pays},
+	dispatch::{DispatchInfo, GetDispatchInfo, Pays},
 };
 use numtoa::*;
 use sp_core::{sr25519, Encode, Pair};
+use sp_runtime::traits::SignedExtension;
 
 #[test]
 fn claim_and_retire_handle_happy_path() {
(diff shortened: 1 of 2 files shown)


pallets/handles/src/tests/handle_retirements_tests.rs
```diff
@@ -53,6 +54,64 @@ fn claim_and_retire_handle_happy_path() {
...
+			alice.public().into(),
+			proof.clone(),
+			payload.clone()
+		));
+		let msa_id = MessageSourceId::decode(&mut &alice.public().encode()[..]).unwrap();
+		let handle = Handles::get_handle_for_msa(msa_id);
+		assert!(handle.is_some());
+
+		// Compose the full display handle from the base handle, "." delimeter and suffix
+		let handle_result = handle.unwrap();
```
enddynayn, line 77: i think there is a function here in handles you can use for this.

pallets/handles/src/tests/handle_retirements_tests.rs
```diff
@@ -53,6 +54,64 @@ fn claim_and_retire_handle_happy_path() {
...
+			&RuntimeCall::Handles(HandlesCall::retire_handle {});
+		let info = DispatchInfo::default();
+		let len = 0_usize;
+		let early_retire_result = HandlesSignedExtension::<Test>::new().validate(
+			&alice.public().into(),
+			call_retire_handle,
+			&info,
+			len,
+		);
+		assert!(early_retire_result.is_err());
```
enddynayn, line 96: Do we need to make sure that we are getting the expected here we want here?

pallets/handles/src/tests/handle_retirements_tests.rs
saraswatpuneet: That is with the suffix index, at this place we are getting handle from state and forming a display handle (ow a consumer of rpc would do)

pallets/handles/src/tests/handle_retirements_tests.rs
saraswatpuneet: We are safe from the point of view of that this is a direction function call so it is expected to return error, but I will see if there is a way to do some comparison

enddynayn (approved):
👍 Nice!

wilwade (approved):
Much better naming and test. Thanks!

PR #1421, opened 2023-04-17
Using Build.rs to generate confusables

### LINKED ISSUE #1353: Replace confusables generator CLI tool with phf_codegen
https://docs.rs/phf_codegen/latest/phf_codegen/

wilwade:
Use build.rs inside of handles-utils to generate the confusables instead of committing it

Closes #1353 

# Discussion
Simple update to make the code gen from confusables run at build time.


pallets/handles/src/handles-utils/src/converter.rs
```diff
@@ -2,14 +2,22 @@
...
+
 use unicode_normalization::{char::is_combining_mark, UnicodeNormalization};
 extern crate alloc;
+use crate::types::HandleSuffix;
 use alloc::{
 	string::{String, ToString},
 	vec::Vec,
 };
 
+/// Will load up with all the confusables at build time.
```
aramikm, line 14: I was thinking about another way of doing this. instead of running this on every build we would just add it to our node cli and when we ran it it will write the `confusables.rs` and we can use it in our pallet. This would be really similar to how weight generation works. We don't need to re-generate this on every build, we only need to generate it when we edited `confusable_characters.txt`

aramikm (approved):
I think this will work but I added an alternative path in a comment.

pallets/handles/src/handles-utils/src/converter.rs
rlaferla: There should be a way to add the data file as a build dependency so it only rebuilds it if the file changed.

aramikm: I mean there are arguments to be made for either case.

For automated generation in case of changes you don't have to think about it. You go and change and everything is automated.

For non automated cli generation my argument is that is allows you decouple the changing of these files from the generating, which makes the changes more intentional. As I mentioned before kinda like weight generation.

wilwade: @rlaferla correct. It only runs if certain files change.

@aramikm There is another PR coming that changes this so that it only changes if a constants file is changed and if that changes or the confusables, then we need to rerun benchmarks anyway.

aramikm: It still needs to generate it the first time if it's building from scratch which might be fine. I see the benefits of doing it like this but this has it's own drawbacks such as more complex code resolving. Having the generated file checked in the github allows to easily see how everything is wired together.

wilwade: Agreed. The next one will start altering the list off of the list of allowed char (savings of 50% in rust code file size!)

So I think that makes it more worthwhile to go this direction (changes are triggered when that allowed list changes).

Here's the draft PR of that off of this one: https://github.com/LibertyDSNP/frequency/pull/1422 if that helps to see that next step.

aramikm: FYI: this generates two files
**➜  ~/source/mrc: find ./ -name confusables.rs**
```
.//target/release/wbuild/frequency-runtime/target/wasm32-unknown-unknown/release/build/handles-utils-10a27716b8ce5877/out/confusables.rs
.//target/release/build/handles-utils-28f38b0ca9ade246/out/confusables.rs
```

johnmela (approved):

saraswatpuneet (approved):
Lgtm

pallets/handles/src/handles-utils/src/converter.rs
```diff
@@ -2,14 +2,22 @@
...
+/// See build.rs
+#[derive(Clone)]
+enum CONFUSABLES {}
```
aramikm, line 17: this looks redundant. doesn't it?

PR #1422, opened 2023-04-17
Remove unneeded confusables

### LINKED ISSUE #1414: [Handles] Reduce WASM size by removing unneeded confusables
Depends on #1353

When building the confusables, we should remove any that are not within the supported ranges so that we get a smaller/faster WASM.

wilwade:
# Goal
The goal of this PR is to remove the unneeded confusables from the confusables tree for a much smaller wasm file.

Saved ~13k in the wasm (when doing a release build, not a production optimized build)

Closes #1414

# Discussion
Moving the const to the constants.rs file is not considered normal good practice, but the other option is moving it to an entire separate package.

# Checklist
- [x] Benchmarks (handles)


pallets/handles/src/handles-utils/src/converter.rs
```diff
@@ -2,14 +2,22 @@
...
+use crate::types::HandleSuffix;
 use alloc::{
 	string::{String, ToString},
 	vec::Vec,
 };
 
+/// Will load up with all the confusables at build time.
+/// See build.rs
+#[derive(Clone)]
+enum CONFUSABLES {}
```
aramikm, line 17: why do we need this? It looks redundant.

wilwade: Checking...

wilwade: Correct! Removed.

pallets/handles/src/handles-utils/build.rs
```diff
@@ -32,17 +39,25 @@ fn main() {
 		// The first character in the line will be the value of each key
 		let normalized_character = original_line_characters.next().unwrap();
 
+		// Skip if the normalized is not in the allowed range
+		if !allowed_char(normalized_character) {
```
aramikm, line 43: So I'm guessing that the first character (representation) is enough to skip the whole set

aramikm (approved):
Looks good!

pallets/handles/src/handles-utils/build.rs
wilwade: Right. If the normalized character is not allowed, it will just be rejected as we test that the canonical only has allowed characters.

saraswatpuneet (approved):
👍

PR #1442, opened 2023-04-21
DRAFT: 1401 Build Frequency runtime without relay chain

### LINKED ISSUE #1401: Build/run Frequency runtime without a relay chain
When running the chain as a solo local chain without a parachain, I cannot upgrade because the local solo chain is still waiting for the relay chain which doesn't exist.

Currently in `runtime/frequency/src/lib.rs`:
```rust
type OnSetCode = cumulus_pallet_parachain_system::ParachainSetCode<Self>;
```

This should be `type OnSetCode = ()` when running without a parachain (aka feature `frequency-rococo-local`) however changing that requires removing the other bits of the parachain pallets and config as well.

All doable with feature flags, but a fair amount of them.

Might also be worth seeing if we want to add the instant sealing wasm to the release builds for wasm upgrades

rlaferla:
# Goal
The goal of this PR is to segment local/parachain with existing feature flags

Closes #1401


node/cli/src/command.rs
```diff
@@ -387,68 +383,96 @@ pub fn run() -> Result<()> {
...
+		}
+	})
+}
 
-				let state_version = Cli::native_runtime_version(&config.chain_spec).state_version();
-				let block: Block = generate_genesis_block(&*config.chain_spec, state_version)
-					.map_err(|e| format!("{:?}", e))?;
-				let genesis_state = format!("0x{:?}", HexDisplay::from(&block.header().encode()));
+#[cfg(any(not(feature = "frequency-rococo-local"), feature = "all-frequency-features"))]
+use codec::Encode;
```
wilwade, line 413: I've found for these shorter ones it is usually better to either go fully qualified or to put the use statement inside the function. However not always the best choice. Depends a lot on the setup.

node/cli/src/command.rs
```diff
@@ -231,6 +226,7 @@ macro_rules! construct_async_run {
 }
 
 /// Parse command line arguments into service configuration.
+#[rustfmt::skip] // This is necessary to workaround a bug in rustfmt with the function returns in the None case.
```
wilwade, line 229: Might be worth putting the returns into a separate function call if that would help remove this. It's a lot of nested code to lose formatting on.

node/cli/src/command.rs
```diff
@@ -387,68 +383,96 @@ pub fn run() -> Result<()> {
...
+			))]
+			return run_parachain(cli);
+		},
+	}
+}
 
-				let parachain_account =
-					AccountIdConversion::<polkadot_primitives::v2::AccountId>::into_account_truncating(&id);
+#[cfg(feature = "frequency-rococo-local")]
+fn run_local(cli: Cli) -> Result<()> {
```
wilwade, line 399: If `run_local` and `run_parachain` were both named the same, you could remove the flag from inside the `run` command.

runtime/frequency/src/lib.rs
wilwade, line 1: The changes here are triggering the change to the metadata that CI is catching. I think we could either setup CI to test against the mainnet version, or I guess I am ok with missing out on these pallets in metadata checking if we have to.

runtime/frequency/src/lib.rs
```diff
@@ -1110,6 +1118,7 @@ construct_runtime!(
 	{
 		// System support stuff.
 		System: frame_system::{Pallet, Call, Config, Storage, Event<T>} = 0,
+		#[cfg(any(not(feature = "frequency-rococo-local"), feature = "all-frequency-features"))]
```
wilwade, line 1121: Does this mean we cannot run as a parachain locally? Or would we now use `frequency-rococo-testnet` to do that?

Makes me wonder if there is a way to do this based on a cli flag like manual sealing is instead of a feature flag

wilwade (commented):
Few questions and thoughts, but looks like we have at least one path forward.

node/cli/src/command.rs
rlaferla: That was existing code.  I can change it if you think it's necessary.

rlaferla:
Closing and re-opening due to CI issues w/integration tests per @saraswatpuneet.

Issue #743, opened 2022-12-05
Code Coverage Improvements


wilwade:
We should take another pass at our code coverage and see where it may be worth improving it.

https://app.codecov.io/gh/LibertyDSNP/frequency

- [x] RPC tests
- [x] Remove `common/primitives/src/benchmarks.rs` from coverage
- [x] Adding simple tests?

PR #1448, opened 2023-04-24
Increase test coverage and Minor RPC Bug

wilwade:
# Goal
The goal of this PR is to improve code coverage (+4.11%) and fix a bug discovered while working on it

Bug: When there was no `msa_id` for a message, the RPC response would be `Some(0)` instead of `None`

Part of #743 

# Discussion
- Removed the runtime-apis from code coverage as they are just traits
- Added tests for various areas
- Moved Frequency-Tx-Payment pallet tests into a folder


pallets/frequency-tx-payment/src/tests/pallet_tests.rs
```diff
@@ -1,5 +1,7 @@
-use super::*;
-use crate::{self as pallet_frequency_tx_payment, mock::*, ChargeFrqTransactionPayment};
+use crate::{
```
shannonwells, line 1: With @rlaferla's test refactor, we decided to organize tests by function and not location, but this can be in a follow-up.

pallets/schemas/src/migration.rs
```diff
@@ -85,27 +67,44 @@ pub mod v1 {
 // ==============================================
 /// Schema migration to v1 for pallet-stateful-storage
 /// This struct derives OnRuntimeUpgrade trait which is used to run the migration (check runtime)
-pub struct SchemaMigration<T: Config>(PhantomData<T>);
-
-impl<T: Config> OnRuntimeUpgrade for SchemaMigration<T> {
-	#[cfg(feature = "try-runtime")]
```
shannonwells, line 91: What's the motivation for removing all this try-runtime code?  It's kind of out of scope for this PR.

pallets/schemas/src/migration.rs
```diff
@@ -85,27 +67,44 @@ pub mod v1 {
...
 
+impl<T: Config> OnRuntimeUpgrade for Migration<T> {
 	fn on_runtime_upgrade() -> Weight {
-		v1::migrate_schemas_to_v1::<T>()
-	}
+		let version = StorageVersion::get::<Pallet<T>>();
+		let mut weight: Weight = Weight::zero();
+
+		if version < 1 {
+			weight = weight.saturating_add(v1::migrate::<T>());
```
shannonwells, line 78: Is this migration not idempotent?

runtime/common/src/constants.rs
```diff
@@ -20,26 +20,26 @@ pub const TOKEN_DECIMALS: u8 = 8;
 
 parameter_types! {
 	/// Clone + Debug + Eq  implementation for u32 types
-	pub const MaxDataSize: u32 = 30;
+	pub const Const30u32: u32 = 30;
```
shannonwells, line 23: This doesn't seem more clear to me.  Is it always going to be 30? Shouldn't it be MaxSchemaGrants or something?

shannonwells (changes requested):
There's a lot of little cleanup things in this PR which is okay, but I question the new constant name.

pallets/schemas/src/migration.rs
wilwade: It was showing up as not tested and I was under the impression (wrong?) that it was unneeded. So trying to standardize the migration

pallets/schemas/src/migration.rs
wilwade: This is more standardization and following a pattern. As far as I know this is the standard setup so that there are
the minimal on init code run.

runtime/common/src/constants.rs
wilwade: Perhaps? The old one collided with a generic type and it wasn't very clear before either. Open to other names. MaxSchemaGrants works for me

pallets/frequency-tx-payment/src/tests/pallet_tests.rs
wilwade: Agreed. This was more for test coverage calculations which needs things under /test or it picks up extra stuff

runtime/common/src/constants.rs
johnmela: I second `MaxSchemaGrants`

johnmela (changes requested):

runtime/common/src/constants.rs
wilwade: Moved over to #1453

johnmela (approved):

pallets/schemas/src/migration.rs
saraswatpuneet: Yep all good, not sure we even need to keep migration logic anymore, since it must have already migrated all schemas on mainnet

saraswatpuneet (approved):
lgtm

shannonwells (approved):
Requested changes removed and pulled into another PR

PR #1452, opened 2023-04-26
Restrict Pays:No call in batches

### LINKED ISSUE #599: Understand Security issues of Pays::No, and pre-dispatch validation for options to enable pallet_utility
Spike: Understand Security issues of Pays::No, and pre-dispatch validation for options to enable pallet_utility

saraswatpuneet:
# Goal
The goal of this PR is to prevent bad actors filling the tx pool with free txs (`Pays::No`) by filtering specific calls from frequency's pallets out espcially `msa` and `handles` which have free extrinsics
 
Closes #599 

# Checklist
- [ ] Chain spec updated
- [ ] Design doc(s) updated
- [x] Tests added
- [ ] Benchmarks added
- [ ] Weights updated


runtime/frequency/src/lib.rs
```diff
@@ -150,6 +148,31 @@ impl Contains<RuntimeCall> for BaseCallFilter {
...
+            RuntimeCall::Msa(pallet_msa::Call::revoke_delegation_by_delegator { .. }) |
+            RuntimeCall::Msa(pallet_msa::Call::revoke_delegation_by_provider { .. }) |
+            RuntimeCall::Msa(pallet_msa::Call::delete_msa_public_key { .. }) |
+            RuntimeCall::Msa(pallet_msa::Call::retire_msa { .. }) |
+            // Block following `Pays::No` from `pallet-handles` calls
+            RuntimeCall::Handles(pallet_handles::Call::retire_handle { .. }) => true,
+            _ => false,
+        }
+	}
+}
```
saraswatpuneet, line 174: This is one of the ways to do it, looking for more patterns and design decisions

saraswatpuneet: The batch will fail if anyone of these transactions match in the list of calls sent to utility

runtime/frequency/src/lib.rs
```diff
@@ -150,6 +148,31 @@ impl Contains<RuntimeCall> for BaseCallFilter {
...
+			RuntimeCall::Utility(pallet_utility::Call::batch { calls, .. }) |
+			RuntimeCall::Utility(pallet_utility::Call::batch_all { calls, .. }) |
+			RuntimeCall::Utility(pallet_utility::Call::force_batch { calls, .. }) =>
+				!calls.iter().any(Self::is_blocked_call),
+			_ => false,
+		}
+	}
+
+	fn is_blocked_call(call: &RuntimeCall) -> bool {
+		match call {
```
enddynayn, line 163: I suggest calling this `is_pays_no_call`.

What would happen if we did not allow any free transaction to be called using the Utility batch?

```
if call.get_dispatch_info().pays_fee == Pays::No {
   true
}

false
```

johnmela: If that works ^ I like it better. That way the blocked call list doesn't require maintenance

saraswatpuneet: @wilwade  would we want to block every `Pays::No` call? I dont think so, then we will end up blocking basic system calls too, like remark etc.
@enddynayn what you think

runtime/frequency/src/lib.rs
```diff
@@ -130,16 +130,14 @@ impl Contains<RuntimeCall> for BaseCallFilter {
...
+				RuntimeCall::Utility(..) => Self::is_utility_call_allowed(call),
 				_ => true,
 			}
 		}
 		#[cfg(feature = "frequency")]
 		{
 			match call {
-				// Utility Calls are blocked. Issue #599
-				RuntimeCall::Utility(..) => false,
+				RuntimeCall::Utility(..) => Self::is_utility_call_allowed(call),
```
saraswatpuneet, line 140: Following implementation is only invoked for utility calls, so it wont affect any other pallet calls

runtime/frequency/src/lib.rs
```diff
@@ -150,6 +148,31 @@ impl Contains<RuntimeCall> for BaseCallFilter {
 	}
 }
 
+impl BaseCallFilter {
+	fn is_utility_call_allowed(call: &RuntimeCall) -> bool {
+		match call {
+			RuntimeCall::Utility(pallet_utility::Call::batch { calls, .. }) |
+			RuntimeCall::Utility(pallet_utility::Call::batch_all { calls, .. }) |
+			RuntimeCall::Utility(pallet_utility::Call::force_batch { calls, .. }) =>
+				!calls.iter().any(Self::is_blocked_call),
```
saraswatpuneet, line 157: this O(n) for now, thinking of ways to optimize this, given this is in wasm

saraswatpuneet: This is  very rudimentary optimization https://github.com/LibertyDSNP/frequency/compare/main..599_optimized_for_wasm

but we can move out to a const BTreeMap to maintain block calls

runtime/frequency/src/lib.rs
```diff
@@ -150,6 +148,32 @@ impl Contains<RuntimeCall> for BaseCallFilter {
 	}
 }
 
+impl BaseCallFilter {
+	fn is_utility_call_allowed(call: &RuntimeCall) -> bool {
+		match call {
+			RuntimeCall::Utility(pallet_utility::Call::batch { calls, .. }) |
+			RuntimeCall::Utility(pallet_utility::Call::batch_all { calls, .. }) |
+			RuntimeCall::Utility(pallet_utility::Call::force_batch { calls, .. }) =>
```
saraswatpuneet, line 156: Doing it for all batch extrinsic in utility pallet, let me know if I missed any

runtime/frequency/src/lib.rs
enddynayn: We are blocking these only when calling batch.

wilwade: @saraswatpuneet I think we do want to. What pays::no call might we want to allow?

wilwade: @enddynayn Ah yes. You are correct. We only want to block them from batches. So I think we do want to block all from batching.

runtime/frequency/src/lib.rs
```diff
@@ -150,6 +148,32 @@ impl Contains<RuntimeCall> for BaseCallFilter {
...
+
+	fn is_batch_call_allowed(call: &RuntimeCall) -> bool {
+		match call {
+			// Block all `FrequencyTxPayment` calls from utility batch
+			// Block all `Pays::No` calls from utility batch
+			RuntimeCall::FrequencyTxPayment(..) => true,
+			_ if Self::is_pays_no_call(call) => true,
+			_ => false,
+		}
+	}
```
saraswatpuneet, line 170: @wilwade  @enddynayn  here is updated version

integration-tests/miscellaneous/utilityBatch.test.ts
```diff
@@ -0,0 +1,45 @@
...
+        goodBatch.push(ExtrinsicHelper.api.tx.balances.transfer(recipient.address, 1000))
+        goodBatch.push(ExtrinsicHelper.api.tx.system.remark("Hello From Batch"))
+        goodBatch.push(ExtrinsicHelper.api.tx.msa.create())
+        const batch = ExtrinsicHelper.executeUtilityBatchAll(sender, goodBatch);
+        const [event, eventMap] = await batch.fundAndSend();
+        assert.notEqual(event, undefined, "should return an event");
+        assert.notEqual(eventMap, undefined, "should return an eventMap");
+    });
+
+    it("should fail to execute ❌ batch with disallowed calls", async function () {
```
wilwade, line 28: This would fail as long as any one of the calls inside the batch is bad. I think you'd need to look over them with "one" call per batch attempt

saraswatpuneet: Added more tests to cover some cases

johnmela (approved):

runtime/frequency/src/lib.rs
```diff
@@ -150,6 +148,44 @@ impl Contains<RuntimeCall> for BaseCallFilter {
...
+
+			// Block all `FrequencyTxPayment` calls from utility batch
+			RuntimeCall::FrequencyTxPayment(..) => false,
+
+			// Block `create_provider` and `create_schema` calls from utility batch
+			RuntimeCall::Msa(pallet_msa::Call::create_provider { .. }) |
+			RuntimeCall::Schemas(pallet_schemas::Call::create_schema { .. }) => false,
+
+			// Block all `Pays::No` calls from utility batch
+			_ if Self::is_pays_no_call(call) => false,
```
johnmela, line 177: Excellent !

integration-tests/miscellaneous/utilityBatch.test.ts
```diff
@@ -0,0 +1,142 @@
...
+        sender = await createAndFundKeypair();
+        recipient = await createAndFundKeypair();
+    });
+
+    it("should successfully execute ✅ batch with allowed calls", async function () {
+        // good batch: with only allowed calls
+        const goodBatch: SubmittableExtrinsic<ApiTypes>[] = [];
+        goodBatch.push(ExtrinsicHelper.api.tx.balances.transfer(recipient.address, 1000))
+        goodBatch.push(ExtrinsicHelper.api.tx.system.remark("Hello From Batch"))
+        goodBatch.push(ExtrinsicHelper.api.tx.msa.create())
```
shannonwells, line 21: Non-blocking:  What's the reasoning for allowing Msa::create but disallowing create_provider?

integration-tests/miscellaneous/utilityBatch.test.ts
```diff
@@ -0,0 +1,142 @@
...
+            error = err;
+            assert.notEqual(error, undefined, " batchAll should return an error");
+        }
+        assert.notEqual(error, undefined, " batchAll should return an error");
+    });
+
+    it("should fail to execute ❌ batch with disallowed calls", async function () {
+        // bad batch: with a mix of allowed and disallowed calls
+        const badBatch: SubmittableExtrinsic<ApiTypes>[] = [];
+        badBatch.push(ExtrinsicHelper.api.tx.balances.transfer(recipient.address, 1000))
```
shannonwells, line 51: Non-blocking:  would be convenient here to indicate which of these is disallowed

integration-tests/miscellaneous/utilityBatch.test.ts
```diff
@@ -0,0 +1,142 @@
...
+        badBatch.push(ExtrinsicHelper.api.tx.system.remark("Hello From Batch"))
+        badBatch.push(ExtrinsicHelper.api.tx.handles.retireHandle())
+        badBatch.push(ExtrinsicHelper.api.tx.msa.retireMsa())
+
+        // forceBatch
+        const forceBatch = ExtrinsicHelper.executeUtilityForceBatch(sender, badBatch);
+        let [ok, eventMap] = await forceBatch.fundAndSend();
+        assert.equal(ok, undefined, "should not return an ok event");
+        assert.equal(eventMap["utility.BatchCompleted"], undefined, "should not return a batch completed event");
+        assert.notEqual(eventMap["utility.BatchCompletedWithErrors"], undefined, "should return a batch completed with error event");
```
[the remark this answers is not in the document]
shannonwells: Do we have a reason to execute `force_batch` / is it executed by other extrinsics such as democracy or collectives calls, the scheduler, or cumulus?

shannonwells (approved):
Looks great!

runtime/frequency/src/lib.rs
```diff
@@ -150,6 +148,44 @@ impl Contains<RuntimeCall> for BaseCallFilter {
 	}
 }
 
+impl BaseCallFilter {
```
enddynayn, line 151: how are we unit testing filters?

runtime/frequency/src/lib.rs
```diff
@@ -130,16 +130,14 @@ impl Contains<RuntimeCall> for BaseCallFilter {
 		#[cfg(not(feature = "frequency"))]
 		{
 			match call {
-				// Utility Calls are blocked. Issue #599
-				RuntimeCall::Utility(..) => false,
+				RuntimeCall::Utility(..) => Self::is_utility_call_allowed(call),
```
enddynayn, line 133: I wonder we could do something like the following:
```
impl Contains<RuntimeCall> for BaseCallFilter {
	fn contains(call: &RuntimeCall) -> bool {
		#[cfg(not(feature = "frequency"))]
		{
			match call {
				RuntimeCall::Utility(utility_call) => Self::is_utility_call_allowed(utility_call),
				_ => true,
			}
		}
...

```
impl BaseCallFilter {
	fn is_utility_call_allowed(call: &pallet_utility::Call::<Runtime>) -> bool {
		match call {
			pallet_utility::Call::batch { calls, .. } |
			pallet_utility::Call::batch_all { calls, .. } |
			pallet_utility::Call::force_batch { calls, .. } =>
				calls.iter().any(Self::is_pays_no_call),
			_ => false,
		}
	}
```

Since  RuntimeCall::Utility(......THIS IS THE CALL FROM UTILITY......), we can pass that into `is_utility_call_allowed`.

It seems we are doing it now is passing the RuntimeCall down and making an additional match statement to filter out utility calls. I think this might be able to allow to remove that additional function.

File: runtime/frequency/src/lib.rs (hunk 8 above)
Comment (enddynayn) [inline runtime/frequency/src/lib.rs:177]: I do not think this is blocking // Block all `Pays::No` calls from the utility batch.

File: runtime/frequency/src/lib.rs (hunk 13 above)
Comment (saraswatpuneet) [reply runtime/frequency/src/lib.rs]: I like it, did some parts of it as we still have to check for various `RuntimeCall` along with pays not so the `is_batch_call_allowed` checks for blocked `RuntimeCalls` calls within batches hence this function would require  generic +match

but good suggestion

File: integration-tests/miscellaneous/utilityBatch.test.ts (hunk 11 above)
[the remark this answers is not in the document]
Comment (saraswatpuneet) [reply integration-tests/miscellaneous/utilityBatch.test.ts:79]: not sure about any current or future cases for `force_batch` but we are protected from its bad usage now, so not sure if we want to not allow it at all

File: runtime/frequency/src/lib.rs [hunk 14]
```diff
@@ -150,6 +150,43 @@ impl Contains<RuntimeCall> for BaseCallFilter {
...
+
+	fn is_batch_call_allowed(call: &RuntimeCall) -> bool {
+		match call {
+			// Block all nested `batch` calls from utility batch
+			RuntimeCall::Utility(pallet_utility::Call::batch { .. }) |
+			RuntimeCall::Utility(pallet_utility::Call::batch_all { .. }) |
+			RuntimeCall::Utility(pallet_utility::Call::force_batch { .. }) => false,
+
+			// Block all `FrequencyTxPayment` calls from utility batch
+			RuntimeCall::FrequencyTxPayment(..) => false,
```
enddynayn, line 171: Nice!

enddynayn (approved):
👍 Awesome! 💯

runtime/frequency/src/lib.rs
wilwade: Looks good to me

integration-tests/miscellaneous/utilityBatch.test.ts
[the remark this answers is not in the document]
wilwade: I am ok with allowing it. While uncommon, I can see reasons why people would want to use it and just allow some transactions to fail. I don't see any security issues with it thanks to the work you put in place.

If it is used by other extrinsics, then that would bypass the call filters, so it is good either way for that reason at least.

wilwade (approved):
Rebase and ship it!

PR #1453, opened 2023-04-26
Decrease use of parameter types

wilwade:
# Goal
The goal of this PR is to reduce and remove code

Split off of #1448 

# Discussion
- Turns out that parameter_types! is not needed as much anymore so removed a bunch of them (based on https://github.com/paritytech/substrate/issues/10383 )
- Removed the need for custom implementations on `MaxSchemaGrantsPerDelegation`
- Aligned the BlockNumber (old u64) in testing with the BlockNumber (u32) in production

### DIFF
--- a/common/primitives/src/msa.rs
+++ b/common/primitives/src/msa.rs
@@ -53,7 +53,7 @@ impl From<DelegatorId> for MessageSourceId {
 }
 
 /// Struct for the information of the relationship between an MSA and a Provider
-#[derive(TypeInfo, RuntimeDebug, Clone, Decode, Encode, PartialEq, MaxEncodedLen, Eq)]
+#[derive(TypeInfo, RuntimeDebug, Clone, Decode, Encode, MaxEncodedLen, Eq)]
 #[scale_info(skip_type_params(MaxSchemaGrantsPerDelegation))]
 pub struct Delegation<SchemaId, BlockNumber, MaxSchemaGrantsPerDelegation>
 where
@@ -63,6 +63,19 @@ where
 	pub revoked_at: BlockNumber,
 	/// Schemas that the provider is allowed to use for a delegated message.
 	pub schema_permissions: BoundedBTreeMap<SchemaId, BlockNumber, MaxSchemaGrantsPerDelegation>,
+}
+
+// Cannot derive the PartialEq without a mess of impl PartialEq for MaxSchemaGrantsPerDelegation
+impl<SchemaId, BlockNumber, MaxSchemaGrantsPerDelegation> PartialEq
+	for Delegation<SchemaId, BlockNumber, MaxSchemaGrantsPerDelegation>
+where
+	SchemaId: PartialEq,
+	BlockNumber: PartialEq,
+	MaxSchemaGrantsPerDelegation: Get<u32>,
+{
+	fn eq(&self, other: &Self) -> bool {
+		self.revoked_at == other.revoked_at && self.schema_permissions == other.schema_permissions
+	}
 }
 
 impl<
@@ -173,7 +186,7 @@ pub trait ProviderLookup {
 	/// Type for block number.
 	type BlockNumber;
 	/// Type for maximum number of schemas that can be granted to a provider.
-	type MaxSchemaGrantsPerDelegation: Get<u32> + Clone + Eq;
+	type MaxSchemaGrantsPerDelegation: Get<u32>;
 	/// Schema Id is the unique identifier for a Schema
 	type SchemaId;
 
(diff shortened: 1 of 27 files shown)


common/primitives/src/msa.rs
```diff
@@ -65,6 +65,19 @@ where
...
+// Cannot derive the PartialEq without a mess of impl PartialEq for MaxSchemaGrantsPerDelegation
+impl<SchemaId, BlockNumber, MaxSchemaGrantsPerDelegation> PartialEq
+	for Delegation<SchemaId, BlockNumber, MaxSchemaGrantsPerDelegation>
+where
+	SchemaId: PartialEq,
+	BlockNumber: PartialEq,
+	MaxSchemaGrantsPerDelegation: Get<u32>,
+{
+	fn eq(&self, other: &Self) -> bool {
+		self.revoked_at == other.revoked_at && self.schema_permissions == other.schema_permissions
```
wilwade, line 77: By adding this one impl here, we can remove a ton of impl elsewhere

runtime/common/src/constants.rs
```diff
@@ -226,6 +189,7 @@ pub const TREASURY_PALLET_ID: PalletId = PalletId(*b"py/trsry");
 
 // https://wiki.polkadot.network/docs/learn-treasury
 // https://paritytech.github.io/substrate/master/pallet_treasury/pallet/trait.Config.html
+// Needs parameter_types! for the Permill and PalletId
```
wilwade, line 192: I added a simple comment each time we use `parameter_types!` as to why we still have to use it

runtime/common/src/constants.rs
```diff
@@ -1,55 +1,21 @@
...
-		Ok(())
-	}
-
-	#[cfg(not(feature = "std"))]
-	fn fmt(&self, _: &mut sp_std::fmt::Formatter) -> sp_std::fmt::Result {
-		Ok(())
-	}
-}
+/// The maximum number of schema grants allowed per delegation
+pub type MaxSchemaGrants = ConstU32<30>;
```
wilwade, line 18: Much better name now

pallets/stateful-storage/src/lib.rs
```diff
@@ -120,7 +120,7 @@ pub mod pallet {
 
 		/// The maximum number of pages in a Paginated storage model
 		#[pallet::constant]
-		type MaxPaginatedPageId: Get<u32>;
+		type MaxPaginatedPageId: Get<u16>;
```
wilwade, line 123: @aramikm Can you validate that this is ok. Everything appears to be u16 for this lower down.

aramikm: This should have been u16. The issue I've seen before was that `Get<u16>` was not implementing all the traits that `Get<u32>` was, so either we had to use Get<u32> or implement those leftover traits that Parity didn't implement. This was a while ago so they might have added those traits to Get<u16> if you don't see any errors

pallets/capacity/src/tests/mock.rs
```diff
@@ -41,14 +40,14 @@ impl frame_system::Config for Test {
 	type RuntimeOrigin = RuntimeOrigin;
 	type RuntimeCall = RuntimeCall;
 	type Index = u64;
-	type BlockNumber = u64;
+	type BlockNumber = u32;
```
shannonwells, line 43: thanks, this has been bugging me

common/primitives/src/msa.rs
aramikm: It's nice being able to do a deep eq check on `BoundedBTreeMap` in Rust by just `==`. In some other languages this would not be as easy.

shannonwells (approved):
Awesome possum and huzzah for thoughtful PRs

aramikm (approved):
Looks great!

enddynayn (approved):
💯 clean!

pallets/handles/src/tests/mock.rs
```diff
@@ -5,16 +5,15 @@ pub use pallet_handles::Call as HandlesCall;
...
 	utils::wrap_binary_data,
 };
 use frame_support::{
 	dispatch::DispatchError,
-	traits::{ConstU16, ConstU32, ConstU64, OnFinalize, OnInitialize},
+	traits::{ConstU16, ConstU32, OnFinalize, OnInitialize},
 };
 use sp_core::{crypto::AccountId32, sr25519, ByteArray, Encode, Pair, H256};
 use sp_runtime::{
-	testing::Header,
```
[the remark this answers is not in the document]
wilwade: The testing one uses u64, and I cannot see a reason not to use ours. In fact it looks like the testing one is more for internal testing?

Issue #1461, opened 2023-05-01
Frequency api-augment and integration tests upgrades


saraswatpuneet:
Current polkadot js package version is 10.4.1

Need too look into
* substrate metadata recent updates
* Fix api-augment and integration tests to be compatible with latest version 10.5+

PR #1462, opened 2023-05-01
pin polkadotjs api to running version

saraswatpuneet:
# Goal
The goal of this PR is to pin the version of polkadot js as latest builds break and needs a careful update.

Pining to last stable version 10.4.1 https://github.com/polkadot-js/api/releases/tag/v10.4.1

Creates #1461 

# Checklist
- [ ] Chain spec updated
- [ ] Design doc(s) updated
- [ ] Tests added
- [ ] Benchmarks added
- [ ] Weights updated


wilwade (approved):
All worked locally with clearing out the node_modules first (on both). So looks good to me if CI passes!

PR #1476, opened 2023-05-05
[chore] Add lint clippy::unwrap_used to modules and/or functions that must not panic

mattheworris:
# Goal
The goal of this PR is 

Closes 

# Discussion

# Checklist
- [ ] Chain spec updated
- [ ] Design doc(s) updated
- [ ] Tests added
- [ ] Benchmarks added
- [ ] Weights updated


pallets/capacity/src/benchmarking.rs
```diff
@@ -11,6 +11,7 @@ fn assert_last_event<T: Config>(generic_event: <T as Config>::RuntimeEvent) {
 	frame_system::Pallet::<T>::assert_last_event(generic_event.into());
 }
 
+#[allow(clippy::expect_used)]
```
rlaferla, line 14: Since this is benchmarking, should this be for the entire file/module?

pallets/capacity/src/types.rs
```diff
@@ -55,6 +55,7 @@ impl<T: Config> StakingAccountDetails<T> {
 	}
 
 	#[cfg(any(feature = "runtime-benchmarks", test))]
+	#[allow(clippy::unwrap_used)]
```
rlaferla, line 58: As this in the runtime, should we be allowing it?  Is there a way to display a warning but allow the build to proceed?  e.g.  we don't have to drop everything and fix it but be aware that we should fix it at some point?

pallets/stateful-storage/src/benchmarking.rs
```diff
@@ -1,3 +1,4 @@
+#![allow(clippy::unwrap_used, clippy::expect_used)]
```
rlaferla, line 1: Perhaps all benchmarking.rs files should use this allow statement for both unwrap and expect.

rlaferla (commented):
Please address my comments but otherwise looks good.

rlaferla (changes requested):
Please address my comments but otherwise looks good.

pallets/capacity/src/benchmarking.rs
mattheworris: I think we should keep the exceptions scope as limited as possible. If the we need to expand the exception to more than one location, it might make sense to make it file/module scope.

pallets/capacity/src/types.rs
mattheworris: The comment for the function says it's for testing only and it has `feature = "runtime_benchmarks"` or `feature = test`, so I thought it was ok.
`make lint` promotes `warnings` to errors, so I'm not sure if we can adjust that.

pallets/stateful-storage/src/benchmarking.rs
mattheworris: I was trying to limit the scope of the lint exceptions.

node/cli/src/export_metadata_cmd.rs
```diff
@@ -38,6 +38,8 @@ pub struct ExportMetadataCmd {
 	pub tmp: bool,
 }
 
+// REVIEW: Should this be refactored or panic is okay here?
+#[allow(clippy::unwrap_used)]
```
JoeCap08055, line 42: I think panic is OK for in this case; this is only invoked from CLI for development purposes.

JoeCap08055 (approved):
Nice! Always good to enable lints that catch actual problems.

pallets/messages/src/lib.rs
```diff
@@ -281,48 +289,50 @@ pub mod pallet {
...
-
-			if Self::add_message(
-				provider_msa_id,
-				Some(maybe_delegator.into()),
-				bounded_payload,
-				schema_id,
-				current_block,
-			)? {
-				Self::deposit_event(Event::MessagesStored {
+			if let Some(schema) = T::SchemaProvider::get_schema_by_id(schema_id) {
```
johnmela, line 292: Great use of `if let`

johnmela (approved):
Assuming tests pass, I think it looks ok

rlaferla (approved):

Issue #1458, opened 2023-04-27
Running Integration tests on Rococo 


p5150j:
As a team we have Integration tests on a "local node" but another tangible value is to run the Integration tests on Rococo testnet to get a real feel for how they will run on a real network. 

In the past, we build the Integration tests with this in mind but haven't kept up with it so we might have some drift and they might be less agnostic to the network type (hard coded local note, etc)

### The goal of this issue is:

- review the Integration tests and make a list of anything that wouldn't work out of the box on Rococo
- add the files that need to be refactored to be agnostic here (or in sub-task issues)
- get the integration tests to run against Rococo network

Ideas to look into:
https://docs.substrate.io/reference/command-line-tools/try-runtime/

johnmela:
* Possible ideas:
        * Export the chain state from Rococo to create a snapshot in time of the chain; build a local instant seal node to run integration tests against the exported chain state
        * Run the tests directly against the running chain
        * Perhaps there is a valid use case for each? 
            * The first option does not require a testnet account with token that may need continual replenishment.  This could be run by engineers during the development process and/or on CI for PR checks
            * The second option is a more realistic test as it will be running against a live chain that is actively producing blocks. This may be more likely to catch race conditions that we wouldn’t normally catch in a sterile local environment that seals after each block. Tests may need modifications for running in this state

johnmela:
I've been working with this repository:
https://github.com/maxsam4/fork-off-substrate

TLDR; I Suspect it's out of date and will not work without some modification.

The long end of it:
It is a little out of date but there is a more recent branch I have been working out of:
https://github.com/maxsam4/fork-off-substrate/tree/dependabot/npm_and_yarn/polkadot/api-10.0.1
The project is supposed to fork the state of a running chain and bootstrap a new chain to run with said state.

It fails on this RPC call:
https://github.com/maxsam4/fork-off-substrate/blob/dependabot/npm_and_yarn/polkadot/api-10.0.1/index.js#L163
with this error:
RpcError: -32601: RPC call is unsafe to be called externally
An error that I can reproduce when calling the RPC against rococo with polkadot.js UI

I would like to ask Parity in the call this week to determine what a modern method of forking a running chain is.

It may be related to https://docs.substrate.io/reference/command-line-tools/node-template/

the `export-state` command provided in https://docs.substrate.io/reference/command-line-tools/node-template/
may help

johnmela:
[...]

PR #1481, opened 2023-05-05
Integration tests on Rococo

johnmela:
# Goal
The goal of this PR is to add the ability for MSA creation integration tests to run against Frequency on Rococo

Related to #1458 

Adds `integration-test-rococo` target to the `Makefile`

# Discussion
This was first tested against a local Frequency build connected to a local relay.
The `createMsa` suite of mocha tests runs successful against the local relay, however there are a few failing against Frequency on Rococo. More testing is needed to determine if this is an environmental difference or related to something else.

### DIFF
--- a/integration-tests/.relay-chain.mocharc.json
+++ b/integration-tests/.relay-chain.mocharc.json
@@ -1,0 +1,16 @@
+{
+    "exit": true,
+    "extension": [
+        "ts"
+    ],
+    "parallel": false,
+    "require": [
+        "scaffolding/rootHooks.ts",
+        "scaffolding/extrinsicHelpers.ts"
+    ],
+    "loader": "ts-node/esm",
+    "spec": [
+        "./{,!(node_modules|load-tests)/**}/createMsa.test.ts"
+    ],
+    "timeout": 200000
+}
--- a/integration-tests/scaffolding/extrinsicHelpers.ts
+++ b/integration-tests/scaffolding/extrinsicHelpers.ts
@@ -5,7 +5,7 @@ import { Compact, u128, u16, u32, u64, Vec, Option, Bytes } from "@polkadot/type
 import { FrameSystemAccountInfo, SpRuntimeDispatchError } from "@polkadot/types/lookup";
 import { AnyNumber, AnyTuple, Codec, IEvent, ISubmittableResult } from "@polkadot/types/types";
 import {firstValueFrom, filter, map, pipe, tap} from "rxjs";
-import {devAccounts, getBlockNumber, log, Sr25519Signature} from "./helpers";
+import {devAccounts, getBlockNumber, log, rococoAccounts, Sr25519Signature} from "./helpers";
 import { connect, connectPromise } from "./apiConnection";
 import { CreatedBlock, DispatchError, Event, SignedBlock } from "@polkadot/types/interfaces";
 import { IsEvent } from "@polkadot/types/metadata/decorate/types";
@@ -139,8 +139,15 @@ export class Extrinsic<T extends ISubmittableResult = ISubmittableResult, C exte
     }
 
     public async fundOperation(source?: KeyringPair, nonce?: number): Promise<void> {
+        let default_source;
+        if (process.env.CHAIN_ENVIRONMENT == "rococo") {
+            default_source = rococoAccounts[0];
+        } else {
+            default_source = devAccounts[0];
+        }
+
         const amount = await this.getEstimatedTxFee();
-        await ExtrinsicHelper.transferFunds(source || devAccounts[0].keys, this.keys, amount).signAndSend(nonce);
+        await ExtrinsicHelper.transferFunds(source || default_source.keys, this.keys, amount).signAndSend(nonce);
     }
 
     public async fundAndSend(source?: KeyringPair, nonce?: number): Promise<ParsedEventResult> {
--- a/integration-tests/scaffolding/extrinsicHelpers.ts
+++ b/integration-tests/scaffolding/extrinsicHelpers.ts
@@ -5,7 +5,7 @@ import { Compact, u128, u16, u32, u64, Vec, Option, Bytes } from "@polkadot/type
 import { FrameSystemAccountInfo, SpRuntimeDispatchError } from "@polkadot/types/lookup";
 import { AnyNumber, AnyTuple, Codec, IEvent, ISubmittableResult } from "@polkadot/types/types";
 import {firstValueFrom, filter, map, pipe, tap} from "rxjs";
-import {devAccounts, getBlockNumber, log, rococoAccounts, Sr25519Signature} from "./helpers";
+import { devAccounts, getBlockNumber, log, getDefaultFundingSource, Sr25519Signature} from "./helpers";
 import { connect, connectPromise } from "./apiConnection";
 import { CreatedBlock, DispatchError, Event, SignedBlock } from "@polkadot/types/interfaces";
 import { IsEvent } from "@polkadot/types/metadata/decorate/types";
@@ -139,15 +139,10 @@ export class Extrinsic<T extends ISubmittableResult = ISubmittableResult, C exte
     }
 
     public async fundOperation(source?: KeyringPair, nonce?: number): Promise<void> {
-        let default_source;
-        if (process.env.CHAIN_ENVIRONMENT == "rococo") {
-            default_source = rococoAccounts[0];
-        } else {
-            default_source = devAccounts[0];
-        }
+        let default_funding_source = getDefaultFundingSource();
 
         const amount = await this.getEstimatedTxFee();
-        await ExtrinsicHelper.transferFunds(source || default_source.keys, this.keys, amount).signAndSend(nonce);
+        await ExtrinsicHelper.transferFunds(source || default_funding_source.keys, this.keys, amount).signAndSend(nonce);
     }
 
     public async fundAndSend(source?: KeyringPair, nonce?: number): Promise<ParsedEventResult> {
--- a/integration-tests/scaffolding/helpers.ts
+++ b/integration-tests/scaffolding/helpers.ts
@@ -53,7 +53,7 @@ export async function getBlockNumber(): Promise<number> {
   return (await ExtrinsicHelper.getLastBlock()).block.header.number.toNumber()
 }
 
-export async function generateAddKeyPayload(payloadInputs: AddKeyData, expirationOffset: number = 5, blockNumber?: number): Promise<AddKeyData> {
+export async function generateAddKeyPayload(payloadInputs: AddKeyData, expirationOffset: number = 100, blockNumber?: number): Promise<AddKeyData> {
   let { expiration, ...payload } = payloadInputs;
   if (!expiration) {
     expiration = (blockNumber || (await getBlockNumber())) + expirationOffset;
--- a/integration-tests/scaffolding/helpers.ts
+++ b/integration-tests/scaffolding/helpers.ts
@@ -111,21 +111,20 @@ export function createKeys(name: string = 'first pair'): KeyringPair {
   return keypair;
 }
 
+export function getDefaultFundingSource() {
+  return process.env.CHAIN_ENVIRONMENT === "rococo" ? rococoAccounts[0] : devAccounts[0];
+}
+
 export async function fundKeypair(source: KeyringPair, dest: KeyringPair, amount: bigint, nonce?: number): Promise<void> {
   await ExtrinsicHelper.transferFunds(source, dest, amount).signAndSend(nonce);
 }
 
 export async function createAndFundKeypair(amount = EXISTENTIAL_DEPOSIT, keyName?: string, source?: KeyringPair, nonce?: number): Promise<KeyringPair> {
-  let default_source;
-  if (process.env.CHAIN_ENVIRONMENT == "rococo") {
-    default_source = rococoAccounts[0];
-  } else {
-    default_source = devAccounts[0];
-  }
+  const default_funding_source = getDefaultFundingSource();
   const keypair = createKeys(keyName);
 
   // Transfer funds from source (usually pre-funded dev account) to new account
-  await fundKeypair((source || default_source.keys), keypair, amount, nonce);
+  await fundKeypair((source || default_funding_source.keys), keypair, amount, nonce);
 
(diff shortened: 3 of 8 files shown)


integration-tests/scaffolding/rootHooks.ts
```diff
@@ -1,19 +1,29 @@
...
     async beforeAll() {
         await ExtrinsicHelper.initialize();
-        for (const uri of ["//Alice", "//Bob", "//Charlie", "//Dave", "//Eve", "//Ferdie"]) {
-            devAccounts.push({
-                uri,
-                keys: createKeys(uri),
-            });
+        let default_source;
+        if (process.env.CHAIN_ENVIRONMENT == "rococo") {
+            const seed_phrase = "Place holder for seed phrase";
```
johnmela, line 12: So as to not expose seed phrases holding token, we may want to use AWS Secrets Manager to store the seed phrase for the funding source account.

johnmela:
For reference
```Create Accounts
    createMsa
      ✔ should successfully create an MSA account (47933ms)
      ✔ should fail to create an MSA for a keypair already associated with an MSA (48147ms)
    addPublicKeyToMsa
      ✔ should fail to add public key if origin is not one of the signers of the payload (MsaOwnershipInvalidSignature) (47922ms)
      ✔ should fail to add public key if new keypair is not one of the signers of the payload (NewKeyOwnershipInvalidSignature) (47907ms)
      1) should fail to add public key if origin does not have an MSA (NoKeyExists)
      ✔ should fail to add public key if origin does not own MSA (NotMsaOwner) (48127ms)
      ✔ should fail if expiration has passed (ProofHasExpired) (47816ms)
      ✔ should fail if expiration is not yet valid (ProofNotYetValid) (47880ms)
      2) should successfully add a new public key to an existing MSA & disallow duplicate signed payload submission (SignatureAlreadySubmitted)
      ✔ should fail if attempting to add the same key more than once (KeyAlreadyRegistered) (47857ms)
      ✔ should allow new keypair to act for/on MSA (47889ms)
    retire/delete keys
2023-05-05 18:24:54        RPC-CORE: submitAndWatchExtrinsic(extrinsic: Extrinsic): ExtrinsicStatus:: 1010: Invalid Transaction: Custom error: 2
2023-05-05 18:24:54        RPC-CORE: submitAndWatchExtrinsic(extrinsic: Extrinsic): ExtrinsicStatus:: 1010: Invalid Transaction: Custom error: 2
      ✔ should disallow retiring MSA belonging to a provider (24237ms)
2023-05-05 18:25:18        RPC-CORE: submitAndWatchExtrinsic(extrinsic: Extrinsic): ExtrinsicStatus:: 1010: Invalid Transaction: Custom error: 3
2023-05-05 18:25:18        RPC-CORE: submitAndWatchExtrinsic(extrinsic: Extrinsic): ExtrinsicStatus:: 1010: Invalid Transaction: Custom error: 3
      ✔ should disallow retiring an MSA with more than one key authorized (24125ms)
2023-05-05 18:25:42        RPC-CORE: submitAndWatchExtrinsic(extrinsic: Extrinsic): ExtrinsicStatus:: 1010: Invalid Transaction: Custom error: 4
2023-05-05 18:25:42        RPC-CORE: submitAndWatchExtrinsic(extrinsic: Extrinsic): ExtrinsicStatus:: 1010: Invalid Transaction: Custom error: 4
      ✔ should fail to delete public key for self (23968ms)
2023-05-05 18:26:06        RPC-CORE: submitAndWatchExtrinsic(extrinsic: Extrinsic): ExtrinsicStatus:: 1010: Invalid Transaction: Custom error: 5
2023-05-05 18:26:06        RPC-CORE: submitAndWatchExtrinsic(extrinsic: Extrinsic): ExtrinsicStatus:: 1010: Invalid Transaction: Custom error: 5
      ✔ should fail to delete key if not authorized for key's MSA (23917ms)
      - should test for 'NoKeyExists' error
      ✔ should delete all other authorized keys (95836ms)
      ✔ should allow retiring MSA after additional keys have been deleted (47936ms)


  15 passing (17m)
  1 pending
  2 failing

  1) Create Accounts
       addPublicKeyToMsa
         should fail to add public key if origin does not have an MSA (NoKeyExists):
     AssertionError [ERR_ASSERTION]: Expected values to be strictly deep-equal:
+ actual - expected

  Comparison {
+   name: 'ProofHasExpired'
-   name: 'NoKeyExists'
  }
      at processTicksAndRejections (node:internal/process/task_queues:95:5)
      at async Context.<anonymous> (msa/createMsa.test.ts:88:13)

  2) Create Accounts
       addPublicKeyToMsa
         should successfully add a new public key to an existing MSA & disallow duplicate signed payload submission (SignatureAlreadySubmitted):
     AssertionError [ERR_ASSERTION]: should reject sending the same signed payload twiced
      at processTicksAndRejections (node:internal/process/task_queues:95:5)
      at async Context.<anonymous> (msa/createMsa.test.ts:146:13)
```

scripts/run_integration_tests.sh
```diff
@@ -30,29 +30,53 @@ function cleanup () {
...
     case "${OPTNAME}" in
-        "s") SKIP_JS_BUILD=1
+        "s")
+            SKIP_JS_BUILD=1
+        ;;
+        "r")
+            TEST_AGAINST_FREQUENCY_ROCOCO=1
+        ;;
+        "l")
+            TEST_AGAINST_LOCAL_RELAY=1
```
johnmela, line 50: For completeness the shell script should return an error if both of these cli options are passed as they are mutually exclusive

integration-tests/.relay-chain.mocharc.json
```diff
@@ -0,0 +1,16 @@
...
+    "parallel": false,
+    "require": [
+        "scaffolding/rootHooks.ts",
+        "scaffolding/extrinsicHelpers.ts"
+    ],
+    "loader": "ts-node/esm",
+    "spec": [
+        "./{,!(node_modules|load-tests)/**}/createMsa.test.ts"
+    ],
+    "timeout": 200000
```
johnmela, line 15: The timeout is increased so that `signAndSend()` has sufficient time to wait on the callback from the chain while the relay validates / seals each block

rlaferla: Is that 3.3 minutes?  Is that too long?

johnmela: It may not be long enough for some tests in the replenishment suite to complete

rlaferla: Interesting.  Why does validation/sealing take so long?  I thought blocks could be produced every 12 seconds or so?

johnmela: This likely will be able to be reduced for Frequency on Rococo. Testing with a local relay overshoots that target often. We can fine tune this

johnmela: Do you see a downside to a longer time out when testing with a Relay?

rlaferla (approved):
g👀d2go!

integration-tests/.relay-chain.mocharc.json
rlaferla: I'm just trying to understand why there is slow performance.  The timeout needs to be set relative to the expected performance.

johnmela: I had to think about it for a minute, here is why: 
Some tests have multiple transaction per test. I am also wondering if the transaction may also not be guaranteed to make it into the current block.
```
  Create Accounts
    createMsa
      ✔ should successfully create an MSA account (47902ms)
```

scripts/run_integration_tests.sh
JoeCap08055: Seems like a custom enum implementation would be overkill & non-standard. I lean toward the KISS principle 😄 -- stick with 0/1 and an appropriately named variable.

JoeCap08055: (However, on further thought, a single CLI option `-c` (`c` for "chain") that takes an argument of "local" or "rococo" might be simpler, and eliminate the need for the mutually-exclusive option check. Since there only two strings to check for, a hand-rolled enum implementation seems like overkill (though it's a pretty cool pattern), but the overall principal is similar.)

scripts/run_integration_tests.sh
```diff
@@ -68,15 +92,15 @@ else
...
+    case ${BLOCK_SEALING} in
+        "instant") ${RUNDIR}/init.sh start-frequency-instant >& frequency.log &
         ;;
-        "start-manual") ${RUNDIR}/init.sh start-frequency-manual >& frequency.log &
+        "manual") ${RUNDIR}/init.sh start-frequency-manual >& frequency.log &
         ;;
     esac
 
-    declare -i timeout_secs=30
+    declare -i timeout_secs=60
```
mattheworris, line 103: Nice, I was going to suggest increasing this timeout, I keep hitting it.

mattheworris (approved):
Looks good!

integration-tests/scaffolding/rootHooks.ts
wilwade: Could this just be an environment variable?

integration-tests/scaffolding/helpers.ts
```diff
@@ -114,11 +115,17 @@ export async function fundKeypair(source: KeyringPair, dest: KeyringPair, amount
...
 }
 
-export async function createAndFundKeypair(amount = EXISTENTIAL_DEPOSIT, keyName?: string, devAccount?: KeyringPair, nonce?: number): Promise<KeyringPair> {
+export async function createAndFundKeypair(amount = EXISTENTIAL_DEPOSIT, keyName?: string, source?: KeyringPair, nonce?: number): Promise<KeyringPair> {
+  let default_source;
+  if (process.env.CHAIN_ENVIRONMENT == "rococo") {
+    default_source = rococoAccounts[0];
+  } else {
+    default_source = devAccounts[0];
+  }
```
wilwade, line 124: Might be worth hiding this if statement inside a function. Feels like it shouldn't be in more than one place.

integration-tests/scaffolding/rootHooks.ts
```diff
@@ -1,19 +1,29 @@
...
+            rococoAccounts.push({
```
wilwade, line 13: Is there a reason this isbn't just pushing into `devAccounts`?

scripts/run_integration_tests.sh
johnmela: I like this idea

integration-tests/scaffolding/rootHooks.ts
johnmela: The reason is for clarity; to differentiate between the common accounts available in a development environment vs the accounts available for testnet.

integration-tests/scaffolding/helpers.ts
johnmela: Agreed. We should probably retrieve the appropriate funding source from a function.

integration-tests/scaffolding/rootHooks.ts
johnmela: It can be if we would prefer to simplify the process. My thinking was that multiple environments could be added (local / testnet / mainnet) for storing funding accounts as well as any future distinct aspects to each env.

If we use an environment variable we would still want to store the seed phrase values secretly. Possibly github actions has a solution?

integration-tests/scaffolding/helpers.ts
```diff
@@ -114,11 +115,17 @@ export async function fundKeypair(source: KeyringPair, dest: KeyringPair, amount
   await ExtrinsicHelper.transferFunds(source, dest, amount).signAndSend(nonce);
 }
 
-export async function createAndFundKeypair(amount = EXISTENTIAL_DEPOSIT, keyName?: string, devAccount?: KeyringPair, nonce?: number): Promise<KeyringPair> {
+export async function createAndFundKeypair(amount = EXISTENTIAL_DEPOSIT, keyName?: string, source?: KeyringPair, nonce?: number): Promise<KeyringPair> {
+  let default_source;
+  if (process.env.CHAIN_ENVIRONMENT == "rococo") {
```
shannonwells, line 120:
```suggestion
  const default_source = process.env.CHAIN_ENVIRONMENT === "rococo" ?
      rococoAccounts[0] : devAccounts[0];
```

shannonwells (approved):
Nothing blocking, mostly code golf although the use of `===` is good form.

integration-tests/scaffolding/extrinsicHelpers.ts
```diff
@@ -139,8 +139,15 @@ export class Extrinsic<T extends ISubmittableResult = ISubmittableResult, C exte
     }
 
     public async fundOperation(source?: KeyringPair, nonce?: number): Promise<void> {
+        let default_source;
+        if (process.env.CHAIN_ENVIRONMENT == "rococo") {
```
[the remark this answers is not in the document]
johnmela: Agreed thanks, I think this logic could go into a function in helpers.ts, then we invoke the function where we need the default funding source

integration-tests/scaffolding/rootHooks.ts
JoeCap08055: I believe most secrets engines (certainly AWS; I think GH Actions, too) can inject the secret as an environment variable. While I'm not generally in favor of this (there's still a risk of exposure, if someone gains access to the pod while it's running), it seems to be the way we're currently doing things. The alternative is to have the code retrieve secrets via the secret engine's API at runtime, but that seems like overkill for e2e tests.

johnmela: A risk for exposure may also be if the env variable is leaked into actions output when injecting the variable. We would need to make sure that doesn't happen. My initial thought was to have the secret pulled at runtime, although I could be wrong about the leaking of environment variable to output. 

Preventing a leak becomes more important if we shift to using a funding account that creates token rather than simply holds and transfers it.

I think these are great options to investigate. I think this should probably be pulled into into a separate issue that creates a manual workflow to kick off e2e tests against Rococo and also investigates injecting the secret as an environment variable vs. pulling from a secret engine at runtime? @JoeCap08055

johnmela: I just checked, it does NOT leak the secrets into the output. The environment variable injected secrets would probably suffice

PR #1492, opened 2023-05-09
Feat/1463 vscode debug config

### LINKED ISSUE #1463: off-chain and on-chain runtime chain specifications don’t match
Basically, the off-chain and on-chain runtime spec names don’t match.  This is preventing native code execution and prevents developers from using the debugger for palette and runtime code.

@rlaferla and @mattheworris in your session please add any details to this that aren't captured above.

mattheworris:
# Goal
The goal of this PR is add VSCode debugger configuration in `.vscode/launch.json`.

Closes #1463 

# Discussion
- Updated `.gitignore` so that `.vscode/launch.json` can now be tracked.
- Updated `init.sh` for target `start-frequency-instant` to modify `--execution=native`, when attaching to a running process, this allows the debugger to stop at breakpoints in the native runtime.

# Checklist -- Not applicable
- [ ] Chain spec updated
- [ ] Design doc(s) updated
- [ ] Tests added
- [ ] Benchmarks added
- [ ] Weights updated


rlaferla (approved):

saraswatpuneet (approved):
💯🥇

scripts/init.sh
```diff
@@ -86,7 +86,7 @@ start-frequency-instant)
     -lruntime=debug \
     --instant-sealing \
     --wasm-execution=compiled \
-    --execution=wasm \
+    --execution=native \
```
aramikm, line 89: Since most of our local testing is done using `start-frequency-instant` I think changing it to always being `native` might be a bit concerning since there might be some problems that only might happen when you ran it as wasm. While there is a low priority this might happen but being able to test something with the exact or close enough configuration as main-net is always valuable.

rlaferla: Unfortunately, the native setting is what enables the use of a debugger. That is a much more important  aspect.  Developers can always switch it to wasm if they need to do additional testing.

aramikm: If the vscode config is already set to use `native` I don't see why we need to set it to `native` in the script. Or are we running the debugger outside of vscode?

rlaferla:
Good point, Aramik.  If we can have native inside VSCode and wasm outside, it’s a win-win!

scripts/init.sh
mattheworris: I changed it to native to make it easier to attach to a running node. It's a command line argument and can be easily changed without needing to compile. 

We can change the makefile back to WASM. Developers needing to attach a debugger can easily modify the arguments.

PR #1518, opened 2023-05-11
update built binaries in CI

### LINKED ISSUE #1512: Releases should include no-relay but not rococo-local binary
Releases should include no-relay but not rococo-local binary.  However, both need to be built and the rococo-local used when comparing the metadata in the workflow.

demisx:
# Goal
The goal of this PR is to update built binaries in CI and where they are being used. See issue description for more detail.

These are the binaries that will be posted in each GitHub release after this PR is merged:

Screenshot 2023-05-12 at 1 03 27 PM

Closes #1512

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -164,11 +164,11 @@ jobs:
             release-file-name-prefix: frequency-rococo-local
           - network: rococo
             spec: frequency-rococo-testnet
-            build-profile: production
+            build-profile: release
             release-file-name-prefix: frequency-rococo
           - network: mainnet
             spec: frequency
-            build-profile: production
+            build-profile: release
             release-file-name-prefix: frequency
           - os: [self-hosted, Linux, X64, build]
             arch: amd64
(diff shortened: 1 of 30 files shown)


.github/workflows/verify-pr-commit.yml
```diff
@@ -93,20 +74,20 @@ jobs:
     strategy:
       fail-fast: true
       matrix:
-        network: [rococo, mainnet]
+        # network: [dev, local, rococo, mainnet]
```
wilwade, line 77: @demisx Is this supposed to be commented out?

.github/workflows/release.yml
```diff
@@ -865,13 +858,19 @@ jobs:
     strategy:
       fail-fast: true
       matrix:
-        network: [local]
         arch: [amd64]
         node: [collator-node-local, instant-seal-node]
         include:
-          - network: local
+          - node: collator-node-local
```
wilwade, line 864: I'd be OK with not publishing this one at all now that I look at it. Fine to do so, but also fine to have less in the release.

wilwade (commented):
One comment to check and an option to remove a release bin

.github/workflows/release.yml
wilwade: Be nice if we could remove it from the build entirely if we don't use it elsewhere

saraswatpuneet (approved):

PR #1520, opened 2023-05-11
New CLI parameter interface for block sealing

### LINKED ISSUE #1517: Change CLI parameters for sealing in preparation for new interval sealing mode
--sealing <SEALING>
          The sealing mode
          
          [default: instant]

          Possible values:
          - instant:
            Author a block immediately upon receiving a transaction into the transaction pool
          - manual:
            Author a block upon receiving an RPC command
          - interval:
            Author blocks at a regular interval specified in seconds

      --sealing-interval <SECONDS>
          The interval in seconds
          
          [default: 120]

      --sealing-allow-empty-blocks
          Allows empty blocks to be created when sealing in manual and interval modes

rlaferla:
# Goal
The goal of this PR is to provide more robust CLI parameters for sealing modes.

Closes #1517

### DIFF
--- a/node/cli/Cargo.toml
+++ b/node/cli/Cargo.toml
@@ -11,7 +11,6 @@ version = "0.0.0"
 [dependencies]
 clap = { version = "4.0.17", features = ["derive"] }
 codec = { package = "parity-scale-codec", version = "3.1.5" }
-chrono = { version = "0.4.24" }
 derive_more = "0.99.17"
 futures = { version = "0.3.21", features = ["thread-pool"] }
 hex-literal = "0.3.4"
--- a/node/cli/Cargo.toml
+++ b/node/cli/Cargo.toml
@@ -11,6 +11,7 @@ version = "0.0.0"
 [dependencies]
 clap = { version = "4.0.17", features = ["derive"] }
 codec = { package = "parity-scale-codec", version = "3.1.5" }
+chrono = { version = "0.4.24" }
 derive_more = "0.99.17"
 futures = { version = "0.3.21", features = ["thread-pool"] }
 hex-literal = "0.3.4"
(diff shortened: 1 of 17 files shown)


node/cli/src/cli.rs
```diff
@@ -79,19 +84,23 @@ pub struct Cli {
 	/// Instant block sealing
 	/// Blocks are triggered to be formed each time a transaction hits the validated transaction pool
 	/// Empty blocks can also be formed using the `engine_createBlock` RPC
-	///
-	/// Can only be used with frequency-no-relay and frequency-rococo-local feature flags
-	#[cfg(any(feature = "frequency-no-relay", feature = "frequency-rococo-local"))]
-	#[clap(long = "instant-sealing")]
-	pub instant_sealing: bool,
```
wilwade, line 86: Is there a way with clap to leave a message that the option has changed?

Aka if someone after this runs ./frequency --instant-sealing it wouldn't run, but would report that it has changed to be --sealing=instant ?

node/cli/src/run_as_localchain.rs
```diff
@@ -1,15 +1,23 @@
...
+	// Temporary error message
+	if cli.sealing == SealingMode::Interval {
+		return Result::<(), _>::Err(sc_cli::Error::Input(String::from(
+			"Interval sealing is not yet implemented.",
+		)))
+	}
+
+	let current_time = Local::now();
+	let formatted_time = current_time.format("%Y-%m-%d %H:%M:%S").to_string();
+	println!("{} NO-RELAY MODE with {} sealing", formatted_time, cli.sealing);
```
wilwade, line 16: Don't we get the time part for free when we use the `log::info!` logging?

wilwade (commented):
One question and I think a log to switch to the info macro.

node/cli/src/cli.rs
rlaferla: I will check to see if it has support for deprecated parameters.  If not, I could trap them and display a warning message.

node/cli/src/run_as_localchain.rs
rlaferla: Normally but info level logs are being filtered so it doesn't display.  Hence, the workaround.  If I could only modify the runner to append additional info.

node/cli/src/cli.rs
rlaferla: Fixed.

node/cli/Cargo.toml
```diff
@@ -11,6 +11,7 @@ version = "0.0.0"
 [dependencies]
 clap = { version = "4.0.17", features = ["derive"] }
 codec = { package = "parity-scale-codec", version = "3.1.5" }
+chrono = { version = "0.4.24" }
```
wilwade, line 14: I think you can drop this dep now.

wilwade (approved):
🎉

mattheworris (changes requested):
Looks good! One update to `--instant-sealing` missed in `init.sh` for `start-frequency-native` when my PR was merged.

mattheworris (approved):

Issue #1482, opened 2023-05-05
Add Rust toolchain install to CI base image


demisx:
Install Rust along with Toolchain in the base CI image and remove those steps from CI, correct

**Tasks**:
- [x] Add rust install to CI base image
- [x] Disable rust install toolchain in CI jobs where possible
- [x] Delete `1482-delete-me.rs` file

p5150j:
ah.. @demisx .. if there were rust dev needs I set it as hi-prior so it wasn't blocking anything. Apologies.

PR #1536, opened 2023-05-18
add rust installation to ci base image

demisx:
# Goal
The goal of this PR is to move rust installation to the CI base image. It also fixes some errors observed after runners have been upgraded to ubuntu `22.04`.

Related #1482

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -175,7 +175,7 @@ jobs:
           # - os: [self-hosted, Linux, ARM64]
           #   arch: arm64
     runs-on: ${{matrix.os}}
-    container: ubuntu:20.04
+    container: ghcr.io/libertydsnp/frequency/ci-base-image
     steps:
       - name: Install Required Packages
         run: |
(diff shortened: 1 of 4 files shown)


.github/workflows/release.yml
```diff
@@ -306,7 +310,7 @@ jobs:
           echo "runtime_filename_${{matrix.network}}=$release_wasm_filename" >> $GITHUB_OUTPUT
       - name: Install srtool-cli
         run: |
-          cargo install --git https://github.com/chevdor/srtool-cli
+          rustup run stable cargo install --git https://github.com/chevdor/srtool-cli
```
mattheworris, line 313: If the self-hosted runner is ubuntu 22.04 and has glibc ^2.3, then we can use a pre-built binary for srtool-cli/srtool, and not need to install the rust toolchain.

wilwade: @mattheworris I think this is just a temp step. We are trying to get docker-in-docker working so we can remove the need to have anything special on the self-hosted runner.

wilwade: That said, yes, it is fine to have both installed on the CI base image and fine to build srtool-cli with stable.

wilwade (approved):

mattheworris (approved):
💯

PR #1538, opened 2023-05-19
Capacity Integration Tests on Rococo

### LINKED ISSUE #1502: Integration tests on Rococo capacity/
Integration tests on Rococo capacity/

https://github.com/LibertyDSNP/frequency/tree/main/integration-tests/capacity

These should run both rococo and local. 
If we think we should skip any of these it will be up to the developers to make the judgment call on how valuable it would be to run on Rococo.

johnmela:
# Goal
The goal of this PR is to allow Message related integration tests to execute successfully both in instant sealing and Frequency on Rococo environments.

Closes #1502

# Discussion
- Reduced the amount of token transfer from the funding source to test providers
- There *could* be some future work in reducing all funding source transfers to ensure that the resulting account balance of testing account drops below the existential balance 
  - This would result in the account artifacts being reaped rather than existing into eternity on the test chain
  - There may, however, be an argument for keeping all test artifacts created by the integration tests; this may result in potential problems or bottlenecks surfacing sooner on testnet so that they may be addressed before a scale up on mainnet


integration-tests/capacity/staking.test.ts
```diff
@@ -13,12 +13,17 @@ import {
 import { firstValueFrom } from "rxjs";
 
 describe("Capacity Staking Tests", function () {
-    const accountBalance: bigint = 200n * DOLLARS;
+    const accountBalance: bigint = 2n * DOLLARS;
```
johnmela, line 16: The account was overfunded which isn't necessarily a problem. I reduced some initial account funding values to reduce the amount of token transferred from funding source on Rococo

integration-tests/capacity/transactions.test.ts
```diff
@@ -7,39 +7,34 @@ import { u8aWrapBytes } from "@polkadot/util";
...
     describe("pay_with_capacity", function () {
         describe("when caller has a Capacity account", async function () {
             let capacityKeys: KeyringPair;
             let capacityProvider: u64;
             let schemaId: u16;
-            const amountStaked = 25n * DOLLARS;
+            const amountStaked = 2n * DOLLARS;
 
-            beforeEach(async function () {
+            before(async function () {
```
johnmela, line 32: Creating a new provider and schema for each test was unnecessary

integration-tests/scaffolding/helpers.ts
```diff
@@ -262,15 +265,89 @@ export async function setEpochLength(keys: KeyringPair, epochLength: number): Pr
...
-  if (createSchemaEvent && ExtrinsicHelper.api.events.schemas.SchemaCreated.is(createSchemaEvent)) {
-    return createSchemaEvent.data.schemaId;
+export async function getOrCreateGraphChangeSchema(): Promise<u16> {
+  if (process.env.CHAIN_ENVIRONMENT === "rococo") {
+    const ROCOCO_GRAPH_CHANGE_SCHEMA_ID: u16 = new u16(ExtrinsicHelper.api.registry, 53);
+    return ROCOCO_GRAPH_CHANGE_SCHEMA_ID;
   } else {
-    assert.fail("failed to create a schema")
+    const [createSchemaEvent, eventMap] = await ExtrinsicHelper
+      .createSchema(devAccounts[0].keys, AVRO_GRAPH_CHANGE, "AvroBinary", "OnChain")
```
mattheworris, line 274: Is it useful to use `getDefaultFundingSource()` here?

integration-tests/scaffolding/helpers.ts
```diff
@@ -262,15 +265,89 @@ export async function setEpochLength(keys: KeyringPair, epochLength: number): Pr
...
 }
 
-export async function createGraphChangeSchema(): Promise<u16> {
-  const [createSchemaEvent, eventMap] = await ExtrinsicHelper
-    .createSchema(devAccounts[0].keys, AVRO_GRAPH_CHANGE, "AvroBinary", "OnChain")
-    .fundAndSend();
-  assert.notEqual(eventMap["system.ExtrinsicSuccess"], undefined);
-  if (createSchemaEvent && ExtrinsicHelper.api.events.schemas.SchemaCreated.is(createSchemaEvent)) {
-    return createSchemaEvent.data.schemaId;
+export async function getOrCreateGraphChangeSchema(): Promise<u16> {
```
mattheworris, line 268: I like these `getOr` functions.

mattheworris (approved):
Looking good!

integration-tests/scaffolding/helpers.ts
johnmela: It may be if it is decided to create schemas on testnet. Since the `getOrCreate` schema helper functions are currently only creating schemas when executing against a development node, I left this alone.

johnmela: I made a mistake in that comment:
The else block of the branch executes when `CHAIN_ENVIRONMENT` is anything but "rococo". Technically we should only reference those `devAccounts` when `CHAIN_ENVIRONMENT` is "local", but the only other `CHAIN_ENVIRONMENT` supported by the testing framework is "local" currently. 


There may be a case to add a "mainnet" environment in the future

.github/workflows/verify-pr-commit.yml
```diff
@@ -647,8 +647,10 @@ jobs:
         working-directory: integration-tests
       - name: Run Integration Tests
         working-directory: integration-tests
-        run: |
-          CHAIN_ENVIRONMENT="dev" TEST_CHAIN_VALIDATION="instant_finality" WS_PROVIDER_URL="ws://127.0.0.1:9944" npm test
+        env:
```
shannonwells, line 650: 🙌

integration-tests/capacity/staking.test.ts
```diff
@@ -5,20 +5,25 @@ import assert from "assert";
...
     const tokenMinStake: bigint = 1n * CENTS;
     let capacityMin: bigint = tokenMinStake / 50n;
 
     before(async function () {
-        await setEpochLength(devAccounts[0].keys, TEST_EPOCH_LENGTH);
+        // Pallet config changes such as modifying the epoch length will
+        // only be modified when running tests against a Frequency node built
+        // for development.
+        if (process.env.CHAIN_ENVIRONMENT === CHAIN_ENVIRONMENT.DEVELOPMENT) {
```
shannonwells, line 24: 👍

integration-tests/scaffolding/helpers.ts
```diff
@@ -262,15 +270,89 @@ export async function setEpochLength(keys: KeyringPair, epochLength: number): Pr
...
+  if (process.env.CHAIN_ENVIRONMENT === CHAIN_ENVIRONMENT.ROCOCO_TESTNET) {
+    const ROCOCO_GRAPH_CHANGE_SCHEMA_ID: u16 = new u16(ExtrinsicHelper.api.registry, 53);
+    return ROCOCO_GRAPH_CHANGE_SCHEMA_ID;
+  } else {
+    const [createSchemaEvent, eventMap] = await ExtrinsicHelper
+      .createSchema(devAccounts[0].keys, AVRO_GRAPH_CHANGE, "AvroBinary", "OnChain")
+      .fundAndSend();
+    assert.notEqual(eventMap["system.ExtrinsicSuccess"], undefined);
```
shannonwells, line 281: Non-blocking, because it's not in scope, but if you change this file...I see a lot of this line and it's always confusing to read. I'd love to see a helper function that is "assertExtrinsicSuccess(eventMap)" or something similar.

shannonwells (approved):
Looks good. Thanks for all this work!

.github/workflows/verify-pr-commit.yml
```diff
@@ -647,8 +647,10 @@ jobs:
...
+          CHAIN_ENVIRONMENT=dev
+          WS_PROVIDER_URL=ws://127.0.0.1:9944
```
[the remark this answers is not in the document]
johnmela: Thank you, @demisx

PR #1547, opened 2023-05-23
remove 3rd party rust toolchain action

### LINKED ISSUE #1546: Remove 3rd party install rust toolchain action
Ideally, we should be able to preinstall 1 toolchain in CI base image and then use it in the CI jobs. However, usage of nightly along stable may be required.

demisx:
# Goal
The goal of this PR is to remove 3rd party rust toolchain install action. The couple of jobs that still run on EKS runner directly will install rust in the same consistent manner as it's done in the CI base image. 

Also, this PR removes `stable` toolchain as everything seems to compile and run under nightly version now.

Closes #1546


wilwade (approved):
❤️

mattheworris (approved):
Looks solid. 🚀

PR #1552, opened 2023-05-23
1472 upgrade to polkadot release v0942

mattheworris:
# Goal
The goal of this PR is 

Closes 

# Discussion

# Checklist
- [ ] Chain spec updated
- [ ] Design doc(s) updated
- [ ] Tests added
- [ ] Benchmarks added
- [ ] Weights updated


pallets/messages/src/rpc/src/lib.rs
```diff
@@ -100,7 +99,7 @@ where
 		'loops: for block_number in from..to {
 			let list: Vec<MessageResponse> = api
 				.get_messages_by_schema_and_block(
-					&at,
+					self.client.info().best_hash,
```
wilwade, line 102: I think this is a function call and not memoized, so we'd need to keep the at so it doesn't get called n times.

pallets/msa/src/rpc/src/lib.rs
```diff
@@ -93,15 +93,14 @@ where
...
-		let at = BlockId::hash(self.client.info().best_hash);
 		let results = delegator_msa_ids
 			.par_iter()
 			.map(|delegator_msa_id| {
 				let api = self.client.runtime_api();
 				// api.has_delegation returns  Result<bool, ApiError>), so _or(false) should not happen,
 				// but just in case, protect against panic
 				let has_delegation: bool = match api.has_delegation(
-					&at,
+					self.client.info().best_hash,
```
wilwade, line 103: Another here that is inside a loop

runtime/common/src/weights/pallet_balances.rs
```diff
@@ -34,45 +34,61 @@ use sp_std::marker::PhantomData;
...
 	fn force_unreserve() -> Weight {
-		Weight::from_ref_time(19_985_000 as u64)
+		Weight::from_parts(19_985_000 as u64, 0)
 			.saturating_add(T::DbWeight::get().reads(1 as u64))
 			.saturating_add(T::DbWeight::get().writes(1 as u64))
 	}
+	/// Storage: System Account (r:999 w:999)
+	/// Proof: System Account (max_values: None, max_size: Some(128), added: 2603, mode: MaxEncodedLen)
+	/// The range of component `u` is `[1, 1000]`.
+	fn upgrade_accounts(u: u32, ) -> Weight {
```
wilwade, line 82: Are we going to need to call this once it is deployed to do the migration?

wilwade (commented):
I know it is still a WIP, but a few notes, and you'll want to trigger all the benchmarks once it is ready to go.

pallets/messages/src/rpc/src/lib.rs
mattheworris: Nice catch, I didn't catch the loop part, focused on the BlockId removal.

runtime/common/src/weights/pallet_balances.rs
mattheworris: I think so, there is a release note about needing to run an upgrade script. I grabbed this code from substrate to get things compiling, I'll check the benchmark output to see if this comes out correct.

wilwade:
Replaced with https://github.com/LibertyDSNP/frequency/pull/1590

PR #1577, opened 2023-06-02
Feat/capacity staking rewards impl

shannonwells:
# Goal
The goal of this PR is to:
1. introduce a Design Document for implementing Staking for Rewards in the Capacity pallet (aka "Provider Boosting") 
2. create some of its needed primitives
3. make the first update to the `stake` extrinsic

# Discussion
Because the decision on the economic model for rewards has not been decided, this treats it as a black box by creating a StakingRewardProvider trait so some of the work that will definitely be needed can proceed.

The implementation Design Document is to be considered a living document, so it is expected to change.

# Checklist
- [ ] Chain spec updated
- [x] Design docs created
- [x] Tests added
- [ ] Weights updated


shannonwells:
Something is weird about the diffs

PR #1591, opened 2023-06-12
Improved documentation for add_public_key_to_msa()

rlaferla:
# Goal

To improve the API level documentation for the add_public_key_to_msa() extrinsic in the MSA pallet.


pallets/msa/src/lib.rs
```diff
@@ -610,12 +610,20 @@ pub mod pallet {
...
 
-		/// Adds a given `new_key` to `msa_id` of the account signing `msa_owner_proof`, which must match the MSA in `add_key_payload`.
-		/// The `new_key` must sign the `add_key_payload` to authorize the addition.
+		/// Adds a new public key to an existing Message Source Account (MSA). This functionality enables the MSA owner to manage multiple keys
+		/// for their account or rotate keys for enhanced security.
 		///
-		/// # Remarks
-		/// * Origin can be same as msa owner.
-		/// * Signatures should be over the [`AddKeyData`] struct
+		/// The `origin` parameter represents the account from which the function is called and can be either the MSA owner's account or a provider's account,
```
JoeCap08055, line 616:
```suggestion
		/// The `origin` parameter represents the account from which the function is called and can be either the MSA owner's account or a delegated provider's account,
```

pallets/msa/src/lib.rs
```diff
@@ -610,12 +610,20 @@ pub mod pallet {
...
+		///  depending on the intended use.
+		///
+		/// The function requires two signatures: `msa_owner_proof` and `new_key_owner_proof`, which serve as proofs of ownership for the existing MSA
```
mattheworris, line 619: Should we describe the data that needs to be signed to create the proofs? 
Also important to know about the proof expiration, is that something useful to mention here?
Maybe in a code example at the end:
```rust
	// pulled this from the unit tests, maybe it can be simplified, but useful here for an example
    let add_new_key_data = AddKeyData {
			msa_id: new_msa_id,
			expiration: 10,
			new_public_key: new_key_pair.public().into(),
		};
    let encode_data_new_key_data = wrap_binary_data(add_new_key_data.encode());

	let owner_signature: MultiSignature = owner_key_pair.sign(&encode_data_new_key_data).into();
	let new_key_signature: MultiSignature = new_key_pair.sign(&encode_data_new_key_data).into();

	// act
	Msa::add_public_key_to_msa(
		test_origin_signed(1),
		owner_key_pair.public().into(),
		owner_signature,
		new_key_signature,
		add_new_key_data
	);
```

mattheworris (commented):
Am I asking for too much context that possibly should be documented somewhere else? 🤔 
Possibly.

pallets/msa/src/lib.rs
```diff
@@ -610,12 +610,20 @@ pub mod pallet {
 			Ok(())
 		}
 
-		/// Adds a given `new_key` to `msa_id` of the account signing `msa_owner_proof`, which must match the MSA in `add_key_payload`.
-		/// The `new_key` must sign the `add_key_payload` to authorize the addition.
+		/// Adds a new public key to an existing Message Source Account (MSA). This functionality enables the MSA owner to manage multiple keys
+		/// for their account or rotate keys for enhanced security.
```
johnmela, line 614: 💯

pallets/msa/src/lib.rs
rlaferla: I think examples should be separate documentation.  I will be discussing this with Jeanette later.

mattheworris (approved):
💯

pallets/msa/src/lib.rs
rlaferla: Where are the docs for this?  I don't recall ever seeing <Binary></Binary>.  Did you mean <Bytes></Bytes>?

pallets/msa/src/lib.rs
```diff
@@ -610,12 +610,20 @@ pub mod pallet {
...
+		///  depending on the intended use.
+		///
+		/// The function requires two signatures: `msa_owner_proof` and `new_key_owner_proof`, which serve as proofs of ownership for the existing MSA
+		/// and the new public key account, respectively.
+		///
+		/// The necessary information for the key addition, such as the new public key and the MSA ID, is provided within the `add_key_payload` parameter.
+		/// Additionally, the `add_key_payload` includes an expiration block number for both proofs, ensuring they are valid and must be greater than
+		/// the current block.
+		///
+		/// By using this function, MSA owners can conveniently add new keys to their accounts, facilitating multi-device usage or key rotation for improved security.
```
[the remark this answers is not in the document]
rlaferla: The wrapping/signing of the payload shouldn't be part of the docs for the function.  Instead, it should be elsewhere.

JoeCap08055 (approved):

pallets/msa/src/lib.rs
```diff
@@ -610,12 +610,18 @@ pub mod pallet {
...
+		/// Adds a new public key to an existing Message Source Account (MSA). This functionality enables the MSA owner to manage multiple keys
+		/// for their account or rotate keys for enhanced security.
 		///
-		/// # Remarks
-		/// * Origin can be same as msa owner.
-		/// * Signatures should be over the [`AddKeyData`] struct
+		/// The `origin` parameter represents the account from which the function is called and can be either the MSA owner's account or a delegated provider's account,
+		///  depending on the intended use.
+		///
+		/// The function requires two signatures: `msa_owner_proof` and `new_key_owner_proof`, which serve as proofs of ownership for the existing MSA
```
shannonwells, line 619: only `msa_owner_proof` serves as a proof of ownership for the existing MSA.  `new_key_owner_proof` proves that the submitter of the extrinsic got the new key owner to sign the payload.  It's really that the extrinsic submitter can't add a new key without both the MSA owner, AND the new key owner's permission.

pallets/msa/src/lib.rs
```diff
@@ -610,12 +610,18 @@ pub mod pallet {
...
 
-		/// Adds a given `new_key` to `msa_id` of the account signing `msa_owner_proof`, which must match the MSA in `add_key_payload`.
-		/// The `new_key` must sign the `add_key_payload` to authorize the addition.
+		/// Adds a new public key to an existing Message Source Account (MSA). This functionality enables the MSA owner to manage multiple keys
+		/// for their account or rotate keys for enhanced security.
 		///
-		/// # Remarks
-		/// * Origin can be same as msa owner.
-		/// * Signatures should be over the [`AddKeyData`] struct
+		/// The `origin` parameter represents the account from which the function is called and can be either the MSA owner's account or a delegated provider's account,
```
shannonwells, line 616: > The `origin` parameter represents the account from which the function is called

While that's technically true, the Origin is the same as any other extrinsic Origin, the submitter of the extrinsic call.

The extrinsic is designed to allow either the MSA Owner (if they pay token for it) or the MSA's Provider to be able to submit the extrinsic, so that new keys can be added via a Provider's dApp (so they _don't_ pay token for it)

pallets/msa/src/lib.rs
```diff
@@ -610,12 +610,18 @@ pub mod pallet {
...
+		///  depending on the intended use.
+		///
+		/// The function requires two signatures: `msa_owner_proof` and `new_key_owner_proof`, which serve as proofs of ownership for the existing MSA
+		/// and the new public key account, respectively.
+		///
+		/// The necessary information for the key addition, such as the new public key and the MSA ID, is provided within the `add_key_payload` parameter.
```
shannonwells, line 622:
```suggestion
		/// The necessary information for the key addition, such as the new public key and the MSA ID, is provided within the `add_key_payload` parameter, of type  [AddKeyData](https://github.com/LibertyDSNP/frequency/blob/5b8790e8c6d2a07ada76357193f3cffc568cf9a3/pallets/msa/src/types.rs#L22) 
```

pallets/msa/src/lib.rs
```diff
@@ -610,12 +610,18 @@ pub mod pallet {
...
+		/// Additionally, the `add_key_payload` includes an expiration block number for both proofs, ensuring they are valid and must be greater than
```
shannonwells, line 623: If you link to AddKeyData you won't need this line, IMO, assuming AddKeyData is adequately documented.

pallets/msa/src/lib.rs
shannonwells: I don't think we should be describing this process just for this extrinsic.  It belongs somewhere else in a more general area, since we have a lot of extrinsics that require signed payloads.

pallets/msa/src/lib.rs
rlaferla: Added link to AddKeyData

pallets/msa/src/lib.rs
rlaferla: The critical word is "respectively" which separates the two.  The overall intent of getting the permission of both the owner of the MSA and the owner of the new key will be added.

PR #1609, opened 2023-06-20
Create version 2 of MSA RPC to return revoked_at block number for schema permission grants

### LINKED ISSUE #1610: [bug] Schema permission RPC response should include revocation block number
### Description

(somewhat simplified, but basically:)
Delegations are a double map on `<grantor id, delegatee id>`
Each delegation stores a BTreeMap with key `schema_id` and value `block_number`, where `block number` is 0 if schema permission grant is in force, and non-zero represents the block number at which the permission for that schema was revoked.
Currently, the only way to grant permission on a schema to an existing delegation is to call `grant_delegation`. `grant_delegation` syncs the list of `schema_id`s provided and revokes permission from any schema not explicitly given in the call. So there is no way to "add" a schema permission; you must first call the RPC `get_granted_schemas_by_msa_id` to request all existing schema permissions for a given delegation, then add the desired schema to the list, and finally call `grant_delegation` with the new list.

However, that flow does not work, because when we retrieve schema permission grants for a delegation, the RPC only returns schema_id, not the block number--so we do not know if the schema permission is currently active or revoked; we only know that it was active at one time.

### Acceptance Criteria
* Created v2 of `get_granted_schemas_by_msa_id`
* v2 of `get_granted_schemas_by_msa_id` should return an array of structures containing `schema_id, revoked_at_block_number`

JoeCap08055:
# Goal
The goal of this PR is to enhance the MSA RPC call `get_granted_schemas_by_msa_id` to return the `revoked_at` block number in addition to just the schema id.

Closes #1610 

# Checklist
- [x] Chain spec updated
- [ ] Design doc(s) updated
- [x] Tests added
- [ ] Benchmarks added
- [ ] Weights updated


pallets/msa/src/rpc/src/tests/mod.rs
```diff
@@ -16,6 +16,7 @@ const SCHEMA_FOR_A: u16 = 1;
 const SCHEMA_FOR_A_AND_B: u16 = 2;
 const SCHEMA_FOR_B: u16 = 3;
 const NOT_EXIST_MSA: u64 = 100;
+const BLOCK_ZERO: BlockNumber = 0;
```
johnmela, line 19: We could possibly move the definition for 
`const BLOCK_ZERO: BlockNumber = 0;`
to `common/primitives/src/msa.rs` and import it here as well as `other_tests.rs`

shannonwells (approved):
Conditional approve: Need to update the RPC rustdoc comment as noted below.

pallets/msa/src/rpc/src/tests/mod.rs
JoeCap08055: Ultimately I would like to see a `tests` module exported from `common`, but I feel that's out of scope here. And I don't want to introduce a new dependency between `pallet-msa` and `pallet-msa-rpc` that doesn't currently exist, so I'm just going to keep BLOCK_ZERO local in both crates.

johnmela: Ah, fair enough. I was thinking of a common declaration somewhere since its used by both the msa rpc tests and general unit tests

pallets/msa/src/tests/other_tests.rs
```diff
@@ -475,8 +478,10 @@ pub fn schema_granted_success_rpc() {
 		let schema_grants = vec![1, 2];
 		assert_ok!(Msa::add_provider(provider, delegator, schema_grants));
 		let schemas_granted = Msa::get_granted_schemas_by_msa_id(delegator, provider);
-		let expected_schemas_granted = vec![1, 2];
-		let output_schemas: Vec<SchemaId> = schemas_granted.unwrap().unwrap();
+		let expected_schemas_granted =
```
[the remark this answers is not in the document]
JoeCap08055: Test added.

johnmela (approved):

PR #1616, opened 2023-06-23
Bugs/provider nonce #1613

### LINKED ISSUE #1613: Verification: Account Nonce is retained for Provider Control Keys without a Balance
Setup:
- Generate Key 1
- Create MSA 1 with Key 1 as Control Key
- Register MSA 1 as Provider A
- Generate Key 2
- Add Key 2 as a control key for Provider A

Action: Send capacity transaction using Key 2

Expected: See that nonce for Key 2 is now 1

shannonwells:
# Goal
The goal of this PR is to address #1613 by adding a new end-to-end test that demonstrates the behavior described.

Closes #1613 

# Discussion
The test verifies that if a Provider adds a new control key via `pallet_msa.add_public_key_to_msa`, and that new key does not have a balance, calling a capacity transaction with that key is an invalid transaction. It fails with `1010: Invalid Transaction: Custom error: 3`

Also fixes a problem with the tests for capacity transactions on the msa pallet, which were not running.  I had to do some restructuring to get them to pass.

It should be documented that if Providers wish to use different control keys to spend their capacity, the keys must have an existential balance, i.e. they must be "real" accounts on the Frequency chain.

# Checklist
- [x] Tests added

### DIFF
--- a/integration-tests/capacity/transactions.test.ts
+++ b/integration-tests/capacity/transactions.test.ts
@@ -19,7 +19,7 @@ import {
     getOrCreateParquetBroadcastSchema, getOrCreateAvroChatMessagePaginatedSchema, CHAIN_ENVIRONMENT
 } from "../scaffolding/helpers";
 
-describe.only("Capacity Transactions", function () {
+describe("Capacity Transactions", function () {
     const FUNDS_AMOUNT: bigint = 50n * DOLLARS;
 
   describe("pay_with_capacity", function () {
--- a/integration-tests/capacity/transactions.test.ts
+++ b/integration-tests/capacity/transactions.test.ts
@@ -19,15 +19,15 @@ import {
     getOrCreateParquetBroadcastSchema, getOrCreateAvroChatMessagePaginatedSchema, CHAIN_ENVIRONMENT
 } from "../scaffolding/helpers";
 
-describe("Capacity Transactions", function () {
-  const FUNDS_AMOUNT: bigint = 100n * DOLLARS;
+describe.only("Capacity Transactions", function () {
+    const FUNDS_AMOUNT: bigint = 25n * DOLLARS;
 
   describe("pay_with_capacity", function () {
     describe("when caller has a Capacity account", async function () {
       let capacityKeys: KeyringPair;
       let capacityProvider: u64;
       let schemaId: u16;
-      const amountStaked = 10n * DOLLARS;
+      const amountStaked = 2n * DOLLARS;
 
       before(async function () {
         capacityKeys = createKeys("CapacityKeys");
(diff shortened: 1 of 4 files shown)


integration-tests/capacity/transactions.test.ts
```diff
@@ -22,555 +22,604 @@ import {
...
+      before(async function () {
+        capacityKeys = createKeys("CapacityKeys");
+        capacityProvider = await createMsaAndProvider(capacityKeys, "CapacityProvider", FUNDS_AMOUNT);
+
+        // Create schemas for testing with Grant Delegation to test pay_with_capacity
+        schemaId = await getOrCreateGraphChangeSchema();
+        assert.notEqual(schemaId, undefined, "setup should populate schemaId");
+      });
+
+      async function assertAddNewKey(addKeyPayload: AddKeyData, newControlKeypair: KeyringPair) {
```
shannonwells, line 41: The formatter kind of messed things up for the diffs.  If you hide the whitespace it'll be easier to read.

integration-tests/capacity/transactions.test.ts
```diff
@@ -22,555 +22,604 @@ import {
...
+          baseHandle: handle_vec,
+          expiration: expiration,
+        };
+        const claimHandlePayload = ExtrinsicHelper.api.registry.createType("CommonPrimitivesHandlesClaimHandlePayload", handlePayload);
+        const claimHandle = ExtrinsicHelper.claimHandle(newControlKeypair, claimHandlePayload);
+        const [_bar, chainEvents2] = await claimHandle.payWithCapacity();
+        assertEvent(chainEvents2, "system.ExtrinsicFailed");
+      });
+
+      describe.only("when capacity eligible transaction is from the msa pallet", async function () {
```
wilwade, line 77:
```suggestion
      describe("when capacity eligible transaction is from the msa pallet", async function () {
```

integration-tests/capacity/transactions.test.ts
```diff
@@ -22,555 +22,604 @@ import {
...
+        const expiration = (await getBlockNumber()) + 10;
+        const handle_vec = new Bytes(ExtrinsicHelper.api.registry, handle);
+        const handlePayload = {
+          baseHandle: handle_vec,
+          expiration: expiration,
+        };
+        const claimHandlePayload = ExtrinsicHelper.api.registry.createType("CommonPrimitivesHandlesClaimHandlePayload", handlePayload);
+        const claimHandle = ExtrinsicHelper.claimHandle(newControlKeypair, claimHandlePayload);
+        const [_bar, chainEvents2] = await claimHandle.payWithCapacity();
+        assertEvent(chainEvents2, "system.ExtrinsicFailed");
```
wilwade, line 74: Question: Does this also trigger any cost at all?

Aka, is this a DoS vector:
- Create provider
- Add tokenless key to provider
- Get small amount of Capacity
- Flood the chain with failing capacity calls

wilwade (dismissed):
Found a `.only` in there, but otherwise the tests look correct. One DoS vector question as well.

integration-tests/capacity/transactions.test.ts
shannonwells: thought I got rid of that one, got it.

integration-tests/capacity/transactions.test.ts
shannonwells: It doesn't even get into the mempool I think but I will double check.  The error is "Invalid Transaction"

enddynayn (approved):
👍 nice!

wilwade (approved):
🚢

PR #1617, opened 2023-06-24
fix: Optimized replenish_all()

rlaferla:
# Goal
The goal of this PR is to optimize the Capacity replenish_all() function

Closes


wilwade (dismissed):
Can you give a bit more context on what is changing here? I'm a bit lost.

rlaferla:
It's a small optimization since the current one called a second function and performed multiple calculations that could be performed in just two.

shannonwells (approved):

saraswatpuneet (approved):

PR #1633, opened 2023-07-11
Increase Capacity Batch limit from 2 -> 10

### LINKED ISSUE #1628: Increase the capacity batch limit to 10
We need to increase the capacity batch limit to 10 as with the implementation of graph calls needed our current batch size of 3 will fall short of what we need.

wilwade:
# Goal
The goal of this PR is to increase the limit on capacity batches from 2 -> 10

Closes #1628

# Discussion
Why? 2 was set originally to be initially careful with how many were allowed. With additional testing, it is clear there is no reason to not increase to at least 10. In think in the end we'll be able to increase to higher, but going slow.

# Checklist
- [x] Weights updated


JoeCap08055 (approved):

saraswatpuneet (approved):
:shipit:

PR #1635, opened 2023-07-12
Add js extensions to imports to support esm fully in api-augment

wilwade:
# Goal
To make sure that the generated api-augment javascript can be used directly with node 16+ and in browsers when using es modules, we needed to use the full path

See: https://nodejs.org/dist/latest-v14.x/docs/api/esm.html#esm_resolution_algorithm

# Discussion
- Added the `.js`
- Added the linter to enforce it


saraswatpuneet (approved):
Lgtm

JoeCap08055 (approved):
💯

Issue #1362, opened 2023-04-10
Switch workflows to new EKS runners


demisx:
The new runners have been successfully tested with Verify PR workflow. We need to initiate the switch.

**Tasks:**
- [x] Switch jobs in "Verify PR" workflow https://github.com/LibertyDSNP/frequency/pull/1363
- [x] (devOps) Resolve [Process completed with exit code 255](https://github.com/LibertyDSNP/frequency/actions/runs/4671666885/jobs/8272944460?pr=1346)
- [x] 4/17+: Switch remaining jobs in "Verify PR" workflow (except failing code coverage)
- [x] Switch "Merge PR" workflow https://github.com/LibertyDSNP/frequency/pull/1459
- [x] Switch "Release" workflow https://github.com/LibertyDSNP/frequency/pull/1460
- [x] Use ci-base-image globally
- [x] Add `v1` label to the legacy runners https://github.com/LibertyDSNP/frequency/pull/1636
- [x] Rename `testing, v2` -> `build, v2` label https://github.com/LibertyDSNP/frequency/pull/1636
- [x] ~~Ask devOps to disable legacy runners~~ Still needed for couple of jobs

Blocked by:
- #1449

PR #1636, opened 2023-07-17
update runner labels

demisx:
# Goal
The goal of this PR is to update runner labels to their long-term stable versions.

Part of #1362


wilwade (approved):
👍

PR #1638, opened 2023-07-19
[stateful storage] add a new extrinsic for ItemizedSignaturePayloadV2

### LINKED ISSUE #1637: [Stateful storage] Remove msa_id form ItemizedSignaturePayload
# Problem

To be able to support signing of new graph key without having the msa created first from inside custodial wallet we will need to remove `msa_id` from `ItemizedSignaturePayload`. This should not have any security negative side-effects due to  checking of the total signature.

# Accepetence criterea
-  For backwards compatibility create a new `ItemizedSignaturePayloadV2`
- Create a new extrinsic to have `ItemizedSignaturePayloadV2` as the param
- Write integration test to check the new extrinsic
- Update spec version

aramikm:
# Goal
The goal of this PR is to allow generating signatures for adding itemized storages before even creating the msa. This is required for custodial wallet flow.

Closes #1637 

# Discussion
- for backwards compatibility reasons instead of just removing the msa_id from the payload I created a v2 for the payload and extrinsic
- benchmarks and stables weights are not changed at all so it's using the old extrinsic one
- I have not been working on the chain for a while so please let me know any any new process or step is missing.

# Checklist
- [x] Chain spec updated
- [x] Tests added


pallets/stateful-storage/src/types.rs
```diff
@@ -85,6 +85,7 @@ pub enum PageError {
 	PageSizeOverflow,
 }
 
+/// Warning: This struct is `deprecated`. please use `ItemizedSignaturePayloadV2` instead
```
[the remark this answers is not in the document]
mattheworris: Looks like Rust has a tag for `deprecated`:
https://doc.rust-lang.org/reference/attributes/diagnostics.html#the-deprecated-attribute

runtime/frequency/src/lib.rs
```diff
@@ -751,6 +751,7 @@ impl GetStableWeight<RuntimeCall, Weight> for CapacityEligibleCalls {
 			RuntimeCall::StatefulStorage(StatefulStorageCall::upsert_page { payload, ..}) => Some(capacity_stable_weights::SubstrateWeight::<Runtime>::upsert_page(payload.len() as u32)),
 			RuntimeCall::StatefulStorage(StatefulStorageCall::delete_page { .. }) => Some(capacity_stable_weights::SubstrateWeight::<Runtime>::delete_page()),
 			RuntimeCall::StatefulStorage(StatefulStorageCall::apply_item_actions_with_signature { payload, ..}) => Some(capacity_stable_weights::SubstrateWeight::<Runtime>::apply_item_actions_with_signature(StatefulStorage::sum_add_actions_bytes(&payload.actions))),
+			RuntimeCall::StatefulStorage(StatefulStorageCall::apply_item_actions_with_signature_v2 { payload, ..}) => Some(capacity_stable_weights::SubstrateWeight::<Runtime>::apply_item_actions_with_signature(StatefulStorage::sum_add_actions_bytes(&payload.actions))),
```
aramikm, line 754: @enddynayn Is this enough to allow capacity transactions or I need to do more?

pallets/stateful-storage/src/types.rs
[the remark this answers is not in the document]
aramikm: I'll add to extrinsic but adding it on top of the struct generated warning messages that are not supressed by 
`#[allow(deprecated)]`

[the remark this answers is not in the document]
aramikm: actually that is expected based on what deprecated does
`The deprecated attribute marks an item as deprecated. rustc will issue warnings on usage of #[deprecated] items. rustdoc will show item deprecation, including the since version and note, if available.`

shannonwells (commented):
Seems fairly straightforward.  I'm wondering if you considered making the extrinsic with Option<MessageSourceId> and if so what would be the reason for not using that.  Is the old version simply not needed?

JoeCap08055 (approved):
lgtm

aramikm:
> Seems fairly straightforward. I'm wondering if you considered making the extrinsic with Option and if so what would be the reason for not using that. Is the old version simply not needed?

Not sure what you mean by `making the extrinsic with option`, would you mind elaborate on that? As for the older version you are correct, it will be removed in future.

runtime/frequency/src/lib.rs
```diff
@@ -249,7 +249,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	spec_name: create_runtime_str!("frequency"),
 	impl_name: create_runtime_str!("frequency"),
 	authoring_version: 1,
-	spec_version: 44,
+	spec_version: 46,
```
mattheworris, line 252: Just a reminder this should be 45

runtime/frequency/src/lib.rs
mattheworris: Do we need to update `frequency-tx-payment/src/capacity_stable_weights.rs`?
If so, do we need to run benchmarks to update the weights?
Should this be:
```suggestion
			RuntimeCall::StatefulStorage(StatefulStorageCall::apply_item_actions_with_signature_v2 { payload, ..}) => Some(capacity_stable_weights::SubstrateWeight::<Runtime>::apply_item_actions_with_signature_v2(StatefulStorage::sum_add_actions_bytes(&payload.actions))),
```

mattheworris (commented):
I had some questions about the weights and pay for capacity, that I'm not sure about.

runtime/frequency/src/lib.rs
mattheworris: Just noticed the comment in the issue that weights haven't changed, so it seems like this comment can be ignored.

aramikm: Well I could copy and paste all the weights and benchmarks but not sure it worths the effort since we are going to remove the old extrinsic in near future

mattheworris: I think I see, if the weights are the same, I agree, no need to copy and paste a duplicate set.
The only other thing I was thinking is that the integration tests do test for every capacity eligible transaction.
Do we want to test the v1 and v2 concurrently?

aramikm: So thanks for reminding to add capacity integration test for this new extrinsic. I'll add that

enddynayn (approved):
Awesome! 👍

integration-tests/capacity/transactions.test.ts
```diff
@@ -336,6 +357,42 @@ describe("Capacity Transactions", function () {
           assert.notEqual(pageUpdateEvent1, undefined, "should have returned a PalletStatefulStorageItemizedActionApplied event");
         });
 
+        it("successfully pays with Capacity for eligible transaction - applyItemActionsWithSignatureV2", async function () {
```
mattheworris, line 360: 👍

PR #1641, opened 2023-07-22
update metadata labels

### LINKED ISSUE #1640: Change CI to use different labels for metadata-changed vs. metadata-version-mismatch
In our standup meeting 7/20, the Frequency team voted on making this change based on confusion that several developers have encountered around the "metadata-mismatch" label.

The proposal is to use two labels:

1.  `metadata-version-not-incremented` - a red label indicating that the metadata has changed but the version has not incremented and thus CI will fail.
2.  `metadata-changed` a yellow label indicating that the metadata changed but the version was incremented, therefore CI will succeed.

demisx:
# Goal
The goal of this PR is to update mismatch labels and the logic when they are applied.

Closes #1640

### DIFF
--- a/.github/workflows/verify-pr-commit.yml
+++ b/.github/workflows/verify-pr-commit.yml
@@ -9,6 +9,8 @@ on:
       - "*-development" # Feature Branches should suffix with -development
 env:
   BIN_DIR: target/release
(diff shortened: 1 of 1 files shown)


rlaferla (approved):
L👀ks great!  Thanks Dmitri!

wilwade (approved):

PR #1647, opened 2023-07-28
feat: change user handle

### LINKED ISSUE #1627: [user handles] handle swapping
Currently, we don't support swapping a handle for a new one. Right now we need to retire & claim to meet this use case. Although this works, it makes for a poor provider/wallet developer experience.

As a provider/wallet, I would like a single extrinsic for swapping a handle + instruction what is the optimal way of listening handle changes on the blockchain. 

## Req
- Should be able to be called with Capacity

## Open Questions
- Should we instead just allow claiming a handle when an existing handle is already claimed? aka swap because a new handle is being claimed instead of throwing an error when a handle is claimed?

## Implementation Notes
- Should require the user signature just like claiming a handle

rlaferla:
# Goal
The goal of this PR is add the ability to change a user handle.

Closes #1627 

- A user who already has claimed a user handle can now change it using the `change_handle()` extrinsic.
- The `change_handle()` extrinsic can be paid directly with tokens or with capacity

Changing a user handle, will retire the existing handle and claim a new one.  The extrinsic emit `HandleRetired` and `HandleClaimed` events.

image


pallets/handles/src/lib.rs
```diff
@@ -417,11 +417,77 @@ pub mod pallet {
...
+			Self::verify_signature_mortality(payload.expiration.into())?;
+
+			// Validation: Verify the payload was signed
+			Self::verify_signed_payload(&proof, &msa_owner_key, payload.encode())?;
+
+			// Get existing handle to retire
+			Self::get_display_name_for_msa_id(msa_id).ok_or(Error::<T>::MSAHandleDoesNotExist)?;
+
+			// Retire old handle
+			let old_display_handle: Vec<u8> = Self::do_retire_handle(msa_id)?;
```
shannonwells, line 482: so nice and clean, we just call do_whatever

pallets/handles/src/tests/handle_change_tests.rs
```diff
@@ -0,0 +1,59 @@
...
+		let (payload, proof) =
+			get_signed_claims_payload(&alice, new_handle.as_bytes().to_vec(), expiry);
+		assert_ok!(Handles::change_handle(
+			RuntimeOrigin::signed(alice.public().into()),
+			alice.public().into(),
+			proof,
+			payload
+		));
+		let changed_handle = create_full_handle_for_index(new_handle, 0);
+		System::assert_last_event(Event::HandleClaimed { msa_id, handle: changed_handle }.into());
```
shannonwells, line 37: Shouldn't this also check for HandleRetired?

pallets/handles/src/lib.rs
```diff
@@ -417,11 +417,77 @@ pub mod pallet {
...
+		pub fn change_handle(
+			origin: OriginFor<T>,
+			msa_owner_key: T::AccountId,
+			proof: MultiSignature,
+			payload: ClaimHandlePayload<T::BlockNumber>,
+		) -> DispatchResult {
+			let _ = ensure_signed(origin)?;
+
+			// Validation: Check for base_handle size to address potential panic condition
+			ensure!(
```
shannonwells, line 463: Since much of this parameter checking is the same as for claim_handle, I think it would work to encapsulate it and call it from both places.  Then you could have unit tests just for that verification and not need to test them in both extrinsics.

pallets/handles/src/tests/handle_change_tests.rs
```diff
@@ -0,0 +1,59 @@
...
+			Handles::change_handle(
+				RuntimeOrigin::signed(alice.public().into()),
+				alice.public().into(),
+				proof,
+				payload
+			),
+			Error::<Test>::MSAHandleDoesNotExist
+		);
+	});
+}
```
shannonwells, line 59: I think this needs tests that exercise the parameter checks too, e.g. the handle size. But if you did the encapsulation you could do unit tests there, as noted above.

shannonwells (commented):
I think this needs a little extra testing

pallets/handles/src/tests/handle_change_tests.rs
rlaferla: We could but you can't claim a handle unless it's retired so it's not strictly necessary.

pallets/frequency-tx-payment/src/capacity_stable_weights.rs
```diff
@@ -215,4 +216,15 @@ impl<T: frame_system::Config> WeightInfo for SubstrateWeight<T> {
 			.saturating_add(T::DbWeight::get().reads(3 as u64))
 			.saturating_add(T::DbWeight::get().writes(3 as u64))
 	}
+	// Storage: Msa PublicKeyToMsaId (r:1 w:0)
+	// Storage: Handles MSAIdToDisplayName (r:1 w:1)
+	// Storage: Handles CanonicalBaseHandleToSuffixIndex (r:1 w:1)
+	// Storage: Handles CanonicalBaseHandleAndSuffixToMSAId (r:0 w:1)
+	fn change_handle(b: u32) -> Weight {
```
wilwade, line 223: I think this needs to be updated from the benchmark run. They had a different number of db writes.

wilwade (commented):
Just one note to update the stable weights with the benchmark generated one. Otherwise looks good to me from a 👀 check.

pallets/frequency-tx-payment/src/capacity_stable_weights.rs
wilwade: This is the handles benchmarks not the f-tx-payment run. It is generated just by copying over the details to stable weights. (Or at least I believe so)

pallets/handles/src/lib.rs
rlaferla: done

pallets/handles/src/tests/handle_change_tests.rs
rlaferla: I added a check for the HandleRetired event.

pallets/frequency-tx-payment/src/capacity_stable_weights.rs
rlaferla: Fixed

pallets/handles/src/tests/handle_change_tests.rs
rlaferla: done

rlaferla: I will also add a check for minimum handle length.

shannonwells (approved):
Cool beans with awesome sauce

PR #1657, opened 2023-08-15
Upgrade to v0.9.43

### LINKED ISSUE #1599: Upgrade to Polkadot Release v0.9.43
The Parity team has published new version of Polkadot: `v0.9.43`

See https://github.com/paritytech/polkadot/releases/tag/v0.9.43

Analysis: https://forum.polkadot.network/t/polkadot-release-analysis-v0-9-43/3158

## Major Impact
- rpc server: break legacy CLI options and remove “backward compatible HTTP server”

## Upgrade Priority

See [Polkadot Release Page](https://github.com/paritytech/polkadot/releases/tag/v0.9.43)

## Host Functions

See [Polkadot Release Page](https://github.com/paritytech/polkadot/releases/tag/v0.9.43)

## Database Migrations

See [Polkadot Release Page](https://github.com/paritytech/polkadot/releases/tag/v0.9.43)

## Runtime Migrations

See [Polkadot Release Page](https://github.com/paritytech/polkadot/releases/tag/v0.9.43)

aramikm:
# Goal
The goal of this PR is upgrade the inner dependencies such as `polkadot`, `substrate` and `cumulus` to v0.9.43

Closes #1599 

Change impact https://forum.polkadot.network/t/polkadot-release-analysis-v0-9-43/3158

# Major inner changes
- [x] removing of web-socket related cli params and merging them with Rpc ones. (Merging ports)
       - everything that was calling port `9933` should be changed to call `9944`
       - `9934` -> `9945`
       - `9935` -> `9946`
       - `9936` -> `9947`
- [x] Upgrading rust tool chain to `rustc 1.71.0-nightly`
- [x] testing runtime upgrade locally
- [ ] dependent issue #1659 

# Discussions
- cli changes will break any tool that is calling those not supported ports
- `ed25519-dalek` vulnerability is fixed in their repo but there is no official releases that have that (only `rc` ones)
- skipped adding sudo pallet weights in benchmarks and copied the substrate weights since using a macro for feature inside another macro for benchmark definitions was not supported and we only use sudo in rococo and local.
- There are some infrastructure changes that might need to be sequenced before deploying this upgrade


docker/docker-compose.yml
```diff
@@ -4,13 +4,12 @@ version: "3"
 services:
   relay_alice:
     container_name: alice
-    image: "parity/polkadot:v0.9.42"
+    image: "parity/polkadot:v0.9.43"
     platform: linux/x86_64
     hostname: relay_alice
     ports:
       - "30335:30335" # p2p port
-      - "9935:9933" # rpc port
```
aramikm, line 12: we can route both ports to `9944` as an alternative. please let me know which one is more preferable

runtime/common/src/weights/pallet_sudo.rs
```diff
@@ -0,0 +1,75 @@
...
+// --header=./HEADER-APACHE2
+// --output=./frame/sudo/src/weights.rs
+// --template=./.maintain/frame-weight-template.hbs
+
+#![cfg_attr(rustfmt, rustfmt_skip)]
+#![allow(unused_parens)]
+#![allow(unused_imports)]
+#![allow(missing_docs)]
+
+use frame_support::{traits::Get, weights::{Weight, constants::RocksDbWeight}};
```
aramikm, line 31: copied this from substrate and I don't think we can currently ran the benchmarks for it. The reason is that we can not add `pallet_sudo` into list of benchmarks due to it only being in certain features (and macro didn't supporting that) and we might need to run it manually to get these value but if there Frequency reference hardware is close enough to the Polkadot then we might be able to just use these values.

wilwade: I think that the copy-paste is fine:

1. Rococo and local only
2. Sudo usually overrides the weight anyway.

scripts/run_benchmarks.sh
```diff
@@ -15,6 +15,7 @@ ALL_EXTERNAL_PALLETS=( \
   pallet_preimage \
   pallet_scheduler \
   pallet_session \
+  pallet_sudo \
```
wilwade, line 18: Per other comment, I think we can skip this, but also if it works, then leave it.

rust-toolchain.toml
```diff
@@ -1,5 +1,5 @@
 [toolchain]
-channel = "nightly-2022-11-15"
+channel = "nightly-2023-05-22"
```
wilwade, line 2: I think we also need to update `.tool-versions`? (Although I dislike trying to convert between the nightly dated version and the semvers)

.github/workflows/verify-pr-commit.yml
```diff
@@ -226,13 +226,10 @@ jobs:
...
         uses: actions/checkout@v3
-        # cargo doc requires a toolchain version ^nightly-2022-11-15 because of a bug that causes a panic
-        # When the toolchain is updated to ^nightly-2023-06-01, this can be removed
       - name: Build Docs
         run: |
-          rustup install nightly-2023-06-01
-          rustup target add wasm32-unknown-unknown --toolchain nightly-2023-06-01
-          RUSTDOCFLAGS="--enable-index-page --check -Zunstable-options" cargo +nightly-2023-06-01 doc --no-deps --features frequency
+          rustup target add wasm32-unknown-unknown
+          RUSTDOCFLAGS="--enable-index-page --check -Zunstable-options" cargo doc --no-deps --features frequency
```
wilwade, line 232: 🎉

.github/workflows/release.yml
wilwade, line 1: One more place to update so the release notes are correct:

`  RUST_TOOLCHAIN: "nightly-2022-11-15" # Match to /rust-toolchain.toml`

wilwade (changes requested):
Found one more spot to update the toolchain and a few small unrelated adjacent changes if you want.

common/primitives/Cargo.toml
```diff
@@ -16,8 +16,8 @@ targets = ['x86_64-unknown-linux-gnu']
 codec = { package = "parity-scale-codec", version = "3.2.2", default-features = false, features = [
```
mattheworris, line 16: Parity has updated their codec.
```suggestion
codec = { package = "parity-scale-codec", version = "3.4.0", default-features = false, features = [
```

node/cli/Cargo.toml
```diff
@@ -28,39 +28,39 @@ pallet-handles = { package = "pallet-handles", path = "../../pallets/handles", d
...
-sc-executor = { git = "https://github.com/paritytech/substrate", default-features = false, branch = "polkadot-v0.9.42" }
-sp-io = { git = "https://github.com/paritytech/substrate", default-features = false, branch = "polkadot-v0.9.42" }
-sp-api = { git = "https://github.com/paritytech/substrate", default-features = false, branch = "polkadot-v0.9.42" }
-sp-core = { git = "https://github.com/paritytech/substrate", optional = true, branch = "polkadot-v0.9.42" }
-sp-inherents = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.42" }
-sp-keyring = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.42" }
-sp-runtime = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.42" }
-sp-timestamp = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.42" }
-try-runtime-cli = { git = "https://github.com/paritytech/substrate", optional = true, branch = "polkadot-v0.9.42" }
+frame-benchmarking-cli = { git = "https://github.com/paritytech/substrate", optional = true, branch = "polkadot-v0.9.43" }
```
mattheworris, line 31: Another `codec` here, search and replace `parity-scale-codec` should find them all.

rust-toolchain.toml
wilwade: Sounds like we should just scrap asdf for rust and remove the notes about the rust asdf from the readme.

mattheworris: Double checking my conversation with Peter, he finds asdf for rust useful, (if rust is installed with rustup, asdf is a noop), so he asked to keep it up to date. .tool-versions is already on 1.69 (same as 0.9.43), so I think no update is needed. The readme could detail rustup and leave asdf as an exercise left to the reader.

mattheworris (changes requested):
I checked out the branch and ran 'make ci-local'.
I ran `make start` and connected with Provider Dashboard and polkadot js ui.
Notes: I needed to do `rustup update`, something failed when I tried to build the first time, also forgot to do `make clean`, after that, all was good.

I think we need to update `parity-scale-codec`, but let me know if I missing something and we don't need to.

mattheworris:
> I checked out the branch and ran 'make ci-local'. I ran `make start` and connected with Provider Dashboard and polkadot js ui. Notes: I needed to do `rustup update`, something failed when I tried to build the first time, also forgot to do `make clean`, after that, all was good.
> 
> I think we need to update `parity-scale-codec`, but let me know if I missing something and we don't need to.

Alternatively, if we are moving to 1.0.0, the codec will need to be upgraded again to 3.6.1, so we could skip this upgrade and do it in 1.0.0.

enddynayn (approved):
👍 beautiful!

docker/parachain-node.dockerfile
```diff
@@ -23,11 +23,10 @@ COPY --from=base /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certificat
 # COPY --chown=frequency target/release/frequency.amd64 ./frequency/frequency
 COPY --chown=frequency target/release/frequency ./frequency/
 
-# 9933 for RPC call
-# 9944 for Websocket
+# 9944 for Websocket and Rpc
 # 30333 for P2P
 # 9615 for Telemetry (prometheus)
-EXPOSE 9933 9944 30333 9615
+EXPOSE 9944 30333 9615
```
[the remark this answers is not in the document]
aramikm: adding in here

.github/workflows/release.yml
aramikm: Good catch!

common/primitives/Cargo.toml
aramikm: Hmm there is a mismatch for this crate between substrate and polkadot repos.
 I've checked the substrate repo `polkadot-v0.9.43` branch and  for `parity-scale-codec` the version is "3.2.2"
but on polkadot repo `release-v0.9.43` branch the version is "3.4.0" and on cumulus its "3.1.5"
Not sure which one would make sense for us. Open to any ideas.

wilwade: With v1.0.0 we'll move to `v3.6.1` for this package, so I think it is fine as is. Added a note to the v1.0.0 upgrade issue.

node/cli/Cargo.toml
aramikm: Please look at previous comments for resolution!

docker/docker-compose.yml
```diff
@@ -4,13 +4,13 @@ version: "3"
...
-      - "9946:9944" # ws port
+      - "9946:9944" # ws and rpc port
+      - "9615:9615" # telemetry (prometheus)
```
wilwade, line 13: On relay_alice and Relay_bob we can skip telemetry as they are not rococo or mainnet. (Either way we'd need to fix that alice and bob are binding to the same local port)

runtime/common/src/weights/mod.rs
```diff
@@ -28,6 +28,7 @@ pub mod pallet_multisig;
 pub mod pallet_preimage;
 pub mod pallet_scheduler;
 pub mod pallet_session;
+pub mod pallet_sudo;
```
wilwade, line 31: Is it a problem that this is here even when the mainnet build happens?

runtime/frequency/src/lib.rs
```diff
@@ -945,6 +945,7 @@ impl pallet_handles::Config for Runtime {
 impl pallet_sudo::Config for Runtime {
 	type RuntimeEvent = RuntimeEvent;
 	type RuntimeCall = RuntimeCall;
+	type WeightInfo = weights::pallet_sudo::SubstrateWeight<Runtime>;
```
wilwade, line 948: Now that I think of it, if we are just using substrate weights, we could map this directly to the weights from the pallet like we used to before moving to our own weights.

wilwade (approved):
A few notes, but I don't think any of them are blocking.

mattheworris (approved):
Nice work!

runtime/common/src/weights/mod.rs
aramikm: if we are going to use the original weights then this is going to be redundant anyway.

runtime/frequency/src/lib.rs
aramikm: Makes sense to use original weights from pallet

PR #808, opened 2022-12-16
[Experiment] Custom checks on extrinsic submit

aramikm:
# Goal
The goal of this PR is experiment of customizing checks of a extrinsic before submitting into the pool

# Custom check example
Screen Shot 2022-12-15 at 5 09 51 PM

PR #1658, opened 2023-08-16
[RPC] Query capacity cost (`FeeDetails`) for a capacity supported transaction 

### LINKED ISSUE #1335: [Capacity Improvements] RPC for capacity tx cost
As a User I want to be able to calculate the cost of a capacity call without making the transaction

Parallel to the `query_fee_details` https://docs.rs/pallet-transaction-payment-rpc/latest/pallet_transaction_payment_rpc/trait.TransactionPaymentApiServer.html#tymethod.query_fee_details

saraswatpuneet:
# Goal
The goal of this PR is to add an RPC to compute cost of a transaction in capacity, useful to know upfront how much capacity my typical transaction will cost, following the same flavor as transaction pallet and reuse the `FeeDetails` type from substrate

Closes #1335 

# Checklist
- [ ] Chain spec updated
- [ ] Design doc(s) updated
- [x] Tests added
- [ ] Benchmarks added
- [ ] Weights updated


saraswatpuneet:
Parity seems to moving `Block::Extrinsics` to `Runtimecall` for inputs at their core APIs. So in this PR I have discarded older api method and using new api method using RuntimeCalls 

Notice updated apiVersion in followed and their respect functions

https://github.com/paritytech/substrate/blob/94be94be6d26becd2395b58ae09ca31f596afe7d/frame/transaction-payment/rpc/runtime-api/src/lib.rs#L39

pallets/frequency-tx-payment/src/lib.rs
```diff
@@ -271,6 +271,42 @@ impl<T: Config> Pallet<T> {
...
+		Extrinsic: sp_runtime::traits::Extrinsic + GetDispatchInfo,
+	>(
+		unchecked_extrinsic: Extrinsic,
+		len: u32,
+	) -> FeeDetails<BalanceOf<T>>
+	where
+		<T as pallet::Config>::RuntimeCall: Dispatchable<Info = DispatchInfo>,
+	{
+		let dispatch_info = <Extrinsic as GetDispatchInfo>::get_dispatch_info(&unchecked_extrinsic);
+		let extrinsic_weight = dispatch_info.weight;
```
enddynayn, line 291: Perhaps, I am wrong but `get_dispatch_info` is not using the stable weights.
Here you need to grab the stable weight.

pallets/frequency-tx-payment/src/lib.rs
```diff
@@ -271,6 +271,42 @@ impl<T: Config> Pallet<T> {
...
+		let weight_fee = Self::weight_to_fee(extrinsic_weight);
+
+		let len_fee = Self::length_to_fee(len);
+		let base_fee = Self::weight_to_fee(CAPACITY_EXTRINSIC_BASE_WEIGHT);
+
+		let adjusted_weight_fee = base_fee.saturating_add(weight_fee).saturating_add(len_fee);
```
enddynayn, line 297: I think there is no `adjusted_weight_fee` with Capacity fee.

aramikm:
I'm not sure if it's related to your issue but I remember creating a custom rpc that was getting an extrinsic and doing some random validation on it. It might be useful for your usecase since you want to calculate weights by passing an extrinsic of something similar https://github.com/LibertyDSNP/frequency/pull/808/files

pallets/frequency-tx-payment/src/lib.rs
saraswatpuneet: so adjusted_weight_fee here as of now is -> `capacity base fee` + `capacity weight fee` + `len fee`, you mean we dont need  to represent it as adjusted weight fee but rather `base_fee = capacity base fee + weight fee + len fee` , while `tip and adjusted_weight_fee =0` ?

Suggestions

pallets/frequency-tx-payment/src/lib.rs
saraswatpuneet: Using stable weights

pallets/frequency-tx-payment/src/lib.rs
saraswatpuneet: @wilwade  any suggestion on this, `FeeDetail` type from transaction pallet has three type of fee how it translate to our fee structure that we want to return?

pallets/frequency-tx-payment/src/lib.rs
```diff
@@ -271,6 +271,46 @@ impl<T: Config> Pallet<T> {
...
+			.expect("A collection of calls is expected at minimum one.");
+
+		let mut calls_weight_sum = Weight::zero();
+		for inner_call in calls {
+			let call_weight = T::CapacityCalls::get_stable_weight(&inner_call).unwrap_or_default();
+			calls_weight_sum = calls_weight_sum.saturating_add(call_weight);
+		}
+
+		let fees = FeeDetails { inclusion_fee: None, tip: Zero::zero() };
+		if calls_weight_sum.is_zero() {
```
enddynayn, line 296: If the call_weigh_sum is zero but the dispatch calls something, do we want to return the cost of the dispatch call?

saraswatpuneet: That might give a false idea that the call they sent is capacity enabled or not, but open to suggestions

enddynayn: small suggestion. Since we are returning here maybe get rid of the else if:
```
if { ... return}
if let Some(weight) = calll... {
}
else {
}
```

enddynayn (approved):
Stupendous 👍

pallets/frequency-tx-payment/src/lib.rs
saraswatpuneet: yep cleaned it up

wilwade (approved):

PR #1661, opened 2023-08-18
Fix Incorrect weight on `set_epoch_length`

### LINKED ISSUE #1624: `set_epoch_length` using wrong weight function
The weight function for `set_epoch_length` [is using the weight function for `unstake`](https://github.com/LibertyDSNP/frequency/blob/38e748920c993fa275d9c21bd2fa0b76189761c0/pallets/capacity/src/lib.rs#L405) instead of its own.

It should use `T::WeightInfo::set_epoch_length()` instead.

wilwade:
# Goal
The goal of this PR is to fix `set_epoch_length`'s weight function

Closes #1624

# Discussion
- Also removes unused static weight function call `grant_schema_permissions` as it was removed.

### DIFF
--- a/pallets/capacity/src/lib.rs
+++ b/pallets/capacity/src/lib.rs
@@ -402,7 +402,7 @@ pub mod pallet {
 		/// ### Errors
 		/// - Returns `Error::MaxEpochLengthExceeded` if `length` is greater than T::MaxEpochLength.
 		#[pallet::call_index(3)]
-		#[pallet::weight(T::WeightInfo::unstake())]
+		#[pallet::weight(T::WeightInfo::set_epoch_length())]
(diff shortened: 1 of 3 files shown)


pallets/capacity/src/lib.rs
```diff
@@ -402,7 +402,7 @@ pub mod pallet {
 		/// ### Errors
 		/// - Returns `Error::MaxEpochLengthExceeded` if `length` is greater than T::MaxEpochLength.
 		#[pallet::call_index(3)]
-		#[pallet::weight(T::WeightInfo::unstake())]
+		#[pallet::weight(T::WeightInfo::set_epoch_length())]
 		pub fn set_epoch_length(origin: OriginFor<T>, length: T::BlockNumber) -> DispatchResult {
```
enddynayn, line 406: ahhh Thanks!

enddynayn (approved):
👍 Nice!

PR #1665, opened 2023-08-28
Frequency Council: Update external propose majority

### LINKED ISSUE #1664: Polkadotjs Assumes a 50% + 1 External Propose Majority
However, Frequency currently has it set to 3/5 a super majority.

This will mean the UI allows it to be closed before the correct majority is reached and will fail with BadOrigin (due to a lack of enough votes).

On Mainnet this is not an issue as the 3/5 is the same as a simple majority as there are 5 council members. (Testnet has 11).

Proposed: We should switch to a simple majority as well.

https://github.com/LibertyDSNP/frequency/blob/1535f15386791f79db8c808d22df7ad7c2883e0f/runtime/frequency/src/lib.rs#L611

```rust
/// A simple-majority of 50% + 1 can have the next scheduled referendum be a straight majority-carries vote.
type ExternalMajorityOrigin = EitherOfDiverse<
  pallet_collective::EnsureProportionAtLeast<AccountId, CouncilCollective, 1, 2>,
  frame_system::EnsureRoot<AccountId>,
>;
```

wilwade:
# Goal
The goal of this PR is to update the majority size required for approving a council motion to approve a proposal for governance (that can be fast tracked).

Closes #1664

# Discussion
- This does NOT change the effective value for Mainnet currently due to the the current size of the Frequency Council.

### DIFF
--- a/runtime/frequency/src/lib.rs
+++ b/runtime/frequency/src/lib.rs
@@ -249,7 +249,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	spec_name: create_runtime_str!("frequency"),
 	impl_name: create_runtime_str!("frequency"),
 	authoring_version: 1,
-	spec_version: 50,
+	spec_version: 51,
 	impl_version: 0,
 	apis: RUNTIME_API_VERSIONS,
 	transaction_version: 1,
@@ -263,7 +263,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	spec_name: create_runtime_str!("frequency-rococo"),
 	impl_name: create_runtime_str!("frequency"),
 	authoring_version: 1,
-	spec_version: 50,
+	spec_version: 51,
 	impl_version: 0,
 	apis: RUNTIME_API_VERSIONS,
 	transaction_version: 1,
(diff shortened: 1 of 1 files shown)


shannonwells (approved):
As discussed last week

runtime/frequency/src/lib.rs
```diff
@@ -607,9 +607,9 @@ impl pallet_democracy::Config for Runtime {
 		frame_system::EnsureRoot<AccountId>,
 	>;
 
-	/// A super-majority of 3/5ths can have the next scheduled referendum be a straight majority-carries vote.
+	/// A simple-majority of 50% + 1 can have the next scheduled referendum be a straight majority-carries vote.
 	type ExternalMajorityOrigin = EitherOfDiverse<
-		pallet_collective::EnsureProportionAtLeast<AccountId, CouncilCollective, 3, 5>,
+		pallet_collective::EnsureProportionAtLeast<AccountId, CouncilCollective, 1, 2>,
```
aramikm, line 612: wouldn't this also work exactly on 50%?

wilwade: Ah it would. I should use `EnsureProportionMoreThan` to make it 50% + 1

Great catch. Fixing.

wilwade: Fixed!

aramikm (approved):
Looks Good!

PR #1284, opened 2023-03-28
RPC for Handles

saraswatpuneet:
# Goal
The goal of this PR is  to add rpcs to handles pallet to do the following

- [x] get handle for a msa
- [x] get next suffixes for a handle

TODO:
- [x] Run full node test for rpc

Closes #1235 

# Discussion

# Checklist
- [ ] Chain spec updated
- [x] Custom RPC OR Runtime API added/changed? Updated js/api-augment.
- [ ] Design doc(s) updated
- [x] Tests added
- [ ] Benchmarks added
- [ ] Weights updated

PR #1672, opened 2023-09-07
Custom RPC: Handle check

shannonwells:
# Goal
The goal of this PR is to implement a handle-checking custom RPC which applies the same validation as used by `claim_handle` extrinsic.  The purpose is to help Providers do a pre-flight check before submitting user signatures. This will help minimize round trips during the new user flow.

Closes #1284 

# Discussion
## To Verify:
Check out this branch and `make start`.

This should return false in the JSON RPC output:
```shell
curl -H "Content-Type: application/json" -d '{"id":1, "jsonrpc":"2.0", "method": "handles_validateHandle", "params": ["Bobbay😀😀"]}' http://localhost:9944/
```
This should return true in the JSON RPC output:
```shell
 curl -H "Content-Type: application/json" -d '{"id":1, "jsonrpc":"2.0", "method": "handles_validateHandle", "params": ["BobbyTables"]}' http://localhost:9944/
```

# Checklist
- [x] Chain spec updated
- [x] Tests added


rlaferla:
So one comment, is that whitespace doesn't seem to be trimmed so I can get a handle of "     "

rlaferla (approved):
Looks good to me.

enddynayn:
do we also want to validate that the handle is available?

shannonwells:
@enddynayn maybe check that later?  Just wanted to get the low-hanging fruit on this one.  I'll defer to @wilwade

saraswatpuneet (approved):
Lgtm, maybe we need to check if we are trimming white spaces in handles ( don't remember) 

👍

shannonwells:
@saraswatpuneet funny you should ask, see the ticket @rlaferla just filed: #1673

Issue #1663, opened 2023-08-23
[Dos] Possibility of dos in a transitive crate called `webpki`


saraswatpuneet:
![image](https://github.com/LibertyDSNP/frequency/assets/61435908/14740ecb-3970-408f-85b9-d073fe5dba3b)

* https://github.com/briansmith/webpki/issues/69
* https://github.com/magic-wormhole/magic-wormhole.rs/issues/204

Immediate fix is to ignore `RUSTSEC-2023-0052` in cargo deny but advisory further suggested to use fork

PR #1685, opened 2023-09-25
Regular updates

### LINKED ISSUE #1503: Integration tests on Rococo handles/
Integration tests on Rococo handles/

https://github.com/LibertyDSNP/frequency/tree/main/integration-tests/handles

These should run both rococo and local. 
If we think we should skip any of these it will be up to the developers to make the judgment call on how valuable it would be to run on Rococo.

wilwade:
# Goal
The goal of this PR is to fix various dependabot issues as well as other various updates

# Discussion
- Update api-augment packages
- Update integration tests packages
- Update integration tests to use Helia instead of js-ipfs
- Cargo updates
- Closes #1503
- Closes: #1663


integration-tests/.relay-chain.mocharc.json
```diff
@@ -10,7 +10,7 @@
     ],
     "loader": "ts-node/esm",
     "spec": [
-        "./{,!(node_modules|load-tests)/**}/{createMsa,addIPFSMessage,staking,transactions}.test.ts"
+        "./{,!(node_modules|load-tests)/**}/{createMsa,addIPFSMessage,staking,transactions,handles}.test.ts"
```
wilwade, line 13: Added support for the handles test in the testnet test runner

integration-tests/messages/ipfs.ts
```diff
@@ -0,0 +1,21 @@
+import { unixfs } from '@helia/unixfs';
+import { createHelia } from 'helia';
+
+export async function ipfsCid(content: string, path: string) {
+  // create a Helia node
+  const helia = await createHelia({ start: false });
```
wilwade, line 6: We don't actually need to start the server to get the CID from it

integration-tests/messages/ipfs.ts
```diff
@@ -0,0 +1,21 @@
+import { unixfs } from '@helia/unixfs';
+import { createHelia } from 'helia';
```
wilwade, line 2: js-ipfs is deprecated. Helia is now the standard.

integration-tests/package.json
```diff
@@ -2,12 +2,13 @@
     "name": "frequency-integration-tests",
     "version": "1.0.0",
     "description": "Integration tests for the Frequency project",
+    "type": "module",
```
wilwade, line 5: Integration tests are now esmodules

integration-tests/tsconfig.json
```diff
@@ -4,7 +4,7 @@
     "compilerOptions": {
         "allowSyntheticDefaultImports": true,
         "esModuleInterop": true,
-        "module": "CommonJS",
+        "module": "esnext",
```
JoeCap08055, line 7: ❤️

JoeCap08055 (approved):
lgtm

integration-tests/messages/ipfs.ts
saraswatpuneet: did you figure out helia? would be great if we can get a refresher

wilwade: Mostly just used https://github.com/ipfs/helia/wiki/Migrating-from-js-IPFS and the other Helia examples.

PR #1694, opened 2023-10-03
Provider Boosting implementation

shannonwells:
# Goal
Implement a Provider Boost feature, whereby token holders may support the network and a specific Provider by means of a custom staking model.  Token holders lock up a certain amount of token, and receive a return in Frequency token for this support.  The token holder chooses a Provider to receive some Capacity, which the Provider may use to pay for chain transactions.

Token holders may still stake for `MaximizedCapacity` and receive no token return. As before, the entire benefit for staking would go to the targeted Provider for this type.

### Support the new staking type, Provider Boost
1. The `stake` transaction now specifies the Staking Type of `MaximizedCapacity` which is stored with `StakingTargetDetails`. 
2. There are 3 new extrinsics:  
    * `provider_boost` which specifies ProviderBoost staking type. 
    * `claim_staking_rewards` which mints and transfers all eligible rewards in token to the staker.
    * `change_staking_target` which basically swaps your Provider Boost staking target from one Provider to another.
4. The `StakingType` determines _how much_ Capacity is generated for the targeted provider.  For `ProviderBoost` type, 50% of the Capacity for `MaximizedCapacity` is generated when using `provider_boost`.
5. The `StakingType` also determines if there is a periodic return in token to the staker.  For `ProviderBoost` type, there is a periodic return.  For `MaximizedCapacity` type, there isn't.

### Issue rewards to Provider Boost accounts
1. Rewards are a fixed amount based on the proportion of staked amount to total staked amount, and capped at 10% per year, or to ~0.0385% per Reward Era.
2. Individual Rewards are calculated based on the number of _complete_ `RewardEras` that the token holder has staked.
3. A `RewardEra` is approximately two weeks of blocks.
6. Reward Pool size is a constant.
7. Rewards are are not compounded.
8. Rewards are minted when they are claimed. All available Rewards are paid out when the `claim_staking_rewards` extrinsic succeeds.
9. Rewards expire after 30 Reward Eras. This is the limit of the Reward Pool History storage and the limit of the individual Provider Boost History storage, so rewards _cannot_ be claimed for earlier eras.  With that said, if for example you ProviderBoost 10,000 FRQCY, and did nothing else for say 50 Reward Eras ("set and forget"), you would not lose everything.  You would simply would not be paid out except for the last 30 Eras.  Only the potential earnings for the 20 eras before that would be lost.

### Allow a staking 'retarget'
A staker may change some or all of their staked token to target a different provider up to 16 times in a `RewardEra` without penalty.  Otherwise, they must wait until the next `RewardEra`.

#### For more details, please see the [Capacity Staking Rewards Implementation](designdocs/capacity_staking_rewards_implementation.md) design doc, which links to the economic model for this feature.

# Checklist
- [x] Chain spec updated
- [x] 6 SECOND BLOCK TIMES:  length of constants adjusted as necessary
- [x] Design doc(s) updated
- [x] Tests added
- [x] Benchmarks added
- [x] Weights updated


aramikm (commented):
I've read the code and had a lot of questions and nits. This is very complex and after spending a good amount of time I'm still confused about some parts or how everything fits together.

js/api-augment/definitions/capacity.ts
```diff
@@ -0,0 +1,31 @@
+export default {
+  rpc: {
+    dummy: { description: 'This API has no custom RPCs', params: [], type: 'undefined' },
+  },
+  types: {
+    RewardEra: 'u32',
+    Balance: 'u128',
+    BlockNumber: 'u32',
```
saraswatpuneet, line 8: nit: `Balance` and `BlockNumber` definition is redundant.

pallets/capacity/src/lib.rs
```diff
@@ -121,7 +127,7 @@ pub mod pallet {
 		/// The maximum number of unlocking chunks a StakingAccountLedger can have.
 		/// It determines how many concurrent unstaked chunks may exist.
 		#[pallet::constant]
-		type MaxUnlockingChunks: Get<u32>;
+		type MaxUnlockingChunks: Get<u32> + Clone;
```
saraswatpuneet, line 130: I won't why we need a clonable type for this

pallets/capacity/src/lib.rs
```diff
@@ -644,18 +949,292 @@ impl<T: Config> Pallet<T> {
...
+			if !staked_amount.is_zero() {
+				let expires_at_era = reward_era.saturating_add(max_history.into());
+				let expires_at_block = Self::block_at_end_of_era(expires_at_era);
+				let eligible_amount = if staked_amount.lt(&previous_amount) {
+					staked_amount
+				} else {
+					previous_amount
+				};
+				let total_for_era =
+					Self::get_total_stake_for_past_era(reward_era, current_era_info.era_index)?;
```
aramikm, line 1111: This can get optimized. Instead of reading the storage per era you should read the storage per chunk.

pallets/capacity/src/lib.rs
```diff
@@ -644,18 +949,292 @@ impl<T: Config> Pallet<T> {
...
+			.into();
+		ensure!(total_to_mint.gt(&Zero::zero()), Error::<T>::NoRewardsEligibleToClaim);
+		let _minted_unused = T::Currency::mint_into(&staker, total_to_mint)?;
+
+		let mut new_history: ProviderBoostHistory<T> = ProviderBoostHistory::new();
+		let last_staked_amount =
+			rewards.last().unwrap_or(&UnclaimedRewardInfo::default()).staked_amount;
+		let current_era = Self::get_current_era().era_index;
+		// We have already paid out for the previous era. Put one entry for the previous era as if that is when they staked,
+		// so they will be credited for current_era.
```
aramikm, line 1223: These comment are more confusing.

designdocs/capacity_staking_rewards_economic_model.md
```diff
@@ -0,0 +1,110 @@
...
+
+[//]: # (A short description of the landscape in which the new system is being built, what is actually being built.)
+[//]: # (It may also say what is not being built, and any assumptions.)
+[//]: # (Example: The proposed feature is a testing library. The context is: the library is for our chosen blockchain. The scope is: this is for a specific repository, so it's not meant to be reused. That means it won't be a separate package, and the code will be tailored for this repo. One might also say that the scope is also limited to developer testing, so it's not meant to be used in CI or a test environment such as a test blockchain network.)
+
+## Problem Statement:
+A system consisting only of providers and coinless users who delegate to providers will tend toward centralization.
+To build a self-sustaining Frequency network where control is decentralized, a variety of economic solutions are needed.  One of these is the ability to stake FRQCY token in return for something; this creates an incentive for participation and involvement with the chain fundamentals and governance.
```
enddynayn: The line “stake FRQCY token in return for something” is a bit too general. Specifying what that “something” is such as rewards or governance rights.. would clarify the incentive for participation.

designdocs/capacity_staking_rewards_economic_model.md
```diff
@@ -0,0 +1,110 @@
...
+3. Stake a minimum amount of FRQCY token to receive [Capacity](https://github.com/LibertyDSNP/frequency/blob/main/designdocs/capacity.md).
+
+There is also a business case for allowing any token holder to lock up its tokens in exchange for a reward - known as _staking_ - while also targeting a Provider to receive some Capacity.
+
+[//]: # (A short description of the landscape in which the new system is being built, what is actually being built.)
+[//]: # (It may also say what is not being built, and any assumptions.)
+[//]: # (Example: The proposed feature is a testing library. The context is: the library is for our chosen blockchain. The scope is: this is for a specific repository, so it's not meant to be reused. That means it won't be a separate package, and the code will be tailored for this repo. One might also say that the scope is also limited to developer testing, so it's not meant to be used in CI or a test environment such as a test blockchain network.)
+
+## Problem Statement:
+A system consisting only of providers and coinless users who delegate to providers will tend toward centralization.
```
enddynayn: Maybe a quick sentence of why this leads to centralization would be nice. Maybe something like:
without a mechanism for broader participation, power and control naturally concentrate among a few providers.

designdocs/capacity_staking_rewards_economic_model.md
```diff
@@ -0,0 +1,110 @@
...
+To build a self-sustaining Frequency network where control is decentralized, a variety of economic solutions are needed.  One of these is the ability to stake FRQCY token in return for something; this creates an incentive for participation and involvement with the chain fundamentals and governance.
+
+How is that so?  Capacity is how Frequency intends Providers to pay for the vast majority of their on-chain messages. If Providers receive some capacity when end users stake, then they lose capacity if those end users unstake. If a Provider's Capacity from Reward staking is significant, this gives Reward Stakers some power over Providers. If a Provider is utilizing all or nearly all their Capacity almost every Epoch -- which they should be if trying to be economical --  then even a small percentage of lost Capacity will literally cost them to replace it.
```
enddynayn: I like this sentence:
> If a Provider's Capacity from Reward staking is significant, this gives Reward Stakers some power over Providers. 
I would add to it  - because their ability to operate **efficiently** depends on this capacity.

designdocs/capacity_staking_rewards_economic_model.md
```diff
@@ -0,0 +1,110 @@
...
+* How does the capacity generated by Maximized Capacity Staking differ from that generated by Rewards Capacity Staking? Maximized is more than Rewards, but by how much?
+* Are rewards compounded?
+
+## Economic Model
+By "economic model" we mean the formulae used to manage staking rewards to achieve the goals of decentralization, economic stabiilty, and sustainability.
+
+Inputs in the economic model to consider could include:
+* Proportion of accounts are staking to total accounts
+* Mean number MSAs per Provider, or the mean compared to median.
+* Proportion of token staked to free balance held by non-Treasury accounts
```
enddynayn: The proportion of tokens staked **compared** to **the** free balance held by non-Treasury accounts.

designdocs/capacity_staking_rewards_economic_model.md
```diff
@@ -0,0 +1,110 @@
...
+* The Capacity-per-Epoch use rate, perhaps averaged over N epochs
+* The length of an Epoch in blocks
+* The length of an Era n blocks
```
enddynayn: The length of an Era **in** blocks

designdocs/capacity_staking_rewards_economic_model.md
```diff
@@ -0,0 +1,110 @@
...
+
+## Economic Model
+By "economic model" we mean the formulae used to manage staking rewards to achieve the goals of decentralization, economic stabiilty, and sustainability.
+
+Inputs in the economic model to consider could include:
+* Proportion of accounts are staking to total accounts
+* Mean number MSAs per Provider, or the mean compared to median.
+* Proportion of token staked to free balance held by non-Treasury accounts
+* The Capacity-per-Epoch use rate, perhaps averaged over N epochs
+* The length of an Epoch in blocks
```
enddynayn: The length of an Epoch, **measured** in blocks.

pallets/capacity/src/lib.rs
```diff
@@ -151,6 +157,37 @@ pub mod pallet {
...
+		/// The ProviderBoostRewardsProvider used by this pallet in a given runtime
+		type RewardsProvider: ProviderBoostRewardsProvider<Self>;
+
+		/// A staker may not retarget more than MaxRetargetsPerRewardEra
+		#[pallet::constant]
+		type MaxRetargetsPerRewardEra: Get<u32>;
+
+		/// The fixed size of the reward pool in each Reward Era.
+		#[pallet::constant]
+		type RewardPoolEachEra: Get<BalanceOf<Self>>;
```
enddynayn, line 181: RewardPoolEachEra > RewardPoolPerEra

e2e/capacity/provider_boost.test.ts
```diff
@@ -0,0 +1,46 @@
...
+    await assert.doesNotReject(boostProvider(fundingSource, booster, provider, 1n * DOLLARS));
+  });
+
+  it('fails when they are a Maximized Capacity staker', async function () {
+    const stakeKeys = createKeys('booster');
+    const provider = await createMsaAndProvider(fundingSource, stakeKeys, 'Provider1', providerBalance);
+    await assert.rejects(boostProvider(fundingSource, stakeKeys, provider, 1n * DOLLARS));
+  });
+
+  it("fails when they don't have enough token", async function () {
```
enddynayn, line 31: To make sure the test is failing for the right reason (not enough tokens), you can update it to check the specific error message or code. This way, you know it’s not just rejecting for some random reason.

I recommend doing this to all these test so we can ensure we have not introduced a bug.

e2e/capacity/provider_boost.test.ts
```diff
@@ -0,0 +1,46 @@
...
+  const providerBalance = 2n * DOLLARS;
+
+  it('happy path succeeds', async function () {
+    const stakeKeys = createKeys('booster');
+    const provider = await createMsaAndProvider(fundingSource, stakeKeys, 'Provider1', providerBalance);
+    const booster = await createAndFundKeypair(fundingSource, 5n * DOLLARS, 'booster');
+    await assert.doesNotReject(boostProvider(fundingSource, booster, provider, 1n * DOLLARS));
+  });
+
+  it('fails when they are a Maximized Capacity staker', async function () {
```
enddynayn, line 25: who is they?

pallets/capacity/src/tests/rewards_provider_tests.rs
```diff
@@ -0,0 +1,268 @@
...
+		let rewards = Capacity::list_unclaimed_rewards(&account).unwrap();
+		assert_eq!(rewards.len(), 5usize);
+		let expected_info: [UnclaimedRewardInfo<BalanceOf<Test>, BlockNumberFor<Test>>; 5] = [
+			UnclaimedRewardInfo {
+				reward_era: 1u32,
+				expires_at_block: 130,
+				staked_amount: 1000,
+				eligible_amount: 0,
+				earned_amount: 0,
+			},
```
wilwade, line 129: Should `UnclaimedRewardInfo` ever be able to return when `earned_amount` and/or `eligible_amount` is zero? Is that not a reward at all?

pallets/capacity/src/types.rs
```diff
@@ -223,3 +246,300 @@ pub fn unlock_chunks_from_vec<T: Config>(chunks: &Vec<(u32, u32)>) -> UnlockChun
...
+			self.retarget_count = 1;
+		} else {
+			self.retarget_count = self.retarget_count.saturating_add(1u32);
+		}
+		Some(())
+	}
+}
+
+/// A trait that provides the Economic Model for Provider Boosting.
+pub trait ProviderBoostRewardsProvider<T: Config> {
```
enddynayn, line 518: It seems that Hash, Balance, and AccountId are not being used in this trait. Instead, we are using a type alias for Balance. We should consider removing these unused associated types to simplify the trait, or alternatively, ensure Self::Balance is consistently used if Balance is required.

example
```
	fn era_staking_reward(
		era_amount_staked: Self::Balance,
		era_total_staked: Self::Balance,
		era_reward_pool_size: Self::Balance,
	) -> Self::Balance;
```

e2e/capacity/change_staking_target.test.ts
```diff
@@ -0,0 +1,44 @@
...
+  createProviderKeysAndId,
+} from '../scaffolding/helpers';
+
+const fundingSource = getFundingSource('capacity-change-staking-target');
+
+describe('Capacity: change_staking_target', function () {
+  const tokenMinStake: bigint = 1n * CENTS;
+  const capacityMin: bigint = tokenMinStake / 50n;
+
+  it('happy path succeeds', async function () {
```
saraswatpuneet, line 20:
```suggestion
  it('successfully stake tokens to a provider, async function () {
```

e2e/capacity/provider_boost.test.ts
```diff
@@ -0,0 +1,46 @@
...
+  createAndFundKeypair,
+  boostProvider,
+} from '../scaffolding/helpers';
+
+const fundingSource = getFundingSource('capacity-provider-boost');
+
+describe('Capacity: provider_boost extrinsic', function () {
+  const providerBalance = 2n * DOLLARS;
+
+  it('happy path succeeds', async function () {
```
saraswatpuneet, line 18:
```suggestion
  it('succeeds in boosting a provider', async function () {
```

pallets/capacity/src/lib.rs
```diff
@@ -151,6 +157,37 @@ pub mod pallet {
 		/// How much FRQCY one unit of Capacity costs
 		#[pallet::constant]
 		type CapacityPerToken: Get<Perbill>;
+
+		/// The number of blocks in a RewardEra
+		#[pallet::constant]
+		type EraLength: Get<u32>;
```
enddynayn, line 163: I think we should use `Get<BlockNumberFor<Self>>` to ensures that your EraLength is compatible with the block number type of the runtime.

enddynayn: I miss read that.

pallets/capacity/src/lib.rs
```diff
@@ -151,6 +157,37 @@ pub mod pallet {
...
+		/// The fixed size of the reward pool in each Reward Era.
+		#[pallet::constant]
+		type RewardPoolEachEra: Get<BalanceOf<Self>>;
+
+		/// the percentage cap per era of an individual Provider Boost reward
+		#[pallet::constant]
+		type RewardPercentCap: Get<Permill>;
+
+		/// The number of chunks of Reward Pool history we expect to store
+		/// MUST be a divisor of [`Self::ProviderBoostHistoryLimit`]
```
enddynayn, line 188: Could we remove this comment by adding the following code that enfoces this at compile time?:

```
   const_assert!(Self::ProviderBoostHistoryLimit::get() % Self::RewardPoolChunkLength::get() == 0);
 ```

File: pallets/capacity/src/types.rs [hunk 20]
```diff
@@ -223,3 +246,300 @@ pub fn unlock_chunks_from_vec<T: Config>(chunks: &Vec<(u32, u32)>) -> UnlockChun
...
+		self.count().eq(&(T::ProviderBoostHistoryLimit::get() as usize))
+	}
+}
+
+/// Struct with utilities for storing and updating unlock chunks
+#[derive(Debug, TypeInfo, PartialEqNoBound, EqNoBound, Clone, Decode, Encode, MaxEncodedLen)]
+#[scale_info(skip_type_params(T))]
+pub struct RetargetInfo<T: Config> {
+	/// How many times the account has retargeted this RewardEra
+	pub retarget_count: u32,
```
saraswatpuneet, line 483: can be atomic?

pallets/capacity/src/lib.rs
```diff
@@ -212,6 +249,36 @@ pub mod pallet {
...
 
+	/// stores how many times an account has retargeted, and when it last retargeted.
+	#[pallet::storage]
+	#[pallet::getter(fn get_retargets_for)]
+	pub type Retargets<T: Config> = StorageMap<_, Twox64Concat, T::AccountId, RetargetInfo<T>>;
+
+	/// Information about the current reward era. Checked every block.
+	#[pallet::storage]
+	#[pallet::whitelist_storage]
+	#[pallet::getter(fn get_current_era)]
```
enddynayn, line 260: The `pallet::getter` macro is being deprecated I suggest using the alternative way `SomeStorage::<T>::get()`

aramikm (approved):
Read the code and it looked good! Only some nits

pallets/capacity/src/lib.rs
```diff
@@ -477,6 +715,36 @@ impl<T: Config> Pallet<T> {
...
+
+		target_details.deposit(*amount, capacity).ok_or(ArithmeticError::Overflow)?;
+		Self::set_target_details_for(staker, *target, target_details);
+
+		let mut capacity_details = Self::get_capacity_for(target).unwrap_or_default();
+		capacity_details.deposit(amount, &capacity).ok_or(ArithmeticError::Overflow)?;
+		Self::set_capacity_for(*target, capacity_details);
+
+		let era = Self::get_current_era().era_index;
+		Self::upsert_boost_history(staker, era, *amount, true)?;
```
enddynayn, line 740: I believe we should avoid writing upsert functions whenever possible. While they can be useful in certain situations, I think we can take a different approach here. We’ve encountered multiple bugs in our code due to this style, so it might be better to consider alternatives.

saraswatpuneet (approved):
great work! some nits
* Read through majority of the code
* CI needs to pass
* Left some comments and questions

pallets/capacity/src/lib.rs
```diff
@@ -644,18 +949,292 @@ impl<T: Config> Pallet<T> {
...
+			.ok_or(ArithmeticError::Overflow)?;
+
+		Self::set_target_details_for(staker, *to_msa, to_msa_target);
+		Self::set_capacity_for(*to_msa, capacity_details);
+		Ok(())
+	}
+
+	/// updates or inserts a new boost history record for current_era.   Pass 'add' = true for an increase (provider_boost),
+	/// pass 'false' for a decrease (unstake)
+	pub(crate) fn upsert_boost_history(
```
enddynayn, line 1024: I think we should avoid writing upsert functions whenever possible. While they can be useful in certain situations, I think we can take a different approach here. We’ve encountered multiple bugs in our code due to this style.

Also, when calling an upsert function, it typically creates or updates fields, but in this case, it’s also removing data, which was something I didn’t expect. This could lead to unexpected behavior and further complicates the logic.

pallets/capacity/src/lib.rs
```diff
@@ -477,6 +715,36 @@ impl<T: Config> Pallet<T> {
...
+
+		let mut capacity_details = Self::get_capacity_for(target).unwrap_or_default();
+		capacity_details.deposit(amount, &capacity).ok_or(ArithmeticError::Overflow)?;
+		Self::set_capacity_for(*target, capacity_details);
+
+		let era = Self::get_current_era().era_index;
+		Self::upsert_boost_history(staker, era, *amount, true)?;
+
+		let reward_pool_total = CurrentEraProviderBoostTotal::<T>::get();
+		CurrentEraProviderBoostTotal::<T>::set(reward_pool_total.saturating_add(*amount));
```
enddynayn, line 743: My understanding is that a user will only be rewarded for completing a full era, meaning they won’t receive rewards for the era in which they initially stake. However, it seems that when a user stakes, their stake affects the total amount that everyone can stake. 

Does this imply that new stakers reduce the rewards for those who have already staked?

e2e/capacity/list_unclaimed_rewards.test.ts
```diff
@@ -0,0 +1,72 @@
...
+    await ExtrinsicHelper.runToBlock(await getNextRewardEraBlock());
+
+    const result = await ExtrinsicHelper.apiPromise.call.capacityRuntimeApi.listUnclaimedRewards(booster.address);
+
+    assert.equal(result.length, 4);
+
+    // This is the era we first boosted in, shouldn't have any rewards
+    assert.equal(result[0].staked_amount.toHuman(), '1.0000 UNIT');
+    assert.equal(result[0].eligible_amount.toHuman(), '0');
+    assert.equal(result[0].earned_amount.toHuman(), '0');
```
wilwade, line 57: @shannonwells, I updated this to match the rust test as this was failing with what I thought was an issue with a fence post, but now I am not so sure. Sometimes we get this result, and sometimes we do not? Thoughts as to why?

pallets/capacity/src/lib.rs
```diff
@@ -212,6 +249,36 @@ pub mod pallet {
 	pub type UnstakeUnlocks<T: Config> =
 		StorageMap<_, Twox64Concat, T::AccountId, UnlockChunkList<T>>;
 
+	/// stores how many times an account has retargeted, and when it last retargeted.
```
enddynayn, line 252: I think some documentation for UnstakeUnlocks would help clarify the purpose of that storage.

pallets/capacity/src/lib.rs
```diff
@@ -477,6 +715,36 @@ impl<T: Config> Pallet<T> {
...
+		staker: &T::AccountId,
+		staking_details: &mut StakingDetails<T>,
+		target: &MessageSourceId,
+		amount: &BalanceOf<T>,
+	) -> Result<BalanceOf<T>, DispatchError> {
+		staking_details.deposit(*amount).ok_or(ArithmeticError::Overflow)?;
+		Self::set_staking_account_and_lock(staker, staking_details)?;
+
+		// get the capacity generated by a Provider Boost
+		let capacity = Self::capacity_generated(T::RewardsProvider::capacity_boost(*amount));
```
enddynayn, line 728: I would recommend renaming this to calculate_capacity instead of capacity_generated, as it calculates how much capacity to issue rather than retrieving capacity that has already been generated.

pallets/capacity/src/lib.rs
```diff
@@ -477,6 +715,36 @@ impl<T: Config> Pallet<T> {
 		Ok(capacity)
 	}
 
+	fn increase_stake_and_issue_boost_capacity(
```
enddynayn, line 718: I’ve noticed a significant overlap between `increase_stake_and_issue_capacity` and `increase_stake_and_issue_boost_capacity`. While I generally don’t mind code repetition if it enhances clarity, I’m a bit concerned in this case that we might accidentally make changes in one function and not the other, potentially leading to unintended bugs.

enddynayn: additionally, is there a reason why the function `set_staking_account_and_lock` inside both `increase_stake_and_issue_capacity` and `increase_stake_and_issue_boost_capacity` is in a different order?

pallets/capacity/src/lib.rs
```diff
@@ -477,6 +715,36 @@ impl<T: Config> Pallet<T> {
...
+		staking_details.deposit(*amount).ok_or(ArithmeticError::Overflow)?;
+		Self::set_staking_account_and_lock(staker, staking_details)?;
+
+		// get the capacity generated by a Provider Boost
+		let capacity = Self::capacity_generated(T::RewardsProvider::capacity_boost(*amount));
+
+		let mut target_details = Self::get_target_for(staker, target).unwrap_or_default();
+
+		target_details.deposit(*amount, capacity).ok_or(ArithmeticError::Overflow)?;
+		Self::set_target_details_for(staker, *target, target_details);
```
enddynayn, line 733: I also was not expecting set_target_details to also delete details perhaps a more descriptive name might be nice here such as `set_or_delete_target_details`.

(minor) Also, a more idiomatic way to rewrite the function in rust
```
fn update_or_delete_target_details(
    staker: &T::AccountId,
    target: MessageSourceId,
    target_details: StakingTargetDetails<BalanceOf<T>>,
) {
    match target_details.amount.is_zero() {
        true => StakingTargetLedger::<T>::remove(staker, target),
        false => StakingTargetLedger::<T>::insert(staker, target, target_details),
    }
}
```

pallets/capacity/src/lib.rs
```diff
@@ -430,26 +649,45 @@ impl<T: Config> Pallet<T> {
...
+	}
+
+	fn ensure_can_boost(
+		staker: &T::AccountId,
+		target: &MessageSourceId,
+		amount: &BalanceOf<T>,
+	) -> Result<(StakingDetails<T>, BalanceOf<T>), DispatchError> {
+		let (mut staking_details, stakable_amount) =
+			Self::ensure_can_stake(staker, *target, *amount, StakingType::ProviderBoost)?;
+		staking_details.staking_type = StakingType::ProviderBoost;
```
enddynayn, line 689: I’m not sure MaximumCapacity is the best default choice since it might not be the most common. If the default isn’t right for all situations, it could cause bugs or unexpected behavior, especially if developers assume it’s the right choice without realizing the potential issues.

https://github.com/frequency-chain/frequency/pull/1694/files#diff-4446b08e1f533bd6e73c4e3ee2ef514e8a7ab7dc63d770aaafcdc5a5d14bf596R74

pallets/capacity/src/lib.rs
```diff
@@ -212,6 +249,36 @@ pub mod pallet {
...
+	pub type ProviderBoostRewardPools<T: Config> =
+		StorageMap<_, Twox64Concat, u32, RewardPoolHistoryChunk<T>>;
+
+	/// How much is staked this era
+	#[pallet::storage]
+	pub type CurrentEraProviderBoostTotal<T: Config> = StorageValue<_, BalanceOf<T>, ValueQuery>;
+
+	/// Individual history for each account that has Provider-Boosted.
+	#[pallet::storage]
+	#[pallet::getter(fn get_staking_history_for)]
```
enddynayn, line 278: I believe the following might be irrelevant if we decide to remove the macro getter functions.

Since we have two types of staking, boost and maximum, it would be helpful to be specific about which type of staking we are referring to. Therefore, I recommend renaming `get_staking_history_for` to `get_boost_staking_history_for`.

e2e/capacity/provider_boost.test.ts
JoeCap08055: For on-chain extrinsic errors, I believe you can check the code. For RPC (ie, pre-dispatch) errors, we get a generic error code, but I like matching the error message against a RegEx (just some keywords or something, so it is not as brittle if the exact wording of the error changes)

e2e/scaffolding/helpers.ts
```diff
@@ -461,19 +487,26 @@ export async function setEpochLength(keys: KeyringPair, epochLength: number): Pr
   }
 }
 
+export async function getNextRewardEraBlock(): Promise<number> {
+  const eraInfo = await ExtrinsicHelper.apiPromise.query.capacity.currentEraInfo();
+  const actualEraLength: number = 50;
```
JoeCap08055, line 492: Shouldn't we be able to get this from the chain?

e2e/capacity/list_unclaimed_rewards.test.ts
wilwade: Found and fixed. It was a Genesis config issue

pallets/capacity/src/lib.rs
aramikm: I will open a PR to remove these getters

pallets/capacity/src/lib.rs
aramikm: I will open a PR to remove these getters

pallets/capacity/src/lib.rs
```diff
@@ -634,22 +953,297 @@ impl<T: Config> Pallet<T> {
...
+			ProviderBoostRewardPools::<T>::get(chunk_idx).unwrap_or(RewardPoolHistoryChunk::new()); // 1r
+
+		// If it is full we are resetting.
+		// This assumes that the chunk length is a divisor of the history limit
+		if new_chunk.is_full() {
+			new_chunk = RewardPoolHistoryChunk::new();
+		};
+
+		if new_chunk.try_insert(era, boost_total).is_err() {
+			// Handle the error case that should never happen
```
enddynayn, line 1212: I agree that this should never happen, but adding a log message here could be helpful:
log::error!("Failed to insert into the provider boost reward pool for era: {}", era);

saraswatpuneet: I wonder if tracing::info is better than log::info for distributed tracing , if at all these places are ever hit

e2e/capacity/list_unclaimed_rewards.test.ts
shannonwells: Note: we've fixed the issue with this test by adding a Genesis config.

js/api-augment/definitions/capacity.ts
```diff
@@ -0,0 +1,31 @@
+export default {
+  rpc: {
+    dummy: { description: 'This API has no custom RPCs', params: [], type: 'undefined' },
+  },
```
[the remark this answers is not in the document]
shannonwells: I think we do need it so that it shows up in the list of items augmenting state calls?

pallets/capacity/src/migration/provider_boost_init.rs
```diff
@@ -0,0 +1,47 @@
...
+
+/// Initialization during runtime upgrade for Provider Boost storage
+pub struct ProviderBoostInit<T: Config>(sp_std::marker::PhantomData<T>);
+
+impl<T: Config> OnRuntimeUpgrade for ProviderBoostInit<T> {
+	fn on_runtime_upgrade() -> Weight {
+		let current_era_info = CurrentEraInfo::<T>::get(); // 1r
+		if current_era_info.eq(&RewardEraInfo::default()) {
+			let current_block = frame_system::Pallet::<T>::block_number(); // Whitelisted
+			let era_index: RewardEra = 1u32.into();
```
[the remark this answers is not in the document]
shannonwells: Maybe I should just be ok with a first era of 0.  I didn't like the idea of "era zero" but there is a block zero, and if it causes problems maybe it's not worth it.

runtime/frequency/src/lib.rs
```diff
@@ -515,6 +522,15 @@ impl pallet_capacity::Config for Runtime {
 	type EpochNumber = u32;
 	type CapacityPerToken = CapacityPerToken;
 	type RuntimeFreezeReason = RuntimeFreezeReason;
+	type EraLength = CapacityRewardEraLength;
+	type ProviderBoostHistoryLimit = ConstU32<30u32>;
+	type RewardsProvider = Capacity;
+	type MaxRetargetsPerRewardEra = ConstU32<16>;
```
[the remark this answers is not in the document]
shannonwells: I'm okay with that!  I was trying to be generous.  Interested in other opinions.

pallets/capacity/src/migration/provider_boost_init.rs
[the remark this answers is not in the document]
shannonwells: I'm working on moving this to zero-indexed.  It's a slog so I put it in a WIP branch (`era-0-indexed`) for now.

pallets/capacity/src/lib.rs
shannonwells: MaximumCapacity needs to be default for backward compatibility.

pallets/capacity/src/lib.rs
shannonwells: I'll look at this again, but at one point it was making the code more complicated to combine the two, given what else has to happen when you boost versus staking MaximizedCapacity, however maybe there are some small optimizations.

pallets/capacity/src/lib.rs
shannonwells: The reward pool total is set at the _beginning_ of each era (in `start_new_reward_era_if_needed`).  So the reward is not changed until the _following_ era, because all staking changes don't go into effect until they've been that way for the entire era.

pallets/capacity/src/lib.rs
shannonwells: @wilwade and I reviewed this one; this history is a ring history, I think that certainly a rename to indicate the record could be deleted is a good idea.  However it would be good to know what bugs have been seen due to this pattern so I/we can know how best to change it.

pallets/capacity/src/lib.rs
```diff
@@ -644,18 +949,292 @@ impl<T: Config> Pallet<T> {
...
+			return Ok(BoundedVec::new());
 		}
+
+		let staking_history =
+			Self::get_staking_history_for(account).ok_or(Error::<T>::NotAProviderBoostAccount)?; // cached read
+
+		let current_era_info = Self::get_current_era(); // cached read, ditto
+		let max_history: u32 = T::ProviderBoostHistoryLimit::get();
+
+		let mut reward_era = current_era_info.era_index.saturating_sub((max_history).into());
```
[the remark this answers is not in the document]
shannonwells: This is because the ring function can hold onto additional data in a chunk, and we don't want to hit old data that isn't valid.

pallets/capacity/src/lib.rs
shannonwells: Suggestions for better wording, or leave it out entirely?

runtime/frequency/src/lib.rs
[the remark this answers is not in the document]
shannonwells: From @wilwade "I think making that a 2 is fine"

e2e/scaffolding/helpers.ts
shannonwells: @wilwade 4u

pallets/capacity/src/lib.rs
```diff
@@ -638,18 +957,293 @@ impl<T: Config> Pallet<T> {
...
+		let cycle: u32 = era_u32 % history_limit.saturating_add(chunk_len);
+		cycle.saturating_div(chunk_len)
+	}
+
+	// This is where the reward pool gets updated.
+	pub(crate) fn update_provider_boost_reward_pool(era: RewardEra, boost_total: BalanceOf<T>) {
+		// Current era is this era
+		let chunk_idx: u32 = Self::get_chunk_index_for_era(era);
+		let mut new_chunk =
+			ProviderBoostRewardPools::<T>::get(chunk_idx).unwrap_or(RewardPoolHistoryChunk::new()); // 1r
```
enddynayn, line 1203: (Minor) I noticed that there’s already a default implementation available, so you could simplify this by calling `ProviderBoostRewardPools::<T>::get(chunk_idx).unwrap_or_default()`.

pallets/capacity/src/lib.rs
```diff
@@ -638,18 +957,293 @@ impl<T: Config> Pallet<T> {
...
+		let era_u32: u32 = era.saturating_sub(One::one()).into();
+
+		// Add one chunk so that we always have the full history limit in our chunks
+		let cycle: u32 = era_u32 % history_limit.saturating_add(chunk_len);
+		cycle.saturating_div(chunk_len)
+	}
+
+	// This is where the reward pool gets updated.
+	pub(crate) fn update_provider_boost_reward_pool(era: RewardEra, boost_total: BalanceOf<T>) {
+		// Current era is this era
```
enddynayn, line 1200: Tracking all the weight calculations within comments can be a bit challenging. It might be more straightforward to set up the initial weight at the top level and then increment the reads or writes as they occur. This approach could improve readability and reduce the risk of errors. For example:
```
 let mut weight = T::DbWeight::get().reads(1); 
 Storage::insert(...)
 weight += T::DbWeight::get().writes(1);
 Storage::get(...)
 weight += T::DbWeight::get().reads(1);
 ```

File: pallets/capacity/src/lib.rs [hunk 40]
```diff
@@ -205,6 +242,50 @@ pub mod pallet {
...
+	#[pallet::storage]
+	#[pallet::whitelist_storage]
+	pub type CurrentEraInfo<T: Config> =
+		StorageValue<_, RewardEraInfo<RewardEra, BlockNumberFor<T>>, ValueQuery>;
+
+	/// Reward Pool history is divided into chunks of size RewardPoolChunkLength.
+	/// ProviderBoostHistoryLimit is the total number of items, the key is the
+	/// chunk number.
+	#[pallet::storage]
+	pub type ProviderBoostRewardPools<T: Config> =
```
enddynayn, line 259: I suggest creating an alias for the u32 type, as it appears to represent the chunk index. Providing an alias could make the code a bit easier to read and understand.

pallets/capacity/src/lib.rs
```diff
@@ -638,18 +957,293 @@ impl<T: Config> Pallet<T> {
...
+
+		// Add one chunk so that we always have the full history limit in our chunks
+		let cycle: u32 = era_u32 % history_limit.saturating_add(chunk_len);
+		cycle.saturating_div(chunk_len)
+	}
+
+	// This is where the reward pool gets updated.
+	pub(crate) fn update_provider_boost_reward_pool(era: RewardEra, boost_total: BalanceOf<T>) {
+		// Current era is this era
+		let chunk_idx: u32 = Self::get_chunk_index_for_era(era);
```
enddynayn, line 1201: Awesome work on the circular queue!

pallets/capacity/src/lib.rs
enddynayn: all of our extrinsic that have upsert have had bugs.

pallets/capacity/src/lib.rs
enddynayn: This seems like an easy fix if we doing a migration. It appears that we can explicitly set it.

pallets/capacity/src/lib.rs
enddynayn: Thanks for taking another look! I get that combining the two functions might complicate things, especially with the differences between boosting and staking MaximizedCapacity. That said, having a single place for the shared logic could help prevent bugs from creeping in if the duplicated code gets updated differently. Happy to help just let me know.

pallets/capacity/src/lib.rs
wilwade: On testnet at least a migration of that scale would need to be multiblock

pallets/capacity/src/lib.rs
wilwade: I don't believe this is for actual tracking, more useful comments for understanding the cost as you read through it.

enddynayn: We can return the total weight directly from each function instead of manually going inside each one to sum them up. This is a common pattern used in Substrate, especially when managing weight in nested functions becomes complicated.

pallets/capacity/src/lib.rs
enddynayn: It sounds like you’ve already encountered an issue where, if a migration wasn’t completed in a previous PR, we end up with three states: Empty, ProviderBoost, and MaximumCapacity. This seems incorrect and could be impacting data integrity, potentially leading to unwanted results.

pallets/capacity/src/lib.rs
enddynayn: By the way, this is a minor comment but we have miscalculated weight on_initialize multiple times.

wilwade: I thought the weights should come from the benchmarks not from here?

enddynayn: Ah, you’re right! I noticed it once but forgot about start_new_reward_era_if_needed. I initially thought we were adding them up manually because of the comments left by each get or insert.

pallets/capacity/src/types.rs
shannonwells: They were being used by a function that was removed, so yea we should get rid of them.

e2e/scaffolding/helpers.ts
```diff
@@ -461,19 +487,26 @@ export async function setEpochLength(keys: KeyringPair, epochLength: number): Pr
...
+  return actualEraLength + eraInfo.startedAt.toNumber() + 1;
+}
+
 export async function getOrCreateGraphChangeSchema(source: KeyringPair): Promise<u16> {
   const existingSchemaId = getGraphChangeSchema();
   if (existingSchemaId) {
     return new u16(ExtrinsicHelper.api.registry, existingSchemaId);
   } else {
-    const { target: createSchemaEvent, eventMap } = await ExtrinsicHelper.createSchemaV3(
+    const op = ExtrinsicHelper.createSchemaV3(
```
shannonwells, line 501: Wondering if this change is real or a rebase mistake.

wilwade: It is fine either way, so OK by me at least.

pallets/capacity/src/lib.rs
aramikm: Not required. Removed

pallets/capacity/src/types.rs
aramikm: done

pallets/capacity/src/types.rs
aramikm: Any reason why you think this needs to be atomic?

pallets/capacity/src/lib.rs
shannonwells: heh, I actually had that about 50 lines up, too

pallets/capacity/src/types.rs
saraswatpuneet: yeah it isnt being shared and modified concurrently ,

e2e/scaffolding/helpers.ts
wilwade: @shannonwells Ended up being this simple, so feel free to grab this.

```suggestion
  const actualEraLength: number = ExtrinsicHelper.api.consts.capacity.eraLength.toNumber();
```

runtime/frequency/src/lib.rs
```diff
@@ -549,6 +565,15 @@ impl pallet_capacity::Config for Runtime {
 	type EpochNumber = u32;
 	type CapacityPerToken = CapacityPerToken;
 	type RuntimeFreezeReason = RuntimeFreezeReason;
+	type EraLength = CapacityRewardEraLength;
+	type ProviderBoostHistoryLimit = ConstU32<30u32>;
```
wilwade, line 569: Shouldn't this refer to the `parameter_types!` value?
```suggestion
	type ProviderBoostHistoryLimit = ProviderBoostHistoryLimit;
```

shannonwells: Yes.

e2e/scaffolding/helpers.ts
shannonwells: 🤦

JoeCap08055 (approved):
- reviewed changes since previous review
- built branch, ran unit & e2e tests

pallets/capacity/src/lib.rs
```diff
@@ -1185,7 +1169,7 @@ impl<T: Config> Pallet<T> {
 		let history_limit: u32 = T::ProviderBoostHistoryLimit::get();
 		let chunk_len = T::RewardPoolChunkLength::get();
 		// Remove one because eras are 1 indexed
-		let era_u32: u32 = era.saturating_sub(One::one()).into();
+		let era_u32: u32 = era;
```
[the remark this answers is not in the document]
shannonwells: fixed

js/api-augment/definitions/capacity.ts
```diff
@@ -0,0 +1,29 @@
...
+          list_unclaimed_rewards: {
+            description: 'List any rewards that can be claimed up to the previous Reward Era',
+            params: [{ name: 'address', type: 'AccountId' }],
+            type: 'Vec<UnclaimedRewardInfo>',
+          },
+        },
+        version: 1,
+      },
+    ],
+  },
```
wilwade, line 28: Don't have to have runtime apis specified anymore! 🎉  (Changed since this was written)
```suggestion
```

pallets/capacity/README.md
```diff
@@ -40,24 +40,45 @@ The Capacity Pallet provides for:
...
+| Get Staking Account Ledger       | Returns information about an account's current staking details                                    | `stakingAccountLedger`      | 1             |
+| Staking Target Ledger            | Returns information about an account's current staking details for a specific target Provider MSA Id | `stakingTargetLedger`       | 1             |
+| Get Unstake Information          | Returns the information about an account's current unstaking details and the unlocking chunks     | `unstakeUnlocks`            | 1             |
+
+### RPCs
+Custom RPCs are not enabled for this pallet. The following RuntimeAPI functions may be accessed by making a state call, for example:
+```javascript
+    const encodedAddr = ExtrinsicHelper.api.registry.createType('AccountId32', booster.address);  // where booster is a polkadot/keyring Keypair type
+    let result = await api.rcp.state.call('CapacityRuntimeApi_list_unclaimed_rewards', encodedAddr);
+    const decodedResult: Vec<UnclaimedRewardInfo> = ExtrinsicHelper.api.registry.createType('Vec<UnclaimedRewardInfo>', result);
```
wilwade, line 75:
```suggestion
    const encodedAddr = api.registry.createType('AccountId32', booster.address);  // where booster is a polkadot/keyring Keypair type
    let result = await api.rcp.state.call('CapacityRuntimeApi_list_unclaimed_rewards', encodedAddr);
    const decodedResult: Vec<UnclaimedRewardInfo> = api.registry.createType('Vec<UnclaimedRewardInfo>', result);
```

runtime/frequency/src/apis.rs
wilwade, line 1: Extra file that needs removing

pallets/capacity/src/lib.rs
```diff
@@ -26,33 +26,42 @@
 	missing_docs
```
wilwade, line 26: Line 4-7 change to add the Runtime API Link:

```
//! ## Quick Links
//! - [Configuration: `Config`](Config)
//! - [Extrinsics: `Call`](Call)
//! - [Runtime API: `CapacityRuntimeApi`](../pallet_capacity_runtime_api/trait.CapacityRuntimeApi.html)
//! - [Event Enum: `Event`](Event)
//! - [Error Enum: `Error`](Error)
```

pallets/capacity/src/lib.rs
```diff
@@ -593,29 +874,52 @@ impl<T: Config> Pallet<T> {
...
 	fn reduce_capacity(
 		unstaker: &T::AccountId,
 		target: MessageSourceId,
 		amount: BalanceOf<T>,
+		staking_type: StakingType,
 	) -> Result<BalanceOf<T>, DispatchError> {
 		let mut staking_target_details = Self::get_target_for(&unstaker, &target)
 			.ok_or(Error::<T>::StakerTargetRelationshipNotFound)?;
+
+		ensure!(amount.le(&staking_target_details.amount), Error::<T>::InsufficientStakingBalance);
```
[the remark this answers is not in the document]
shannonwells: I think this was addressed by the fix that the audit highlighted.  Other than that, last time I tried it I could not make that case happen.

pallets/capacity/src/lib.rs
shannonwells: Pushed to an update. See #2207

pallets/capacity/src/lib.rs
shannonwells: Pushed to an update. See #2207

pallets/capacity/src/lib.rs
```diff
@@ -255,47 +336,96 @@ pub mod pallet {
...
+		CollectionBoundExceeded,
+		/// This origin has nothing staked for ProviderBoost.
+		NotAProviderBoostAccount,
 	}
 
 	#[pallet::hooks]
 	impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {
 		fn on_initialize(current: BlockNumberFor<T>) -> Weight {
 			Self::start_new_epoch_if_needed(current)
+				.saturating_add(Self::start_new_reward_era_if_needed(current))
```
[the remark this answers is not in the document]
shannonwells: I'm making an issue for all these as follow-ups. See #2207

Issue #1696, opened 2023-10-04
Remove without_storage_info from pallets


aramikm:
# Description
Based on the discussions in https://substrate.stackexchange.com/questions/3891/benchmarks-command-error/3894#3894 we should remove `without_storage_info` usages so that we can get accurate PoV benchmarks

Currently its been used in `Messages` and `Schema` pallets

# Acceptence Criterea
- Remove `without_storage_info` from `Messages` and `Schema` pallets and fix all compile issues
- Do migrations from unbounded to bounded types for these pallets
- Rerun benchmarks to include PoV size

# Additional details
Examples for migration and other changes from substrate:
https://github.com/paritytech/polkadot-sdk/issues/323

PR #1702, opened 2023-10-06
Removed without_storage_info from messages and schema pallets

aramikm:
# Goal
The goal of this PR is to remove `without_storage_info` which is necessary to be able to calculate max encoded length for PoV.

Closes 
#1696 

# Changes
- Since we can not submit a `MaxEncodedLen` PoV for messages due to oversized PoV we would need to use `measured` until we reworked the messages pallet.
- Reduced the `MessagesMaxPayloadSizeBytes` from 50K to 3K to reduce the chance of Oversized PoV even for measured mode until the rework of messages pallet.
- Reduced the `MessagesMaxPerBlock` from 7000 to 200 to reduce the chance of Oversized PoV even for measured mode until the rework of messages pallet.

# Checklist
- [X] Chain spec updated
- [X] Weights updated

# Notes
I checked all the submitted messages on rococo and main-net and there are no messages that violate these restrctions
Here is list of all messages on rococo and there are none on main-net. The biggest one has less than 700 bytes

[messages.txt](https://github.com/LibertyDSNP/frequency/files/13033855/messages.txt)

# Merge order
This most probably will get merged after node upgrade PR unless something changes


saraswatpuneet (approved):
Based on description and code review ✅

For off ipfs type messages we would not exceed pov, so I think it's safe to go with it for time being

shannonwells (approved):
I downloaded the branch and ran make test and make start as well.

The two comments I have are non-blocking but that python script does deserve some explanation/comments at the top if you want to add it. It looks pretty generic looking, so I can't tell what it's meant to be used for.

pallets/messages/src/tests/mock.rs
```diff
@@ -83,26 +84,34 @@ parameter_types! {
...
-		MaxMessagePayloadSizeBytes {}
+		MessagesMaxPayloadSizeBytes {}
+	}
+}
+
+impl Encode for MessagesMaxPayloadSizeBytes {}
+
+impl MaxEncodedLen for MessagesMaxPayloadSizeBytes {
+	fn max_encoded_len() -> usize {
+		MessagesMaxPayloadSizeBytes::get() as usize
```
JoeCap08055, line 114: Is this right? Wouldn't the max encoded length of the constant be the max encoded length of a `usize` variable? ie, 4 or 8 bytes (or whatever the SCALE-encoded size of a `usize` is), not the value itself. Or am I misunderstanding here?

scripts/python/get_messages.py
```diff
@@ -0,0 +1,52 @@
+import requests
+
+# Define the endpoint URL and parameters
+# url = 'https://frequency-rococo-rpc.dwellir.com'
+url = 'https://frequency-rpc.dwellir.com'
+
+
+
+# Open the file for appending
+with open('results_mainnet.txt', 'a') as f:
```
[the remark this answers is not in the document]
JoeCap08055: ➕

pallets/messages/src/tests/mock.rs
aramikm: Great catch!

runtime/common/src/constants.rs
```diff
@@ -252,19 +263,39 @@ pub type CollatorKickThreshold =
...
 
 impl Clone for MessagesMaxPayloadSizeBytes {
 	fn clone(&self) -> Self {
 		MessagesMaxPayloadSizeBytes {}
 	}
 }
+
+impl sp_std::fmt::Debug for MessagesMaxPayloadSizeBytes {
+	#[cfg(feature = "std")]
+	fn fmt(&self, _: &mut sp_std::fmt::Formatter) -> sp_std::fmt::Result {
```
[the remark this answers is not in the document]
aramikm: Good Catch!

pallets/messages/src/weights.rs
```diff
@@ -18,13 +18,13 @@
 //! Autogenerated weights for pallet_messages
 //!
 //! THIS FILE WAS AUTO-GENERATED USING THE SUBSTRATE BENCHMARK CLI VERSION 4.0.0-dev
-//! DATE: 2023-10-10, STEPS: `20`, REPEAT: `10`, LOW RANGE: `[]`, HIGH RANGE: `[]`
+//! DATE: 2023-07-07, STEPS: `20`, REPEAT: `10`, LOW RANGE: `[]`, HIGH RANGE: `[]`
```
enddynayn, line 21: Why is the date here outdated?

enddynayn: ahh, I think you might want to add another commit to rerun the benchmarks.

enddynayn:
👍 looks great!

JoeCap08055 (approved):

enddynayn (approved):
stupendous!  💯

PR #1704, opened 2023-10-09
Updates documentation for the conditions of MSA retirement on retire_msa()

rlaferla:
Updates documentation for the conditions of MSA retirement on `retire_msa()`


JoeCap08055 (approved):

runtime/frequency/src/lib.rs
```diff
@@ -249,7 +249,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	spec_name: create_runtime_str!("frequency"),
 	impl_name: create_runtime_str!("frequency"),
 	authoring_version: 1,
-	spec_version: 54,
+	spec_version: 55,
```
JoeCap08055, line 252: Why the spec bump?

rlaferla: The metadata comparison in the GH action is showing a difference and is requiring it.

enddynayn (approved):

runtime/frequency/src/lib.rs
wilwade: This is because extrinsic documentation is part of the metadata.

PR #1735, opened 2023-10-23
feat: add VERSION to ci-base-image

mattheworris:
# Goal
The goal of this PR is to add a version to the `ci-base-image` container when it is published to DockerHub.
`IMAGE_VERSION` is added to the publishing jobs and stored as a `LABEL` in the container itself.
`IMAGE_VERSION` will be baselined as `1.0.0`
`docker inspect` can be used to check the version of a container and should be in-sync with the tag version.

A follow on PR will be used to update the github workflows to use the version number.

# Discussion
The current workflows pull using the defaults, which is the `latest` tag. 
Adding a version will allows us better traceability and better maintenance moving forward.


.github/workflows/publish-dev-ci-base-image.yml
```diff
@@ -48,3 +49,4 @@ jobs:
           file: tools/ci/docker/ci-base-image.dockerfile
           tags: |
             ${{steps.repo_slug.outputs.result}}/${{env.IMAGE_NAME}}:${{env.BRANCH_NAME}}
+            ${{steps.repo_slug.outputs.result}}/${{env.IMAGE_NAME}}:${{env.BRANCH_NAME}}-${{env.VERSION}}
```
mattheworris, line 52: Only publish the `image-version` tag with `merge-pr.yml`

wilwade (approved):

.github/workflows/merge-pr.yml
```diff
@@ -81,6 +81,7 @@ jobs:
     name: Publish CI Base Image
     env:
       IMAGE_NAME: ci-base-image
+      IMAGE_VERSION: 1.0.0
```
enddynayn, line 84: nice

enddynayn (approved):
👍

PR #198, opened 2022-06-21
chore(*): refactor messages

enddynayn:
Refactor "add" extrinsic by creating a helper function, add_messages,
that is used to store messages.

issue-143

PR #1738, opened 2023-10-25
 Messages v2 included migration

aramikm:
# Goal
The goal of this PR is to propose and implement messages v2 compatible with PoV

Closes #198 

# Discussion
- Refactored Messages to minimize used PoV
- Added storage migration (single block)

# Migration Details
- Based on data used in rococo and main-net and calculations we don't need to do a multi-block migration. (only around 15%) of the block is being used.
- Was not able to test with upgrading on local due to getting errors when running relay nodes
- Was able to successfully run try-run-time cli tool against rococo

# Checklist
- [x] Chain spec updated
- [x] Design doc(s) updated
- [x] Tests added
- [x] Benchmarks added
- [x] Weights updated


wilwade (commented):
Read through it. It looks correct and good to me.

As you mentioned in the PR description, it doesn't yet have the rpc changes, but that should be fairly simple changes.

I really like that the only interface changes are the storage queries and for messages we have the custom paginated RPC to handle that.

I at least see no reason not to proceed.

designdocs/message_storage_v2.md
```diff
@@ -0,0 +1,68 @@
...
+ and tertiary keys using [StorageNMap](https://paritytech.github.io/substrate/master/frame_support/storage/trait.StorageNMap.html) data structure provided in Substrate.
+
+### Main Storage types
+- **MessagesV2**
+    - _Type_: `StorageNMap<(BlockNumber, SchemaId, u16), Message>`
+    - _Purpose_: Main structure To store all messages for a certain block number and schema id and
+      index
+- **BlockMetadata**
+    -  _Type_: `StorageValue<_, BlockMetadata`
+    - _Purpose_: Stores temporary (Whitelisted) metadata for that block
```
saraswatpuneet, line 27: nit: term  `BlockMeta' maybe too generic ? It's pretty much message attributes

designdocs/message_storage_v2.md
```diff
@@ -0,0 +1,68 @@
...
+    pub msa_id: u64,                //  Message source account id (the original source of the message)
+    pub index: u16,		    //  Stores index of message in block to keep total order
+}
+
+pub struct BlockMetadata<MaxMessagesPerBlock>
+{
+    ///  schema and the number of messages for that schema in current block
+    pub schema_counts: BoundedVec<SchemaCount, MaxMessagesPerBlock>,
+    ///  Stores index of message in block to keep total order.
+    pub total_index: u16
```
saraswatpuneet, line 46: Is it total count of messages in a block ?

aramikm: yes, it is used for assigning a unique index to each message based on the order that the chain got the message in a block.

designdocs/message_storage_v2.md
aramikm: It is but it's related to the current bock. maybe `BlockMessageMetadata` ? Open to other suggestions

designdocs/message_storage_v2.md
```diff
@@ -0,0 +1,68 @@
...
+
+### On Chain Structure
+Following is a proposed data structure for storing a Message on chain.
+```rust
+// NOT CHANGED
+pub struct Message<AccountId> {
+    pub payload: Vec<u8>,		    //  Serialized data in a user-defined schemas format
+    pub provider_key: AccountId,	    //  Signature of the signer
+    pub msa_id: u64,                //  Message source account id (the original source of the message)
+    pub index: u16,		    //  Stores index of message in block to keep total order
```
enddynayn, line 38: Remind me how we currently using this index?

enddynayn: total ordering

pallets/messages/src/lib.rs
```diff
@@ -117,29 +124,27 @@ pub mod pallet {
...
+		StorageValue<_, BlockMetadata<T::MaxMessagesPerBlock>, ValueQuery>;
 
-	/// A temporary storage for getting the index for messages
-	/// At the start of the next block this storage is set to 0
 	#[pallet::storage]
-	#[pallet::whitelist_storage]
-	#[pallet::getter(fn get_message_index)]
-	pub(super) type BlockMessageIndex<T: Config> = StorageValue<_, u16, ValueQuery>;
+	#[pallet::getter(fn get_messages_v2)]
+	pub(super) type MessagesV2<T: Config> = StorageNMap<
```
enddynayn, line 138: nice!

pallets/messages/src/lib.rs
```diff
@@ -344,28 +351,39 @@ impl<T: Config> Pallet<T> {
...
-				let index = BlockMessageIndex::<T>::get();
-				let msg = Message {
-					payload, // size is checked on top of extrinsic
-					provider_msa_id,
-					msa_id,
-					index,
-				};
+		let mut metadata = MessageBlockMetadata::<T>::get();
+		ensure!(
+			u32::from(metadata.total_index) < T::MaxMessagesPerBlock::get(),
```
enddynayn, line 356: With this new structure why do we have a maximum number of messages per block?

wilwade: I think we had a max before as well... Yup:

```
#[pallet::storage]
	#[pallet::getter(fn get_messages)]
	pub(super) type Messages<T: Config> = StorageDoubleMap<
		_,
		Twox64Concat,
		BlockNumberFor<T>,
		Twox64Concat,
		SchemaId,
		BoundedVec<Message<T::MessagesMaxPayloadSizeBytes>, T::MaxMessagesPerBlock>,
		ValueQuery,
	>;
```

enddynayn: I noticed that but we changed the data structure. I am curious why we still need it with this new data structure.

enddynayn: If my understanding is wrong about why the pov grew then could we have left messages as is and only updated schemas to reduce the pov size of reading them?

aramikm: We have to update all of them to bring down PoV

aramikm: About the `MaxMessagesPerBlock` check in that particular line we don't need it, only argument for adding it was to exactly get the same response of the old extrinsic. But i can be removed if everyone onboard

pallets/messages/src/weights.rs
```diff
@@ -56,78 +56,78 @@ pub trait WeightInfo {
...
+	/// Proof: `Messages::MessagesV2` (`max_values`: None, `max_size`: Some(3125), added: 5600, mode: `MaxEncodedLen`)
 	fn add_ipfs_message() -> Weight {
 		// Proof Size summary in bytes:
-		//  Measured:  `6183`
-		//  Estimated: `18558`
-		// Minimum execution time: 36_000_000 picoseconds.
-		Weight::from_parts(174_485_000, 18558)
-			.saturating_add(T::DbWeight::get().reads(3_u64))
+		//  Measured:  `915`
+		//  Estimated: `77893`
```
shannonwells, line 88: what is making the estimates so far off?

pallets/messages/src/lib.rs
```diff
@@ -394,12 +412,14 @@ impl<T: Config> Pallet<T> {
...
-					.into_inner()
-					.iter()
-					.map(|msg| msg.map_to_response(block_number_value, schema_payload_location))
-					.collect(),
+			_ => {
+				let mut messages: Vec<_> =
+					<MessagesV2<T>>::iter_prefix_values((block_number, schema_id))
+						.map(|msg| msg.map_to_response(block_number_value, schema_payload_location))
+						.collect();
+				messages.sort_by(|a, b| a.index.cmp(&b.index));
```
shannonwells, line 420: nice one

pallets/messages/src/lib.rs
```diff
@@ -117,29 +124,27 @@ pub mod pallet {
...
-		BlockNumberFor<T>,
-		Twox64Concat,
-		SchemaId,
-		BoundedVec<Message<T::MessagesMaxPayloadSizeBytes>, T::MaxMessagesPerBlock>,
-		ValueQuery,
-	>;
+	#[pallet::whitelist_storage]
+	#[pallet::getter(fn get_block_metadata)]
+	pub(super) type MessageBlockMetadata<T: Config> =
+		StorageValue<_, BlockMetadata<T::MaxMessagesPerBlock>, ValueQuery>;
```
enddynayn, line 134: I think we should consider using 
https://crates.parity.io/frame_support/storage/bounded_btree_set/struct.BoundedBTreeSet.html

I think it will simplify the code a little.

pallets/messages/src/lib.rs
enddynayn: I think if we can get rid of code that is not needed it's probably good code hygiene to remove it.

pallets/messages/src/lib.rs
```diff
@@ -344,28 +351,39 @@ impl<T: Config> Pallet<T> {
...
-				};
+		let mut metadata = MessageBlockMetadata::<T>::get();
+		ensure!(
+			u32::from(metadata.total_index) < T::MaxMessagesPerBlock::get(),
+			Error::<T>::TooManyMessagesInBlock
+		);
+		let mut found = false;
+		let mut current_index: MessageIndex = 0;
+		for schema_count in metadata.schema_counts.iter_mut() {
+			if schema_count.schema_id == schema_id {
```
saraswatpuneet, line 362: Nit: You can use `find`
metadata.schema_counts.iter_mut().find(.....

node/cli/src/command.rs
```diff
@@ -371,22 +371,24 @@ pub fn run() -> Result<()> {
...
-						.map_err(|e| sc_cli::Error::Service(sc_service::Error::Prometheus(e)))?;
-				Ok((
-					cmd.run::<Block, ExtendedHostFunctions<
-						sp_io::SubstrateHostFunctions,
-						<ExecutorDispatch as NativeExecutionDispatch>::ExtendHostFunctions,
-					>>(),
-					task_manager,
-				))
+
+			type HostFunctions =
```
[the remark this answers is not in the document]
aramikm: The `try-runtime` would not compile if we don't add it. it was simply not included in previous upgrades.

pallets/messages/src/lib.rs
```diff
@@ -344,28 +351,39 @@ impl<T: Config> Pallet<T> {
...
+				found = true;
+				current_index = schema_count.count;
+				schema_count.count = schema_count.count.saturating_add(1);
+			}
+		}
+		if !found {
+			metadata
+				.schema_counts
+				.try_push(SchemaCount { count: 1, schema_id })
+				.map_err(|_| Error::<T>::TooManyMessagesInBlock)?;
```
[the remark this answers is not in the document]
enddynayn: The whitelist would still remain what we would still have an index and that index would be a counter for each message. I think this would work because we never access messages on the chain. In the case that we did would provide all the keys required to access that message. 

There are tradeoffs in that we might be pushing logic to off-chain, however, since the messages themselves contain an index relative ordering for each schema message is still possible.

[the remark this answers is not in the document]
aramikm: Applied

pallets/messages/src/lib.rs
aramikm: Not applicable anymore

pallets/messages/src/lib.rs
aramikm: No applicable anymore

pallets/messages/src/migration/v2.rs
```diff
@@ -0,0 +1,133 @@
...
+		pub provider_msa_id: MessageSourceId,
+		///  Message source account id (the original source).
+		pub msa_id: Option<MessageSourceId>,
+		///  Stores index of message in block to keep total order.
+		pub index: u16,
+	}
+
+	/// old permanent storage for messages mapped by block number and schema id.
+	#[storage_alias]
+	pub(crate) type Messages<T: Config> = StorageDoubleMap<
```
shannonwells, line 40: How does storage alias know what prefix to use for the old storage? Going back over this I expected something like `#[crate::storage_alias(pallet_name)]`

aramikm: It gets it from the name of the Storage in this case `Messages<T: Config>`. This name should stay the same

aramikm: I tried it with try-run-time and it worked, will add a test to double verify.

shannonwells: I think it will look for the same prefix since it's still inside the crate.

pallets/messages/src/weights.rs
aramikm: It's because of the mode of benchmark changed from `Measured` to `MaxEncodedLen`. This increase is temporary since the PR for Schema changes is not merged we are not seeing the effect of that here yet.

enddynayn (approved):
🚀 marvelous! 👍

pallets/messages/src/migration/v2.rs
aramikm: Added a test and worked as expected. As mentioned the `storage-name `comes from the name of type and the `pallet-name` comes from the generic params Pallet<T>

mattheworris (approved):
Nice work!
[read-through the code, make ci-local passed]

JoeCap08055 (approved):

PR #1741, opened 2023-10-25
fix: update container to ci-base-image:1.0.0

### LINKED ISSUE #1734: ci-base-image should be versioned
When building the container for `ci-base-image` a version should be used.
This is helpful to isolate changes from all of CI when the image needs to be updated.
The following steps must be followed in order to affect the necessary changes.

- [x] Add `VERSION` to `env` when building the container in `publish-dev-ci-base-image.yml`
- [x] Add `LABEL version=${VERSION} to `ci-base-image.dockerfile`
- [x] Add `${{env.VERSION}}` to the image tag in `publish-dev-ci-base-image.yml`
- [x] Commit change to new branch to update the `VERSION` to `1.0.0`
- [x] Update workflows that use `ci-base-image` to use version `1.0.0`

**Future changes to `ci-base-image.dockerfile` should increment the `VERSION` in all github workflows and the container itself.**

mattheworris:
# Goal
This is part 2 of Issue #1734.
This PR will update the github workflows to reference the new versioned `ci-base-image`.

Closes #1734 
# Discussion
- Why not use a variable for `container.image` so that future changes to `ci-base-image:1.0.0` are easier?
  - In github jobs, the global `env` context is not available. Each job can set it's own `env` which can be passed to its actions and containers. However, at time of writing, there was no easy way to share a variable across jobs.
  - The container version is most likely to change during a polkadot-sdk upgrade. Notes have been added here: https://github.com/LibertyDSNP/frequency/wiki/Upgrading-to-a-new-Polkadot-Release

# Checklist
- [x] Tests
- `spec_version` was modified to trigger CI (`verify-pr-commit.yml`) and then reverted.
- `merge-pr.yml`, `rococo.yml`, and `release.yml`, are not easily tested, but little risk is expected with these changes.

### DIFF
--- a/.github/workflows/merge-pr.yml
+++ b/.github/workflows/merge-pr.yml
@@ -11,7 +11,7 @@ jobs:
   changes:
     name: Determine Changed Files
     runs-on: ubuntu-20.04
-    container: ghcr.io/libertydsnp/frequency/ci-base-image
+    container: ghcr.io/libertydsnp/frequency/ci-base-image:1.0.0
     outputs:
       ci-base-image: ${{steps.filter.outputs.ci-base-image}}
     steps:
@@ -28,7 +28,7 @@ jobs:
   publish-js-api-augment-rc:
     name: Merge - Publish JS API Augment Release Candidate
     runs-on: [self-hosted, Linux, X64, build, v2]
-    container: ghcr.io/libertydsnp/frequency/ci-base-image
+    container: ghcr.io/libertydsnp/frequency/ci-base-image:1.0.0
     steps:
       - name: Check Out Repo
         uses: actions/checkout@v4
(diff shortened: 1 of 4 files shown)


.github/workflows/merge-pr.yml
```diff
@@ -11,7 +11,7 @@ jobs:
   changes:
     name: Determine Changed Files
     runs-on: ubuntu-20.04
-    container: ghcr.io/libertydsnp/frequency/ci-base-image
+    container: ghcr.io/libertydsnp/frequency/ci-base-image:1.0.0
```
wilwade, line 14: Would it be worth making this an Actions variable instead? I guess an update to the version always requires a code change anyway, but it would be easier to roll back. Hmm... Simple and explicit is perhaps better.

wilwade (approved):

.github/workflows/merge-pr.yml
wilwade: I wouldn't worry about it then. This a good solution to 🚢 I think.

PR #1743, opened 2023-10-25
schemas: PoV compatible changes

### LINKED ISSUE #1742: Schemas optimized for time and PoV
# Description
We need to redo schema pallet so it is optimized for Both time and PoV

# Acceptence Criterea
- [ ]  Create Design Doc for new storages
- [ ]  Modify existing pallet to work with new storages
- [ ]  Modify Tests
- [ ]  Run benchmarks
- [ ]  Add Migration
- [ ]  Migration Tests

aramikm:
# Goal
The goal of this PR is to split schemas and the model into 2 separate storages so that we can limit the size of PoV being accessed from other pallets.

Closes #1742 

# Checklist
- [x] Chain spec updated
- [x] Design doc(s) updated
- [x] Tests added
- [x] Benchmarks added
- [x] Weights updated


pallets/schemas/src/lib.rs
```diff
@@ -184,16 +188,24 @@ pub mod pallet {
 	pub(super) type CurrentSchemaIdentifierMaximum<T: Config> =
 		StorageValue<_, SchemaId, ValueQuery>;
 
+	/// Storage for message schema info struct data
+	/// - Key: Schema Id
+	/// - Value: [`SchemaInfo`](SchemaInfo)
+	#[pallet::storage]
+	#[pallet::getter(fn get_schema_info)]
+	pub(super) type SchemaInfos<T: Config> =
```
enddynayn, line 196: Minor and not a blocker. How about `SchemasInfo` since it's not about information about one schema? Or just `SchemaInfo`. I kinda like `SchemaInfo` without the s.

enddynayn: ahhh I see. There is a struct called SchemaInfo. dang.

pallets/schemas/src/lib.rs
```diff
@@ -184,16 +188,24 @@ pub mod pallet {
...
+
 	/// Storage for message schema struct data
 	/// - Key: Schema Id
-	/// - Value: [`Schema`](Schema)
+	/// - Value: [`BoundedVec`](BoundedVec<T::SchemaModelMaxBytesBoundedVecLimit>)
 	#[pallet::storage]
-	#[pallet::getter(fn get_schema)]
-	pub(super) type Schemas<T: Config> = StorageMap<
+	#[pallet::getter(fn get_schema_payload)]
+	pub(super) type SchemaPayloads<T: Config> = StorageMap<
```
enddynayn, line 204: minor: Ok. I think we can drop the `s` here and call it `SchemaPayload`.

enddynayn (approved):
awesome! 👍

node/cli/src/command.rs
```diff
@@ -371,22 +371,22 @@ pub fn run() -> Result<()> {
 
 		#[cfg(feature = "try-runtime")]
 		Some(Subcommand::TryRuntime(cmd)) => {
```
enddynayn, line 373: I have not used this. For my learning, can you explain what this does?

aramikm: This allows trying to simulate a runtime upgrade (which include storage migrations) connecting to a chain. Some more details in https://paritytech.github.io/try-runtime-cli/try_runtime/

designdocs/schema_v2.md
```diff
@@ -0,0 +1,41 @@
+# On Chain Message Storage
```
JoeCap08055, line 1: Might be nice to update this doc with some numbers, ie before/after PoV weights. Maybe add a placeholder section for now?

pallets/schemas/src/lib.rs
```diff
@@ -85,6 +87,8 @@ pub use weights::*;
 
 mod serde;
 
+const LOG_TARGET: &str = "runtime::schemas";
```
JoeCap08055, line 90: 💯

pallets/schemas/src/lib.rs
```diff
@@ -383,7 +395,7 @@ pub mod pallet {
 			<CurrentSchemaIdentifierMaximum<T>>::set(n);
 		}
 
-		/// Build the [`Schema`] and insert it into storage
+		/// Build the [`SchemaInfo`] and insert it into storage
```
JoeCap08055, line 398: This description is not quite correct. We add _both_ the `SchemaInfo` and the `SchemaPayload` to storage; it's just now they're separate storage.

pallets/schemas/src/lib.rs
```diff
@@ -398,23 +410,46 @@ pub mod pallet {
...
+			let schema_info = SchemaInfo { model_type, payload_location, settings: set_settings };
 			<CurrentSchemaIdentifierMaximum<T>>::set(schema_id);
-			<Schemas<T>>::insert(schema_id, schema);
+			<SchemaInfos<T>>::insert(schema_id, schema_info);
+			<SchemaPayloads<T>>::insert(schema_id, model);
 			Ok(schema_id)
 		}
 
 		/// Retrieve a schema by id
 		pub fn get_schema_by_id(schema_id: SchemaId) -> Option<SchemaResponse> {
```
JoeCap08055, line 421: Hmm, I'm thinking there are use cases where you _only_ want to get the model/payload, and not the other info... perhaps add another method & RPC `get_schema_model_by_id`. Then we could decide whether we need this one, depending on our analysis of use cases. I think in a lot of use cases you only need one or the other.

JoeCap08055:
Looks pretty good. Perhaps add/update e2e tests for all pallets to use the new RPC?

pallets/schemas/src/tests/other_tests.rs
```diff
@@ -698,3 +703,59 @@ fn create_schema_with_append_only_setting_and_non_itemized_should_fail() {
...
+				v2::old::Schema {
+					model_type: ModelType::AvroBinary,
+					payload_location: PayloadLocation::OnChain,
+					settings: SchemaSettings::all_disabled(),
+					model: BoundedVec::try_from(fields.as_bytes().to_vec())
+						.expect("should have value"),
+				},
+			);
+		}
+		let old_schema_1 = v2::old::Schemas::<Test>::get(1u16).expect("should have value");
```
shannonwells, line 737: nit/non-blocking: Though I trust that they're there, this test would be improved if it verified that all three schemas were in old::Schemas.

pallets/schemas/src/lib.rs
shannonwells: +1

shannonwells (approved):
Looking good now

pallets/schemas/src/lib.rs
aramikm: I think either we need to keep all of them plural or singular. Based on what we have in some pallets we have singulars like in Msa or handles and in other places plural like in Messages and time-release

pallets/schemas/src/lib.rs
aramikm: I think there is confusion going on about RPCs and methods for this. Right now we have an RPC that returns the `SchemaResponse` which is the combination of payload and metadata. I don't think there is any need for an RPC that only returns payload since we don't need to optimize queries that we expose to RPC since all of it is considered off-chain processing. Adding an exposed method that only returns payload would only make sense if there was any need to query that on-chain which currently we don't.

JoeCap08055: True, maybe that's the subject of a separate issue/PR that is focused on RPC network efficiency for client interactions.

enddynayn (approved):
👍 fantastic!

designdocs/schema_v2.md
```diff
@@ -39,3 +39,8 @@ pub struct SchemaInfo {
     pub settings: SchemaSettings,
 }
 ```
+### Expected PoV improvements
+This PoV improvement would not extrinsic weights in this pallet, but it would directly affect any
```
shannonwells, line 43:
```suggestion
This PoV improvement would not _____? extrinsic weights in this pallet, but it would directly affect any
```

mattheworris (approved):
Solid. 
:shipit: 

[Read through code, ran tests, 
caveat: I don't have much experience with storage migration, but nothing smells off]

designdocs/schema_v2.md
```diff
@@ -0,0 +1,46 @@
...
+    pub payload_location: PayloadLocation,
+    /// additional control settings for the schema
+    pub settings: SchemaSettings,
+}
+```
+### Expected PoV improvements
+This PoV improvement would not affect extrinsic weights in this pallet, but it would directly affect any
+pallet that is dependent on **Schemas** pallet. Some of these pallets are **Messages** and
+**Stateful-Storage**. After these changes we are expecting see to see around 30-60KiB decrease in PoV
+for `MaxEncodedLen` mode.
```
saraswatpuneet, line 46: Nice

saraswatpuneet (approved):
Nice work!

pallets/schemas/src/lib.rs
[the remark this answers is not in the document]
aramikm: So I've looked into [CountedStorageMap](https://github.com/paritytech/polkadot-sdk/blob/18257373b313fcfad4f7a2ab2cb02cb8bfc5b8f1/substrate/frame/support/src/storage/types/counted_map.rs#L198-L203) again. The implementation is very generic and not efficient. For every insert it checks if the key exists and then does the increment. Since for our use-case we know that every key would be new so keeping them separate is more efficient.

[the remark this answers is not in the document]
wilwade: Sounds good. Let's keep it as is for now.

PR #1745, opened 2023-10-30
Interval Sealing support for `engine_createBlock`

wilwade:
# Goal
The goal of this PR is enable `engine_createBlock` for Interval Sealing mode.

Effectively it uses the should make empty block coming down the command instead of making a special exception for interval sealing.

Note: This does reduce the logs in other modes when `engine_createBlock` is called with not making an empty block and no transactions are in the pool.

# Test
- `make start-interval` (or the new 1 second make command `make start-interval-1`)
- Wait for it to start
- `make local-block` now works (it will would before this PR)


node/service/src/block_sealing.rs
```diff
@@ -251,9 +247,8 @@ pub async fn run_seal_command<B, BI, CB, E, C, TP, SC, CS, CIDP, P>(
 	while let Some(command) = commands_stream.next().await {
 		match command {
 			EngineCommand::SealNewBlock { create_empty, finalize, parent_hash, sender } =>
-				if sealing_mode != SealingMode::Interval ||
-					sealing_create_empty_blocks ||
-					pool.ready().count() > 0
+				// To keep the logs quieter, don't even attempt to create empty blocks unless there are transactions in the pool
+				if create_empty || pool.ready().count() > 0
```
wilwade, line 251: This is the important logic change!

scripts/init.sh
```diff
@@ -120,6 +120,34 @@ start-frequency-interval)
...
+  if [ "$2" == "purge" ]; then
+    echo "purging parachain..."
+    rm -rf $parachain_dir
+  fi
+
+  ./target/debug/frequency \
+    --dev \
+    -lruntime=debug \
+    --sealing=interval \
+    --sealing-interval=1 \
```
wilwade, line 139: Just a duplicate of the other one with the one second time set.

saraswatpuneet (approved):

node/service/src/block_sealing.rs
rlaferla: How is this different from specifying to not create empty blocks?

enddynayn:
Can you please add clarification to the commit message? I am a little unclear on what problem we are trying to solve.

wilwade:
> Can you please add clarification to the commit message? I am a little unclear on what problem we are trying to solve.

I'm not sure how to make it more clear than supporting the engine_createBlock rpc in the interval sealing mode. Ideas?

enddynayn:
> @enddynayn
> 
> > Can you please add clarification to the commit message? I am a little unclear on what problem we are trying to solve.
> 
> I'm not sure how to make it more clear than supporting the engine_createBlock rpc in the interval sealing mode. Ideas?

We did not have it before why do we need it now?

wilwade:
> > @enddynayn
> > > Can you please add clarification to the commit message? I am a little unclear on what problem we are trying to solve.
> > 
> > 
> > I'm not sure how to make it more clear than supporting the engine_createBlock rpc in the interval sealing mode. Ideas?
> 
> We did not have it before why do we need it now? What triggers this to be needed? Is this a bug? Are we adding a new test that requires this for some test we plan to add in the future?

Ah:
1. Foremost: It was expected to work, but didn't.
2. Trigger: I am thinking about using interval sealing for e2e tests (which requires `engine_createBlock` to work)

Or to put another way, I expected to be able to run the e2e tests against interval sealing and was surprised when it didn't work.

aramikm (approved):
Looks good! 

I think our current **interval-mode**  does not capture all the scenarios that can happen when submitting an extrinsic since it starts an interval with submission. Ideally we would like an independent block interval that would start independently and only create a block (if any extrinsic inside). this would allow us to be able to get more realistic result let's say for extrinsic that got submitted close to the end of current interval and they might get included in the next one and ...

PR #1757, opened 2023-11-03
Stateful storage pov optimization

### LINKED ISSUE #1782: Copy of Messages V2 optimized for time and PoV
# Description
We need to redo messages pallet so it is optimized for Both time and PoV

# Acceptence Criterea
- [x]  Create Design Doc for new storages
- [x]  Modify existing pallet to work with new storages
- [x]  Modify Tests
- [x]  Run benchmarks
- [x]  Add Migration
- [x]  Migration Tests

aramikm:
# Goal
The goal of this PR is to evaluate and minimize PoV consumption by `stateful-storage` pallet.

Closes #1782 

# Discussion
- refactored benchmarks to calculate the max of time and PoV
- decreased `additional-trie-layers` number
- decreased  `MaxItemizedPageSizeBytes ` from 64KiB to around 10KiB (will allow around 292 of 32 bytes itemized public keys)

# Improvements
- by applying mentioned changes the PoV for `apply_item_actions` got reduced from **45KB** to **15KB**
- by applying mentioned changes the PoV for `upsert` and `delete` pages got reduced from **12KB** to **6KB**

# Future improvements
- by using PoV clawback we can further reduce the PoV sizes. A [PR](https://github.com/paritytech/polkadot-sdk/pull/1462) to track

# Checklist
- [x] Benchmarks added
- [x] Weights updated


scripts/run_benchmarks.sh
```diff
@@ -183,7 +183,7 @@ function run_benchmark() {
   --chain="frequency-bench" \
   --heap-pages=4096 \
   --wasm-execution=compiled \
-  --additional-trie-layers=20 \
+  --additional-trie-layers=5 \
```
aramikm, line 186: Question: why `5`?

Answer: 

So my understanding is that during PoV benchmarks we have two steps of computation
1- PoV from Main root node to the root of desired storage map.
2- PoV for storageMap

`additional-trie-layers` is for the first step and it's the LOG16(total number of storage items). By default this number should be between 2-3 and we don't need to set that. But since we are using child trees and it seems like child-trees are not getting benchmarked the same way that StorageMaps might have, **to be cautious** we are setting this to 5 which means it should be fine for at least 1_000_000 child trees. in case the number of child trees passed 1 Million this might need to be bumped to 6.

 for second step if `worst-case-map-values` is not set it will use 1_000_000 as the number of items in map.

JoeCap08055 (commented):
Looks pretty good. A couple of questions:
* Do we really want to run _all_ benchmarks? Seems like we could omit unrelated benchmark files.
* Am I reading it wrong, or did some of the weights in `stateful-storage` _increase_?

aramikm:
> Looks pretty good. A couple of questions:
> 
> * Do we really want to run _all_ benchmarks? Seems like we could omit unrelated benchmark files.

Yes we do since changing the  ` --additional-trie-layers` to 5 decreases PoV for all pallets across the board

> * Am I reading it wrong, or did some of the weights in `stateful-storage` _increase_?

So PoV for all of them should be decreased. 
Since the underlying logic is the same the ref-times should not have a significant change. Other than that I improved benchmarks for itemized related extrinsics to be more accurate.

Please let me know which one you think increased more than regular benchmark fluctuations.

shannonwells (approved):
Non-blocking comments only - 

I did not look through all of the weights files but I did look through about a dozen and yes it seems that those changes are within the normal fluctuation of actual weights and error margins, whereas PoV is dramatically reduced.  I do wonder about the child trie estimate though, how did you come up with the estimate of 1M child tries and why do you think that is a good estimate for stateful storage?  

Is there a drawback of giving enough ceiling for 10M tries i.e. would it be too costly to overestimate the PoV, so does it make more sense to increase only when needed?

aramikm:
> I did not look through all of the weights files but I did look through about a dozen and yes it seems that those changes are within the normal fluctuation of actual weights and error margins, whereas PoV is dramatically reduced. I do wonder about the child trie estimate though, how did you come up with the estimate of 1M child tries and why do you think that is a good estimate?
>

The idea is that since child tries are different from regular StorageMaps in the sense that StorageMaps are bottom-heavy. Meaning, the number of nodes from `root of the tree` to the `root of the storage` is usually less than the number of nodes between the `root of the storage` to the `last inserted node in map` on average (Considering we have a map with around at least 1 million items). In child tries the path from `root of the tree` to the `root of a child trie` would require more node traversal compared to the path from `root of child trie` to a `member` (Considering each child trie's inserted items are low).
In another words for regular StorageMaps, step # 1 of PoV calculation mentioned in one of comments are much smaller  compared to step # 2 but this is not the case for child tries.

Having the above explanation to make sure our PoV calculations are considering this fact we need to ensure that step # 1 of PoV calculations are going to consider the max number of existing child tries for benchmarks. Since on a balanced tree in substrate we will need to access Log16(`X`) nodes for `X` inserted items, if we consider having 1_000_000 child tries for benchmarks we will get `Log16(1_000_000) ~ 5`


> Is there a drawback of giving enough ceiling for 10M tries i.e. would it be too costly to overestimate the PoV, so does it make more sense to increase only when needed?

A number of things to consider

- it would be less efficient, a `+1` change in that parameter would increase all PoVs in all pallets by around `+600` bytes.
- We would probably need to only isolate this --additional-trie-layers = 5 or more for only benchmarks running for stateful-storage in future as an optimization. It doesn't need to affect all pallets.
- This is my current understanding of how everything work based on a number of experiments. To 100% make sure how they handle (or don't handle) child tires we will need to dive deep into the PoV calculation code.

aramikm:
1 Million child trees was a guesstimate based on the number of users on chain. Considering we would need to update our benchmark params every 4-6 months if we reached that point can bump it along with other benchmark calculations like database access benchmarks and etc which should be based on the data inside merkle tree.

PR #1765, opened 2023-11-07
E2E Testing fixes and cleanup

### LINKED ISSUE #1731: e2e-tests inconsistent
**Description**
e2e tests seem to pause when starting a node and then running `make e2e-tests`.
When running `make e2e-tests` it sometimes outputs `1010: Invalid Transaction: Transaction is outdated`.

wilwade:
# Goal
The goal of this PR is the final misc pieces of getting the e2e tests less flaky. Does this fix all of it? No. Most? Yes.

Closes #1731 

# Discussion
- Correct batch error asserting
- Ids used as "bad" should use close to the max
- Use Immortal Eras due to issues with speed and [AncientBirthBlock](https://substrate.stackexchange.com/questions/10411/fast-block-production-causes-mortality-checking-failing-with-ancientbirthblock)
- Cleanup before funding
- Move assertAddNewKey
- Add new assertHasMessage

### DIFF
--- a/e2e/capacity/transactions.test.ts
+++ b/e2e/capacity/transactions.test.ts
@@ -38,7 +38,8 @@ import {
   generatePaginatedUpsertSignaturePayloadV2,
   generatePaginatedDeleteSignaturePayloadV2,
   getCapacity,
-  getTestHandle
+  getTestHandle,
+  assertHasMessage
 } from "../scaffolding/helpers";
 import { FeeDetails } from "@polkadot/types/interfaces";
 import { ipfsCid } from "../messages/ipfs";
@@ -240,8 +241,7 @@ describe("Capacity Transactions", function () {
               page_size: 999
             }
           );
-          const response: MessageResponse = get.content[get.content.length - 1];
-          assert.equal(response.payload, "0xdeadbeef", "payload should be 0xdeadbeef");
+          assertHasMessage(get, x => x.payload.isSome && x.payload.toString() === "0xdeadbeef");
         });
       });
 
--- a/e2e/capacity/transactions.test.ts
+++ b/e2e/capacity/transactions.test.ts
@@ -110,6 +110,7 @@ describe("Capacity Transactions", function () {
         before(async function () {
           capacityKeys = createKeys("CapacityKeys");
           capacityProvider = await createMsaAndProvider(fundingSource, capacityKeys, "CapacityProvider", FUNDS_AMOUNT);
+          // Stake enough for all transactions
           await assert.doesNotReject(stakeToProvider(fundingSource, capacityKeys, capacityProvider, stakedForMsa));
         })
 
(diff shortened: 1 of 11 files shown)


e2e/capacity/transactions.test.ts
```diff
@@ -60,16 +62,6 @@ describe("Capacity Transactions", function () {
         assert.notEqual(schemaId, undefined, "setup should populate schemaId");
       });
 
-      async function assertAddNewKey(capacityKeys: KeyringPair, addKeyPayload: AddKeyData, newControlKeypair: KeyringPair) {
```
wilwade, line 63: Moved to helpers

e2e/capacity/transactions.test.ts
```diff
@@ -254,8 +245,6 @@ describe("Capacity Transactions", function () {
         before(async function () {
           capacityKeys = createKeys("CapacityKeys");
           capacityProvider = await createMsaAndProvider(fundingSource, capacityKeys, "CapacityProvider", FUNDS_AMOUNT);
-        })
-        beforeEach(async function () {
           await assert.doesNotReject(stakeToProvider(fundingSource, capacityKeys, capacityProvider, amountStaked));
```
wilwade, line 248: No need to stake before each

e2e/miscellaneous/utilityBatch.test.ts
```diff
@@ -44,11 +43,10 @@ describe("Utility Batch Filtering", function () {
         const batchAll = ExtrinsicHelper.executeUtilityBatchAll(sender, badBatch);
         try {
             await batchAll.fundAndSend(fundingSource);
+            assert.fail("batchAll should have caused an error");
         } catch (err) {
-            error = err;
-            assert.notEqual(error, undefined, " batchAll should return an error");
+            assert.notEqual(err, undefined, " batchAll should return an error");
         }
-        assert.notEqual(error, undefined, " batchAll should return an error");
```
wilwade, line 51: This is the best way to do an error assertion.

e2e/msa/msaKeyManagement.test.ts
```diff
@@ -155,7 +157,7 @@ describe("MSA Key management", function () {
             assert.notEqual(event, undefined, 'should have added public key');
 
             // Cleanup
-            await assert.doesNotReject(ExtrinsicHelper.deletePublicKey(keys, thirdKey.publicKey).signAndSend());
+            await assert.doesNotReject(ExtrinsicHelper.deletePublicKey(keys, thirdKey.publicKey).signAndSend('current'));
```
wilwade, line 160: Important fix. `deletePublicKey` is a free transaction, so doesn't use up a nonce.

e2e/scaffolding/rootHooks.ts
```diff
@@ -15,20 +15,21 @@ export const mochaHooks = {
...
   async afterAll() {
+    const testSuite = this.test.parent.suites[0].title;
+    console.log("Starting ROOT hook shutdown", testSuite)
     try {
       // Any key created using helpers `createKeys` is kept in the module
       // then any value remaining is drained here at the end
       const rootAddress = getRootFundingSource().keys.address;
       await drainFundedKeys(rootAddress);
-      await ExtrinsicHelper.api.disconnect();
-      await ExtrinsicHelper.apiPromise.disconnect();
```
wilwade, line 29: No need to wait for this

mattheworris (approved):
LGTM
- Read through code
- Executed `make e2e-tests` multiple times against a running node (`make start`)

e2e/miscellaneous/utilityBatch.test.ts
JoeCap08055: Just curious, any objection to using `assert.rejects`?
https://nodejs.org/api/assert.html#assertrejectsasyncfn-error-message

Also, I wonder if at some point (not now) we should switch these to Jest (as all of our other Node projects are using it for testing)?

JoeCap08055: Or, if we want to keep using Mocha here, I'd be in favor of eventually swapping out the standard Node `assert` package for something a bit more capable. But definitely not a priority item.

JoeCap08055 (approved):

e2e/miscellaneous/utilityBatch.test.ts
wilwade: Ah right. Assert rejects likely would be better. I was just updating it to still match the existing pattern, but that would be better. Next time.

wilwade: Re: Jest

We might be able to, but I wasn't willing to make the switch yet.

PR #1779, opened 2023-11-13
chore: Replace Currency->fungible migration for capacity pallet

### LINKED ISSUE #942: [Capacity] replace currency trait with fungible trait
After talking speaking with Parity they are going to replace the currency trait with fungible trait.

### LINKED ISSUE #1532: Update .vscode/launch.json to source map the rust standard library.
Update .vscode/launch.json to source map the rust standard library.

mattheworris:
# Goal
The goal of this PR is to replace the `Currency` trait with the `fungible` trait in the `capacity` pallet.

Closes #942 
Closes #1532 

# Discussion
The following Parity issues/PRs were used as references for changes:
[Deprecate Currency - PR 12951](https://github.com/paritytech/substrate/pull/12951) (Explanation of necessary changes.)
[FRAME: Move pallets over to use fungible traits](https://github.com/paritytech/polkadot-sdk/issues/226) (Issue to track Parity's efforts to update their pallets.)
[pallet vesting / update to use fungible](https://github.com/paritytech/polkadot-sdk/pull/1760/files) (Example of some necessary changes.)
[Fungibles: migrate Democracy pallet](https://github.com/paritytech/polkadot-sdk/pull/1861) (Example of storage migration for Locks->Freezes.)

# Changes
- `.cargo-deny.toml` Added "multiformats" for `cid` crate to fix warning.
-  `vscode/settings.json` Added script that sets up the source map for the rust std library files for debugging.
-  `Cargo.lock` cargo updated.
-  `e2e/package-lock.json` npm updated.
-  Replaced traits as needed: Use `tokens::fungible::` 
    - `InspectFungible` for `balance()` and `reducible_balance()`
    - `InspectFreeze` for `balance_frozen()`
    - `Mutate` for `set_balance()`, `mint_into()`
    - `MutateFreeze` for `set_freeze()`, and `thaw()`
- Added `pub enum FreezeReason` to support `freezes`
- Updated error handling as `set_freeze()` and `thaw()` can fail, so errors needed to be propagated.
- Updated mocks to set MaxLocks, MaxReserves, MaxFreezes, and MaxHolds to 50, to match Balances pallet config.
- Updated runtime pallet configs to use BalancesMaxXXXXXs
- Updated tests with `.expect()` where `set_freeze()` or `thaw()` can fail.
- `FreezeIdentifier` and `RuntimeFreezeReason` configured with defaults.
- Updated time-release pallet to propagate errors from set_freeze/thaw.
- Added v3 migration to Capacity.

# Storage Migrations
The value of `BalancesMaxFreezes` has been updated, which will impact the storage of the Balances pallet by changing `T::MaxFreezes`, see the code here: [substrate/frame/balances/src/lib.rs:480](https://github.com/paritytech/polkadot-sdk/blob/release-polkadot-v1.1.0/substrate/frame/balances/src/lib.rs#L480)

```rust
	/// Freeze locks on account balances.
	#[pallet::storage]
	pub type Freezes<T: Config<I>, I: 'static = ()> = StorageMap<
		_,
		Blake2_128Concat,
		T::AccountId,
		BoundedVec<IdAmount<T::FreezeIdentifier, T::Balance>, T::MaxFreezes>,
		ValueQuery,
	>;
```
The previous value of `T::MaxFreezes` was `0` so no data could be stored in `Freezes`, therefore no storage migration for `Freezes` is needed for this change. Even if there was data in storage, it would only need to be migrated if `T::MaxFreezes` is *decreased*.

However, the current chain has data in `Locks` that needs to migrated to `Freezes`. Testing has shown that these `Locks` will no longer be accessible once the new traits are in place. 

The `Balances` pallet is configured to store account data using the `System` pallet. Therefore, these two pallets must be included when using `try-runtime` for testing. 

The migration for `Capacity` will access its storage to determine which accounts have `Locks` that need to be translated to `Freezes`. Then, the old `Currency` trait is used to remove the `Locks` and the new `fungible` trait is used to set the `Freeze`. 

# How to Review
- [ ] Read through [Deprecate Currency - PR 12951](https://github.com/paritytech/substrate/pull/12951) to understand context and check that Currency traits were properly replaced with fungible traits.
- [ ] Check impact of changing to `set_freeze()` and `thaw()` which can now fail and make sure all error states are propagated correctly without possibility for `panic`
- [ ] Check if `balance()` is used correctly, or should be changed to `reducible_balance()`. The calculations evaluating the Existential Deposit (ED) have been updated and Parity comments indicate that `reducible_balance()` is most likely the value needed.
- [ ] Ensure that the migration weights calculations are correct. (Please let me know if you would like to walk through the migration code path together).

# How to Test Runtime Migrations
[Install the CLI version of try-runtime](https://paritytech.github.io/try-runtime-cli/try_runtime/#installation), then run try-runtime to test the migration against Frequency Rococo:
```bash
cargo build --release --features frequency-rococo-testnet,try-runtime && \
try-runtime --runtime ./target/release/wbuild/frequency-runtime/frequency_runtime.wasm on-runtime-upgrade live --uri wss://rpc.rococo.frequency.xyz:443 -pallet  Capacity  --pallet Balances --pallet System
```
Alternatively, you can use the non-release version for faster compiles:
```bash
cargo build --features frequency-rococo-testnet,try-runtime && \
try-runtime --runtime ./target/debug/wbuild/frequency-runtime/frequency_runtime.wasm on-runtime-upgrade live --uri wss://rpc.rococo.frequency.xyz:443 -pallet  Capacity --pallet Balances --pallet System
```
You should see output like this:
```bash
[2023-12-16T17:03:57Z INFO  runtime::capacity] migrated 344
[2023-12-16T17:03:57Z INFO  runtime::capacity] migrated 345
[2023-12-16T17:03:57Z INFO  runtime::capacity] migrated 346
[2023-12-16T17:03:57Z INFO  runtime::capacity] migrated 347
[2023-12-16T17:03:57Z INFO  runtime::capacity] 🔄 migration finished  
[2023-12-16T17:03:57Z INFO  runtime::capacity] Migration calculated weight = Weight { ref_time: 78200000000, proof_size: 0 }
[2023-12-16T17:03:57Z INFO  runtime::capacity] ✅ migration post_upgrade checks passed

// end of Capacity v2 migration, v3 migration follows

[2023-12-16T17:03:57Z INFO  runtime::capacity] Running pre_upgrade...
[2023-12-16T17:03:57Z INFO  runtime::capacity] Finish pre_upgrade for 347 records
[2023-12-16T17:03:57Z INFO  runtime::capacity] 🔄 Capacity Locks->Freezes migration started
[2023-12-16T17:03:57Z INFO  runtime::capacity] 🔄 migrated account 0xac4e66be328c4d235be27fcb0af8ccbe12dce375236f9ccc5516780522bc8870, amount:1500000000
[2023-12-16T17:03:57Z INFO  runtime::capacity] 🔄 migrated account 0x3048b40a2e7185e510b695c7ba15f31218a1d4a501c6a71596e4f60317fc180f, amount:800000000
[2023-12-16T17:03:57Z INFO  runtime::capacity] 🔄 migrated account 0x663ea97a6b40a51ed79f3328f9629423e5525c84bb178e63a854e4ce497fde25, amount:900000000
[2023-12-16T17:03:57Z INFO  runtime::capacity] 🔄 migrated account 0x1488b10710285607257cbaa5f2ee273b50613358c50e443b8e43790ba7e15705, amount:900000000
[2023-12-16T17:03:57Z INFO  runtime::capacity] 🔄 migrated account 0x283ac37222b6e34afb4b187e41e8e828c2e2d2b7f5bc027a2bf98eccbca76d47, amount:200000000
[2023-12-16T17:03:57Z INFO  runtime::capacity] 🔄 migrated account 0x1a038c038ece2448a22096697a35905a98cc14138666432a1767142e67f76414, amount:900000000
[2023-12-16T17:03:57Z INFO  runtime::capacity] 🔄 migrated account 0xa848c35ccb975405ceec7d275d2cbbfe0df577f54a4e87caee5a4bf64f68ad01, amount:500000000
[2023-12-16T17:03:57Z INFO  runtime::capacity] 🔄 migrated account 0x2aa39545e21212127c97ce1059988eb7f8dabe94f61e023bde1b85fd9625a638, amount:1000000
[2023-12-16T17:03:57Z INFO  runtime::capacity] 🔄 migrated account 0x9404948a1b48c6818117f5803e5b868515fc88d1bcbd6fa27a1083089e5b5604, amount:900000000
[2023-12-16T17:03:57Z INFO  runtime::capacity] 🔄 migrated account 0xa0e340ec2c110a262b3f730c1cd333d52d6bb9fcb0b149bdd7216fb8cd4b2468, amount:100000000
[2023-12-16T17:03:57Z INFO  runtime::capacity] 🔄 migrated account 0x28d9dadd32d1864fa4aa553a74ff82637acb131079bfec5b2777df7dea4a2e5d, amount:900000000
[2023-12-16T17:03:57Z INFO  runtime::capacity] 🔄 migrated account 0x4c54252a95ae9e334be70375d14d51d4d89a6ae2deeaf9ca47091d797f319620, amount:200000000
[2023-12-16T17:03:57Z INFO  runtime::capacity] 🔄 migrated account 0xe2133699b4fdc8444b98f9fddc7039374302a9a41e9ce53eafab113fd236c739, amount:200000000
```

The total weight calculated for the Capacity migration on testnet:
```bash
[2023-12-18T14:50:36Z INFO  runtime::capacity] total accounts migrated from locks to freezes: 347
[2023-12-18T14:50:36Z INFO  runtime::capacity] 🔄 Capacity Locks->Freezes migration finished
[2023-12-18T14:50:36Z INFO  runtime::capacity] Capacity Migration calculated weight = Weight { ref_time: 260375000000, proof_size: 0 }
[2023-12-18T14:50:36Z INFO  runtime::capacity] ✅ migration post_upgrade checks passed
```

The total weight calculated for the Capacity migration on main net:
```bash
[2023-12-18T14:57:04Z INFO  runtime::capacity] 🔄 Capacity Locks->Freezes migration finished
[2023-12-18T14:57:04Z INFO  runtime::capacity] Capacity Migration calculated weight = Weight { ref_time: 1625000000, proof_size: 0 }
[2023-12-18T14:57:04Z INFO  runtime::capacity] ✅ migration post_upgrade checks passed
```

# Upgrade Notes

1. `scripts/upgrade_accounts.py` should be executed to ensure that all accounts have been upgraded before running the migration. 

# Checklist
- [x] Chain spec updated
- [ ] Design doc(s) updated
- [x] Tests added
- [ ] Benchmarks added
- [x] Weights updated


pallets/time-release/src/lib.rs
```diff
@@ -430,11 +445,11 @@ impl<T: Config> Pallet<T> {
 	}
 
 	fn update_lock(who: &T::AccountId, locked: BalanceOf<T>) {
-		T::Currency::set_lock(RELEASE_LOCK_ID, who, locked, WithdrawReasons::all());
+		let _ = T::Currency::set_freeze(&FreezeReason::NotYetVested.into(), who, locked);
 	}
 
 	fn delete_lock(who: &T::AccountId) {
-		T::Currency::remove_lock(RELEASE_LOCK_ID, who);
+		let _ = T::Currency::thaw(&FreezeReason::NotYetVested.into(), who);
```
enddynayn, line 452: Let's consider adding an integration test for this and a unit test. We might also be silently failing.

pallets/capacity/src/lib.rs
```diff
@@ -465,35 +482,39 @@ impl<T: Config> Pallet<T> {
 		let mut capacity_details = Self::get_capacity_for(target).unwrap_or_default();
 		capacity_details.deposit(&amount, &capacity).ok_or(ArithmeticError::Overflow)?;
 
-		Self::set_staking_account(&staker, staking_account);
+		Self::set_staking_account(&staker, staking_account)?;
```
enddynayn, line 485: Since we can fail here we can probably add some new unit tests around this.

e2e/capacity/staking.test.ts
```diff
@@ -45,6 +45,7 @@ describe('Capacity Staking Tests', function () {
 
       // Confirm that the tokens were locked in the stakeKeys account using the query API
       const stakedAcctInfo = await ExtrinsicHelper.getAccountInfo(stakeKeys.address);
+      console.log("stakedAcctInfo.data.frozen", stakedAcctInfo.data.frozen.toHuman());
```
enddynayn, line 48: rm

pallets/capacity/src/lib.rs
mattheworris: This code is now outdated. I will have a look at the new code and see if any new tests are still needed.

pallets/capacity/src/lib.rs
```diff
@@ -100,6 +101,16 @@ pub mod pallet {
...
+	/// A reason for freezing funds.
+	#[pallet::composite_enum]
+	pub enum FreezeReason {
+		/// The account has staked tokens to the Frequency network.
+		#[codec(index = 0)]
+		Staked,
+		/// Funds are currently locked and are not yet liquid.
+		NotYetVested,
+	}
+
```
JoeCap08055, line 113: Based on [this](https://paritytech.github.io/polkadot-sdk/master/frame_support/attr.pallet.html#composite-enum-palletcomposite_enum-optional), it looks like we do not need to duplicate all of the variants of this enum in each pallet it appears in; we only need the variants specific to the pallet, similar to `#[pallet::error]` and `#[pallet::event]`

pallets/time-release/src/lib.rs
```diff
@@ -88,13 +89,31 @@ pub mod module {
...
+	/// A reason for freezing funds.
+	#[pallet::composite_enum]
+	pub enum FreezeReason {
+		#[codec(index = 0)]
+		/// Funds are staked.
+		Staked,
+		/// Funds are currently locked and are not yet liquid.
+		NotYetVested,
+	}
+	// use crate::pallet_capacity::FreezeReason;
```
JoeCap08055, line 101: https://github.com/LibertyDSNP/frequency/pull/1779/files#r1417914929

JoeCap08055 (commented):
As mentioned in the walkthrough, maybe a deeper review of whether this has any storage impact that would require a migration, especially since we seem to be at the forefront of users of this trait w/capacity & time-release.

pallets/capacity/src/lib.rs
mattheworris: Thanks @JoeCap08055, based on those docs, I've refactored FreezeReason in both pallets.

pallets/time-release/src/tests.rs
```diff
@@ -557,18 +615,23 @@ fn alice_time_releases_schedule() {
...
 }
 
 fn set_balance<T: Config>(who: &T::AccountId, balance: BalanceOf<T>) {
-	let deposit_result = T::Currency::deposit_creating(who, balance.saturated_into());
-	let actual_deposit = deposit_result.peek();
-	assert_eq!(balance, actual_deposit);
+	let _ = T::Currency::mint_into(&who, balance.saturated_into());
+	assert_eq!(
+		T::Currency::balance(who).saturated_into::<u64>(),
+		balance.saturated_into::<u64>() + 100u64
```
mattheworris, line 633: Why is this 100 here?

JoeCap08055: Existential balance, I'd guess?

mattheworris: ED is what I was thinking when I put it there, _but_ this runs in the tests and so ED is more like `1`.
The real reason is that it was refactored incorrectly.  The new code works a bit differently.

mattheworris:
> Does `time-release` not also need a migration?

Yes, time-release will also need to migrate its Locks to Freezes.

pallets/capacity/src/migration/v3.rs
```diff
@@ -0,0 +1,130 @@
...
+	}
+
+	#[storage_alias]
+	/// alias to StakingAccountLedger storage
+	pub(crate) type StakingAccountLedger<T: Config> = StorageMap<
+		Pallet<T>,
+		Twox64Concat,
+		<T as frame_system::Config>::AccountId,
+		StakingDetails<T>,
+	>;
```
JoeCap08055, line 42: I've lost track--is this before or after @shannonwells staking ledger changes for PB?

mattheworris: > I've lost track--is this before or after @shannonwells staking ledger changes for PB?

Currency->fungible is after the staking ledger changes.

pallets/capacity/src/migration/v3.rs
```diff
@@ -0,0 +1,159 @@
...
+{
+	fn on_runtime_upgrade() -> Weight {
+		let on_chain_version = Pallet::<T>::on_chain_storage_version(); // 1r
+
+		if on_chain_version.lt(&3) {
+			log::info!(target: LOG_TARGET, "🔄 Capacity Locks->Freezes migration started");
+			let mut maybe_count = 0u32;
+			StakingAccountLedger::<T>::iter()
+				.map(|(account_id, staking_details)| (account_id, staking_details.active))
+				.for_each(|(staker, mut amount)| {
```
aramikm, line 53: nit: Having a `mut` variable in foreach looks very confusing. Why not just define a new variable like below?

`let total_freeze = amount.saturating_add(Pallet::<T>::get_unlocking_total_for(&staker));`

pallets/capacity/src/migration/v3.rs
```diff
@@ -0,0 +1,159 @@
...
+			pallet_balances::Pallet::<T>::set_lock(
+				STAKING_ID,
+				&account,
+				amount,
+				WithdrawReasons::all(),
+			);
+			// Confirm lock exists
+			assert_eq!(
+				pallet_balances::Pallet::<T>::locks(&account).get(0),
+				Some(&BalanceLock { id: STAKING_ID, amount: 50u64, reasons: Reasons::All })
```
aramikm, line 138: nit: if amount is 50 why duplicating in multiple places?

pallets/time-release/src/migration/v2.rs
```diff
@@ -0,0 +1,207 @@
...
+		Pallet<T>,
+		Blake2_128Concat,
+		<T as frame_system::Config>::AccountId,
+		BoundedVec<ReleaseScheduleOf<T>, <T as Config>::MaxReleaseSchedules>,
+		ValueQuery,
+	>;
+}
+
+/// The OnRuntimeUpgrade implementation for this storage migration
+pub struct MigrationToV3<T, OldCurrency>(sp_std::marker::PhantomData<(T, OldCurrency)>);
```
aramikm, line 60:
```suggestion
pub struct MigrationToV2<T, OldCurrency>(sp_std::marker::PhantomData<(T, OldCurrency)>);
```

pallets/time-release/src/migration/v2.rs
```diff
@@ -0,0 +1,207 @@
...
+	#[test]
+	fn migration_works() {
+		ExtBuilder::build().execute_with(|| {
+			StorageVersion::new(2).put::<Pallet<Test>>();
+			// Create some data in the old format
+			// Grab an account with a balance
+			let account = 200;
+			let amount = 50;
+
+			pallet_balances::Pallet::<Test>::set_lock(
```
aramikm, line 179: Don't you need to add something into v1::ReleaseSchedules first for the test since that is the main storage that you are iterating on?

aramikm (commented):
Added some comments. Would also need to run the migration using try-runtime on rococo and mainnet to ensure it would fit in a block

pallets/capacity/src/migration/v3.rs
```diff
@@ -0,0 +1,159 @@
...
+		StakingType::MaximumCapacity,
+	};
+	use pallet_balances::{BalanceLock, Reasons};
+
+	type MigrationOf<T> = MigrationToV3<T, pallet_balances::Pallet<T>>;
+
+	#[test]
+	fn migration_works() {
+		new_test_ext().execute_with(|| {
+			StorageVersion::new(2).put::<Pallet<T>>();
```
enddynayn, line 123: why do we set it here?

pallets/capacity/src/migration/v3.rs
```diff
@@ -0,0 +1,159 @@
...
+impl<T: Config, OldCurrency> OnRuntimeUpgrade for MigrationToV3<T, OldCurrency>
+where
+	T: Config,
+	OldCurrency: 'static + LockableCurrency<T::AccountId, Moment = BlockNumberFor<T>>,
+	OldCurrency::Balance: IsType<BalanceOf<T>>,
+{
+	fn on_runtime_upgrade() -> Weight {
+		let on_chain_version = Pallet::<T>::on_chain_storage_version(); // 1r
+
+		if on_chain_version.lt(&3) {
```
enddynayn, line 48: lets use STORAGE_VERSION from lib
```lib.rs
pub const STORAGE_VERSION: StorageVersion = StorageVersion::new(2);
```

pallets/capacity/src/migration/v3.rs
```diff
@@ -0,0 +1,159 @@
...
+			log::info!(target: LOG_TARGET, "🔄 Capacity Locks->Freezes migration started");
+			let mut maybe_count = 0u32;
```
enddynayn, line 50: maybe_count -> accounts_migrated_count

pallets/capacity/src/migration/v3.rs
```diff
@@ -0,0 +1,159 @@
...
+pub struct MigrationToV3<T, OldCurrency>(sp_std::marker::PhantomData<(T, OldCurrency)>);
+impl<T, OldCurrency> MigrationToV3<T, OldCurrency>
+where
+	T: Config,
+	OldCurrency: 'static + LockableCurrency<T::AccountId, Moment = BlockNumberFor<T>>,
+	OldCurrency::Balance: IsType<BalanceOf<T>>,
+{
+	/// Translate capacity staked locked deposit to frozen deposit
+	pub fn translate_lock_to_freeze(account_id: T::AccountId, amount: OldCurrency::Balance) {
+		OldCurrency::remove_lock(STAKING_ID, &account_id); // 1r + 1w
```
enddynayn, line 30: this seems more than 1r and 1r
can you please double-check this:
https://github.com/LibertyDSNP/frequency/pull/1779/files#diff-2fd628e078dc6cd044753cc133baa49c80e80c3f1a3d2cbe54396956179b5447R29

enddynayn:
Thanks for the awesome commit message!

pallets/capacity/src/lib.rs
```diff
@@ -100,19 +101,35 @@ pub mod pallet {
...
+	/// A reason for freezing funds.
+	/// Creates a freeze reason for this pallet that is aggregated by `construct_runtime`.
+	#[pallet::composite_enum]
+	pub enum FreezeReason {
+		/// The account has staked tokens to the Frequency network.
+		Staked,
+	}
+
+	/// the storage version for the v2 migration
+	pub const STORAGE_VERSION_V2: StorageVersion = StorageVersion::new(2);
```
enddynayn, line 113: why do we need two of these?

pallets/capacity/src/lib.rs
```diff
@@ -562,12 +579,18 @@ impl<T: Config> Pallet<T> {
...
+		let account_balance = T::Currency::balance(&staker);
 		account_balance
 			.saturating_sub(T::MinimumTokenBalance::get())
 			.min(proposed_amount)
 	}
 
+	// Calculates the total amount of tokens that are currently unlocked for the given staker.
+	pub(crate) fn get_unlocking_total_for(staker: &T::AccountId) -> (BalanceOf<T>, Weight) {
+		let unlocks = Self::get_unstake_unlocking_for(staker).unwrap_or_default();
+		(unlock_chunks_total::<T>(&unlocks), T::DbWeight::get().reads(1))
```
enddynayn, line 591: why do we return the weights here?

pallets/capacity/src/lib.rs
mattheworris: Capacity has to run 2 migrations in series. We could just write `2`, I thought `STORAGE_VERSION_V2` would be useful in more places.

pallets/capacity/src/lib.rs
mattheworris: Yes, this is only used for migrations.
Yes, I will work on moving it there.

As written, it returns the weight to keep track of the read.
However, as the number of reads remains constant here, the calling function could be refactored without the return.

enddynayn: I think it might be a good idea to move it to the migrations file since it's not used anywhere but there. if we were to reuse it in the lib.rs we would not need to return the weights since it would be caught be the benchmarks.

aramikm (commented):
Added some comments. I'm a bit worried about the total weight on test-net since it's around 40% of the block capacity + time-release changes alone and if another migration comes on top of those then it would be getting closer to the limit.

scripts/upgrade_accounts.py
```diff
@@ -0,0 +1,99 @@
...
+    # Using the public endpoint can get you rate-limited.
+    # url="wss://kusama-rpc.polkadot.io",
+    # These Parity internals are not limited.
+    # url="wss://polkadot-try-runtime-node.parity-chains.parity.io:443"
+)
+
+print(f"Connected to {chain.name}: {chain.chain} v{chain.version}")
+
+sender_uri = os.getenv('SENDER_URI', '//Alice')
+sender = Keypair.create_from_uri(sender_uri)
```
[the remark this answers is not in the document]
mattheworris: If more than 90% of the accounts need to be upgraded, then it is a free transaction. See here:
https://github.com/paritytech/polkadot-sdk/blob/f93f461acc24022bf7ef07c5913e620121776708/substrate/frame/balances/src/lib.rs#L685

pallets/capacity/src/lib.rs
mattheworris: `STORAGE_VERSION_V2` has been moved to v2.rs to avoid confusion.

enddynayn (approved):
Nice! Some small comments but no blockers. Amazing work.

pallets/capacity/src/migration/v3.rs
```diff
@@ -0,0 +1,208 @@
...
+		//   1 write set Account: update_locks->try_mutate_account->ensure_upgraded: if account is *not* already upgraded.
+		//     We could avoid this write by calling the upgrade script before running the migration,
+		//     which would ensure that all accounts are upgraded.
+		// 1 write set Account: update_locks->try_mutate_account: set account data
+		// 1 read get Locks: update_locks: set existed with `contains_key`
+		// 1 write set Locks: update_locks->Locks::remove: remove existed
+		OldCurrency::remove_lock(STAKING_ID, &account_id);
+
+		// 1 read get Freezes: set_freeze: set locks
+		// 1 read get Locks: update_freezes: Locks::get().iter()
```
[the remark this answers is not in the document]
mattheworris: `update_freezes` is a common function that runs for both setting and clearing `Freezes`. It also does look at the `Locks` as it takes the max of `Locks` or `Freezes`.

pallets/capacity/src/migration/v3.rs
```diff
@@ -0,0 +1,208 @@
...
+			&FreezeReason::Staked.into(),
+			&account_id,
+			amount.into(),
+		)
+		.unwrap_or_else(|err| {
+			log::error!(target: LOG_TARGET, "Failed to freeze {:?} for account 0x{:?}, reason: {:?}", amount, HexDisplay::from(&account_id.encode()), err);
+		});
+
+		log::info!(target: LOG_TARGET, "🔄 migrated account 0x{:?}, amount:{:?}", HexDisplay::from(&account_id.encode()), amount.into());
+		T::DbWeight::get().reads_writes(6, 4)
```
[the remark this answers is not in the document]
mattheworris: Yes, the counts are updated to (5, 6), the extra writes are added, one of the reads is cached. Comments are also updated.

shannonwells (commented):
This one is blocking because of the referenced PR:
> Freezes are essentially the same as locks, except that they overlap with Holds. Since Holds are designed to be infallibly slashed, this means that _any logic using a Freeze must handle the possibility of the frozen amount being reduced, potentially to zero._ **A permissionless function should be provided in order to allow bookkeeping to be updated in this instance.**

This PR adds no such permissionless function. If I'm reading this correctly, the frozen amount is somehow reduced to zero particularly through slashing, then the unlocks will need to be reduced or cleared.  I think this is a nonzero possibility because a lot of Providers, who will be staking, will also be running their own nodes.  Maybe they would use different collator keys than staking keys, and maybe they wouldn't.  I'm not sure what such a function would look like or by what mechanism it would be called.

Re: `[reducible_balance](https://paritytech.github.io/polkadot-sdk/master/frame/traits/tokens/fungible/trait.Inspect.html#tymethod.reducible_balance)` - based on my reading, this is <= `balance`, and the only reason ever to use this is in operations where we don't need to keep the system account alive, and there are no instances of that here.

- [x] Read through https://github.com/paritytech/substrate/pull/12951 to understand context and check that Currency traits were properly replaced with fungible traits.
- [x]  Check impact of changing to set_freeze() and thaw() which can now fail and make sure all error states are propagated correctly without possibility for panic
- [x]  Check if balance() is used correctly, or should be changed to reducible_balance(). The calculations evaluating the Existential Deposit (ED) have been updated and Parity comments indicate that reducible_balance() is most likely the value needed.
- [x]  Ensure that the migration weights calculations are correct.

shannonwells (approved):
As agreed in the standup parking lot, let's create a new issue for the question I raised, which includes writing some tests to check the behavior.

aramikm (approved):
Looks great. Tremendous effort. It is recommended that we release these heavier migration included PR's separately so that it would not overflow the block size on testnet.

PR #1781, opened 2023-11-13
feat: change instant-seal node docker image to use startup script

### LINKED ISSUE #1780: GitHub Actions cannot use frequencychain/instant-seal-node with interval sealing
The [frequencychain/instant-seal-node](https://hub.docker.com/r/frequencychain/instant-seal-node) allows for the default `CMD ["--sealing=instant"]` to be overridden in the standard Docker fashion. However, the `service` construct in GitHub Workflows do not allow for overriding the `CMD` of an image, making it impossible to run this image in anything other than instant-seal mode for use in CI pipelines.

JoeCap08055:
# Goal
The goal of this PR is to allow the use of interval sealing mode in CI pipelines using GitHub workflows.

The following considerations were taken into account:
* Since GitHub workflows do not allow overriding the `CMD` of a Docker image, and overriding `ENTRYPOINT` in GitHub likewise does not allow for passing arguments to the new entrypoint, the simplest option left is to use environment variables.
* Since it was deemed undesirable to modify the frequency binary itself to honor environment variables, the approach taken was to create a new image, `frequencychain/standalone-node`. The `ENTRYPOINT` of the image is a shell script instead of the `frequency` binary directly. The shell script takes into account the following variables:
    * `SEALING_MODE`: `instant`, `interval`, or `manual`
    * `SEALING_INTERVAL`: number of seconds between blocks in `interval` sealing mode
    * `CREATE_EMPTY_BLOCKS`: `true` means pass `--sealing-create-empty-blocks` to allow interval sealing mode to form empty blocks as in a live chain.

Additional arguments passed to the script are passed through to the underlying `frequency` command-line, just as the `CMD` parameter in the old image. Existing use cases of this image should therefore be unaffected.
* The existing `frequencychain/instant-seal-node` image is now deprecated and will eventually be discontinued/removed.

Closes #1780 

# Discussion

# Checklist
- [x] Doc(s) updated
- [x] Tested new image for all sealing modes: `instant`, `interval`, `manual`


docker/instant-seal-node.dockerfile
wilwade, line 1: We should likely go ahead and change the name as well. Which likely means we should create a new image and just deprecate this one over time. Thoughts?

JoeCap08055: I thought about that, but I was trying to limit the scope, as changing the name of the image would affect any pipelines that are pulling it currently.

JoeCap08055: ...unless we duplicate instead of rename, and just eventually deprecate the older `instant-seal-node` image

scripts/frequency-start.sh
```diff
@@ -0,0 +1,32 @@
...
+then
+    CREATE_EMPTY_BLOCKS="--sealing-create-empty-blocks"
+fi
+
+exec /frequency/frequency \
+    --dev \
+    -lruntime=debug \
+    --no-telemetry \
+	--no-prometheus \
+	--port=30333 \
```
wilwade, line 23:
```suggestion
```

docker/instant-seal-node.overview.md
```diff
@@ -39,6 +39,17 @@ Note: Docker `--rm` removes the volume when stopped.
 docker run --rm -p 9944:9944 -p 30333:30333 frequencychain/instant-seal-node:<version.tag>
```
wilwade, line 39: Note in the old one how to switch to the new one.

docker/instant-seal-node.overview.md
```diff
@@ -1,5 +1,12 @@
 # Frequency Collator Node in Local Only Sealing Mode
 
+## Deprecation Note
+
+This image has been deprecated and will be removed in a future Frequency release. The new image, [frequencychain/standalone-node](./standalone-node.overview.md), replaces it. Current usage patterns for this image are fully supported by the new image, as well as additional enhancements.
+##
```
wilwade, line 6: ?
```suggestion
```

docker/standalone-node.dockerfile
```diff
@@ -0,0 +1,35 @@
...
+COPY --from=base /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certificates.crt
+# For local testing only
+# COPY --chown=frequency target/x86_64-unknown-linux-gnu/debug/frequency ./frequency/frequency
+COPY --chown=frequency target/release/frequency ./frequency/
+COPY --chown=frequency scripts/frequency-start.sh ./frequency
+RUN chmod +x ./frequency/frequency ./frequency/frequency-start.sh
+
+# 9944 for RPC call
+# 9615 for Telemetry (prometheus)
+EXPOSE 9944 9615
```
wilwade, line 31: Telemetry is disabled in the script, but perhaps?
```suggestion
EXPOSE 9944
```

docker/standalone-node.overview.md
```diff
@@ -0,0 +1,77 @@
+# Frequency Collator Node in Local Only Sealing Mode
+
+Runs just one collator node that will not connect to any other nodes.
+Defaults to running in instant sealing mode where a block will be triggered when a transaction enters the validated transaction pool.
+A "collator node" is a Frequency parachain node that is actively collating (aka forming blocks to submit to the relay chain, although in this case without a relay chain).
+
+### Quick Run
+
+```sh
+docker run --rm -p 9944:9944 -p 30333:30333 frequencychain/standalone-node:<version.tag>
```
wilwade, line 10:
```suggestion
docker run --rm -p 9944:9944 frequencychain/standalone-node:<version.tag>
```

wilwade (commented):
Few comments, but otherwise looks safe to ship

docker/instant-seal-node.overview.md
```diff
@@ -1,5 +1,12 @@
 # Frequency Collator Node in Local Only Sealing Mode
 
+## Deprecation Note
+
+This image has been deprecated and will be removed in a future Frequency release. The new image, [frequencychain/standalone-node](./standalone-node.overview.md), replaces it. Current usage patterns for this image are fully supported by the new image, as well as additional enhancements.
```
aramikm, line 5: I think this should be communicated to other teams such as **custodial wallet team** if they are using `instant-seal-node` for tests.

aramikm (approved):
Looks good!

scripts/frequency-start.sh
```diff
@@ -0,0 +1,31 @@
...
+then
+    SEALING_INTERVAL="--sealing-interval=${SEALING_INTERVAL}"
+fi
+
+if [ "${CREATE_EMPTY_BLOCKS}" = true ]
+then
+    CREATE_EMPTY_BLOCKS="--sealing-create-empty-blocks"
+fi
+
+exec /frequency/frequency \
```
shannonwells, line 18: Noting the path - is this script intended -only- for dockerfiles then?

JoeCap08055: Yes, only for dockerfiles. You thinking maybe to put it in the `docker` directory?

docker/standalone-node.overview.md
JoeCap08055: Missed that one, thanks.

scripts/frequency-start.sh
wilwade: I think moving it to the docker directory is a good idea.

JoeCap08055: Done.

wilwade (approved):

PR #1763, opened 2023-11-07
Feat/separate unlock chunks #1753

shannonwells:
# Goal
The goal of this PR is to separate unstaking unlock chunks into their own storage and do a storage migration which includes staking type, for convenience of the planned Provider Boost feature so that another storage migration is not needed for that feature.  There are multiple advantages to this separation discussed in the linked issue.

Closes #1753 .  Please see the linked issue for more details about this proposed change.

# Discussion

A stake will require 1 extra read of unstaking unlocks from before, in order to set the account token lock correctly.  This was not noted in the original issue.

As noted in the original issue, an unstake will require 1 extra read and 1 extra write from before.

Effectively changes the name of `StakingAccountDetails` to just `StakingDetails` while splitting the storage.

`StakingAccountLedger` now stores `StakingDetails`.

The migration uses the `translate` function of the `IterableStorageMap` trait.

The implementation design document was updated to make it more clear that details of actual ProviderBoost staking rewards are undecided and unspecified.

Since token locks are being touched, please review with an eye toward bugs with token locks and test coverage.

## To Verify
[Install the CLI version of try-runtime](https://paritytech.github.io/try-runtime-cli/try_runtime/#installation), then run try-runtime to test the migration against Frequency Rococo:
```shell
cargo build --release --features frequency-rococo-testnet,try-runtime && \
try-runtime --runtime ./target/release/wbuild/frequency-runtime/frequency_runtime.wasm on-runtime-upgrade live --uri wss://rpc.rococo.frequency.xyz:443 -p Capacity
```

You should see at something like:
```
[2023-11-15T01:32:16Z INFO  runtime::capacity] 🔄 StakingAccountLedger migration started
[2023-11-15T01:32:16Z INFO  runtime::capacity] migrated 1
[2023-11-15T01:32:16Z INFO  runtime::capacity] migrated 2
...
[2023-11-15T01:32:16Z INFO  runtime::capacity] migrated 335
[2023-11-15T01:32:16Z INFO  runtime::capacity] migrated 336
[2023-11-15T01:32:16Z INFO  runtime::capacity] 🔄 migration finished
[2023-11-15T01:32:16Z INFO  runtime::capacity] Migration calculated weight = Weight { ref_time: 75725000000, proof_size: 0 }
[2023-11-15T01:32:16Z INFO  runtime::capacity] ✅ migration post_upgrade checks passed
```
The errors afterward are unrelated to the migration, having to do with a runtime difference between local and rococo, and can be ignored.

# Checklist
[...]

PR #1784, opened 2023-11-16
adding schema names support

### LINKED ISSUE #1693: Schema "Names"
## Context

Schemas use a u16 identifier to assist with space constraints. While useful, it requires human mapping outside of the chain to interpret the specific chain to human meaning. This must be done for each chain and can be an issue for local development chains.

However any human interpretation of a schema requires human validation. A goal is that schemas can be produced by anyone. These two are in conflict with each other.

Either way, currently schemas must be approved by governance by the Frequency Council or Referenda.

It is considered reasonable that it is a future problem and while schemas could be created without issue, naming might still require a governance approval to ensure that there are not confusions possible.

For additional context and discussion see the Frequency Community Call on 2023-09-28: https://discord.com/channels/969001918460469250/969308206562164736/1157148627987333261

## Goal

Add some way for a name to be dereferenced on-chain to the schema id.

### Secondary Goal

Include versioning in some form.

### Example Names/Versions

Pieces:
- Namespace
- Descriptor
- Version

Possible Examples:
- `dsnp.broadcast.v1`
- `dsnp.org/broadcast/1` (uri style)
- `dsnp:broadcast#1` (DID-ish style)
- Others?

## Solution Criteria
- Must be in ASCII
- Must be able to call an RPC with the full name(s) and result in the schema ids
- Must be immutable. Once set, a full name should always resolve to the same id

## Non-Requirements
- Existing calls that take Schema Ids should STILL take schema Ids (Must be dereferenced first)

aramikm:
# Goal
The goal of this PR is to add support for schema names and versions.

Closes #1693

# Design
Storing the names in a map such that names are the key and points to the list of different versions and their schema ids for that name
 
SchemaNames drawio

# what is added?
- added new storage `SchemaNameToIds` which allows getting schema ids and versions from a schema name
- added new `create_schema_v3`, `propose_to_create_schema_v2` and `create_schema_via_governance_v2` extrinsics  which support adding an optional name during creation
- added new `propose_to_create_schema_name` and `create_schema_name_via_governance` to assign a name to an schema without a name
- Added migration to assign dsnp schema names to well-known schemas in rococo and mainnet
- Added an RPC which returns the versions and schema ids for a certain schema name or namespace

# name limitations
- only supporting ascii character of [a-z] [A-Z] or -
- name format is : `<namespace>.<descriptor>`
- namespace min size 3 characters
- descriptor min size is 1 character
- name max size 32 characters
- namespace must not start or end with -
- descriptor must not start or end with -

# Checklist
- [x] Chain spec updated
- [X] Tests added
- [X] Benchmarks added
- [X] Weights updated


common/primitives/src/schema.rs
```diff
@@ -14,6 +14,9 @@ use utils::*;
 /// Schema Id is the unique identifier for a Schema
 pub type SchemaId = u16;
 
+/// Schema version number
+pub type SchemaVersion = u8;
```
aramikm, line 18: **Question:** `u8` or something else?

Since the schema ids are u16 it mades sense to use a simple `u8` but open to changes

pallets/schemas/src/lib.rs
```diff
@@ -209,6 +233,21 @@ pub mod pallet {
 		OptionQuery,
 	>;
 
+	/// Storage for message schema info struct data
+	/// - Key: Schema Id
+	/// - Value: [`SchemaInfo`](SchemaInfo)
+	#[pallet::storage]
+	#[pallet::getter(fn get_schema_id)]
+	pub(super) type SchemaNamesToId<T: Config> = StorageDoubleMap<
```
aramikm, line 241: **Question:** `StorageDoubleMa` or `StorageNMap`?

name format: `<namespace>.<descriptor>`
`StorageNMap` will allow us to do prefix queries on namespaces. but since we are hashing on namespace and descriptor separately it might be easier to create unbalanced trie but on the other hand since these are going through governanace maybe that is not an issue.

open to suggestions

pallets/schemas/src/lib.rs
```diff
@@ -402,19 +485,46 @@ pub mod pallet {
...
 			settings: BoundedVec<SchemaSetting, T::MaxSchemaSettingsPerSchema>,
+			schema_name_version: Option<(SchemaNamePayload, SchemaVersion)>,
 		) -> Result<SchemaId, DispatchError> {
 			let schema_id = Self::get_next_schema_id()?;
+			let (schema_name, version) = match schema_name_version {
+				Some((name_payload, schema_version)) => {
+					let parsed_name = SchemaName::try_parse::<T>(name_payload)?;
+					(parsed_name, schema_version)
+				},
+				None => {
```
aramikm, line 496: **Question:** maybe we should remove deprecated extrinsics instead of auto assign a name?

pallets/schemas/src/types.rs
```diff
@@ -18,3 +39,56 @@ pub struct SchemaInfo {
...
+impl SchemaName {
+	/// parses and verifies the request and returns the SchemaName type if successful
+	pub fn try_parse<T: Config>(payload: SchemaNamePayload) -> Result<SchemaName, DispatchError> {
+		// check if all ascii
+		let mut str = String::from_utf8(payload.into_inner())
+			.map_err(|_| Error::<T>::InvalidSchemaNameEncoding)?;
+		ensure!(str.is_ascii(), Error::<T>::InvalidSchemaNameEncoding);
+
+		// to canonical form
+		str = str.to_lowercase();
```
aramikm, line 61: Question: `lowercase` or `uppercase` for canonical?

pallets/schemas/src/lib.rs
```diff
@@ -381,7 +422,49 @@ pub mod pallet {
...
+		/// * [`Error::ExceedsMaxSchemaModelBytes`] - The schema's length is greater than the maximum schema length
+		/// * [`Error::InvalidSchema`] - Schema is malformed in some way
+		/// * [`Error::SchemaCountOverflow`] - The schema count has exceeded its bounds
+		/// * [`Error::InvalidSetting`] - Invalid setting is provided
+		///
+		#[pallet::call_index(5)]
+		#[pallet::weight(T::WeightInfo::create_schema_v3(model.len() as u32 + settings.len() as u32))]
+		pub fn create_schema_v3(
+			origin: OriginFor<T>,
+			schema_name: SchemaNamePayload,
```
aramikm, line 452: Question: use combined name like this or maybe have two separate params for namespace and descriptor?

pallets/schemas/src/types.rs
```diff
@@ -1,13 +1,34 @@
...
+pub const NAMESPACE_MIN: usize = 3;
+/// The maximum size of a namespace in schema
+pub const NAMESPACE_MAX: u32 = 15;
+/// The minimum size of a schema descriptor
+pub const DESCRIPTOR_MIN: usize = 1;
+/// The maximum size of a schema descriptor
+pub const DESCRIPTOR_MAX: u32 = 16;
+/// separator character
+pub const SEPARATOR_CHAR: char = '.';
+const AUTO_ASSIGNED_NAMESPACE: &str = "auto_assigned";
```
aramikm, line 30: **Question:** what should this be?

currently we are using `auto_assign.<current_id>` for older extrinsics

pallets/schemas/src/lib.rs
JoeCap08055: I don't think we should auto-assign a name, but I do think we should auto-increment the version if a name is supplied.
Actually, I think `schema_version` should not be part of the payload, but be an auto-increment--however, this might involve an extra storage read to get the last version. I don't think that's too bad, though, since creating a schema will not be a heavily-used extrinsic.

Will we still support unnamed schemas? I don't necessarily see any reason not to...

pallets/schemas/src/types.rs
JoeCap08055: Should we take advantage of some of the functions developed for the `handles` pallet to do schema name normalization?

common/primitives/src/schema.rs
JoeCap08055: `u8` seems fine--in the rare cases where the number of versions would exceed that, you'd just have to allocate a new name

JoeCap08055 (commented):
Overall, looks pretty good. I have some comments about the overall direction of this, perhaps they are too big for this PR and would need to be a follow-up:

* Regarding schema namespace: maybe it would be a good idea to have the concept of a 'restricted' namespace that can only be acted upon by either (a) a particular MSA/provider, or (b) governance. For instance:
    * `dsnp.*` or `frequency.*` namespace would require governance
    * `mewe.*` namespace could only be affected by MeWe's provider ID
☝🏻 this seems like it would need to be a separate body of work

* Regarding versions: since versions are just a monotonically increasing `u8`, it seems unnecessary to take the version as input. It's not like we're using semver; and there's no benefit to allowing 'holes' in the schema version, so why not skip the input and just bump the version?
    * also, do we want the ability to associate an existing schema with a name? (one benefit of this would be in the migration; it would not be necessary to migrate existing schemas--that could be done post-migration, given an `associate_schema_with_name` extrinsic). For that matter, if we did that, we could separate the creation & naming into two separate steps, which would not require modifying the existing extrinsics. Sure, it would require 2 transactions, but since schema creation is a relatively rare activity, and mostly anyone creating schemas is paying with capacity, I don't see that as a problem.
    * if, given the above, would we want to allow the same schema to be registered under multiple names?
    
* Since the rest of the chain will still operate using just the `u16` schema_id, I think `schema_name` should be optional in all schema creation extrinsics, and we shouldn't auto-assign a name if one is not provided.

pallets/schemas/src/lib.rs
wilwade: Hmm... I like the idea of `nmap`. Clarifies the requirements around each piece and the name becomes more `namespace.descriptor.version` One can get any superset of that.

pallets/schemas/src/lib.rs
wilwade: I think unnamed schemas are ok. Testing schemas etc... don't need an official name.

Either way I think we will need a way for governance to name (although not re-name) an existing schema.

wilwade: Auto incrementing versions would be good and I'm ok with schema creation being a fairly expensive operation

pallets/schemas/src/types.rs
```diff
@@ -18,3 +39,56 @@ pub struct SchemaInfo {
...
+pub struct SchemaName {
+	/// namespace or domain of the schema
+	pub namespace: BoundedVec<u8, ConstU32<NAMESPACE_MAX>>,
+	/// name or descriptor of this schema
+	pub descriptor: BoundedVec<u8, ConstU32<DESCRIPTOR_MAX>>,
+}
+
+impl SchemaName {
+	/// parses and verifies the request and returns the SchemaName type if successful
+	pub fn try_parse<T: Config>(payload: SchemaNamePayload) -> Result<SchemaName, DispatchError> {
```
wilwade, line 54: I think we could just limit the schema name parts to just lowercase `a-z` and `-`. While we can fix stuff for the submitter, as this comes through governance we can just force the input to be correct

wilwade (commented):
Few notes on some details, but I think in general this is a good working approach.

common/primitives/src/schema.rs
```diff
@@ -130,6 +133,16 @@ impl SchemaSettings {
...
+/// RPC Response form a schema name query
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+#[derive(Clone, Encode, Decode, PartialEq, Debug, TypeInfo, Eq)]
+pub struct SchemaVersionResponse {
+	/// The version for this schema
+	pub schema_version: SchemaVersion,
+	/// The unique identifier for this Schema
+	pub schema_id: SchemaId,
+}
+
```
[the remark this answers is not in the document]
wesbiggs: Big +1 from me

pallets/schemas/src/types.rs
aramikm: So I think for schema names based on what it is mentioned in the issue we only support ascii characters. I further simplified the input to only ascii alphanumeric characters but open to changes if somebody feels it's unreasonable

pallets/schemas/src/lib.rs
aramikm: Ok then, I'll take a look into auto-incrementing the version approach. 
If we consider having a name as **optional** for schemas then I think for existing well known schemas we can apply names during migration but to allow assigning names for future unassigned schemas we will need a new extrinsic (only accessible via governance)

pallets/schemas/src/types.rs
saraswatpuneet: some validation steps from handles can be used given the schema creator can add random chars?

aramikm: So we only support `[a-z] [A-Z]` and `-` and `.` in naming so I'm not sure how much the handles sytuff would be helpful but will take a look.

e2e/schemas/createSchema.test.ts
```diff
@@ -128,4 +128,80 @@ describe('#createSchema', function () {
...
+
+  it('should fail to create schema with invalid namespace length v3', async function () {
+    const f = ExtrinsicHelper.createSchemaV3(keys, AVRO_GRAPH_CHANGE, 'AvroBinary', 'OnChain', [], 'a.b');
+    await assert.rejects(f.fundAndSend(fundingSource), {
+      name: 'InvalidSchemaNamespaceLength',
+    });
+  });
+
+  it('get version rpc should return all schemas using the same name', async function () {
+    const f = ExtrinsicHelper.createSchemaV3(keys, AVRO_GRAPH_CHANGE, 'AvroBinary', 'OnChain', [], 'my-namespace.bob');
```
aramikm, line 179: do we need to make these random in case it's being run against testnet?

pallets/schemas/src/lib.rs
```diff
@@ -381,11 +445,204 @@ pub mod pallet {
...
+			T::ProposalProvider::propose_with_simple_majority(proposer, proposal)?;
+			Ok(())
+		}
+
+		/// Create a schema by means of council approval
+		///
+		/// # Events
+		/// * [`Event::SchemaCreated`]
+		///
+		/// # Errors
```
wilwade, line 494: Lots of new events and errors due to the name updates.

The list of errors is getting long, so not sure we keep that correctly up to date.

pallets/schemas/src/lib.rs
```diff
@@ -381,11 +445,204 @@ pub mod pallet {
...
+		}
+
+		/// Adds a given schema to storage. The schema in question must be of length
+		/// between the min and max model size allowed for schemas (see pallet
+		/// constants above). If the pallet's maximum schema limit has been
+		/// fulfilled by the time this extrinsic is called, a SchemaCountOverflow error
+		/// will be thrown.
+		///
+		/// # Events
+		/// * [`Event::SchemaCreated`]
```
wilwade, line 541: Same here. SchemaNameCreated event is missing.

pallets/schemas/src/lib.rs
```diff
@@ -381,11 +445,204 @@ pub mod pallet {
...
+			schema_name: SchemaNamePayload,
+		) -> DispatchResult {
+			T::CreateSchemaViaGovernanceOrigin::ensure_origin(origin)?;
+
+			ensure!(
+				schema_id <= Self::get_current_schema_identifier_maximum(),
+				Error::<T>::SchemaIdDoesNotExist
+			);
+			let parsed_name = SchemaName::try_parse::<T>(schema_name, true)?;
+			SchemaNameToIds::<T>::try_mutate(
```
wilwade, line 632: Hmmm... I was thinking schema names should be immutable, even by governance (Although technically governance can always mutate state directly).

So I was expecting `try_insert`. What are your thoughts on mutability?

pallets/schemas/src/lib.rs
```diff
@@ -524,7 +796,37 @@ pub mod pallet {
...
+				model,
+				model_type,
+				payload_location,
+				settings,
+				schema_name.clone(),
+			)?;
+			Ok((schema_id, schema_name))
+		}
+
+		/// a method to return all versions of a schema name with their schemaIds
```
wilwade, line 816: Do we want to add a note that this should only be used by RPCs?

pallets/schemas/src/migration/v3.rs
```diff
@@ -0,0 +1,137 @@
...
+		(5, b"dsnp.update".to_vec()),
+		(6, b"dsnp.profile".to_vec()),
+		(7, b"dsnp.public-key".to_vec()),
+		(8, b"dsnp.public-follows".to_vec()),
+		(9, b"dsnp.private-follows".to_vec()),
+		(10, b"dsnp.private-connections".to_vec()),
+	])
+}
+/// get known schema names for rococo
+#[cfg(not(feature = "frequency"))]
```
wilwade, line 31: While this is good for testing, this breaks local upgrades. We need to either limit it to Rococo, skip it, or don't throw errors if the schema is not found.

pallets/schemas/src/rpc/src/lib.rs
```diff
@@ -55,6 +55,10 @@ pub trait SchemasApi<BlockHash> {
 	/// validates a schema model and returns `true` if the model is correct.
```
wilwade, line 55: Should the response for `schemas_getBySchemaId` also include the name now? I think that will cause any issues as this is all json at this point right?

runtime/frequency/src/lib.rs
```diff
@@ -222,7 +218,7 @@ pub type Executive = frame_executive::Executive<
 	AllPalletsWithSystem,
 	(
 		pallet_messages::migration::v2::MigrateToV2<Runtime>,
-		pallet_schemas::migration::v2::MigrateToV2<Runtime>,
+		pallet_schemas::migration::v3::MigrateToV3<Runtime>,
```
wilwade, line 221: Should we delete the v2 rust file?

wilwade (commented):
Few questions and such, but looking good. I'll try to take some time to run locally, but wanted to get a few notes in first.

pallets/schemas/src/types.rs
aramikm: I'm closing this comment but upon review if you have specific concern or might have a better way to do something please add comments

js/api-augment/definitions/schemas.ts
```diff
@@ -60,8 +76,18 @@ export default {
...
+              {
+                name: "schema_name",
+                type: "Vec<u8>",
+              },
+            ],
+            type: "Option<Vec<SchemaVersionResponse>>",
+          },
         },
-        version: 1,
+        version: 2,
```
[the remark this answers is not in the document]
aramikm: I think based on the details mentioned in [here](https://paritytech.github.io/polkadot-sdk/master/polkadot_primitives/runtime_api/index.html) it makes sense to only have the latest version in the client side.  I think if there was a change on the runtime-api which is not backwards compatible maybe we might need to have both versions in the client side but right now it looks like all the changes are backwards-compatible

pallets/schemas/src/lib.rs
```diff
@@ -167,6 +175,33 @@ pub mod pallet {
 
 		/// Invalid setting for schema
 		InvalidSetting,
+
+		/// Invalid schema name encoding
+		InvalidSchemaNameEncoding,
```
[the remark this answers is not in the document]
aramikm: I can generalize error names and group them together if desired. Please contribute to this discussion by adding a comment or emoji about your preference.

pallets/schemas/src/lib.rs
aramikm: Thanks for the reminder. Updated the docs.

pallets/schemas/src/lib.rs
aramikm: Updated

pallets/schemas/src/lib.rs
aramikm: Inside `try_mutate` we are always adding the new schema id to the list via `schema_version_id.add::<T>(schema_id)?;` . Since we have an array of schema ids attached to the same name (with different versions) we have to use `mutate` , another not nice alternative is to use `get` and `set`.

pallets/schemas/src/migration/v3.rs
aramikm: I've modified this logic but we don't through any errors if the schema id is not found.

pallets/schemas/src/rpc/src/lib.rs
aramikm: So we don't have the access from SchemaId to SchemaName. We only have the mapping from name to ids.

runtime/frequency/src/lib.rs
aramikm: It depends! I didn't remove it since it documents the historical data changes and since it's not part of the code it would not be included in the wasm but I think there are benefits for having it part of the codebase

pallets/schemas/src/lib.rs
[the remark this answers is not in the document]
wilwade: Let's leave it for this one, and I'll add it generally to the topic list for discussion at a Community Call.

runtime/frequency/src/lib.rs
shannonwells: Why is the migrate to v2 line still in, has it not been run on mainnet yet?

e2e/schemas/createSchema.test.ts
```diff
@@ -128,4 +133,77 @@ describe('#createSchema', function () {
     assertExtrinsicSuccess(eventMap);
     assert.notEqual(createSchemaEvent, undefined);
   });
+
+  it('should successfully create a schema v3 with name', async function () {
```
shannonwells, line 137: what happens if you try to create a schema named ".bob"?

pallets/schemas/src/types.rs
```diff
@@ -18,3 +48,126 @@ pub struct SchemaInfo {
...
+	/// name or descriptor of this schema
+	pub descriptor: SchemaDescriptor,
+}
+
+#[derive(Clone, Encode, Decode, PartialEq, Debug, TypeInfo, Eq, MaxEncodedLen, Default)]
+/// A structure defining name of a schema
+pub struct SchemaVersionId {
+	/// the index of each item + 1 is considered as their version.
+	/// Ex: the schemaId located in `ids[2]` is for version number 3
+	pub ids: BoundedVec<SchemaId, ConstU32<MAX_NUMBER_OF_VERSIONS>>,
```
shannonwells, line 66: In #1763 I also have UnlockChunks as a struct with one BoundedVec field.  @enddynayn suggests it would be better for storage to simply store the BoundedVec as a type alias, and have static fns that operate on the alias.  His feeling was that unless there's a plan to add more fields later, making this a struct with methods adds to storage overhead.  I had been persuaded, but now I'm  waffling since I feel the encapsulation is appropriate and the overhead may be negligible.  Your thoughts?

aramikm: I'm 99% sure based on how scale encoding works there is no overhead

pallets/schemas/src/lib.rs
```diff
@@ -304,6 +355,10 @@ pub mod pallet {
 		///
 		#[pallet::call_index(2)]
 		#[pallet::weight(T::WeightInfo::propose_to_create_schema(model.len() as u32))]
+		#[allow(deprecated)]
+		#[deprecated(
+			note = "please use `propose_to_create_schema_v2` since `propose_to_create_schema` has been deprecated."
```
shannonwells, line 360: One of my concerns about this changeset is potential confusion over the use of v2 versus v3.  It's clear - to me, from reading this PR - that there is a v2 of the interface/extrinsics but the schema structure is v3.  Can we skip over "v2" and align the extrinsic names with the schema version?

pallets/schemas/src/lib.rs
```diff
@@ -381,11 +445,229 @@ pub mod pallet {
...
+		/// Propose to create a schema.  Creates a proposal for council approval to create a schema
+		///
+		#[pallet::call_index(5)]
+		#[pallet::weight(
+			match schema_name {
+				Some(_) => T::WeightInfo::propose_to_create_schema_v2(model.len() as u32),
+				None => T::WeightInfo::propose_to_create_schema(model.len() as u32)
+			}
+		)]
+		pub fn propose_to_create_schema_v2(
```
shannonwells, line 464: re: comment above, I can seem people reading this extrinsic as "this is to create a v2 schema, where is the one to propose creating a v3 schema?"

e2e/schemas/createSchema.test.ts
aramikm: It would fail. Since the number of test cases were high I've tested all those in unit tests. End to end tests usually are not supposed to cover all the test-cases specially if it's covered in the other tests.

pallets/schemas/src/lib.rs
aramikm: I think there are 2 versions that we talking about.

- Version of a specific extrinsic
- Storage Migration 

They are independent of each other and I'm not sure mixing them is a great idea but open to all suggestions. We can discuss it as a parking lot.

shannonwells (approved):
All my comments have been addressed and I have nothing blocking to add.

JoeCap08055 (approved):
lgtm, pending resolution of others' comments

enddynayn:
Great job! I believe this feature will be extremely helpful for users in discovering schemas. Here's a suggestion: Instead of deprecating create_schema_v2 to include the schema_name option, I think it's better to only introduce a new extrinsic for this purpose. Since this feature primarily benefits schema creators as a convenience, creating a new extrinsic should be enough. This approach would simplify the code and wouldn't necessitate updates for users.

enddynayn (approved):
great! 👍

pallets/schemas/src/lib.rs
aramikm: so it's important to note these versions are the versions for extrinsics not the underlying structure. We do not have a schema v2 or schema v3 structure. I think that would make sense if we were doing lazy migrations only.

pallets/schemas/src/types.rs
```diff
@@ -18,3 +48,126 @@ pub struct SchemaInfo {
...
+
+		// check if alphabetic or - or separator character
+		ensure!(
+			str.chars().all(|c| c.is_ascii_alphabetic() || c == '-' || c == SEPARATOR_CHAR),
+			Error::<T>::InvalidSchemaNameCharacters
+		);
+
+		// split to namespace and descriptor
+		let chunks: Vec<_> = str.split(SEPARATOR_CHAR).collect();
+		ensure!(
```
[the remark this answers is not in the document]
aramikm: we have the test for the second one
`TestCase { input: r#"hjs."#, expected: Error::<Test>::InvalidSchemaDescriptorLength },`

I'll add the ".foo" test case, but that class of error is also taken care of by following test which returns the same error
```
			TestCase {
				input: r#"aa.sdhks"#,
				expected: Error::<Test>::InvalidSchemaNamespaceLength,
			},
```

pallets/schemas/src/lib.rs
```diff
@@ -524,7 +821,38 @@ pub mod pallet {
...
+				settings,
+				schema_name.clone(),
+			)?;
+			Ok((schema_id, schema_name))
+		}
+
+		/// a method to return all versions of a schema name with their schemaIds
+		/// Warning: Must only get called from RPC, since the number of DB accesses is not deterministic
+		pub fn get_schema_versions(schema_name: Vec<u8>) -> Option<Vec<SchemaVersionResponse>> {
+			let bounded_name = BoundedVec::try_from(schema_name).ok()?;
```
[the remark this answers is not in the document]
aramikm: From my understanding since this is an RPC call which only gets executes on the receiving node I don't think it makes any difference in terms of optimization

PR #1794, opened 2023-11-29
Ci Update: Use v2 runners and grcov

wilwade:
# Goal
The goal of this PR is to get Code Coverage results reporting using the internal runners (faster, and more consistent results).

Result: https://app.codecov.io/gh/LibertyDSNP/frequency/tree/ci-update%2Fcode-cov-grcov/

# Discussion
- grcov reads the results a bit different than llvm-cov, but llvm cov was always running out of memory
- Lots of exclusions to get better results, but most of them are the same as from llvm-cov


JoeCap08055 (approved):
lgtm.
😢 seems `gcov` has no option for a config file instead of CLI params 🤷🏻‍♂️

enddynayn (approved):
Nice 👍

PR #1805, opened 2023-12-05
nonce: increase for free tnx and existing accounts

### LINKED ISSUE #1766: Prepaid (free) transactions should still increment the nonce for accounts that have an Account
Note: This is not a security issue. This is about expectations.

Correct: Given a tokenless account, executing one of the free transactions does not change the nonce from 0 to 1 because there is no nonce storage.

Strange: Given an account with tokens, executing one of the free transactions does not change the nonce, even with nonce storage.

This second case is unexpected. When sending a transaction on an account with the existential deposit, the nonce incrementing is expected, while if you have an account that gets reaped, then the nonce is expected to remain at 0.

Action:
- [ ] Update the `check_nonce.rs` file to use the `AccountInfo` information in addition to `info.pays_fee` to see if we will increment the nonce. Options/Ideas:
    - `nonce.is_zero()`?
    - Something using `AccountData`?
    - `account.providers > 0`?
- [ ] Update the e2e tests that call free transactions and remove the `'current'` flag on the signAndSend so that they increment the autononce system.

Relevant Rust Code:
```rust
if info.pays_fee == Pays::Yes {
    frame_system::Account::<T>::insert(who, account);
}
```

aramikm:
# Goal
The goal of this PR is to allow increase of nonce for free transactions on existing accounts

Closes #1766 

# Checklist
- [x] Test added
- [x] e2e tests updated

### DIFF
--- a/runtime/common/src/extensions/check_nonce.rs
+++ b/runtime/common/src/extensions/check_nonce.rs
@@ -98,11 +98,16 @@ where
 			.into())
 		}
 
+		// Is this an existing account ?
+		let existing_account =
+			account.providers > 0 || account.consumers > 0 || account.sufficients > 0;
+
 		// Increment account nonce by 1
 		account.nonce += T::Nonce::one();
 
-		// Only create or update the token account if the caller is paying
-		if info.pays_fee == Pays::Yes {
+		// Only create or update the token account if the caller is paying or
+		// account already exists
+		if info.pays_fee == Pays::Yes || existing_account {
 			frame_system::Account::<T>::insert(who, account);
 		}
 
--- a/e2e/handles/handles.test.ts
+++ b/e2e/handles/handles.test.ts
@@ -58,7 +58,7 @@ describe('🤝 Handles', function () {
       await ExtrinsicHelper.runToBlock(currentBlock + expirationOffset + 1); // Must be at least 1 > original expiration
 
       const retireHandle = ExtrinsicHelper.retireHandle(msaOwnerKeys);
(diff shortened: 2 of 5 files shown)


runtime/common/src/extensions/check_nonce.rs
```diff
@@ -98,11 +98,16 @@ where
 			.into())
 		}
 
+		// Is this an existing account ?
+		let existing_account =
+			account.providers > 0 || account.consumers > 0 || account.sufficients > 0;
```
aramikm, line 103: based on the logic on here if all three of these feilds are zero then the account will be reaped
https://github.com/paritytech/polkadot-sdk/blob/master/substrate/frame/system/src/lib.rs#L1246-L1247.

I could have used `nonce > 0` but that seems to be the side effect of an account existing not the real logic that gets checked.

Still if somebody feels strongly about adding `nonce > 0 ` that can also being added here

JoeCap08055: Is this logic copied from somewhere else? Might be good to include a reference to the source as a comment.

wilwade (approved):
Looks great to me!

- Reviewed issue
- Reviewed code
- Tested locally
- Matches all my expectations

runtime/common/src/extensions/check_nonce.rs
aramikm: This is extracted from the logic mentioned in this comment but will add it in the source also so we know where this came from.

saraswatpuneet (approved):
Nice work!

PR #1806, opened 2023-12-05
minor schema name change

aramikm:
# Goal
The goal of this PR is to change some schema names based on recommendations from dsnp foundation.

Related to #1693 

# Discussion
- renamed `dsnp.public-key` to `dsnp.public-key-key-agreement`

### DIFF
--- a/pallets/schemas/src/migration/v3.rs
+++ b/pallets/schemas/src/migration/v3.rs
@@ -22,7 +22,7 @@ pub fn get_known_schemas() -> BTreeMap<SchemaId, Vec<u8>> {
 		(4, b"dsnp.reaction".to_vec()),
 		(5, b"dsnp.update".to_vec()),
 		(6, b"dsnp.profile".to_vec()),
-		(7, b"dsnp.public-key".to_vec()),
+		(7, b"dsnp.public-key-key-agreement".to_vec()),
 		(8, b"dsnp.public-follows".to_vec()),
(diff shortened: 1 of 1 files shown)


enddynayn (approved):
Nice 👍

pallets/schemas/src/migration/v3.rs
```diff
@@ -22,7 +22,7 @@ pub fn get_known_schemas() -> BTreeMap<SchemaId, Vec<u8>> {
 		(4, b"dsnp.reaction".to_vec()),
 		(5, b"dsnp.update".to_vec()),
 		(6, b"dsnp.profile".to_vec()),
-		(7, b"dsnp.public-key".to_vec()),
+		(7, b"dsnp.public-key-key-agreement".to_vec()),
```
JoeCap08055, line 25: Is it intentional to duplicate `-key-key` here? Or should it just be one `-key`?
Also, I realize this comes from DSNP guidance, but `public-key-*` seems a bit confusing; shouldn't it reference the fact that it's the graph encryption key, NOT a Frequency control key, ie `public-graph-key-agreement` or something?

wesbiggs: The attestation updates to DSNP will introduce `public-key-assertion-method`, for what it's worth.

JoeCap08055 (approved):

saraswatpuneet (approved):

PR #1810, opened 2023-12-09
offchain: reverse mapping from msa to keys

### LINKED ISSUE #418: RPC to Get the List of Keys Associated with an MSA

aramikm:
# Goal
This PR adds the feature of being able to get all the keys for a certain MSA and generally implements the infra structure to be able to do indexing of desired data on offchain DB.

Closes #418 

# Diagram

reverse_mapping

# Discussions
- Offchain worker gets the finalized block with an http call to it's own rpc using `http://localhost:9944`. This means that the node should expose RPCs (at least locally). This is based on https://github.com/paritytech/substrate/issues/6742. There might be a way to be able to pass finalized block into offchain worker without RPC call but was not able to identify it so far.
- Offchain worker runs by default the following condition so we need to explicitly disallow them in cases that we don't want them to run by default 
 `- when-authority: Only enable the offchain worker when running as a validator (or collator, if this is a parachain node)` 
- There is an inherent assumption that accessing offchain indexed data after finalization should prevent us from dealing with fork situations since the canonical block should be the last block getting imported for all finalized ones. More info in https://substrate.stackexchange.com/questions/10757/offchain-indexing-storage-overwrites-during-forks
- We might want to disallow 'offchain_localStorageSet' RPC to ensure the indexed data would not be altered easily.

# Checklist
- [X] Chain spec updated
- [x] Update running scripts
- [X] Tests added


e2e/.load-test.mocharc.json
```diff
@@ -4,6 +4,6 @@
   "parallel": false,
   "require": ["scaffolding/globalHooks.ts", "scaffolding/rootHooks.ts", "scaffolding/extrinsicHelpers.ts"],
   "loader": "ts-node/esm",
-  "spec": ["./load-tests/**/*.test.ts"],
+  "spec": ["./msa/**/createMsa.test.ts"],
```
aramikm, line 7: temporary: to test adding/removing keys while indexing initial state

e2e/msa/createMsa.test.ts
aramikm, line 1: temporary: to test adding/removing keys while indexing initial state

e2e/scaffolding/helpers.ts
aramikm, line 1: temporary: to test adding/removing keys while indexing initial state

e2e/scaffolding/known_keys.ts
aramikm, line 1: temporary: to test adding/removing keys while indexing initial state

pallets/msa/src/keys.rs
aramikm, line 1: temporary: to test adding/removing keys while indexing initial state

pallets/msa/src/lib.rs
```diff
@@ -401,6 +422,81 @@ pub mod pallet {
...
+		pub _config: PhantomData<T>,
+	}
+
+	impl<T: Config> sp_std::default::Default for GenesisConfig<T> {
+		fn default() -> Self {
+			Self { accounts: 400_000, _config: Default::default() }
+		}
+	}
+	#[pallet::genesis_build]
+	impl<T: Config> BuildGenesisConfig for GenesisConfig<T> {
```
aramikm, line 470: temporary: to test adding/removing keys while indexing initial state

pallets/msa/src/tests/creation_tests.rs
```diff
@@ -7,15 +7,68 @@ use sp_weights::Weight;
...
+use sp_std::{cmp::*, collections::btree_map::BTreeMap};
+
 use common_primitives::{
 	msa::{DelegatorId, MessageSourceId, ProviderId},
 	node::BlockNumber,
 	utils::wrap_binary_data,
 };
 
+#[test]
+pub fn iteration_test() {
```
aramikm, line 22: experimental: will be removed

pallets/msa/src/offchain_storage.rs
```diff
@@ -0,0 +1,464 @@
...
+
+/// number of blocks to explore when trying to find the block number from block hash
+const NUMBER_OF_BLOCKS_TO_EXPLORE: u32 = 1000;
+
+/// HTTP request deadline in milliseconds
+pub const HTTP_REQUEST_DEADLINE_MILLISECONDS: u64 = 2000;
+
+/// LOCAL RPC URL and port
+/// warning: this should be updated if rpc port is set to anything different from 9944
+pub const RPC_FINALIZED_BLOCK_REQUEST_URL: &'static str = "http://localhost:9944";
```
aramikm, line 67: It would have been great if we could pass this as an input param but not sure if it's possible but open to suggestions.

wilwade: We should see if we cannot derive it, while `9944` is the default, it uses the `rpc-port` cli arg.

rlaferla:
You should add a testing instructions section to the PR.  Should we start with:  `make start-interval-with-offchain` ?

pallets/msa/src/offchain_storage.rs
```diff
@@ -0,0 +1,464 @@
...
+		StorageValueRef::persistent(LAST_PROCESSED_BLOCK_STORAGE_NAME);
+
+	// setting current_block-1 as the last processed so that we start indexing from current_block
+	let target_block: BlockNumberFor<T> =
+		current_block_number.saturating_sub(BlockNumberFor::<T>::one());
+	last_processed_block_storage.set(&target_block);
+}
+
+/// Add a key into MSA offchain DB
+fn add_key_to_offchain_msa<T: Config>(key: T::AccountId, msa_id: MessageSourceId) {
```
wilwade, line 254: Suggestion: I was confused by having both this and the `process_offchain_events`

- Idea 1: Could we just construct the event and run it through `process_offchain_events`?
- Idea 2: Rename / comment this to better reflect that this is dealing with it from on-chain storage.

wilwade (commented):
A few comments from reviewing the code, but most are suggestions and may be disregarded if you disagree.

common/primitives/src/offchain.rs
```diff
@@ -0,0 +1,132 @@
+use crate::msa::MessageSourceId;
+use numtoa::NumToA;
+use parity_scale_codec::Decode;
+use sp_runtime::offchain::storage::{StorageRetrievalError, StorageValueRef};
+use sp_std::{fmt::Debug, vec, vec::Vec};
+
+/// Lock expiration timeout in in milli-seconds for msa pallet per msa account
+pub const MSA_ACCOUNT_LOCK_TIMEOUT_EXPIRATION: u64 = 50;
```
JoeCap08055, line 8: nit: `u64` seems overkill for this. Also, would be nice to have units in the name, ie `_MS`
```suggestion
pub const MSA_ACCOUNT_LOCK_TIMEOUT_EXPIRATION_MS: u32 = 50;
```

pallets/msa/src/offchain_storage.rs
```diff
@@ -0,0 +1,464 @@
...
+const BLOCK_EVENT_KEY: &[u8] = b"frequency::block_event::msa::";
+
+/// Offchain index for MSA events count
+const BLOCK_EVENT_COUNT_KEY: &[u8] = b"frequency::block_event::msa::count::";
+
+/// Lock expiration timeout in in milli-seconds for initial data import msa pallet
+const MSA_INITIAL_LOCK_TIMEOUT_EXPIRATION: u64 = 3000;
+
+/// Lock expiration block for initial data import msa pallet
+const MSA_INITIAL_LOCK_BLOCK_EXPIRATION: u32 = 30;
```
JoeCap08055, line 36: what are the units for this--sec or msec? Add units to const name.

pallets/msa/src/offchain_storage.rs
```diff
@@ -0,0 +1,464 @@
...
+pub const MSA_INITIAL_INDEXED_STORAGE_NAME: &[u8; 25] = b"Msa::ofw::initial-indexed";
+
+/// Lock name for last processed block number events
+const LAST_PROCESSED_BLOCK_LOCK_NAME: &[u8; 35] = b"Msa::ofw::last-processed-block-lock";
+
+/// lst processed block storage name
+pub const LAST_PROCESSED_BLOCK_STORAGE_NAME: &[u8; 30] = b"Msa::ofw::last-processed-block";
+
+/// Lock expiration timeout in in milli-seconds for last processed block
+const LAST_PROCESSED_BLOCK_LOCK_TIMEOUT_EXPIRATION: u64 = 5000;
```
JoeCap08055, line 51:
```suggestion
const LAST_PROCESSED_BLOCK_LOCK_TIMEOUT_EXPIRATION_MS: u32 = 5000;
```

pallets/msa/src/offchain_storage.rs
```diff
@@ -0,0 +1,464 @@
...
+
+/// Lock expiration for last processed block
+const LAST_PROCESSED_BLOCK_LOCK_BLOCK_EXPIRATION: u32 = 20;
```
JoeCap08055, line 54: units?

pallets/msa/src/offchain_storage.rs
```diff
@@ -0,0 +1,464 @@
...
+
+/// number of previous blocks to check to mitigate offchain worker skips processing any block
+const NUMBER_OF_PREVIOUS_BLOCKS_TO_CHECK: u32 = 5u32;
+
+/// number of blocks to explore when trying to find the block number from block hash
+const NUMBER_OF_BLOCKS_TO_EXPLORE: u32 = 1000;
+
+/// HTTP request deadline in milliseconds
+pub const HTTP_REQUEST_DEADLINE_MILLISECONDS: u64 = 2000;
```
JoeCap08055, line 63: is `u64` necessary?

node/service/src/rpc/mod.rs
```diff
@@ -82,7 +84,7 @@ where
 	module.merge(TransactionPayment::new(client.clone()).into_rpc())?;
```
rlaferla, line 84: Can `client.clone()` be performed once?

```
let client_clone = client.clone();
module.merge(MessagesHandler::new(client_clone).into_rpc())?;
module.merge(SchemasHandler::new(client_clone).into_rpc())?;
```

node/service/src/service.rs
```diff
@@ -262,9 +266,39 @@ async fn start_node_impl(
...
+				enable_http_requests: true,
+				custom_extensions: |_| vec![],
+			});
+
+		// Spawn a task to handle off-chain notifications.
+		// This task is responsible for processing off-chain events or data for the blockchain.
+		task_manager.spawn_handle().spawn(
+			"offchain-workers-runner",
+			"offchain-work",
+			offchain_workers.run(client.clone(), task_manager.spawn_handle()).boxed(),
```
rlaferla, line 291: Consider cloning the client once and using it in both places.

rlaferla, line 291: Can this throw an error?  If so, can it be caught/logged?

`    log::error!("Error running off-chain workers: {}", err);
`

pallets/msa/src/offchain_storage.rs
```diff
@@ -0,0 +1,464 @@
...
+
+/// LOCAL RPC URL and port
+/// warning: this should be updated if rpc port is set to anything different from 9944
+pub const RPC_FINALIZED_BLOCK_REQUEST_URL: &'static str = "http://localhost:9944";
+/// request body for getting last finalized block from rpc
+pub const RPC_FINALIZED_BLOCK_REQUEST_BODY: &[u8; 78] =
+	b"{\"id\": 10, \"jsonrpc\": \"2.0\", \"method\": \"chain_getFinalizedHead\", \"params\": []}";
+
+/// stores the event into offchain DB using offchain indexing
+pub fn offchain_index_event<T: Config>(event: &Event<T>, msa_id: MessageSourceId) {
```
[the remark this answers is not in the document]
JoeCap08055: I'd be for one of those. Worth a discussion on what is the better model:
* One generic `MsaKeyModified` event:
    * Pro: It's generic :smile:
    * Con: bigger footprint in the block (added enum) (but probably only the size of a `u8`)
    * Con(?): Impacts more extrinsics
    * Con(?): clients have to decode the entire event payload to determine the action, if they're filtering on Add/Update/Delete.
* Just do `PublicKeyDeletedV2`
    * Pro: Limited scope of changes to chain code
    * Pro: Minimal client impact
    * Con(?): Generic event might be better design-wise
    
(of course, neither is strictly necessary, works as-is)

pallets/msa/src/offchain_storage.rs
```diff
@@ -0,0 +1,464 @@
...
+		match event {
+			Event::MsaCreated { msa_id, key } =>
+				Some(Self::IndexedMsaCreated { msa_id: *msa_id, key: key.clone() }),
+			Event::PublicKeyAdded { msa_id, key } =>
+				Some(Self::IndexedPublicKeyAdded { msa_id: *msa_id, key: key.clone() }),
+			Event::PublicKeyDeleted { key } =>
+				Some(Self::IndexedPublicKeyDeleted { msa_id, key: key.clone() }),
+			_ => None,
+		}
+	}
```
JoeCap08055, line 234: Should we check that the `msa_id` passed in matches the `msa_id` in the event in the case of `Event::MsaCreated` and `Event::PublicKeyAdded`?

pallets/msa/src/rpc/src/lib.rs
```diff
@@ -53,39 +55,61 @@ pub trait MsaApi<BlockHash, AccountId> {
...
+	pub fn new(client: Arc<C>, offchain: Option<OffchainDB>) -> Self {
+		Self { client, offchain: Arc::new(RwLock::new(offchain)), _marker: Default::default() }
+	}
+}
+
+/// Errors that occur on the client RPC
+#[derive(Debug)]
+pub enum MsaOffchainRpcError {
+	/// Error acquiring lock
+	ErrorAcquiringLLock,
```
JoeCap08055, line 88:
```suggestion
	ErrorAcquiringLock,
```

scripts/init.sh
```diff
@@ -126,6 +135,8 @@ start-frequency-interval)
     --rpc-external \
     --rpc-cors all \
     --rpc-methods=Unsafe \
+    --sealing-create-empty-blocks \
```
JoeCap08055, line 138: Probably not a big deal given the limited scope/use of this script, but worth noting that this is a significant change in behavior for `start-frequency-interval` which will now keep forming empty blocks every `${interval}` seconds. For anyone that tends to leave their dev chain running, this means that older blocks will roll off the history unless you manually tweak the params to enable archiving.

Again, not a big deal as it's a dev script, but worth noting.

js/api-augment/definitions/msa.ts
```diff
@@ -47,14 +36,24 @@ export default {
       ],
       type: "Option<Vec<SchemaGrantResponse>>",
     },
+    getKeysByMsaId: {
```
mattheworris, line 39: If I run `make start-interval-with-offchain`, should I be able to access this via polkadot UI->Developer->RPC calls->msa?

node/service/src/rpc/mod.rs
aramikm: Unfortunately that is not how Rust works. In passing the client we are moving the instance inside the constructor so we need a clone. And since client is of type `Arc` cloning is not expensive at all.

node/service/src/service.rs
aramikm: Not possible based on how Rust works!

aramikm: based on what I can see spawn gets a task of `Future<Output = ()> ` type. So it does not return a Result type and the inner code provided by substrate doesn't seem to have a panic scenario that can be handled from anywhere inside here.

js/api-augment/definitions/msa.ts
aramikm: You can access that RPC directly by http call without having `with-offchain` but it would be empty or error out. My impression is that to be able to see the rpc from polkadot UI I think we need to submit PR on polkadot UI once in a while to be able to access them via UI.

common/primitives/src/offchain.rs
aramikm: Not my choice. `from_millis(millis: u64)` only have u64 as param. Thanks for suggestion I'll rename it to reflect the type.

pallets/msa/src/offchain_storage.rs
aramikm: Yes

pallets/msa/src/offchain_storage.rs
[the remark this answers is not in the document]
aramikm: the bigger question is that which execution strategy we would like it to use from following. My impression was that we might want to use the wasm one since we can update them without node upgrades but it might be a bit slower and might inflate the wasm. I think the default right now is `native` by default but I might be wrong. 

```
      --execution-offchain-worker <STRATEGY>
          The means of execution used when calling into the runtime while using an off-chain worker

          Possible values:
          - native:           Execute with native build (if available, WebAssembly otherwise)
          - wasm:             Only execute with the WebAssembly build
          - both:             Execute with both native (where available) and WebAssembly builds
          - native-else-wasm: Execute with the native build if possible; if it fails, then execute with WebAssembly

```

scripts/init.sh
aramikm: I think it's more closer to reality to have this as the default for `start-frequency-interval`, but as you said might not work as some previously expected so we can maybe parking lot it to get more opinions.

pallets/msa/src/offchain_storage.rs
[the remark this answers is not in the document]
mattheworris: Taken with a grain of salt, I found this in the Substrate docs:

> However, offchain workers run in their own WebAssembly execution environment

This makes sense to me, as the option to run the node in native mode has been removed, so I'd be surprised if the OCW could run in native mode. I'm still trying to confirm the execution strategy for OCW, but I think `wasm` is the way forward.

mattheworris (commented):
Read through all the code changes, lots of good work to absorb.
Ran all the tests, read docs about offchain workers.

js/api-augment/definitions/msa.ts
JoeCap08055: Correct--extrinsic changes are visible immediately as they rely on on-chain information, but knowledge of RPCs is built into the polkadot-js JS package, so you have to wait for a Frequency release and _then_ the next PolkadotJS release after that to access new RPCs in the UI.

node/service/src/service.rs
saraswatpuneet: yeah spawn will create a tokio task and run it until it ends , i think task_manager sent to offchain_worker manages that, just a thought

pallets/msa/src/rpc/src/tests/mod.rs
```diff
@@ -175,11 +178,58 @@ async fn get_granted_schemas_by_msa_id_with_no_delegation() {
...
+
+	let result = api.get_keys_by_msa_id(NOT_EXIST_MSA);
+
+	assert_eq!(true, result.is_ok());
+	let response = result.unwrap();
+	assert_eq!(None, response);
+}
+
+#[tokio::test]
+async fn get_keys_by_msa_id_with_value_should_work() {
```
shannonwells, line 217: nit: would prefer this test retrieving multiple account ids for an MSA, even though you check that those are all pushed to the db in the offchain_storage test.
```
#[tokio::test]
async fn get_keys_by_msa_id_with_value_should_work() {
	let msa_id: MessageSourceId = 10;
	let accounts = vec![Public([1u8; 32]), Public([2u8; 32]), Public([5u8; 32])];
	let client = Arc::new(TestApi {});
	let mut db = TestPersistentOffchainDB::new();
	db.set(
		sp_offchain::STORAGE_PREFIX,
		&get_msa_account_storage_key_name(msa_id),
		&accounts.encode(),
	);
	let api = MsaHandler::<TestApi, Block, TestPersistentOffchainDB>::new(client, Some(db));

	let result = api.get_keys_by_msa_id(msa_id);

	assert_eq!(true, result.is_ok());
	let response = result.unwrap();
	assert_eq!(true, response.is_some());
	let expected_response = vec![
		KeyInfoResponse { msa_id, key: accounts[0] },
		KeyInfoResponse { msa_id, key: accounts[1] },
		KeyInfoResponse { msa_id, key: accounts[2] },
	];
	assert_eq!(, response.unwrap());
}
```

shannonwells (approved):
Looks good, works for me, I have only non-blocking comments.

* I ran the tests locally,
* I ran make start-with-offchain, and CURLed the new RPC to retrieve 3 keys that were associated with an MSA Id.
* I ran make start and retried the RPC to ensure it didn't crash and got the OffchainIndexingNotEnabled error.
* I altered and ran a test which initialized the Offchain storage with "None" for offchain and got same error.

js/api-augment/definitions/msa.ts
```diff
@@ -47,14 +36,24 @@ export default {
...
+      description: "Fetch Keys for an MSA Id",
+      params: [
+        {
+          name: "msa_id",
+          type: "MessageSourceId",
+        },
+      ],
+      type: "Option<Vec<KeyInfoResponse>>",
```
JoeCap08055, line 47: hmm, thought I left this comment before but I can't seem to find it.
Since `KeyInfoResponse` is only used by this RPC, it seems like it would be more efficient not to duplicate `msa_id` in each array item in the response. So perhaps the definition should be:
```
KeyInfoResponse: {
      keys: "Vec<AccountId>",
      msa_id: "MessageSourceId",
}
```

and then the RPC return value would just be `Option<KeyInfoResponse>` instead of `Option<Vec<KeyInfoResponse>>`

node/service/src/service.rs
```diff
@@ -262,9 +266,39 @@ async fn start_node_impl(
 		})
 		.await?;
 
+	// Start off-chain workers if enabled
+	if parachain_config.offchain_worker.enabled {
+		log::info!("OFFCHAIN WORKER is Enabled!");
+		let offchain_workers =
+			sc_offchain::OffchainWorkers::new(sc_offchain::OffchainWorkerOptions {
+				runtime_api_provider: client.clone(),
+				is_validator: validator,
```
saraswatpuneet, line 275: I see the offchain workers have access to runtime API via the client, we can use that to do any runtime call ?

pallets/msa/src/offchain_storage.rs
aramikm: created another issue to track this https://github.com/LibertyDSNP/frequency/issues/1846

pallets/msa/Cargo.toml
```diff
@@ -29,6 +29,13 @@ sp-std = { workspace = true }
 sp-weights = { workspace = true }
 # Frequency related dependencies
 common-primitives = { default-features = false, path = "../../common/primitives" }
+serde = { workspace = true, features = ["derive"] }
+serde_json = { workspace = true, features = [
+	"alloc",
+] }
+hex = { version = "0.4", default-features = false, features = [
```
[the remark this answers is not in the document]
aramikm: fixed

pallets/msa/src/rpc/Cargo.toml
```diff
@@ -22,6 +22,9 @@ sp-api = { workspace = true }
 sp-std = { workspace = true }
 sp-blockchain = { workspace = true }
 sp-runtime = { workspace = true }
+sp-core = { workspace = true }
+sp-offchain = { workspace = true }
+parking_lot = "0.12.1"
```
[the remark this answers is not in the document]
aramikm: fixed

pallets/msa/src/offchain_storage.rs
```diff
@@ -0,0 +1,464 @@
...
+use sp_std::{collections::btree_map::BTreeMap, fmt::Debug, vec, vec::Vec};
+
+/// Block event storage prefix
+const BLOCK_EVENT_KEY: &[u8] = b"frequency::block_event::msa::";
+
+/// Offchain index for MSA events count
+const BLOCK_EVENT_COUNT_KEY: &[u8] = b"frequency::block_event::msa::count::";
+
+/// Lock expiration timeout in in milli-seconds for initial data import msa pallet
+const MSA_INITIAL_LOCK_TIMEOUT_EXPIRATION: u64 = 3000;
```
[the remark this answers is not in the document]
aramikm: renamed

pallets/msa/src/offchain_storage.rs
aramikm: renamed

pallets/msa/src/offchain_storage.rs
aramikm: renamed

scripts/init.sh
aramikm: removing this so that it does not violate the current expectations

node/service/src/service.rs
aramikm: since offchain worker is part of the wasm it can access the runtime function but to apply changes to on-chain data it should call one of the following functions
```
Signed Transactions
Unsigned Transactions w/ Signed Payloads
Unsigned Transactions
```

js/api-augment/definitions/msa.ts
aramikm: applied

pallets/msa/src/offchain_storage.rs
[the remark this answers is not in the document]
aramikm: I think having the generic event idea effort is outside the scope of this PR.
One reason to keep the type of regular events separate from indexed events is that in case we would like to do offchain indexing on other things we can not always guarantee all the data would be in the event and I don't think it's an extendable approach.

pallets/msa/src/offchain_storage.rs
aramikm: it doesn't matter tbh. We are always using the msa provided by event (you can tell by `*msa`) if it exists and only use the param one for delete which don't have any msa_id.

pallets/msa/src/lib.rs
```diff
@@ -401,6 +413,61 @@ pub mod pallet {
...
+					},
+				};
+			log::info!(
+				"last finalized block number {:?} and hash {:?}",
+				finalized_block_number,
+				last_finalized_hash
+			);
+
+			match offchain_index_initial_state::<T>(finalized_block_number) {
+				LockStatus::Locked => {
```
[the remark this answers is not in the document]
aramikm: In general should not be a problem since It's only a few blocks that we are going to do some extra rpc calls until all the state is imported

pallets/msa/src/offchain_storage.rs
JoeCap08055: Ah, I see... nit: maybe rename the function input parameter to be distinct from `msa_id` inside the event? It caused me some confusion.

mattheworris (approved):
🚢 it!

saraswatpuneet (approved):

JoeCap08055 (approved):

wilwade:
Just a comment that I have confirmed that this is disabled by default:

```
      --enable-offchain-indexing <ENABLE_OFFCHAIN_INDEXING>
          Enable Offchain Indexing API, which allows block import to write to Offchain DB. Enables a runtime to write directly to a offchain workers DB during block import

          [default: false]
```

JoeCap08055 (approved):
* Ran unit tests
* Ran e2e tests
* Tested new RPC against local chain

JoeCap08055:
note: Should create an issue to add e2e test cases for the new RPC. (Unfortunately, cannot call new RPC from the e2e test suite until after we've released a new Frequency JS package _and_ Polkadot then releases a new PolkadotJS library

aramikm:
@wilwade everything should work from definitions perspective.

JoeCap08055:
> > (Unfortunately, cannot call new RPC from the e2e test suite until after we've released a new Frequency JS package _and_ Polkadot then releases a new PolkadotJS library
> 
> We should be able to write an e2e test as the e2e tests build the js against the active code base. @JoeCap08055 / @aramikm is something not working with the changes to `js/api-augment/definitions/msa.ts`

Maybe I'm misremembering something, but even though we build `@frequencychain/api-augment` locally, for RPCs it still requires that package to be integrated into the `@polkadot/*` build in order for them to be really callable; otherwise we get the types but not the RPC itself. I seem to remember this from working on stateful storage initially. Or am I wrong?

wilwade:
> Or is that maybe only to have it show up in the UI?

Right. PolkadotJS requires the upgrade before it can be called there unless you manually build and install the augmented package directly.

aramikm:
I think for RPCs to show up in UI we would need to open a PR on PolkadotJs UI repo.

wilwade:
It will pick up our updates occasionally, but if we wanted them sooner, we'd have to do a PR.

JoeCap08055:
> > Or is that maybe only to have it show up in the UI?
> 
> Right. PolkadotJS requires the upgrade before it can be called there unless you manually build and install the augmented package directly.

Right; now I remember--I was prepping for the stateful-storage demo & wanted the RPCs to show up in the UI, but apparently it requires more than just building & installing our package; there's actually a code update that gets done by PolkadotJS when they release to pull in our latest definitions. It took me down a rabbit hole to figure out how it gets done, and I gave up 🤷🏻

PR #1844, opened 2024-01-19
fix: MSA owner should be able to invoke non-signature methods on SignatureRequired schemas directly

### LINKED ISSUE #1793: RequiredSignature schemas should work on all extrinsics on stateful-pallet if called from owner
Situation: If a user uses their on key for the transaction, a signature is still required.

Example: A user with tokens wants to remove some stateful storage data.

List of Extrinsics this might apply to and could need a version bump:
- `apply_item_actions_with_signature`
- `upsert_page_with_signature`
- `delete_page_with_signature`

Questions:
- Is this a desired change?
- What is the cost of the change?
- Should `delete_page_with_signature` be a free transaction if from the MSA owner?
- Is this a larger problem? (See Ideas C, D, E)

Ideas on how to make it happen:
- A. Allow for a `0x` signature value when it is from the owner, but do not change the signature.
- B. Create a separate set of extrinsics specifically for use with the owner as origin
- C. Create a new set of extrinsics to be used either by the owner origin, or via a wrapper extrinsics that reads the signature to make a new owner origin that is then passed to the "correct" extrinsic with the "correct" origin
- D. Create a new MSA owner origin in the MSA pallet and then route all sorts of stuff through a new extrinsic and all signature "required" extrinsics now go through that instead of via a separate signature. 
- E. Create a new proxy origin in a new pallet and then route all sorts of stuff through a new extrinsic and all signature "required" extrinsics now go through that instead of via a separate signature. It could also hold on to the signature registry and some things could require use of that.

JoeCap08055:
# Goal
The goal of this PR is to make it possible for MSA owners to call non-signature stateful-storage methods directly without requiring a separate payload signature.

Adds e2e test scenarios to validate this correct behavior for both signature and non-signature stateful-storage extrinsics

Closes #1793 

# Discussion
## Testing
* Run e2e tests for the `stateful-storage` pallet against an unmodified frequency chain; should result in ~ 6 errors
* Run e2e tests against a locally-built chain from this branch; errors should be resolved

# ~TODO~
- [x] Add Rust unit tests

# Checklist
- [ ] Chain spec updated
- [ ] Design doc(s) updated
- [x] Tests added
- [ ] Benchmarks added
- [ ] Weights updated


wilwade (commented):
These aren't the signature require extrinsics, but I think all of your tests should pass without any changes to the lib.rs due to https://github.com/LibertyDSNP/frequency/blob/c92466510fb78a4885a8977d1d7444916ff77465/pallets/stateful-storage/src/lib.rs#L762-L763

JoeCap08055:
> These aren't the signature require extrinsics, but I think all of your tests should pass without any changes to the lib.rs due to
> 
> https://github.com/LibertyDSNP/frequency/blob/c92466510fb78a4885a8977d1d7444916ff77465/pallets/stateful-storage/src/lib.rs#L762-L763

That's right; these _aren't_ the signature required extrinsics. That's not the point of the PR; in fact, I pointed out in the original issue [here](https://github.com/LibertyDSNP/frequency/issues/1793#issuecomment-1900669371) that the problem seems to have been mis-stated. The signature-required extrinsics always worked correctly with a signature on a `SignatureRequired` schema, but I believe the non-signature extrinsics should also work for said schemas if invoked by the schema owner.

I verified that against the current Frequency release, my new e2e tests _fail_; with these modifications, they pass.

wilwade:
> That's right; these _aren't_ the signature required extrinsics. That's not the point of the PR; in fact, I pointed out in the original issue [here](https://github.com/LibertyDSNP/frequency/issues/1793#issuecomment-1900669371) that the problem seems to have been mis-stated. The signature-required extrinsics always worked correctly with a signature on a `SignatureRequired` schema, but I believe the non-signature extrinsics should also work for said schemas if invoked by the schema owner.

Ah I see now. Sorry about that. I missed connecting this with the other. Calling the normal one instead of the `with_signature` is a nice way around this issue.

Will re-review.

.github/workflows/verify-pr-commit.yml
```diff
@@ -184,7 +184,7 @@ jobs:
         uses: actions/checkout@v4
       - name: Set Up Cargo Deny
         run: |
-          cargo install --force cargo-deny
+          cargo install --force --locked cargo-deny
```
wilwade, line 187: 👍

e2e/stateful-pallet-storage/handleItemized.test.ts
```diff
@@ -104,7 +105,7 @@ describe('📗 Stateful Pallet Storage', function () {
       const add_actions = [add_action];
       const fake_schema_id = new u16(ExtrinsicHelper.api.registry, 65_534);
       const itemized_add_result_1 = ExtrinsicHelper.applyItemActions(
-        providerKeys,
+        delegatorKeys,
```
wilwade, line 108: (you might have already thought about and been waiting before doing this bit)

We'll want to have separate tests for the delegation test vs the direct user test.

Or perhaps with the tests in `e2e/stateful-pallet-storage/handleSignatureRequired.test.ts` should these be reverted to just be testing the delegation with `providerKeys`?

e2e/stateful-pallet-storage/handlePaginated.test.ts
```diff
@@ -142,7 +143,7 @@ describe('📗 Stateful Pallet Storage', function () {
       const payload_1 = new Bytes(ExtrinsicHelper.api.registry, 'Hello World From Frequency');
       const fake_schema_id = new u16(ExtrinsicHelper.api.registry, badSchemaId);
       const paginated_add_result_1 = ExtrinsicHelper.upsertPage(
-        providerKeys,
+        delegatorKeys,
```
wilwade, line 146: Same here as with `e2e/stateful-pallet-storage/handleItemized.test.ts` Perhaps we should revert to provider here if the delegator calls are covered by `e2e/stateful-pallet-storage/handleSignatureRequired.test.ts`?

Or do we need to duplicate these?

pallets/stateful-storage/src/lib.rs
```diff
@@ -273,8 +273,14 @@ pub mod pallet {
 		) -> DispatchResult {
 			let key = ensure_signed(origin)?;
 			let is_pruning = actions.iter().any(|a| matches!(a, ItemAction::Delete { .. }));
-			Self::check_schema_for_write(schema_id, PayloadLocation::Itemized, false, is_pruning)?;
-			Self::check_msa_and_grants(key, state_owner_msa_id, schema_id)?;
+			let caller_msa_id = Self::check_msa_and_grants(key, state_owner_msa_id, schema_id)?;
+			let is_signed = caller_msa_id == state_owner_msa_id;
```
wilwade, line 277: (reminder before switching out of draft) We should note in the docs of the `_with_signature` and perhaps also of these how this works to cross reference it.

wilwade (commented):
Now that I understand what is going on, this is a great fix! A few notes, but my guess is that those were mostly already on your todo list for before it gets out of draft mode

e2e/stateful-pallet-storage/handlePaginated.test.ts
JoeCap08055: The itemized and paginated extrinsics are handled in 2 separate files. Similar to the other test referenced, I had to change the caller here because the order of tests in the extrinsic has changed--this test would always have failed for an invalid delegate, but that was not the error thrown because it was checking the schemaId first; now it checks the delegation first and was throwing a different error, so I tweaked the test to pass for the delegate so we could get the desired `InvalidSchemaId` error.

e2e/stateful-pallet-storage/handleItemized.test.ts
wilwade: I guess it doesn't impact with who is sending it, I was worried that there was a logical difference between the paths for provider vs delegate, but I guess we are safe with either one making the call. The error is the same.

pallets/stateful-storage/src/lib.rs
JoeCap08055: Updated docs.

wilwade (approved):
I haven't run the code locally, but this captures a nice simple solution to the problem with minimal code changes. (and a lot of new tests)

- Read code
- Validated it solves the original issue

aramikm (approved):
Looks good, added some nits. Spec version needs to get updated!

- Ran the local chain
- Ran unit & e2e tests

shannonwells (approved):
I only reviewed the code, had read the related issue + discussion fairly well.  Looks like an elegant solution to me. One naming suggestion but NBD.

pallets/stateful-storage/src/lib.rs
```diff
@@ -296,8 +312,10 @@ pub mod pallet {
 		) -> DispatchResult {
 			let provider_key = ensure_signed(origin)?;
 			ensure!(page_id <= T::MaxPaginatedPageId::get(), Error::<T>::PageIdExceedsMaxAllowed);
-			Self::check_schema_for_write(schema_id, PayloadLocation::Paginated, false, false)?;
-			Self::check_msa_and_grants(provider_key, state_owner_msa_id, schema_id)?;
+			let caller_msa_id =
+				Self::check_msa_and_grants(provider_key, state_owner_msa_id, schema_id)?;
+			let is_signed = caller_msa_id == state_owner_msa_id;
+			Self::check_schema_for_write(schema_id, PayloadLocation::Paginated, is_signed, false)?;
```
saraswatpuneet, line 318: `is_state_owner`?

enddynayn (approved):
wonderful! 👍

saraswatpuneet (approved):

PR #1851, opened 2024-01-23
Bug transacting with capacity should not require tokens to cover transaction cost

### LINKED ISSUE #1850: Bug: Transacting with Capacity Should Not Require Tokens to Cover Transaction Cost
Scenario:

1. Use `Account X` to create `Provider X`
2. `Account X` has balance of 200 UNITs
3. `Account X` Stakes to `Provider X` 199 UNITs (The max allowed as the system forces you to retain at least 1 UNIT when staking the max amount)
4. `Account X` then does some things that remove most/all of the 1 UNIT that is unfrozen.
5. `Account X` has minimal unfrozen tokens
6. `Account X` then tries to do a Capacity Transaction on `Provider X` that without Capacity would take MORE than the minimal unfrozen token balance

- Expected: Transaction should be successful as there is enough Capacity to cover the transaction
- Actual: The transaction is rejected with `1010 Invalid Transaction - Inability to pay some fees (e.g. account balance too low)`

mattheworris:
# Goal
The goal of this PR is to fix a frequency-tx-payment bug to allow capacity holders to transact when their free balance is less than the minimum balance, or Existential Deposit (ED).

Closes #1850

# Discussion

- [x] `withdraw_fee` check updated to use `total_balance` instead of `balance`, which ensures that if an account has capacity (`frozen` > 0) but has no available balance (`free == frozen`) transactions are still successful
- [x] Ensures that additional provider keys that have no ED cannot transact successfully.
- [x] Add e2e-test that successfully claims a handle with 0 available balance

# How to Test
- Ensure that e2e-tests complete successfully

# Checklist
- [ ] Chain spec updated
- [ ] Design doc(s) updated
- [x] Tests added
- [ ] Benchmarks added
- [ ] Weights updated


e2e/capacity/transactions.test.ts
```diff
@@ -576,6 +576,52 @@ describe('Capacity Transactions', function () {
         });
       });
 
+      describe('when capacity eligible transaction is from the handles pallet and balance less than ED', function () {
```
JoeCap08055, line 579: nit: description implies that the test case is unique to the handles pallet, when in fact it is for any capacity-eligible transaction (just using `claimHandle` as the test case)

```suggestion
      describe('when capacity eligible transaction and balance less than ED', function () {
```

JoeCap08055 (approved):

enddynayn (approved):
superb! 👍

aramikm (approved):
Looks good to me!

aramikm:
Only need to increase spec version

PR #1873, opened 2024-02-12
Investigate fungible trait Holds and Freezes overlap

### LINKED ISSUE #1819: Investigate fungible trait Holds and Freezes overlap.
As referenced here: [Deprecate Currency; introduce holds and freezing into fungible traits](https://github.com/paritytech/substrate/pull/12951) 
and implemented in the following PRs: #1779 #1818

The following comment indicates that we need to investigate some test cases where `Holds` are slashed and ensure that the affected pallets handle their bookkeeping for `Freezes` correctly:

>Freezes are essentially the same as locks, except that they overlap with Holds. Since Holds are designed to be infallibly slashed, this means that any logic using a Freeze must handle the possibility of the frozen amount being reduced, potentially to zero. A permissionless function should be provided in order to allow bookkeeping to be updated in this instance.

# Discussion

- [x] Implement e2e tests that exercise the overlap between `Holds` and `Freezes`
  - [x] Build a test where an account with a `Freeze` and a `Hold` and the `Hold` is slashed
  - [x] Ensure the capacity pallet handles the logic correctly without any exploits
     - [x] Check if amount staked is reduced to correlate with the amount slashed. Users should not be able to continue receiving the same capacity when tokens are slashed.
  - [ ] Ensure that a user can stake token to network and participate in governance with the same token amount (frozen amount). 
  - [ ] Ensure the time-release pallet handles the logic correctly without any exploits

# Changes
- Add e2e test with simultaneous `Hold` and `Freeze`.

# How to Test
- Confirm that the new e2e tests pass.

mattheworris:
# Goal
The goal of this PR is to investigate the overlap of `hold` and `frozen` in the frame system account, used by the balances pallet and the fungible trait (which replaced the Currency trait).

Closes #1819

# Discussion
Investigation indicates that it is not currently possible to create a situation where `frozen` overlaps with `hold` [currently indicated by `reserved` in the frame system account] and that `hold` can be slashed. 

- Use the `Democracy` pallet to create a proposal that would result in some tokens being placed on `hold`. The tokens on `hold` would always be returned to the user at some point in the future, they are not able to be slashed, as near as I can tell.
- Use the `treasury` pallet to create a spend proposal that would result in a minimum of 100 tokens being used as a bond and placed on `hold`. These tokens will be slashed if the proposal is rejected. We can simulate a rejection using `sudo` and see that the tokens are slashed.
- Validators and their Nominators can have their stake slashed. This is done using `hold`. 

However, experimentation with `hold` shows that conditions where `hold` and `frozen` overlap cause the transactions to fail. 

Tests were added to make sure this behavior does not change in the future.

# Changes
Two tests were added: 
1. Create a treasury spend proposal and then attempt to stake with an amount that would overlap the `hold`. This transaction fails.
2. Create a stake and then attempt to create a treasury spend proposal that would require an amount of tokens that would overlap with the existing stake. This transaction fails.

# How to Test

- Confirm that the e2e tests pass.

# Checklist
- [ ] Chain spec updated
- [ ] Design doc(s) updated
- [x] Tests added
- [ ] Benchmarks added
- [ ] Weights updated


shannonwells (approved):
Looks great, thanks. #themoreyouknow

enddynayn:
> Experimentation with hold shows that conditions where hold and frozen overlap cause the transactions to fail.
Interestingly, the changes will allow to slash hold but yet we validate to prevent that. I am curious to learn what new feature makes this case useful.

enddynayn (approved):
👍 Thanks!

wilwade (approved):
🚢

PR #1877, opened 2024-02-13
Switch self-hosted runners to ubicloud github runners

saraswatpuneet:
# Goal
The goal of this PR is to switch the self hosted runners in github actions (except for benchmarks as they would run on self-hosted) to `ubicloud-standard-4`, code coverage is an exception as it requires higher verticals hence using `ubicloud-standard-8` only for code coverage jobs

- Release Workflow Test: https://github.com/LibertyDSNP/frequency/actions/runs/8900764362

Tests
- [x] Run benchmarks
- [x] Check e2e tests
- [x] Check rust tests

# Checklist
- [ ] Chain spec updated
- [ ] Design doc(s) updated
- [ ] Tests added
- [ ] Benchmarks added
- [ ] Weights updated


.github/workflows/release.yml
```diff
@@ -73,7 +73,7 @@ jobs:
   run-all-benchmarks:
     needs: create-release-branch
     name: Run All Benchmarks
-    runs-on: [self-hosted, Linux, X64, benchmark]
+    runs-on: ubicloud-standard-2
```
wilwade, line 76: I think this one has to stay as it is a benchmark run

.github/workflows/merge-pr.yml
```diff
@@ -64,7 +64,7 @@ jobs:
 
   calc-code-coverage:
     name: Merge - Calculate Code Coverage
-    runs-on: [self-hosted, Linux, X64, build, v2]
+    runs-on: ubicloud-standard-8
```
saraswatpuneet, line 67: This is heavy, needs high memory so falling on standard-8 for code coverage jobs

.cargo/config.toml
```diff
@@ -41,3 +41,6 @@ rustflags = [
 	"-Dclippy::unwrap_used",
 	"-Dclippy::expect_used",
 ]
+
+[net]
+git-fetch-with-cli = true
```
saraswatpuneet, line 46: the random network issues on ubicloud warrants this check 

https://doc.rust-lang.org/cargo/reference/config.html#netgit-fetch-with-cli

.github/workflows/verify-pr-commit.yml
```diff
@@ -371,7 +371,8 @@ jobs:
           RUST_LOG=debug SRTOOL_TAG="1.70.0" srtool build \
             --build-opts="'--features on-chain-release-build,no-metadata-docs,${{matrix.features}}'" \
             --profile=${{matrix.build-profile}} \
-            --package=${{matrix.package}}
+            --package=${{matrix.package}} \
+            --root
```
saraswatpuneet, line 375: needed for permissions

saraswatpuneet:
@furkansahin  we still get some intermittent issues (seems to be network but not sure), please this job

https://github.com/LibertyDSNP/frequency/actions/runs/8051209010/job/21988632422

furkansahin:
Hey @saraswatpuneet ,
I have cloned your repository and will debug this further. I should have more to share with you in the next 2 business days. A couple of questions that would help me to pinpoint the issues would be;
1. Did you have this kind of issue before? 
2. I can see that you are switching from self hosted, is your self hosted instance located in a cloud provider? Also, what would be the location (US, europe?)

Best

saraswatpuneet:
hey @furkansahin , thank you for taking a look, we havent faced this issue before, either on self hosted or github provider runner. this is new to us too.

Regarding where we are hosting our runner, @wilwade  can add details to it below.

wilwade:
@furkansahin 

It currently uses US based AWS runners. They are `c7i.4xlarge` and `c7a.4xlarge`.

furkansahin:
Hey @saraswatpuneet , 
I can see what the problem is and reproduce it. However, since it fails randomly, it is difficult to catch the bug on the fly and reason about it. I have tried bunch of changes here in my fork https://github.com/furkansahin/frequency/actions/workflows/verify-pr-commit.yml. The problem seems like the `srtool build` is pulling bunch of repositories, the biggest being `[https://github.com/paritytech/polkadot-sdk`](https://github.com/paritytech/polkadot-sdk). During fetch, the connection is dropped, sometimes. This is not expected and I believe it has something to do with the location of our runners being in Europe. 
In your self hosted runners, do you use caching? I believe the amount of pull in your self hosted runner may not be as high as it is in ubicloud, is that assumption correct?
I can see that you have added `git-fetch-with-cli = true` to overcome the networking issue but it did not help. Is that correct? Or, do you think it actually solved something but this issue is separate? 
I am at the Devworld Conference in Amsterdam at the moment, so, unfortunately this is all I could find so far.
Best,
Furkan

furkansahin:
Hey @saraswatpuneet , @wilwade 
I have looked into this more and here is my analysis and a solution proposal for you. 
Analysis: Looking into the networking logs in detail (tcpdump + wireshark), it is correct that when cargo fetches the repositories, there are some packet drops in between. While some packet drops are totally acceptable and harmless, in this instance, they cause random connection disruption. This mostly happens when the global traffic is high (US daytime). You would not see that in your self hosted instances or regular github runners because the repositories sit closer to those. In Ubicloud's scenario, our runners are provisioned either in Germany of Finland. 
We are continuously improving our networking stack and I will look further into this problem how we can setup our networking in a way these issues would not happen. In the meantime, if you would give us a try, I found a way that makes the Verify Build Runtime jobs to pass.

Solution proposal: Unfortunately, this is a band aid proposal. I realised `git-fetch-with-cli = true` is not really necessary, but instead, you can use;
```toml
[net]
retry = 5
```

This way, cargo will retry if the fetch fails up to 5 times. In fact, in my last 27 `Verify Build Runtime for X` jobs, it had to try 3 times for only 1 of the jobs. You can see the latest one here https://github.com/furkansahin/frequency/actions/runs/8140985585. 

I understand the inconvenience this issue may cause and would completely understand any decisions you make regarding this PR. However, I really hope you will consider giving us the opportunity to demonstrate our commitment to the overall customer experience.

Best,
Furkan

furkansahin:
Hey @wilwade thanks for giving Ubicloud another try. I wanted to note that; currently, there are some issues with Microsoft's Linux repositories. Several people have reported the same problem to GitHub/Microsoft.
https://github.com/actions/runner-images/issues/9733
https://github.com/microsoft/linux-package-repositories/issues/130

We are actively monitoring this issue.
As a temporary solution, they recommend adding the following step until the Microsoft team resolves the issue with the package repository.

      - run: |
         sudo rm /etc/apt/sources.list.d/microsoft-prod.list
         sudo apt-get update

The issue manifests itself like this;
E: Failed to fetch https://packages.microsoft.com/ubuntu/22.04/prod/dists/jammy/InRelease  Clearsigned file isn't valid, got 'NOSPLIT' (does the network require authentication?)
E: The repository 'https://packages.microsoft.com/ubuntu/22.04/prod jammy InRelease' is no longer signed.
Failed to install browsers
Error: Installation process exited with code: 100

Best

.cargo/config.toml
```diff
@@ -41,3 +41,6 @@ rustflags = [
 	"-Dclippy::unwrap_used",
 	"-Dclippy::expect_used",
 ]
+
+[net]
+retry = 5
```
wilwade, line 46: Needed due to the location of the ubicloud runners

.github/workflows/common/reset-self-hosted-runner/action.yml
wilwade, line 1: Unused

.github/workflows/release.yml
```diff
@@ -17,6 +17,9 @@ on:
         description: "Release version (v#.#.#[-rc#])"
         required: true
 
+permissions:
+  contents: write
```
wilwade, line 21: This allows us to get rid of the special `GHA_GIT_COMMIT` token

.github/workflows/merge-pr.yml
```diff
@@ -12,7 +12,6 @@ jobs:
   changes:
     name: Determine Changed Files
     runs-on: ubuntu-22.04
-    container: ghcr.io/libertydsnp/frequency/ci-base-image:latest
```
wilwade, line 15: You'll see a lot of places where I removed the use of the ci-base-image. It really is only needed for things that use rust.

pallets/schemas/src/tests/other_tests.rs
```diff
@@ -490,7 +490,7 @@ fn create_schema_v3_requires_valid_schema_size() {
 				expected: (Error::<Test>::LessThanMinSchemaModelBytes, 3),
 			},
 			TestCase {
-				input: r#"{"id": "long", "title": "I am a very very very long schema", "properties": "just way too long to live a long life", "description": "Just a never ending stream of bytes that goes on for a minute too long"}"#,
+				input: r#"{"id": "long", "title": "I am a very very very very long schema", "properties": "just way too long to live a long life", "description": "Just a never ending stream of bytes that goes on for a minute too long"}"#,
```
wilwade, line 493: Triggering rust code changes

.github/workflows/verify-pr-commit.yml
```diff
@@ -366,35 +361,30 @@ jobs:
...
+          tag: "1.77.0"
       - name: Check Deterministic WASM Build Exists
         if: steps.cache-wasm.outputs.cache-hit != 'true'
-        run: file ${{env.WASM_DIR}}/${{env.BUILT_WASM_FILENAME}}
+        run: |
+          file ${{ steps.srtool_build.outputs.wasm }}
       - name: Install Subwasm
         run: |
-          cargo install --locked --git https://github.com/chevdor/subwasm --tag v0.19.1 --force
+          cargo install --locked --git https://github.com/chevdor/subwasm --tag v0.20.0 --force
```
wilwade, line 382: Highlighting version bump in subwasm

.github/workflows/verify-pr-commit.yml
```diff
@@ -366,35 +361,30 @@ jobs:
...
-            --package=${{matrix.package}}
+        uses: ./.github/workflows/common/srtool
+        env:
+          BUILD_OPTS: "--features on-chain-release-build,no-metadata-docs,${{matrix.features}}"
+        with:
+          profile: ${{matrix.build-profile}}
+          package: ${{matrix.package}}
+          chain: ${{matrix.chain}}
+          runtime_dir: ${{ matrix.runtime-dir }}
+          tag: "1.77.0"
```
wilwade, line 375: Highlighting version bump in srtool

mattheworris (approved):
- Read through code changes
- Read through comment thread

🚢 it!

JoeCap08055 (approved):
Read through changes and comments... if we're satisfied with the solutions from ubicloud  for the network drops and memory issues, :shipit:

saraswatpuneet (commented):
@wilwade  thank you for taking this ahead, great work and looks good to go
:ship:

wilwade (approved):

.github/workflows/release.yml
wilwade: This is correct. There is an option to control it such that the permission is only on specific steps instead the entire workflow, but I figured this is the best 1:1 swap for now.

PR #1880, opened 2024-02-14
fix: Update cargo install commands to use --locked flag

### LINKED ISSUE #1879: `cargo install` should use `--locked`
When using `cargo install` to install development tools, `--locked` should be used.
This requires `cargo` to use the crates that are specified in the tool's `Cargo.lock` file.

If `--locked` is not specified, `cargo` will use the latest version of the crate, and this can cause dependency issues.

mattheworris:
# Goal
The goal of this PR is prevent `cargo` dependency issues by using `--locked` with all `cargo install` commands.

Closes #1879

# Discussion
When using cargo install to install development tools, `--locked` should be used.
This requires cargo to use the crates that are specified in the tool's Cargo.lock file.

If `--locked` is not specified, cargo will use the latest version of the crate, and this can cause dependency issues.

# Changes

- add `--locked` to all `cargo install` commands in the github workflows

# How to Test

- confirm that CI passes
- Find any instances of `cargo install` that have not been updated to include `--locked` [searched for `cargo.*install` and updated all instances found]

# Checklist
- [ ] Chain spec updated
- [ ] Design doc(s) updated
- [ ] Tests added
- [ ] Benchmarks added
- [ ] Weights updated

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -319,7 +319,7 @@ jobs:
           echo "runtime_filename_${{matrix.network}}=$release_wasm_filename" >> $GITHUB_OUTPUT
       - name: Install srtool-cli
         run: |
-          cargo install --git https://github.com/chevdor/srtool-cli
+          cargo install --locked --git https://github.com/chevdor/srtool-cli
           srtool --version
       - name: Build Deterministic WASM
         run: |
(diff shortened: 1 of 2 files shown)


wilwade (approved):
🙇

saraswatpuneet (approved):

shannonwells (approved):
Yes.  Good.

PR #1948, opened 2024-04-22
Update reward pool on `provider_boost` or `unstake` #1699

### LINKED ISSUE #1699: Update `StakingRewardInfo.total_tokens_staked` on `stake` or `untake`
## Problem
In order to be able to calculate and pay out rewards, the chain must know how much in total token is staked at the end of a Reward Era.

## Solution
Every time there is a stake or unstake, update StakingRewardInfo.total_tokens_staked.

shannonwells:
# Goal
The goal of this PR is to update the StakingRewardPool on a `provider_boost` or `unstake` extrinsic call.

Closes #1699 

# Discussion
Also includes a bunch of lint changes

# Checklist
- [x] Design doc(s) updated
- [x] Tests added
- [x] Benchmarks corrected and updated
- [x] Weights updated


pallets/capacity/src/tests/mock.rs
```diff
@@ -140,8 +143,8 @@ impl StakingRewardsProvider<Test> for TestStakingRewardsProvider {
 	type RewardEra = TestRewardEra;
 	type Hash = Hash; // use what's in common_primitives::node
 
-	fn reward_pool_size(total_staked: BalanceOf<Test>) -> BalanceOf<Test> {
-		total_staked.div(10u64)
+	fn reward_pool_size(_total_staked: BalanceOf<Test>) -> BalanceOf<Test> {
```
shannonwells, line 146: To reflect new economic model behavior of having a constant RewardPool amount.

pallets/capacity/src/benchmarking.rs
```diff
@@ -125,7 +139,47 @@ benchmarks! {
 		assert_last_event::<T>(Event::<T>::EpochLengthUpdated {blocks: epoch_length}.into());
 	}
 
+	change_staking_target {
```
aramikm, line 142: Since the weights are not generated for this it is not applied on top of `change_staking_target` function. You should be able to generate the weight locally for that function to be able to add it on top of your function and then later it will be rewritten with correct weights from CI if generated

pallets/capacity/src/benchmarking.rs
```diff
@@ -125,7 +139,47 @@ benchmarks! {
...
+	verify {
+		assert_last_event::<T>(Event::<T>::StakingTargetChanged {
+			account: caller,
+			from_msa,
+			to_msa,
+			amount: restake_amount.into()
+		}.into());
+	}
+
+	provider_boost {
```
aramikm, line 166: Question: I can see that we have weights for `provider_boost` without the benchmarks since you just added it in this PR. Where does those weights come from?

pallets/capacity/src/tests/mock.rs
aramikm: Might also worth adding that as a comment in the code for the reference

aramikm (approved):
Added some questions and comments

- Read the code
- Ran the tests
- Not able to verify the arithmetic changes compared to what was expected since don't have enough context about it.

pallets/capacity/src/benchmarking.rs
shannonwells: thanks for the reminder... it's been a minute!

pallets/capacity/src/benchmarking.rs
shannonwells: I can't remember but since this is a months-old branch that has been rebased against the previous which was again a months-old branch that had to be rebased against main, I think I must have run benchmarks locally to get those weights, and it got clobbered in the rebase.

pallets/capacity/src/lib.rs
```diff
@@ -767,6 +771,10 @@ impl<T: Config> Pallet<T> {
 		CapacityLedger::<T>::insert(target, capacity_details);
 	}
 
+	fn set_reward_pool(era: <T>::RewardEra, new_reward_pool: &RewardPoolInfo<BalanceOf<T>>) {
+		StakingRewardPool::<T>::set(era, Some(new_reward_pool.clone()));
+	}
```
saraswatpuneet, line 776: Should we mutate here ? set is cool assuming it serves dual purpose of remove/update but need to check

saraswatpuneet (approved):
Just a nit

pallets/capacity/src/lib.rs
shannonwells: I'm following our previous pattern of `let mut foo = Capacity::get_a_stored_thing(key)` followed by a set.  We've been moving away from mutate.

pallets/capacity/src/benchmarking.rs
shannonwells: Done

pallets/capacity/src/benchmarking.rs
shannonwells: Also fixed these issues.  See newly generated weights, and change_staking_target now uses its correct weight function.

PR #1956, opened 2024-04-29
Implement rewards calculation formula #1941

### LINKED ISSUE #1941: Set constants and reward calculation for Provider Boost Rewards formula
## Constants values and Formula
Set the values and formula as specified in the Spreadsheet for the economics model

1. Rewards Era length
2. Rewards cap
3. Rewards claim limit in # of RewardEras
4. Capacity modifier (% Capacity gained through Provider Boosting)

shannonwells:
# Goal
The goal of this PR is to implement (but not really use yet) the chosen formula for calculation of a reward in a single Provider Boost Reward Era.

Closes #1941

# Discussion
I have tried here to balance changes to the trait as outlined in the Design Doc versus simplicity and maintainability. The simplest and most maintainable way to implement the formula was to add another function to the `RewardsProvider` trait which performs a calculation for a single era.

* The chosen economic model sets the Reward Pool tokens per era at a fixed amount.  
* The cap on Provider Boost rewards is also a fixed percentage.  So two more constants were added to this pallet rather than burying them in the trait implementation as hard-coded numbers.  These values are then used by the Capacity pallet implementation of `StakingRewardsProvider`.
* A similar `TestRewardsProvider` is implemented in the mocks.

## To Verify
`make test` should pass and `make start` should also build and run.

# Checklist
- [x] Chain spec updated
- [x] Design doc(s) updated
- [x] Tests added


common/primitives/src/node.rs
```diff
@@ -36,9 +36,6 @@ pub type Signature = MultiSignature;
 /// Index of a transaction in the chain.
 pub type Index = u32;
 
-/// the type used for the index of the Staking Reward time period
```
shannonwells, line 39: unused

pallets/capacity/src/tests/rewards_provider_tests.rs
```diff
@@ -26,43 +25,13 @@ fn test_staking_reward_total_happy_path() {
...
-fn test_payout_eligible() {
+fn payout_eligible_happy_path() {
 	new_test_ext().execute_with(|| {
 		let is_eligible = Capacity::payout_eligible(1);
 		assert!(!is_eligible);
 	})
 }
 
-#[test]
-fn test_reward_pool_size_happy_path() {
```
shannonwells, line 37: no longer valid

designdocs/provider_boosting_implementation.md
```diff
@@ -167,7 +167,12 @@ pub trait StakingRewardsProvider<T: Config> {
     /// Returns whether the claim passes validation.  Accounts must first pass `payoutEligible` test.
     /// Errors: None
     fn validate_staking_reward_claim(account_id: T::AccountID, proof: Hash, payload: StakingRewardClaim<T>) -> bool;
-}
+
+    /// Calculate a reward for a single era based on a chosen economic model
+    fn staking_reward_for_era(amount_staked: BalanceOf<T>,
+		total_staked: BalanceOf<T>,
+		reward_pool_size: BalanceOf<T>,
+    ) -> BalanceOf<T>}
```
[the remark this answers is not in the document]
shannonwells: I agree it's not the best name. We're actually not only computing rewards for the current era.  This would be called whenever staking rewards were claimed, so it would be computing rewards for each historical era they've been staking for.  

I'm also out of ideas.

pallets/capacity/src/lib.rs
```diff
@@ -1109,13 +1117,8 @@ impl<T: Config> StakingRewardsProvider<T> for Pallet<T> {
...
 	// and the other determined factors of the current economic model
-	fn reward_pool_size(total_staked: BalanceOf<T>) -> BalanceOf<T> {
-		if total_staked.is_zero() {
-			return BalanceOf::<T>::zero()
-		}
-
-		// For now reward pool size is set to 10% of total staked token
-		total_staked.checked_div(&BalanceOf::<T>::from(10u8)).unwrap_or_default()
+	fn reward_pool_size(_total_staked: BalanceOf<T>) -> BalanceOf<T> {
+		T::RewardPoolEachEra::get()
```
aramikm, line 1121: Question: Since this is a constant and it seems like based on the nature of `RewardPool` it would change and evolve. What might be the consequences of changing this in future? Since this is constant we are not able to dictate the exact time of the change so what happens if it changes between  two eras?

pallets/capacity/src/lib.rs
```diff
@@ -1141,6 +1144,21 @@ impl<T: Config> StakingRewardsProvider<T> for Pallet<T> {
...
+		amount_staked: BalanceOf<T>,
+		total_staked: BalanceOf<T>,
+		reward_pool_size: BalanceOf<T>,
+	) -> BalanceOf<T> {
+		let capped_reward = T::RewardPercentCap::get().mul(amount_staked);
+
+		let proportional_reward = reward_pool_size
+			.saturating_mul(amount_staked)
+			.checked_div(&total_staked)
+			.unwrap_or_else(|| Zero::zero());
```
aramikm, line 1158: This is going to do integer division and if we would like to round to closer integer maybe we need to use something  else or implement it differently.

runtime/frequency/src/lib.rs
```diff
@@ -493,6 +496,9 @@ impl pallet_capacity::Config for Runtime {
 	type StakingRewardsPastErasMax = ConstU32<30u32>;
 	type RewardsProvider = Capacity;
 	type MaxRetargetsPerRewardEra = ConstU32<16>;
+	// Value determined by desired inflation rate limits for chosen economic model
+	type RewardPoolEachEra = ConstU128<{ currency::CENTS.saturating_mul(172_602_740u128) }>;
```
aramikm, line 500: Question: Where does this magic number come from?

mattheworris (approved):
- Read through changes
- Executed tests

🚢 it

pallets/capacity/src/lib.rs
JoeCap08055: I got the impression from some comments elsewhere that having this be a constant is a temporary solution just to get the logic of how to compute rewards into the codebase. Not sure if that's actually the case or if I'm misremembering something I heard, but either way might be good to elaborate in a comment.

designdocs/provider_boosting_implementation.md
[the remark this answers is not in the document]
JoeCap08055: Are any of the arguments meant to be era-specific? If so, maybe change their names, like `era_amount_staked`, `era_total_staked`, etc.

pallets/capacity/src/lib.rs
shannonwells: There were some placeholders before, yes, that was when it had not been determined how it would be calculated.  But, officially now, the Reward Pool size is a fixed constant for the purposes of setting inflation rates based on calculations and estimates  (referring you back to the spreadsheet, DM me if you need the link again).  If the reward pool size changes between two eras, it won't matter because that historical data is stored in `StakingRewardPool`, and that will be used to calculate the reward for that era.  Now if the formula changes... that's a whole other ball of worms [sic].  We'll just have to cross that bridge when if and when we get to it.

runtime/frequency/src/lib.rs
shannonwells: Spreadsheet referenced above - I'm not sure it's okay to share in a public repo.  DM me for the link.

pallets/capacity/src/lib.rs
shannonwells: let's chat about that

designdocs/provider_boosting_implementation.md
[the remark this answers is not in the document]
shannonwells: How about: 

```rust
	/// Calculate the staking reward for a single era.  We don't care about the era number,
	/// just the values.
	fn era_staking_reward(
		era_amount_staked: BalanceOf<T>, // how much individual staked for a specific era
		era_total_staked: BalanceOf<T>,  // how much everyone staked for the era
		era_reward_pool_size: BalanceOf<T>, // how much token in the reward pool that era
	) -> BalanceOf<T>;
```

pallets/capacity/src/lib.rs
shannonwells: ok, Wil was the deciding factor of not caring about rounding down a planck, so it'll stay the way it is & I added a note in our chat about the exact method.

pallets/capacity/src/tests/rewards_provider_tests.rs
```diff
@@ -72,3 +41,27 @@ fn test_staking_reward_total_era_out_of_range() {
...
+		total_staked: u64,    // by all boosters
+		amount_staked: u64,   // by a single booster
+		expected_reward: u64, // for a single era
+	}
+	let test_cases: Vec<TestCase> = vec![
+		TestCase { total_staked: 1_000_000, amount_staked: 0, expected_reward: 0 }, // shouldn't happen, but JIC
+		TestCase { total_staked: 1_000_000, amount_staked: 30, expected_reward: 0 }, // rounded result
+		TestCase { total_staked: 1_000_000, amount_staked: 150, expected_reward: 1 }, // it rounds the result
+		TestCase { total_staked: 1_000_000, amount_staked: 1000, expected_reward: 4 }, // hits the cap starting with this example
+		TestCase { total_staked: 1_000_000, amount_staked: 11000, expected_reward: 42 }, // > 0.0038% of total, reward = 11000*.0038
```
shannonwells, line 59: I guess this got moved, but the main change to this test was to add the last test case.

pallets/capacity/src/lib.rs
```diff
@@ -1106,24 +1114,20 @@ impl<T: Config> StakingRewardsProvider<T> for Pallet<T> {
 	type AccountId = T::AccountId;
 	type RewardEra = T::RewardEra;
 	type Hash = T::Hash;
+	type Balance = BalanceOf<T>;
```
shannonwells, line 1117: Not only were we not using the defined types but we could also use a new one for Balance.

pallets/capacity/src/tests/change_staking_target_tests.rs
```diff
@@ -48,14 +48,14 @@ fn do_retarget_happy_path() {
 		assert_ok!(Capacity::do_retarget(&staker, &from_msa, &to_msa, &to_amount));
 
 		// expect from stake amounts to be halved
-		assert_capacity_details(from_msa, 1, 300, 1);
+		assert_capacity_details(from_msa, 15, 300, 15);
```
shannonwells, line 51: These changed due to the new formula/values

JoeCap08055 (approved):
Looks great!

pallets/capacity/src/lib.rs
```diff
@@ -203,6 +203,14 @@ pub mod pallet {
...
 		#[pallet::constant]
 		type MaxRetargetsPerRewardEra: Get<u32>;
+
+		/// The fixed size of the reward pool in each Reward Era.
+		#[pallet::constant]
+		type RewardPoolEachEra: Get<BalanceOf<Self>>;
+
+		/// the percentage cap per era of an individual Provider Boost reward
+		#[pallet::constant]
+		type RewardPercentCap: Get<Permill>;
```
wilwade, line 213: I wonder if these should eventually be government settable values instead of constants, but either way it is better this way for now.

runtime/frequency/src/lib.rs
wilwade: It is ~4.5% annually of the initial token mint (1 billion).

`14/365 * 1000000000 * 0.045 = 1726027`

@shannonwells would it be worth putting in the calculation directly? Either way, we can comment it as such that in this or a future pr.

wilwade (approved):
Two notes, but non-blocking.

- Mostly reviewed constants and `era_staking_reward`

PR #1963, opened 2024-05-07
upsert staking history #1699

shannonwells:
# Goal
The goal of this PR is to add and use storage for individual staking history so that rewards can be calculated and paid out.

Relates to #1699 

# Discussion
* Uses a [BoundedBTreeMap](https://docs.rs/bounded-collections/0.1.9/bounded_collections/bounded_btree_map/struct.BoundedBTreeMap.html#) for storage.
* Includes a new struct for this plus needed methods.
* Is used ONLY for provider boost staking
* The BTree key is for the RewardEra so staking changes within an era will update that entry.

Checking for whether a <del>stake</del> Provider Boost has been done earlier than the history retention limit (in number of RewardEras) must be left for when rewards are claimed.  The record needs to be kept around because it may still be valid for some eras.  Examples:

1.  Stake an amount, wait until +1 > history limit then claim.
    - You have only 1 entry
    - There is no mechanism to automatically reap it
    - Need to keep the record around because we should pay out remaining eras.
2. Stake an amount, change stake every era
    - No problem because if you are staking/unstaking every era the limit will be hit automatically.
3. Stake an amount, wait limit/2, unstake full amount, wait until +1 > history limit then claim
    - Not much different from example 1.  As long as there is nonzero stake within the era limit, the chain still needs to pay out those eras.
4. Say history limit is 10.  Stake an amount in era 10.  
    - Unstake everything in era 12, so era 11 is eligible for a reward.
    - But try to claim in era 22.
    - you can't since are no valid eras to claim rewards for any more (history limit + 1), however, on this claim attempt since there is no more staking, all boost history storage can be deleted.

# Checklist
- [x] Tests added


pallets/capacity/src/lib.rs
```diff
@@ -662,12 +667,9 @@ impl<T: Config> Pallet<T> {
 		target: &MessageSourceId,
 		amount: &BalanceOf<T>,
 	) -> Result<(StakingDetails<T>, BalanceOf<T>), DispatchError> {
-		// FIXME: if one is boosting additional amounts, this will fail.
```
shannonwells, line 665: Fixed.

pallets/capacity/src/lib.rs
```diff
@@ -719,11 +721,11 @@ impl<T: Config> Pallet<T> {
 			Self::get_reward_pool_for_era(era).ok_or(Error::<T>::EraOutOfRange)?;
 		reward_pool.total_staked_token = reward_pool.total_staked_token.saturating_add(*amount);
 
-		// TODO:  add boost history record for era when boost history is implemented
 		Self::set_staking_account_and_lock(staker, staking_details)?;
 		Self::set_target_details_for(staker, *target, target_details);
 		Self::set_capacity_for(*target, capacity_details);
 		Self::set_reward_pool(era, &reward_pool);
+		Self::upsert_boost_history(staker, era, *amount, true)?;
```
wilwade, line 728: I ran the weights locally on this. It bumps the boost db cost to 6 writes and 9 reads. `create_sponsored_account_with_delegation` is 7 and 10, but it is one of the highest on the chain.(`add_public_key_to_msa` is tied-ish at 7 and 8)

We just have a lot of db work that has to happen on `provider_boost`. 

I don't know of any way to reduce it, but something to be aware of.

PS, might be worth running the benchmarks on this branch so it is clear the cost of the changes.

wilwade (approved):
Nothing blocking, but a few comments

- Reviewed code
- Pulled
- Ran benchmarks locally (Note: Required typo fix to run `make benchmark-capacity`)

pallets/capacity/src/lib.rs
```diff
@@ -913,6 +914,14 @@ impl<T: Config> Pallet<T> {
 
 		capacity_details.withdraw(actual_capacity, actual_amount);
 
+		if staking_type.eq(&ProviderBoost) {
+			// update staking history
+			let era = Self::get_current_era().era_index;
+			let mut pbh =
+				Self::get_staking_history_for(unstaker).ok_or(Error::<T>::NotAStakingAccount)?;
+			pbh.subtract_era_balance(&era, &actual_amount);
```
saraswatpuneet, line 922: nit: this check can move above get_current_era , not sure if that function makes a read so we might save one one read

saraswatpuneet, line 922: nit: this check can move above get_current_era , not sure if that function makes a read so we might save one one read

pallets/capacity/src/lib.rs
```diff
@@ -913,6 +914,14 @@ impl<T: Config> Pallet<T> {
 
 		capacity_details.withdraw(actual_capacity, actual_amount);
 
+		if staking_type.eq(&ProviderBoost) {
+			// update staking history
+			let era = Self::get_current_era().era_index;
+			let mut pbh =
+				Self::get_staking_history_for(unstaker).ok_or(Error::<T>::NotAStakingAccount)?;
```
saraswatpuneet, line 921: nit: this check can move above get_current_era , not sure if that function makes a read so we might save one one read

saraswatpuneet (approved):
Lgtm, I agree with the sentiments around need for some db optimization later, but non blocking 👍🏽

pallets/capacity/src/lib.rs
```diff
@@ -194,6 +194,8 @@ pub mod pallet {
 		type EraLength: Get<u32>;
 
 		/// The maximum number of eras over which one can claim rewards
+		/// Note that you can claim rewards even if you no longer are boosting, because you
+		/// may claim rewards for past eras up to the history limit.
```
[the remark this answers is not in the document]
shannonwells: This is the same as example 4.  My understanding is that no, she can NOT claim that reward.  She has 10 eras to claim it and then the ability to claim rewards expires.  If she tries, then that is when the (future) code in the claim extrinsic would compare it to the history limit, and delete the record since not only is there nothing staked but what she did stake is far beyond the history limit.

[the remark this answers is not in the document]
wilwade: Is that bad from the cleanup perspective? Alice has no reason to make the call that would cleanup the out of date history.

Or perhaps a more general question, why not allow it to be claimed? Or is the issue we don't want to retain the information about prior total staking amounts in `StakingRewardPool`, however I don't think that is currently getting any data removed.

pallets/capacity/src/lib.rs
```diff
@@ -1036,6 +1046,26 @@ impl<T: Config> Pallet<T> {
...
+	) -> Result<(), DispatchError> {
+		let mut boost_history = Self::get_staking_history_for(account).unwrap_or_default();
+
+		let upsert_result = if add {
+			boost_history.add_era_balance(&current_era, &boost_amount)
+		} else {
+			boost_history.subtract_era_balance(&current_era, &boost_amount)
+		};
+		ensure!(upsert_result.is_some(), Error::<T>::EraOutOfRange);
+		ProviderBoostHistories::<T>::set(account, Some(boost_history));
```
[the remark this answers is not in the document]
shannonwells: One thing - unstaking from a boosted target doesn't need a separate extrinsic, since at some point it was decided not to support staking to the same target in two different ways.  Once the account is found it will know what type of staking was done.

Yes, to your last point.  If you did a double key map then that would make it harder to put limits on staking history when looking for the oldest record.  You would still need to iterate over all the ones in there on a new `provider_boost` to ensure the history limit is respected, so it doesn't save you anything.

The `BoundedBTreeMap` is specifically supposed to be a balance of memory use and performance between a BTreeMap and a BoundedVec, and it gives you convenience methods like `first_key_value` and `last_key_value` if you need/want meaningfully sortable keys.

So here is another example:
1.  Say the history limit is 10, and I stake 100 in era 10.
2.  Over the next 20 eras I stake 10 another 9 times. I have filled up my history limit, and I still have the first record.
3.  I go to claim in era 31, but the rewards for eras from 11-19 have expired.  I get only the rewards from eras 20-29 (maybe there is an off-by-one error, but you get the picture)

pallets/capacity/src/lib.rs
shannonwells: There is other benchmark work to do for on_initialize after adding the reward pool / era updates.  I will add this task to that issue.

pallets/capacity/src/lib.rs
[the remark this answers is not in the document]
shannonwells: The problem is the overall RewardPool history retention.  We can't keep that history around forever.  Without that, there's no way of calculating old rewards.  If Alice has no reason to make the call that would clean up the out of date history, it's possible there could be some periodic check that will do it, perhaps run manually, or even on a runtime upgrade.

However I'm curious as to how much of a problem you anticipate that would be since it's literally in people's interest to claim their rewards before they expire.

pallets/capacity/src/lib.rs
shannonwells: NB: This particular PR doesn't bump the provider boost branch but 1r and 1w.  It was already at 5w and 8r.

pallets/capacity/src/lib.rs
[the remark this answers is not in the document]
wilwade: True, I'm mostly just thinking through consequences, but I see as you say that `start_new_reward_era_if_needed` has the removal, so I had just missed that piece. All good, thanks!

PR #1966, opened 2024-05-09
Chores/update capacity benchmarks #1949

### LINKED ISSUE #1949: Check maximum weight of on_initialize when starting a new era.
We need to get a maximum weight for `start_new_era_if_needed` so we know whether it's possible to overflow a block on a new reward era.  I don't think it's going to be problem but we should have the data.

# Acceptance Criterion
- [ ] Update test/benchmark for the entire on_initialize weight when starting a new era.

shannonwells:
# Goal
The goal of this PR is to update the benchmark for `on_initialize` to include the weight when a new RewardEra must be created.

Closes #1949

# Discussion
Separates the benchmark for on_initialize into two, `start_new_epoch_if_needed` and `start_new_reward_era_if_needed`.  Those functions return the WeightInfo for their respective benchmarks.

For `start_new_reward_era_if_needed`, fills up the reward pool history and sets up the block and era to ensure that a new RewardEra will be created during the benchmark.

Results:  The new weights are a lot larger, 19M as compared to 4M,  6 reads as compared to 2, 4 writes as compared to 1, so there's some optimization work to be done.  Filed new Issue #1967 

# Checklist
- [x] Benchmarks updated
- [x] Weights updated


aramikm (commented):
I don't think this is correct. If you look at the `on_initialize` it consists of `start_new_epoch_if_needed` weights + `start_new_reward_era_if_needed` weights and inside each of them you are reusing `T::WeightInfo::on_initialize()` which makes this whole calculation kinda recursive!?

I think one way to fix this is to benchmark `start_new_epoch_if_needed` and `start_new_reward_era_if_needed` separately and maybe add them to get the weights for `on_initialize`

shannonwells:
You are correct, I'd completely overlooked that it was doing that, and I'm not sure why it was set up that way.

aramikm (approved):
Looks great! Ship it!

saraswatpuneet (approved):

PR #1972, opened 2024-05-16
Feat/check unclaimed rewards 1969

### LINKED ISSUE #1578: Change unstake logic due to rewards payouts requirements
## Story
To make tracking staking rewards easier, the Frequency chain needs to count and pay out rewards only for full staking RewardEras.

## Acceptance Criteria
<del>First, so Frequency doesn't have to consider prorated rewards, Thaw periods start at the next RewardEra, not immediately.</del>
**Update 2024-05-10**: Thaw periods are irrelevant for the purposes of considering rewards.  The unstake will reduce the staking amount and this will be reflected in the staking history.

Next, Discuss and decide on the following
* Unstaking fails unless:
    * neither of the unlock chunks is maxed out, AND 
        * The staking account is for Maximized Staking, OR
        * The staking account is for Rewards, AND 
             * the account can be paid out OR
             * there are no unclaimed rewards.
* Implement the logic decided upon

For clarification, some example cases of the proposal above.

- [x] My account is _Maximized Staking_.  An `unstake` call behaves the same as before.
- My account is _Provider Boosting Staking_:
- [x] My `unlocking` chunks is full.  Unstake call **fails**.
- [x] I have unclaimed rewards. An `unstake` call **fails**.
- [x] There are no unclaimed rewards and `unlocking` chunks is not full.  An `unstake` call **succeeds and adds the amount to the unlock chunks as normal**.
    
## Blocked by
* #1567 
* #1568 
* #1561

### LINKED ISSUE #1969: [Feature]: Check for unclaimed rewards
Need functionality to tell if there are any unclaimed rewards so that it can be determined if an unstake can succeed or whether it's time to claim rewards.

### Assumptions:
* claiming rewards is all or nothing - everything is claimed, or the `claim_rewards` transaction fails because required conditions are not met.
* rewards cannot be claimed for eras past the configured history limit (StakingRewardsPastErasMax).

### Returns:  
`BoundedVec<UnclaimedRewardInfo<T: Config>, T::StakingRewardsPastErasMax>`

shannonwells:
# Goal
The goal of this PR is to implement `list_unclaimed_rewards`, and also one that is lighter weight, `has_unclaimed_rewards`, which returns a `bool` and which `unstake` extrinsic uses.  Unstake now fails if there are any unclaimed rewards.

**Update: put back to draft since I realized I need to update the unstake benchmark for worst case data storage.  Ok to review otherwise.**

Closes #1969 
Closes #1578 

# Discussion
Two issues
1.  The smaller issue, originally in the design doc, one could not unstake without claiming all rewards. Now I am questioning whether we should require that, but I can still see reasons to support it, if only as a nudge to make sure people get all they're due.

2.  With that said, I think the lightweight `has_unclaimed_rewards` ("You've Got Unclaimed Rewards!") would be very useful to providers, and we should make a custom RPC out of it.

3.  I looked at different ways of making this more efficient in terms of storage reads/writes.

The current way stores Reward Pool history as a `StorageMap` by `RewardEra`.  This means on getting a list unclaimed rewards or actually claiming the rewards - if there are any - it's possible that all the past RewardPoolInfos will be accessed up to the history limit (31).  In addition there are 4 more reads needed.  The size of the entire pool history is about 12k bits since it's 3 * u128 (in production) * 31 items.  If someone boosted and then left it for the limit of `RewardEras` without claiming anything, then checked for unclaimed or claimed rewards, the number of reads would be the maximum.

However, at every boost and unstake, only a single `RewardPoolInfo` record is updated.

If to save only the number of read/writes for checking and claiming rewards, the entire Reward Pool history is stored as a single record, that is a lot of data to read and write on every boost/unstake.  This is also a big lift when a new Reward Era begins in `on_initialize`, because instead of 2 smaller records (the new one and the oldest one) the entire data set would also be read/written.

With the current Provider Boost economics, the Reward Pool amount is a constant, so `total_reward_pool` isn't needed.  Also, nobody has asked for any feature for the `unclaimed_rewards` field.   That field was put in the the design last year, well before the economics model was decided and there had to be some guesswork about it.

I think the middle ground solution is, RewardPool storage should be changed in another PR to have only the one value of `total_staked`, but keep it as key/value `StorageMap` instead of changing it to a `StorageValue` for a `BoundedVec`, i.e. the entire Reward Pool history.

# Checklist
- [x] Feature implemented
- [x] Tests added
- [x] Weights updated

---
ClaimingRewards


pallets/capacity/src/tests/mock.rs
```diff
@@ -208,7 +208,7 @@ impl pallet_capacity::Config for Test {
 	type CapacityPerToken = TestCapacityPerToken;
 	type RewardEra = TestRewardEra;
 	type EraLength = ConstU32<10>;
-	type StakingRewardsPastErasMax = ConstU32<5>;
+	type StakingRewardsPastErasMax = ConstU32<6>; // 5 for claiming rewards, 1 for current reward era
```
shannonwells, line 211: This change was needed, and it broke some other tests.

pallets/capacity/src/tests/mock.rs
```diff
@@ -235,7 +235,7 @@ pub fn new_test_ext() -> sp_io::TestExternalities {
 	let mut ext = sp_io::TestExternalities::new(t);
 	ext.execute_with(|| {
 		System::set_block_number(1);
-		set_era_and_reward_pool_at_block(1, 0, 0);
+		set_era_and_reward_pool(1, 1, 0);
```
shannonwells, line 238: it made more sense to start at era 1 and have that start at block 1 instead of block 0, because block number is also 1.

pallets/capacity/src/types.rs
```diff
@@ -361,11 +362,33 @@ impl<T: Config> ProviderBoostHistory<T> {
 		Some(self.count())
 	}
 
-	/// Return how much is staked for the given era.  If there is no entry for that era, return None.
-	pub fn get_staking_amount_for_era(&self, reward_era: &T::RewardEra) -> Option<&BalanceOf<T>> {
+	/// A wrapper for the key/value retrieval of the BoundedBTreeMap.
```
shannonwells, line 365: better name, better explanation.

pallets/capacity/src/types.rs
```diff
@@ -361,11 +359,33 @@ impl<T: Config> ProviderBoostHistory<T> {
...
+	pub(crate) fn get_entry_for_era(&self, reward_era: &T::RewardEra) -> Option<&BalanceOf<T>> {
 		self.0.get(reward_era)
 	}
 
+	/// Returns how much was staked during the given era, even if there is no explicit entry for that era.
+	/// If there is no history entry for `reward_era`, returns the next earliest entry's staking balance.
+	///
+	/// Note there is no sense of what the current era is; subsequent calls could return a different result
+	/// if 'reward_era' is the current era and there has been a boost or unstake.
+	pub(crate) fn get_amount_staked_for_era(&self, reward_era: &T::RewardEra) -> BalanceOf<T> {
```
aramikm, line 372: nit: so based on my understanding, if the reward_era exists it would return the balance for that but if it doesn't it would return the sum of all balances before that. But this seems to be confusing inside the same function.

aramikm (commented):
Some comments and questions!

pallets/capacity/src/lib.rs
```diff
@@ -407,18 +408,19 @@ pub mod pallet {
 		CannotChangeStakingType,
 		/// The Era specified is too far in the past or is in the future (15)
 		EraOutOfRange,
-		/// Rewards were already paid out for the specified Era range
-		IneligibleForPayoutInEraRange,
```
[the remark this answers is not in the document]
shannonwells: Normally, that is true, but remember this particular error is not in production/main branch; otherwise I would have just marked it as deprecated.  However, in comparing it to main, I realized that "NoThawedTokenAvailable" somehow got moved to the bottom, and that is definitely not backward compatible.  So I am moving it now.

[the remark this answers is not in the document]
shannonwells: Also thanks for flagging this - just to be sure, I just double checked the other error enums which I did rename, I think they all essentially retain the same meaning and are in the same order.  Let me know if you disagree.

pallets/capacity/src/types.rs
shannonwells: It will not return the sum of all balances. It will return whatever is the nearest previous balance if it doesn't have the reward_era key.  If you look in the test for this function it will show how it should work.

However you helped find a bug in that I forgot to implement an important detail for `list_unclaimed_rewards`. A booster should be eligible for rewards only for amounts staked during an entire era. Otherwise you could stake late in an era, unstake early the next era, and still earn full rewards. 

I've fixed this issue, updated the test, and I've also renamed a returned field from `staked_amount` to `eligible_amount` which I hope is more clear.  Please have a look at the test `check_for_unclaimed_rewards_has_eligible_rewards`  and maybe that will help understanding it together.

pallets/capacity/src/lib.rs
```diff
@@ -1062,6 +1068,80 @@ impl<T: Config> Pallet<T> {
...
+
+	pub(crate) fn has_unclaimed_rewards(account: &T::AccountId) -> bool {
+		let current_era = Self::get_current_era().era_index;
+		match Self::get_staking_history_for(account) {
+			Some(provider_boost_history) => {
+				match provider_boost_history.count() {
+					0usize => false,
+					// they staked before the current era, so they have unclaimed rewards.
+					1usize => provider_boost_history.get_entry_for_era(&current_era).is_none(),
+					_ => true,
```
wilwade, line 1080: If the `provider_boost_history` only has values that are not claimable due to age, this would still return true I believe.

pallets/capacity/src/lib.rs
```diff
@@ -1062,6 +1068,80 @@ impl<T: Config> Pallet<T> {
...
+				}
+			},
+			None => false,
+		} // 1r
+	}
+
+	// this could be up to 35 reads.
```
wilwade, line 1087: What about making the storage of past eras a map of bounded vecs of for example size `10`?

Thus this key-value data structure: `<floor(block number /10)>: <10 blocks worth of data>`

It limits the size of the write when the epoch changes, but for 99.9% of users would limit the data to pull when claiming.

wilwade (commented):
Some quick thoughts

pallets/capacity/src/lib.rs
shannonwells: So long as they have a non-zero history entry, that is correct and that is what you want. They will only be able to claim rewards for the last N eras where N is the maximum history retention.  I added a test that illustrates this, and subsequently found a bug where if you boost, and then wait a number of eras past the history retention, the first element of unclaimed rewards is always zeroed out for eligible balance and reward, which is wrong.

Because of this question, I also realized that if e.g. they unstaked everything with only a single entry for the current era, then the provider_boost_history should be deleted.  I added functionality for this and testing also.

There are other things that will need to be done to the history when an account claims rewards, but that will come in when the actual `claim_rewards` extrinsic is implemented.

pallets/capacity/src/lib.rs
shannonwells: I don't quite understand what you mean. And also, are you talking about individual boost history, or overall RewardPool history?

wilwade: I was thinking the the overall history: `StakingRewardPool`. If the issue is that having each era separate creates too many reads, and having all of them in one is too much to read every time, then perhaps we could find a chunking setup that balances between the two. Chunking on the `era % 10` for example.

shannonwells: Ok, as discussed previously, I'll try this idea out in a separate branch after this one is merged.

pallets/capacity/src/lib.rs
```diff
@@ -284,47 +348,86 @@ pub mod pallet {
...
 		/// Staker attempted to stake to an invalid staking target.
 		InvalidTarget,
 		/// Capacity is not available for the given MSA.
-		InsufficientBalance,
+		InsufficientCapacityBalance,
 		/// Staker is attempting to stake an amount below the minimum amount.
-		InsufficientStakingAmount,
-		/// Staker is attempting to stake a zero amount.
+		StakingAmountBelowMinimum,
+		/// Staker is attempting to stake a zero amount.  DEPRECATED
```
wilwade, line 383: It is still used by `ensure_can_stake`, but you can tag it with a rust `deprecated` attribute as well

```suggestion
                /// Staker is attempting to stake a zero amount.
		/// #[deprecated(since = "1.13.0", note = "Use ?? instead")]

```

pallets/README.template.md
```diff
@@ -0,0 +1,49 @@
...
+| -------------------------------- | ------ | ------- | ------------------------------------------------------------------------------------------------------------- | ------------- |
+| `{extrinsic}`<br />{Description} | Tokens | Free    | [`{Event}`](https://frequency-chain.github.io/frequency/{pallet_name}/pallet/enum.Event.html#variant.{Event}) | 1             |
+
+See [Rust Docs](https://frequency-chain.github.io/frequency/{pallet_name}/pallet/struct.Pallet.html) for more details.
+
+### State Queries
+
+| Name      | Description         | Query                    | Runtime Added |
+| --------- | ------------------- | ------------------------ | ------------- |
+| {Query 1} | {Query Desctiption} | `{queryCallInCamelCase}` | 1             |
```
enddynayn, line 35: Desctiption -> Description

pallets/capacity/src/types.rs
```diff
@@ -75,27 +76,46 @@ impl<T: Config> Default for StakingDetails<T> {
...
 	pub fn deposit(&mut self, amount: Balance, capacity: Balance) -> Option<()> {
 		self.amount = amount.checked_add(&self.amount)?;
 		self.capacity = capacity.checked_add(&self.capacity)?;
-
 		Some(())
 	}
 
 	/// Decrease an MSA target Staking total and Capacity amount.
-	pub fn withdraw(&mut self, amount: Balance, capacity: Balance) {
+	/// If the amount would put you below the minimum, zero out the amount.
```
enddynayn, line 101: nice!

pallets/capacity/src/lib.rs
```diff
@@ -666,18 +949,208 @@ impl<T: Config> Pallet<T> {
...
+
+			let past_eras_max = T::StakingRewardsPastErasMax::get();
+			let entries: u32 = StakingRewardPool::<T>::count(); // 1r
+
+			if past_eras_max.eq(&entries) {
+				let earliest_era =
+					current_era_info.era_index.saturating_sub(past_eras_max.into()).add(One::one());
+				StakingRewardPool::<T>::remove(earliest_era); // 1w
+			}
+			CurrentEraInfo::<T>::set(new_era_info); // 1w
```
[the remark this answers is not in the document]
shannonwells: I'll move it anyway since I'm adding the deprecation line.

pallets/README.template.md
shannonwells: this wasn't my change, sorry, it was from conflicting stuff from main.

PR #1973, opened 2024-05-16
rpc: expose missing nonces in future pool

### LINKED ISSUE #1974: [Feature]: New RPC to get missing nonces stuck in the future pool
# Description
Since dealing with nonces is challenging and complex we would like to implement an RPC which would  expose any missing nonces in future pool so that the client can submit new transactions using these nonces and un-stuck the transactions that are waiting for these missing nonces.

aramikm:
# Goal
The goal of this PR is to expose missing nonces in future pool which causes problems when submitting a lot of transactions from the same account.

Closes #1974 

## Discussions
- I had to remove node RPC test for events, since the setup fundamentally didn't work with the Pool and I added the removed test into e2e side.

# Checklist
- [X] Chain spec updated
- [X] Tests added


node/service/src/rpc/frequency_rpc.rs
```diff
@@ -11,43 +11,130 @@ use common_primitives::rpc::RpcEvent;
...
+			current_nonce += One::one();
+			current_tag = (account.clone(), current_nonce.clone()).encode();
+		}
+	}
+
+	let mut result = vec![];
+	let mut my_in_future: Vec<_> = pool
+		.futures()
+		.into_iter()
+		.filter_map(|x| match x.provides().get(0) {
```
saraswatpuneet, line 111: I think rayon can be really useful here `par_iter`

aramikm: I tried that but looked more challenging than expected and got the following error
`doesn't satisfy _: IntoParallelRefIterator<'_>`

Anyway since the logic inside the iterator is simple enough I'm not too worried about the need to optimize at this point.

node/service/src/rpc/tests/mod.rs
aramikm, line 1: I had to remove this since the setup fundamentally didn't work with the Pool and added the removed test into e2e side.

node/service/src/rpc/tests/mod.rs
```diff
@@ -1,55 +0,0 @@
-mod rpc_mock;
-
-use rpc_mock::*;
```
aramikm, line 3: I had to remove this since the setup fundamentally didn't work with the Pool and added the removed test into e2e side.

shannonwells (approved):
I'd say it's ready to go.  Interfaces look good and I'm sure this will be useful.  I:
* reviewed the code and comments
* downloaded the branch and ran `make test` and tests all passed.

node/service/src/rpc/frequency_rpc.rs
```diff
@@ -11,43 +11,130 @@ use common_primitives::rpc::RpcEvent;
...
+
+	for future_nonce in my_in_future {
+		while current_nonce < future_nonce {
+			result.push(current_nonce.clone());
+			current_nonce += One::one();
+
+			// short circuit if we reached the limit
+			if result.len() == MAX_RETURNED_MISSING_NONCE_SIZE {
+				return result
+			}
```
JoeCap08055, line 132: code golf: any benefit to using `BoundedVec` for the result array, instead of this check?

JoeCap08055 (approved):
Nice addition!

saraswatpuneet (approved):

enddynayn (approved):
Code LGTM!

node/service/src/rpc/frequency_rpc.rs
aramikm: Even though I could use it I didn't since I didn't want the result to error out if it was out of bounds and I just wanted to restrict the result size without having extra complexity.

aramikm:
> LGTM. However, I’m a bit unclear on how this will benefit the client. It might be more effective for the client to keep track of transactions that have not yet been processed.

This will allow them to see if there are any transactions that are submitted using incorrect nonces. Tracking submitted transactions is not enough considering there might be more than 1 instances using the same account submitting to the chain.

PR #1971, opened 2024-05-15
Benchmark fixes

wilwade:
# Goal
The goal of this PR is to fix an issue with running benchmarks and update our benchmarks with some tests.

Part of #1853

## Issues

- When I added the Proxy pallet, the copied weight file worked, but the generated version didn't.
- Then when fixing that, the benchmark runners had issues. So updated the benchmark workflow by removing it from the commit commenting system to a manual run. Using that will also automatically trigger verify CI actions. (GitHub actions only run on Manually triggered actions to prevent loops)

# Discussion
- The additional changes to the templates is from Polkadot-SDK
- Actual change is small, as most of this are the generated weights code
- Fix issue with the proxy_pallet weights reference
- Update the benchmark running system to not use special Github tokens

PR #1975, opened 2024-05-21
New benchmark workflow

wilwade:
# Goal
The goal of this PR is to create a new flow for benchmarks so that they are easier to run and do not require a separate github token.

To test this must be in main as `workflow_dispatch` doesn't work without being in `main`, but here is a run that was triggered using defaults with a few other tweaks: https://github.com/frequency-chain/frequency/actions/runs/9174859575

# Discussion
- The PR #1971 will remove the old system from the `.github/workflows/verify-pr-commit.yml` file so that only this one is used
- Once PR #1971 is merged, I will also update the wiki with the new instructions


saraswatpuneet (approved):
lgtm, check that example run and looks correct

PR #1985, opened 2024-05-28
Output the compressed wasm, not the uncompressed

wilwade:
# Goal
The goal of this PR is use the compressed wasm instead of the uncompressed wasm in the releases.

When reviewing the v1.12.0 release, it was discovered that the srtool output for the wasm instead of the compressed wasm was being used.

This corrects that error by using the correct output from the srtool workflow

image


mattheworris (approved):
:shipit:

enddynayn (approved):
👍 great!

PR #2022, opened 2024-06-11
All delegations rpc

### LINKED ISSUE #2019: [Feature]: RPC for Getting All Delegation Information for an MSA
## Details

I want to be able to get the delegation information for all providers for a specific user.

Currently there exists only a query way to access this information `api.query.msa.delegatorAndProviderToDelegation` and making sure the second option is `None`. However, direct access to this state with an optional is not supported by all libraries.

There exists the runtime and custom API that mirrors this, but it does not have the optional: `grantedSchemaIdsByMsaId(delegator_msa_id, provider_msa_id)`

We should also provide `grantedSchemaIdsByMsaId(delegator_msa_id)` that returns the data indexed by provider id.

Example Response (based on Mainnet MSA Id 5500):
```javascript
[
  {
    provider_id: 2,
    permissions: [
      {
        schema_id: 6
        revoked_at: 0
      }
      {
        schema_id: 7
        revoked_at: 0
      }
      {
        schema_id: 8
        revoked_at: 0
      }
      {
        schema_id: 9
        revoked_at: 0
      }
      {
        schema_id: 10
        revoked_at: 0
      }
    ]
  }
]
```

## Possible Related Issues
- #420 Original RPC PR

aramikm:
# Goal
The goal of this PR is to expose an RPC which would return all providers for an msa

Closes #2019 

# Checklist
- [x] Chain spec updated
- [x] Tests added

# Verification
- Calling the RPC returns expected result

Screenshot 2024-06-11 at 4 19 22 PM


js/api-augment/definitions/msa.ts
```diff
@@ -100,50 +114,17 @@ export default {
                 type: 'ProviderId',
               },
             ],
-            type: 'Option<Vec<SchemaId>>',
```
aramikm, line 103: The version 2 was used in frontend side even-though it was not bumped in the node side. I bumped it to 2 so this version in here will stay the same as 2. I don't think there are any reasons to keep version 1 around but I'm open to counter-arguments to why that might be necessary.

enddynayn (approved):
great! 👍

pallets/msa/src/rpc/src/lib.rs
```diff
@@ -56,6 +58,13 @@ pub trait MsaApi<BlockHash, AccountId> {
 		provider_msa_id: ProviderId,
 	) -> RpcResult<Option<Vec<SchemaGrant<SchemaId, BlockNumber>>>>;
 
+	/// Retrieve the list of all delegations for a MsaId
+	#[method(name = "msa_getAllGrantedDelegationsByMsaId")]
+	fn get_all_granted_delegations_by_msa_id(
+		&self,
+		delegator_msa_id: DelegatorId,
+	) -> RpcResult<Vec<DelegationResponse<SchemaId, BlockNumber>>>;
+
```
saraswatpuneet, line 67: nit: do we want to take a an optional blocknumber as input in order to check delegations at specific block(s)

wilwade: I'd say no. If you care about blocks, then you can interact directly with chain state.

pallets/msa/src/runtime-api/src/lib.rs
```diff
@@ -29,7 +29,7 @@ sp_api::decl_runtime_apis! {
 	/// - MUST be incremented if anything changes
 	/// - Also update in js/api-augment
 	/// - See: https://paritytech.github.io/polkadot/doc/polkadot_primitives/runtime_api/index.html
-	#[api_version(1)]
+	#[api_version(2)]
```
wilwade, line 32: I think we can leave this v1 and just tag the `get_all_granted_delegations_by_msa_id` with `#[api_version(2)]`

That way anyone using v1 doesn't have to change because `get_granted_schemas_by_msa_id` doesn't change.

See the example here: https://paritytech.github.io/polkadot-sdk/master/polkadot_primitives/runtime_api/index.html#runtime-api-versioning-crash-course

js/api-augment/definitions/msa.ts
wilwade: Related to my other comment, I think we want to keep v1 around, but here are the only reasons I can think of to have it here:

1. To stop those annoying messages of "api not..." from Polkadotjs
2. To limit the breaking points for upgrades.

`#1` there is actually the most convincing for me 😆

wilwade (commented):
A note about versions, but otherwise looks good from a quick review.

common/primitives/src/msa.rs
```diff
@@ -62,6 +72,16 @@ pub struct SchemaGrant<SchemaId, BlockNumber> {
 	pub revoked_at: BlockNumber,
 }
 
+impl<SchemaId, BlockNumber> PartialEq for DelegationResponse<SchemaId, BlockNumber>
+where
+	SchemaId: PartialEq,
+	BlockNumber: PartialEq,
+{
+	fn eq(&self, other: &Self) -> bool {
+		self.provider_id == other.provider_id && self.permissions == other.permissions
```
JoeCap08055, line 81: Does the Rust `==` operator on `Vec` test equality of the elements in a Set-like fashion? I didn't think it did... if not, do we want something like that?

runtime/frequency/src/lib.rs
```diff
@@ -1404,9 +1404,19 @@ impl_runtime_apis! {
 		}
 
 		fn get_granted_schemas_by_msa_id(delegator: DelegatorId, provider: ProviderId) -> Option<Vec<SchemaGrant<SchemaId, BlockNumber>>> {
-			match Msa::get_granted_schemas_by_msa_id(delegator, provider) {
+			match Msa::get_granted_schemas_by_msa_id(delegator, Some(provider)) {
+				Ok(res) => match res.into_iter().next() {
+					Some(delagation) => Some(delagation.permissions),
```
JoeCap08055, line 1409:
```suggestion
					Some(delegation) => Some(delegation.permissions),
```

pallets/msa/src/lib.rs
```diff
@@ -1321,28 +1322,41 @@ impl<T: Config> Pallet<T> {
...
-		let schema_permissions = provider_info.schema_permissions;
-		if schema_permissions.is_empty() {
-			return Err(Error::<T>::SchemaNotGranted.into());
-		}
+		let mut result = vec![];
+		for (provider_id, provider_info) in delegations {
+			let schema_permissions = provider_info.schema_permissions;
+			if schema_permissions.is_empty() {
+				return Err(Error::<T>::SchemaNotGranted.into());
+			}
```
JoeCap08055, line 1341: Returning an error here made sense when this could only be called for a specific provider, but when called with `None`, I don't think we would want to error here...

shannonwells (approved):
* reviewed code
* downloaded branch and ran local chain
* ran e2e tests locally
* tested the new RPC testing positive and negative results

mattheworris (approved):
- Read through code.
🚢 it!

pallets/msa/src/runtime-api/src/lib.rs
aramikm: Unfortunately it doesn't work as documentation suggests. There might be some kind of bug with `impl_runtime_apis` macro implementation that doesn't compile when I add `#[api_version(2)]` on top of the new method. 
This generate two version for RuntimeApi but it doesn't let me to implement the V2 due to different compile errors. I tried a bunch of things but none of them seemed to work. Please feel free to take a stab at it but as of now we have to keep the implementation as is.

js/api-augment/definitions/msa.ts
aramikm: I was not able to see that error you mentioned inside Polkadotjs when I ran it. But regardless it seems to be need to keep the implementation as is since runtime doesn't let me to implement multiple versions

runtime/frequency/src/lib.rs
```diff
@@ -344,7 +344,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	spec_name: create_runtime_str!("frequency-testnet"),
 	impl_name: create_runtime_str!("frequency"),
 	authoring_version: 1,
-	spec_version: 82,
+	spec_version: 83,
```
claireolmstead, line 347: why do we have to increase?

common/primitives/src/msa.rs
aramikm: It works, since if the inner type doesn't include the partialEq or Eq it doesn't compile.

runtime/frequency/src/lib.rs
aramikm: So everytime we change the code inside runtime (wasm) we would need to increase this. The reason is **seamless runtime upgrades**. For nodes to upgrade the version of their runtime, we should increase it after any change so that the nodes figure out that their version is an older one and they have to upgrade to the newer one.

pallets/msa/src/lib.rs
aramikm: Sure, applied the change

saraswatpuneet (approved):
Lgtm

PR #2037, opened 2024-06-18
Passkey types

### LINKED ISSUE #2027: [Feature]: Passkey - Implement proxy extrinsic
### Feature Description

Implementing the main extrinsic which would process a p256 signed transaction and dispatch it on behalf of the signer account id.

# Acceptence Criterea
- Implement extrinsic payload types
- Implement unsigned extrinsic that dispatches the inner call with owner account as origin
- Implement tests
- Implement benchmarks and incorporate weights
- implement `ValidateUnsigned` with an empty body

aramikm:
# Goal
The goal of this PR is to implement passkey types and extrinsics

Closes #2027 

# Discussion
-  The benchmark for calculating overhead is going to be implemented after all validation is done.
- Do we need any events for observability?

# Checklist
- [x] Tests added


pallets/passkey/src/lib.rs
```diff
@@ -69,7 +73,7 @@ pub mod module {
 	}
 
 	#[pallet::event]
-	#[pallet::generate_deposit(fn deposit_event)]
+	// #[pallet::generate_deposit(fn deposit_event)]
```
aramikm, line 76: Question: We we want to add an event in the extrinsic for additional visibility?

pallets/passkey/src/types.rs
```diff
@@ -0,0 +1,60 @@
...
+}
+
+/// Inner Passkey call
+#[derive(Encode, Decode, TypeInfo, MaxEncodedLen, PartialEq, RuntimeDebugNoBound, Clone)]
+#[scale_info(skip_type_params(T))]
+pub struct PasskeyCall<T: Config> {
+	/// account id which is the origin of this call
+	pub account_id: T::AccountId,
+	/// account nonce
+	pub account_nonce: T::Nonce,
```
aramikm, line 55: We might have other fields in addition to nonce. If that was the case we might put then inside their own struct.

saraswatpuneet (approved):
great work! lgtm with nits

pallets/passkey/src/lib.rs
```diff
@@ -83,10 +87,33 @@ pub mod module {
...
+				payload.passkey_call.account_id,
+			));
+			payload.passkey_call.call.dispatch(main_origin)
+		}
+	}
+
+	#[pallet::validate_unsigned]
+	impl<T: Config> ValidateUnsigned for Pallet<T> {
+		type Call = Call<T>;
+		fn validate_unsigned(
```
JoeCap08055, line 112: Did we decided which way we're going with the design--Signed or Unsigned? I didn't see an update on that question.

aramikm: I think the assumption is that we are going with unsigned for now.

pallets/passkey/src/lib.rs
aramikm: I'll add the event then

JoeCap08055 (approved):
lgtm

mattheworris (approved):
- Read through changes
- Executed unit tests (all passed)
LGTM 🔥

pallets/passkey/src/lib.rs
```diff
@@ -83,10 +90,40 @@ pub mod module {
...
+			if result.is_ok() {
+				Self::deposit_event(Event::TransactionExecutionSuccess {
+					account_id: transaction_account_id,
+				});
+			}
+			result
+		}
+	}
+
+	#[pallet::validate_unsigned]
```
enddynayn, line 119: I believe it would be beneficial for us to merge into a feature branch. I’m not clear on why we are merging directly into the main branch, as we haven’t done this before. Could you please explain the reason for this deviation?

PR #2039, opened 2024-06-22
passkey: nonce check

### LINKED ISSUE #2030: [Feature]: PassKey - Nonce check validation
### Feature Description

We avoid replay attacks we will need to implement nonce checks and incrementing nonce for owner account

## Acceptence Criterea
- Implement Nonce check against owner account
- Add tests
- Add benchmarks

## Notes
- It is preferable if we can already reuse these implementations

aramikm:
# Goal
The goal of this PR is to verify that the nonce is correct for the account_id and increase the nonce. And to restrict the transactions that can be called inside a passkey

Closes #2030 

# Checklist
- [x] Chain spec updated
- [x] Tests added
- [x] Benchmarks added
- [x] Weights updated


pallets/passkey/src/lib.rs
```diff
@@ -127,53 +131,115 @@ pub mod module {
...
-					}
-				},
-				_ => InvalidTransaction::Call.into(),
-			}
+impl<T: Config> Pallet<T> {
+	fn filter_valid_calls(call: &Call<T>) -> Result<PasskeyPayload<T>, TransactionValidityError> {
+		match call {
+			Call::proxy { payload }
+				if T::PasskeyCallFilter::contains(&payload.clone().passkey_call.call) =>
+				return Ok(payload.clone()),
```
JoeCap08055, line 153: I've been away from Rust for too long... didn't realize you could do this (embed an "if" into a match arm)

saraswatpuneet (commented):
Great work! We can try running benchmarks on this PR to test, if not already done. Otherwise looks good me.

I would wait for others to approve so we get more  :eyes:

pallets/passkey/src/mock.rs
```diff
@@ -74,9 +75,35 @@ impl pallet_balances::Config for Test {
 	type RuntimeHoldReason = ();
 }
 
+pub struct MockPasskeyCallFilter;
+
+impl Contains<RuntimeCall> for MockPasskeyCallFilter {
+	fn contains(call: &RuntimeCall) -> bool {
+		match call {
+			RuntimeCall::System(frame_system::Call::remark { .. }) |
```
enddynayn, line 83: I was not expecting to see remark here. I am curious to know why we included it?

pallets/passkey/src/lib.rs
```diff
@@ -127,53 +131,115 @@ pub mod module {
...
+			_ => return Err(InvalidTransaction::Call.into()),
+		}
+	}
+
+	fn validate_nonce(payload: &PasskeyPayload<T>) -> TransactionValidity {
+		let who = &payload.passkey_call.account_id.clone();
```
enddynayn, line 159: I believe we could reuse the nonce validation from the extension and call that function instead of duplicating the code here. While duplicating code can be acceptable in some cases, it would be beneficial to keep both validations always up to date.

Additionally, it might be a good idea to figure out how to integrate this approach, as it seems we will need to draw from other extensions like validating and processing payments.

pallets/passkey/src/mock.rs
aramikm: It's not easy creating non frame calls (like Balance::transfer) from inside benchmarks and that's why we are using only remark inside benchmarks. When we are running benchmarks as test we are consuming MockPasskeyCallFilter which means we need to have remark also as part of the valid calls.

pallets/passkey/src/lib.rs
aramikm: I already tried that approach. After spending more than 2 hours and the code getting much more complicated than it should have been. I was not able to make it work. One of the trait constraints didn't work as I was hoping and I had to switch to this approach in the end.
 
If anybody would like to spend some time and try it they are welcome to.

pallets/passkey/src/mock.rs
enddynayn: Sounds reasonable. It seems that we might benefit from a feature flag instead?

aramikm: I had it featured flagged first but removed it to simplify. I can add it again if you think that would be better.

pallets/passkey/src/lib.rs
aramikm: It would be great if you can make it work.

pallets/passkey/src/mock.rs
enddynayn: I believe it might be better to use a feature flag since we added this primarily to make benchmarking and testing easier. As far as I know, we only intend to allow users to transfer balances for now.

pallets/passkey/src/lib.rs
enddynayn: Ok, Cool. Ill take a look.

enddynayn: @aramikm Here is an example of using the extension directly. This approach passes all current tests. Please note that I have not yet explored the details of what we need from it extensively. However, at first glance, it appears to meet our requirements.

https://github.com/frequency-chain/frequency/pull/2045

pallets/passkey/src/lib.rs
```diff
@@ -124,56 +129,133 @@ pub mod module {
 	}
 
 	#[pallet::validate_unsigned]
-	impl<T: Config> ValidateUnsigned for Pallet<T> {
+	impl<T: Config> ValidateUnsigned for Pallet<T>
+	where
+		<T as frame_system::Config>::RuntimeCall: Dispatchable<Info = DispatchInfo>,
```
enddynayn, line 134: nice!

pallets/passkey/src/lib.rs
```diff
@@ -124,56 +129,133 @@ pub mod module {
...
+	}
+
+	pub fn validate(
+		&self,
+		who: &T::AccountId,
+		call: <T as frame_system::Config>::RuntimeCall,
+		info: &DispatchInfo,
+		len: usize,
+	) -> TransactionValidity {
+		let some = CheckNonce::<T>::from(self.0);
```
enddynayn, line 171: We should probably rename this variable to nonce.

pallets/passkey/src/lib.rs
```diff
@@ -124,56 +129,133 @@ pub mod module {
...
-			if verified {
-				Ok(())
+		let some_call: &<T as Config>::RuntimeCall = &payload.passkey_call.call;
+		let info = &some_call.get_dispatch_info();
+
+		let passkey_nonce = PasskeyNonce::<T>::new(nonce);
+		passkey_nonce.validate(who, some_call.clone().into(), info, 0usize)
+	}
+
+	fn pre_dispatch_nonce(payload: &PasskeyPayload<T>) -> Result<(), TransactionValidityError> {
```
enddynayn, line 200: I am considering whether we could also call pre_dispatch here. What are your thoughts on this?

aramikm: we should use the pre_dispatch from signed extension. that is one of the remaining things that I'm working on

pallets/passkey/src/benchmarking.rs
```diff
@@ -1,23 +1,72 @@
...
+		},
+		passkey_call: call,
+	};
+	payload
+}
+
 benchmarks! {
-	proxy {
-		let caller: T::AccountId = whitelisted_caller();
+	where_clause {  where <T as frame_system::Config>::RuntimeCall: Dispatchable<Info = DispatchInfo> }
```
aramikm, line 53: @enddynayn This one liner missing was the main issue

pallets/passkey/src/mock.rs
aramikm: Clsoing this since I added it as part of AC for https://github.com/frequency-chain/frequency/issues/2032

pallets/passkey/src/lib.rs
aramikm: Many thanks

saraswatpuneet (approved):
Amazing work!

enddynayn (approved):
Fantastic! 💯

mattheworris (approved):
- Read through changes
Nice work!
🚢 it!

PR #2040, opened 2024-06-24
#2029: Passkey withdraw token fees for tx

### LINKED ISSUE #2029: [Feature]: Passkey - Implement tx payment fee charge
### Feature Description

Since Unsigned Extrinsic does not charge any fees for the transaction we need to implement it 

## Acceptence Criterea
- Calculate the fee for the transaction and check if the owner account has enough deposit to cover it
- Charge the owner account the calculated transaction fee
- Add tests
- Add benchmarks

## Notes
Related logic is implemented inside [`ChargeFrqTransactionPayment`](https://github.com/frequency-chain/frequency/blob/d25b833d77a9968fa5d01d7e21e948e4d0ee16d9/pallets/frequency-tx-payment/src/lib.rs#L443)

saraswatpuneet:
# Goal
The goal of this PR is to enable `pallet_transaction_pallet` to withdraw tx fee, this is needed as our extrinsic is `unsigned`, thereby does not gets charged by default

Closes  #2029 
Creates #2051 

# Details
- [x] Refactor some common types from `frequency-tx-payment` to `common-primitives` 
- [x] Add a withdraw fee function to be called from Validate Unsigned to charge for tx fee
- [x] Add unit test

# Discussion

# Checklist
- [ ] Chain spec updated
- [ ] Design doc(s) updated
- [x] Tests added
- [ ] Benchmarks added
- [ ] Weights updated


pallets/frequency-tx-payment/src/lib.rs
```diff
@@ -67,58 +56,6 @@ pub(crate) type CapacityBalanceOf<T> = <CapacityOf<T> as Nontransferable>::Balan
...
-			InitialPayment::Token(_) => write!(f, "Imbalance"),
-		}
-	}
-
-	#[cfg(not(feature = "std"))]
-	fn fmt(&self, _: &mut sp_std::fmt::Formatter) -> sp_std::fmt::Result {
-		Ok(())
-	}
-}
-
```
saraswatpuneet, line 121: This is pure refactor

pallets/passkey/src/lib.rs
```diff
@@ -173,6 +207,36 @@ pub mod module {
...
+				who,
+				runtime_call,
+				&info,
+				fee,
+				tip,
+			) {
+				Ok(initial_payment) => Ok((fee, InitialPayment::Token(initial_payment))),
+				Err(_) => Err(InvalidTransaction::Payment.into()),
+			}
+		}
```
saraswatpuneet, line 239: charge for the passkey `proxy` tx , fee is computed and withdrawn from account

pallets/passkey/src/lib.rs
```diff
@@ -124,15 +130,27 @@ pub mod module {
...
-	impl<T: Config> ValidateUnsigned for Pallet<T> {
+	impl<T: Config> ValidateUnsigned for Pallet<T>
+	where
+		<T as frame_system::Config>::RuntimeCall: From<Call<T>>
+			+ IsSubType<Call<T>>
+			+ Dispatchable<Info = DispatchInfo, PostInfo = PostDispatchInfo>,
+	{
 		type Call = Call<T>;
 		fn validate_unsigned(_source: TransactionSource, call: &Self::Call) -> TransactionValidity {
+			Self::charge_fee(call)?;
```
saraswatpuneet, line 141: Should we charge the fee first or after validation?

common/primitives/src/payment.rs
```diff
@@ -0,0 +1,68 @@
+use frame_support::DefaultNoBound;
```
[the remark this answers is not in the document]
enddynayn: I noticed that we are moving these to the primitives. However, I wonder if we could avoid this. The pallet consuming these types already needs to call functions as well. I appreciate the portability that the frequency transaction pallet currently offers. It might be beneficial to maintain these within the frequency-payment-pallet to keep the pallet more portable and reduce dependencies. What are your thoughts on this?

pallets/passkey/src/lib.rs
```diff
@@ -173,6 +203,40 @@ pub mod module {
...
+		/// * `Err(InvalidTransaction::Payment)` if the fee cannot be withdrawn
+		fn withdraw_token_fee(
+			who: &T::AccountId,
+			call: &Call<T>,
+		) -> Result<(BalanceOf<T>, InitialPayment<T>), TransactionValidityError> {
+			let tip = Zero::zero();
+			let info = call.get_dispatch_info();
+			let len = call.using_encoded(|b| b.len());
+			let fee = pallet_transaction_payment::Pallet::<T>::compute_fee(len as u32, &info, tip);
+			if fee.is_zero() {
```
enddynayn, line 222: I believe we can remove this logic for now. Also, I believe that calling the withdraw fee already does this check.
https://github.com/paritytech/polkadot-sdk/blob/master/substrate/frame/transaction-payment/src/payment.rs#L182

aramikm (commented):
The overall looks good but I think we need tests for these
- The fee doesn't get removed from caller if validation fails (assuming that's what happens with regular calls)
- The fee get's removed successfully from the account when the call is successful (might need to check the pre_disptach)

if not possible to do via unit tests we will need to add to the list of required e2e tests

saraswatpuneet:
@aramikm  added some tests you suggested here https://github.com/frequency-chain/frequency/pull/2040/commits/42fcca4edb7538b07e8fb7a725aabc7bed5eb3e8

pallets/passkey/src/lib.rs
```diff
@@ -77,6 +87,10 @@ pub mod module {
 
 		/// Filters the inner calls for passkey which is set in runtime
 		type PasskeyCallFilter: Contains<<Self as Config>::RuntimeCall>;
+
+		/// Helper Curreny method for benchmarking
+		#[cfg(feature = "runtime-benchmarks")]
+		type Currency: ReservableCurrency<Self::AccountId>;
```
enddynayn, line 93: I believe the trait we are looking for is the Fungible trait. The ReservableCurrency is typically used to set aside balances for users.

pallets/passkey/src/lib.rs
```diff
@@ -40,6 +47,8 @@ mod tests;
 pub mod weights;
 pub use weights::*;
 
+#[cfg(feature = "runtime-benchmarks")]
+use frame_support::traits::{Currency, ReservableCurrency};
```
enddynayn, line 51: I believe the Currency trait is also being deprecated in favor of the [Fungible](https://paritytech.github.io/polkadot-sdk/master/frame_support/traits/tokens/fungible/index.html) trait.

pallets/passkey/src/lib.rs
```diff
@@ -131,28 +145,42 @@ pub mod module {
...
+			);
+			tx_charge.validate()?;
 			let nonce_check = PasskeyNonce::new(payload.passkey_call.clone());
 			nonce_check.validate()
 		}
 
 		fn pre_dispatch(call: &Self::Call) -> Result<(), TransactionValidityError> {
+			let payload = Self::filter_valid_calls(&call)?;
+			let tx_charge = ChargeTransactionPayment::<T>(
+				payload.passkey_call.account_id.clone(),
```
enddynayn, line 167: I suggest swapping the order of these two dispatch calls:

1.	Filter valid calls
2.	Validate signatures
3.	Increase nonce
4.	Payment

It seems that we are calling validate_unsigned inside pre_dispatch, which might be redundant. The order reflects the order of our [SignedExtra](https://github.com/frequency-chain/frequency/blob/main/runtime/frequency/src/lib.rs#L290).

pallets/passkey/src/lib.rs
```diff
@@ -131,28 +145,42 @@ pub mod module {
...
+				call.clone(),
+			);
+			tx_charge.pre_dispatch()?;
+
```
enddynayn, line 171: Additionally, it seems that we are calling validate_unsigned inside pre_dispatch, which might be redundant.

enddynayn:
Looking good!

pallets/passkey/src/lib.rs
saraswatpuneet: Yeah i was wondering why we need pre_dispatch here, the actual trait `ValidatedUnsigned` provides a default implementation of `pre_dispatch` that calls `validate_unsigned` so it feels redundant!

https://github.com/paritytech/polkadot-sdk/blob/18228a9bdbe5b560a39a31810fdf2e3fba59a40d/substrate/primitives/runtime/src/traits.rs#L1781


Thought?

enddynayn: It appears that if we only write validate_unsigned, it will be called again during pre-dispatch, which seems reasonable. However, since we are not using the default implementation and are instead calling pre_dispatch, which in turn calls validate, it seems fine to avoid calling validate_unsigned again.

Additionally, there are multiple ways to approach this, and I wanted to highlight that we are performing multiple state transitions. It’s important to note that calling it twice might unintentionally result in some stateful transactions.

pallets/passkey/src/lib.rs
saraswatpuneet: Fixed the order for both `pre_dispatch` and `validate_unsigned`

pallets/passkey/src/lib.rs
```diff
@@ -131,28 +146,40 @@ pub mod module {
...
 			Self::validate_signatures(&payload)?;
-
 			let nonce_check = PasskeyNonce::new(payload.passkey_call.clone());
-			nonce_check.validate()
+			nonce_check.validate()?;
+			let tx_charge = ChargeTransactionPayment::<T>(
+				payload.passkey_call.account_id.clone(),
+				call.clone(),
+			);
+			tx_charge.validate()
```
aramikm, line 162: I think we might want to return the nonce validate result since that has the dependency graph in it. But it would nice to take a look how the signed extension does this

pallets/passkey/src/lib.rs
```diff
@@ -131,28 +146,40 @@ pub mod module {
...
 			let payload = Self::filter_valid_calls(&call)?;
+			Self::validate_signatures(&payload)?;
 			let nonce_check = PasskeyNonce::new(payload.passkey_call.clone());
-			nonce_check.pre_dispatch()
+			nonce_check.pre_dispatch()?;
+			let tx_charge = ChargeTransactionPayment::<T>(
+				payload.passkey_call.account_id.clone(),
+				call.clone(),
+			);
+			tx_charge.pre_dispatch()
```
aramikm, line 174: Same as the validate, I think we would need to check the signed extensions to see how they result the results of these.

pallets/passkey/src/tests.rs
```diff
@@ -225,17 +535,17 @@ fn validate_unsigned_with_correct_nonce_should_work() {
...
-			Ok(ValidTransaction {
-				priority: 0,
-				requires: vec![Encode::encode(&(who.clone(), 1u64))],
-				provides: vec![Encode::encode(&(who, 2u64))],
-				longevity: TransactionLongevity::max_value(),
-				propagate: true,
-			})
-		);
+		assert!(v.unwrap().priority > 0);
+		// assert_eq!(
```
aramikm, line 539: The actual result of validation would also have an effect on this test

aramikm (commented):
Added some comments

pallets/passkey/src/tests.rs
saraswatpuneet: Yep i am exploring how signed extension does it before re-activating this this

pallets/passkey/src/lib.rs
saraswatpuneet: i think it wont matter in pre_dispatch since it returns an empty result in pre_dispatch

pallets/passkey/src/lib.rs
saraswatpuneet: Figured from substrate that we need to combine the validity from all three checks via `ValidTransaction::combine_with`

here is how signed extensions are validated

https://github.com/paritytech/polkadot-sdk/blob/18ed309a37036db8429665f1e91fb24ab312e646/substrate/primitives/runtime/src/traits.rs#L1641

pallets/passkey/src/lib.rs
```diff
@@ -131,28 +146,46 @@ pub mod module {
...
 			let nonce_check = PasskeyNonce::new(payload.passkey_call.clone());
-			nonce_check.validate()
+			let nonce_validity = nonce_check.validate()?;
+			let tx_charge = ChargeTransactionPayment::<T>(
+				payload.passkey_call.account_id.clone(),
+				call.clone(),
+			);
+			let tx_payment_validity = tx_charge.validate()?;
+			let valid_tx = valid_tx
+				.combine_with(signature_validity)
```
enddynayn, line 165: Nice!

aramikm (approved):
Looks Great!

shannonwells (approved):
One non-blocking comment.

pallets/passkey/src/lib.rs
```diff
@@ -232,3 +265,67 @@ where
...
+		From<Call<T>> + Dispatchable<Info = DispatchInfo, PostInfo = PostDispatchInfo>,
+{
+	/// Validates the transaction fee paid with tokens.
+	pub fn pre_dispatch(&self) -> Result<(), TransactionValidityError> {
+		let info = &self.1.get_dispatch_info();
+		let len = self.0.using_encoded(|c| c.len());
+		let runtime_call: <T as frame_system::Config>::RuntimeCall =
+			<T as frame_system::Config>::RuntimeCall::from(self.1.clone());
+		let who = self.0.clone();
+		self.withdraw_token_fee(&who, &runtime_call, info, len, Zero::zero())?;
```
enddynayn, line 286: I have a question regarding the code. Could you please explain why we don’t call pre_dispatch here?

pallets/passkey/src/lib.rs
```diff
@@ -232,3 +265,67 @@ where
...
+		Ok(())
+	}
+
+	/// Validates the transaction fee paid with tokens.
+	pub fn validate(&self) -> TransactionValidity {
+		let info = &self.1.get_dispatch_info();
+		let len = self.0.using_encoded(|c| c.len());
+		let runtime_call: <T as frame_system::Config>::RuntimeCall =
```
enddynayn, line 294: Same here why don't we call:

```
pallet_transaction_payment::ChargeTransactionPayment::<T>::validate(....)?
```

pallets/passkey/src/lib.rs
saraswatpuneet: `ChargeTransactionPayment` does not expose those functions `SignedExtension` trait does so we cant directly use it here

Hence we had to do this redundant implementation of check nonce and this

enddynayn: I think SignedExtension is implemented on ChargeTransactionPayment. I think we might have a name collision?

saraswatpuneet: Will give it try

saraswatpuneet: At first glance, it is duable but breaks benchmarks, figuring that out as a separate work

https://github.com/frequency-chain/frequency/issues/2051

yay?  Nay?

enddynayn: Yay.

enddynayn (approved):
👍 superb! 💯

PR #2042, opened 2024-06-25
Update npm packages to use Polkadotjs 12

wilwade:
# Goal
The goal of this PR is to update to use the new Polkadotjs/api v12

Only changes are in the package.json and package-lock.json files

### DIFF
--- a/e2e/package.json
+++ b/e2e/package.json
@@ -20,27 +20,27 @@
   "dependencies": {
     "@frequency-chain/api-augment": "file:../js/api-augment/dist/frequency-chain-api-augment-0.0.0.tgz",
     "@helia/unixfs": "^3.0.6",
-    "@polkadot/api": "11.2.1",
-    "@polkadot/types": "11.2.1",
+    "@polkadot/api": "12.0.1",
+    "@polkadot/types": "12.0.1",
     "@polkadot/util": "12.6.2",
-    "helia": "^4.2.3",
+    "helia": "^4.2.4",
     "multiformats": "^13.1.1",
     "rxjs": "^7.8.1",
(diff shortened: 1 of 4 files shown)


e2e/package.json
```diff
@@ -20,27 +20,27 @@
   "dependencies": {
     "@frequency-chain/api-augment": "file:../js/api-augment/dist/frequency-chain-api-augment-0.0.0.tgz",
     "@helia/unixfs": "^3.0.6",
-    "@polkadot/api": "11.2.1",
-    "@polkadot/types": "11.2.1",
+    "@polkadot/api": "12.0.1",
```
mattheworris, line 23: Mentally noting this update, as it may force similar updates downstream in the gateway micro-services.

mattheworris (approved):
🚢 it

e2e/package.json
JoeCap08055: We'll also want to see how this does with frequency-scenario-template. Note [this commit](https://github.com/AmplicaLabs/frequency-scenario-template/commit/9f7a22261bfd0aa79a97cf48dd2bca53fd6ac809) with the comment:
```
- [x] constrain `@frequency-chain/api-augment` package to `1.11.x`
versions until the upstream dependencies of >= `1.12.x` (namely
`@polkadot` packages >= 11.x) are better supported in our ecosystem
```

This affects the Gateway microservices downstream (it's a dev dependency for chain setup & e2e tests, but the package dependency tree causes issues with the app).

Would be nice to do some comprehensive testing & work out all of the downstream problems once and for all.

claireolmstead (approved):

saraswatpuneet (approved):
Lgtm

PR #2050, opened 2024-07-02
Fix proxy e2e test funding source

wilwade:
# Goal
The goal of this PR is to fix a typo of reusing the same funding source for e2e tests

Also sorts the funding sources for easier reading.


enddynayn (approved):
great! 👍

e2e/capacity/transactions.test.ts
```diff
@@ -49,7 +49,7 @@ describe('Capacity Transactions', function () {
   describe('pay_with_capacity', function () {
     describe('when caller has a Capacity account', function () {
       let schemaId: u16;
-      const amountStaked = 2n * DOLLARS;
+      const amountStaked = 3n * DOLLARS;
```
wilwade, line 52: This fixes the failing e2e tests. The weight change pushed something over that was only on the edge of working.

PR #2057, opened 2024-07-03
Development Genesis Config has Mainnet Schemas

### LINKED ISSUE #2041: [Feature]: Standalone Docker Image can have Mainnet Schemas Deployed
## Details

When developing an application on Frequency, you often want to have the schemas from mainnet deployed to make use of.

This proposes that there is either a new image, or an option on the standalone-node image to have mainnet schemas be deployed at launch.

### Options
- Launch time pull: Pulls the mainnet schemas when the image is first launched
- Build time: Create a separate tags:
    - `frequencychain/standalone-node:v1.12.0-schemas{n}` where n is the max schema id
    - `frequencychain/standalone-node:latest-schemas`
    - Requires CI to look for new schemas every (hour?) and build new images when schemas are created on mainnet (or triggered with `workflow_dispatch`)

## AC
- [ ] Image should have all mainnet schemas deployed
- [ ] Image should have all mainnet schema names correctly set
- [ ] Tag should be published (if needed) to docker hub
- [ ] Image Readme should include the additional details of the new options

wilwade:
# Goal
The goal of this PR is to add the Schemas in at Genesis from Mainnet when on a development chain

Closes #2041

# Discussion
- Adds a tool `tools/genesis-schemas` that downloads the schemas from mainnet and stuffs them into a json file
- GenesisConfig for Schemas Pallet that loads from a json file schemas to start with at genesis

## How to Test
- You can test the pulling yourself with:
    - `cd tools/genesis-data && npm i && npm run schemas` or use `make genesis-schemas`
- Otherwise just run `make start` and see that it has the same schemas as mainnet with the same names.

# Checklist
- [x] Tests added


.github/workflows/release.yml
```diff
@@ -906,16 +906,12 @@ jobs:
       fail-fast: true
       matrix:
         arch: [amd64]
-        node: [collator-node-local, instant-seal-node, standalone-node]
+        node: [collator-node-local, standalone-node]
```
wilwade, line 909: Removing the deprecated "instant-seal" which was replaced with "standalone-node" a while ago now.

common/primitives/src/handles.rs
```diff
@@ -4,6 +4,7 @@ use crate::utils::*;
 use frame_support::BoundedVec;
 use parity_scale_codec::{Decode, Encode};
 use scale_info::TypeInfo;
+#[cfg(feature = "std")]
 use serde::{Deserialize, Serialize};
```
wilwade, line 8: Noticed this should have the cfg on it when I used another that needed it removed.

common/primitives/src/schema.rs
```diff
@@ -18,8 +17,19 @@ pub type SchemaId = u16;
 pub type SchemaVersion = u8;
 
 /// Types of modeling in which a message payload may be defined
-#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
```
wilwade, line 21: Moved to no_std

common/primitives/src/schema.rs
```diff
@@ -28,8 +38,19 @@ pub enum ModelType {
 }
 
 /// Types of payload locations
-#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
```
wilwade, line 31: Moved to no_std

common/primitives/src/schema.rs
```diff
@@ -44,8 +65,19 @@ pub enum PayloadLocation {
 /// Support for up to 16 user-enabled features on a collection.
 #[bitflags]
 #[repr(u16)]
-#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
```
wilwade, line 47: Moved to no_std

node/cli/src/cli.rs
```diff
@@ -81,14 +81,6 @@ pub struct Cli {
 	#[clap(raw = true)]
 	pub relay_chain_args: Vec<String>,
 
-	#[cfg(feature = "frequency-no-relay")]
-	#[clap(long = "instant-sealing", help = "Deprecated.  Use --sealing=instant instead.")]
-	pub instant_sealing: bool,
-
-	#[cfg(feature = "frequency-no-relay")]
-	#[clap(long = "manual-sealing", help = "Deprecated.  Use --sealing=manual instead.")]
-	pub manual_sealing: bool,
```
wilwade, line 90: Long deprecated

pallets/passkey/src/types.rs
```diff
@@ -19,7 +19,7 @@ pub type PasskeyAuthenticatorData = BoundedVec<u8, ConstU32<128>>;
 /// Passkey ClientDataJson type
 /// Note: The `challenge` field inside this json MUST be replaced with `CHALLENGE_PLACEHOLDER`
 /// before submission to the chain
-/// https://w3c.github.io/webauthn/#dictdef-collectedclientdata
+/// <https://w3c.github.io/webauthn/#dictdef-collectedclientdata>
```
wilwade, line 22: Fix doc warning

pallets/schemas/src/tests/mock.rs
```diff
@@ -117,7 +117,7 @@ impl pallet_schemas::Config for Test {
 	// is actually allowed.
 	type SchemaModelMaxBytesBoundedVecLimit = SchemaModelMaxBytesBoundedVecLimit;
 	type MaxSchemaRegistrations = MaxSchemaRegistrations;
-	type MaxSchemaSettingsPerSchema = ConstU32<1>;
+	type MaxSchemaSettingsPerSchema = ConstU32<2>;
```
wilwade, line 120: Needed to be one higher for the test from mainnet schema with 2

pallets/schemas/src/tests/other_tests.rs
```diff
@@ -1050,3 +1050,31 @@ fn propose_to_create_schema_name_happy_path() {
...
+	ext.execute_with(|| {
+		System::set_block_number(1);
+		let res = SchemasPallet::get_current_schema_identifier_maximum();
+
+		// We should have at least 10
+		assert!(res >= 10);
+
+		// Check that the first schema exists
+		let res = SchemasPallet::get_schema_by_id(1);
+		assert!(res.is_some());
```
wilwade, line 1078: Isn't the best test, but as it is pushed this through code tested elsewhere, I figured this was good enough.

resources/genesis-schemas.json
```diff
@@ -0,0 +1,75 @@
+[
```
wilwade, line 1: Generated from tools/genesis-data

tools/genesis-data/getSchema.mjs
```diff
@@ -0,0 +1,89 @@
+// Only way to silence PolkadotJS API warnings we don't want
+console.warn = () => {};
+
+import { ApiPromise, WsProvider } from "@polkadot/api";
+
+const options = {
```
wilwade, line 6: Instead of loading in the entire api-augment, I just copied the minimum needed.

tools/state-copy/copy.mjs
```diff
@@ -0,0 +1,73 @@
+import { ApiPromise, WsProvider } from "@polkadot/api";
```
wilwade, line 1: The state copy tool isn't used, so I am happy to remove, but it works and feels useful enough to leave it.

pallets/schemas/src/lib.rs
```diff
@@ -239,20 +239,60 @@ pub mod pallet {
...
+					model,
+					schema.model_type,
+					schema.payload_location,
+					settings,
+					parsed_name,
+				)
+				.expect("Failed to set Schema in Genesis!");
+			}
+
+			CurrentSchemaIdentifierMaximum::<T>::put(self.initial_schemas.len() as u16);
```
aramikm, line 295: This looks redundant unless I'm missing something since `add_schema` already sets the correct value.

aramikm (approved):
Looks good. Just one small comment.

pallets/schemas/src/lib.rs
wilwade: Thank you. Removing.

tools/genesis-data/schemas.mjs
```diff
@@ -0,0 +1,17 @@
...
+  try {
+    await getSchemas(SOURCE_URL);
+    process.exit(0);
+  } catch (error) {
+    console.error("Error:", error);
+    process.exit(1);
+  }
+}
+
+main();
```
JoeCap08055, line 17: Best practice is probably to have at least a `.catch()` or `.then()` on this. Usually I just move the `try...catch` from the main function into a `.catch(...)` on the `main()` invocation

JoeCap08055:
I think there's one problem with this: if you start/stop/start a node with the same chainstorage directory, but change the `genesis-schemas.json` file in between the start/stop, the genesis hash changes and all historical blocks go to hell...

Steps to reproduce:
* Trim down `genesis-schemas.json` by removing one or more schemas
* `make start`
* Using available tools, record the genesis hash, submit some txns to the chain to form at least one new block
* Stop node
* `make genesis-schemas`
* `make start`
* Blocks formed previously technically still exist, but cannot be retrieved; genesis hash is now different

Do we care about this scenario? Only going to happen in local dev, I guess, but we should have a way to make sure that we don't restart a node with a different genesis config than it was first started with (unless the `chain-storage` directory is deleted first)

tools/state-copy/copy.mjs
shannonwells: I Agree With This Comment

shannonwells (approved):
I reviewed the code and ran the genesis schema script, and compared the chain state of local against main.  All looks good to me.

wilwade:
> I think there's one problem with this: if you start/stop/start a node with the same chainstorage directory, but change the `genesis-schemas.json` file in between the start/stop, the genesis hash changes and all historical blocks go to hell...

From my test, this is only due to Polkadotjs. If you reload Polkadotjs/app after restarting the chain, the issues resolve. I've run into this with Polkadotjs/app before.

@JoeCap08055 Can you confirm this fixes/was tried?

> Thinking this could be an issue when we incorporate this into the Docker image for dev...

You cannot rebuild the binary, so this at least cannot cause issues outside of the normal restart.

JoeCap08055:
> > I think there's one problem with this: if you start/stop/start a node with the same chainstorage directory, but change the `genesis-schemas.json` file in between the start/stop, the genesis hash changes and all historical blocks go to hell...
> 
> From my test, this is only due to Polkadotjs. If you reload Polkadotjs/app after restarting the chain, the issues resolve. I've run into this with Polkadotjs/app before.
> 
> @JoeCap08055 Can you confirm this fixes/was tried?
> 
> > Thinking this could be an issue when we incorporate this into the Docker image for dev...
> 
> You cannot rebuild the binary, so this at least cannot cause issues outside of the normal restart.

Confirmed--reloading the Polkadot.js app/page did the trick. I guess it's really more about proper user education, since it's mainly for dev purposes. We should also be very intentional about how we integrate this into the `dsnp/instant-seal-node-with-deployed-schemas` image (or, given this, do we even need to maintain 2 separate images any more?)

wilwade:
@JoeCap08055 
> or, given this, do we even need to maintain 2 separate images any more

No-ish. For most developers, they would be able to use the Frequency standalone-node image directly.

If someone wanted to have a docker image that would have additional, not-yet-mainnet schemas, then they would need to have a separate image.

So I think we can deprecate that image on docker hub at least.

JoeCap08055:
> If someone wanted to have a docker image that would have additional, not-yet-mainnet schemas, then they would need to have a separate image.
> 
> So I think we can deprecate that image on docker hub at least.

Might be nice to add a fetch target for Paseo schemas as well, to give devs an easy option for that. Beyond that, for testing new, un-deployed schemas, it's easy enough to hand-edit the JSON file with any additional schemas for testing.

JoeCap08055 (approved):
One code-style comment & one feature suggestion, non-blocking. Otherwise, tested & looks good.

wilwade:
> Might be nice to add a fetch target for Paseo schemas as well

I'm actually wondering if we want to force Paseo to also match mainnet. Either by occasionally wiping all schemas or just using setStorage with sudo on Testnet. Effectively, just always have them all match mainnet, but numbers > than mainnet are all yours.

Thoughts?

JoeCap08055:
> > Might be nice to add a fetch target for Paseo schemas as well
> 
> I'm actually wondering if we want to force Paseo to also match mainnet. Either by occasionally wiping all schemas or just using setStorage with sudo on Testnet. Effectively, just always have them all match mainnet, but numbers > than mainnet are all yours.
> 
> Thoughts?

Seems out of scope for this PR, and worth further discussion, but pretty low priority. Would have been more important before schema names...

Issue #2032, opened 2024-06-13
[Feature]: Passkey - Finishing checks and cleanup and documentation


aramikm:
### Feature Description

In this issue we would like to re-check all assumptions and done the finishing clean-up and adding left-over documentations

### Acceptance Criteria
- [x] Filter inner calls just like what we do for token accounts
- [x]  All modules and implementations should have concise and readable documents
- [x] All combinations of transactions and validations should get checked to ensure there are no known leftover vulnerabilities
- [x] Finishing cleanup and refactoring if required
- [x] Readme matches the Pallet Readme template

PR #2070, opened 2024-07-11
Passkey e2e

saraswatpuneet:
# Goal
The goal of this PR is to add e2e tests for passkey pallet

Details
-  Added 3 Negative tests and One happy path test

Part of #2032 

# Checklist
- [ ] Chain spec updated
- [ ] Design doc(s) updated
- [x] Tests added
- [ ] Benchmarks added
- [ ] Weights updated


e2e/package.json
```diff
@@ -20,6 +20,7 @@
   "dependencies": {
     "@frequency-chain/api-augment": "file:../js/api-augment/dist/frequency-chain-api-augment-0.0.0.tgz",
     "@helia/unixfs": "^3.0.6",
+    "@noble/curves": "^1.4.2",
```
saraswatpuneet, line 23: this one provides `secp256r1` which is PassKey,

aramikm (approved):
Looks great!

e2e/passkey/passkeyProxy.test.ts
```diff
@@ -0,0 +1,76 @@
...
+      const { passKeyPrivateKey, passKeyPublicKey, passkeySignature } = createPassKeyAndSignAccount(accountPKey);
+      const accountSignature = fundedKeys.sign(u8aWrapBytes(passKeyPublicKey));
+      const passkeyCall = await createPassKeyCall(accountPKey, nonce, accountSignature, transferCalls);
+      const passkeyPayload = await createPasskeyPayload(passKeyPrivateKey, passKeyPublicKey, passkeyCall, true);
+
+      const passkeyProxy = ExtrinsicHelper.executePassKeyProxy(fundedKeys, passkeyPayload);
+      assert.rejects(passkeyProxy.fundAndSendUnsigned(fundingSource));
+    });
+
+    it('should transfer small balance from fundedKeys to receiverKeys', async function () {
```
enddynayn, line 61: It might not be a bad idea to also verify the nonce increased.

enddynayn (approved):
awesome 👍 !

e2e/passkey/passkeyProxy.test.ts
saraswatpuneet: Its a good suggestion, I will incorporate this feedback.

e2e/passkey/passkeyProxy.test.ts
```diff
@@ -0,0 +1,76 @@
...
+      const accountPKey = fundedKeys.publicKey;
+      const nonce = await getNonce(fundedKeys);
+      const transferCalls = ExtrinsicHelper.api.tx.balances.transferAllowDeath(receiverKeys.publicKey, 100n);
```
shannonwells, line 64: Any particular reason to use transferAllowDeath here instead of transferKeepAlive?

pallets/passkey/src/types.rs
```diff
@@ -171,6 +171,9 @@ impl VerifiablePasskeySignature {
 
 		// prepare signing payload which is [authenticator || sha256(client_data_json)]
 		let mut passkey_signature_payload = self.authenticator_data.to_vec();
+		if passkey_signature_payload.len() < 37 {
+			return Err(PasskeyVerificationError::InvalidProof);
```
saraswatpuneet, line 175: Based on documentation for `authenticator_data` this should be atleast 37 bytes

e2e/passkey/passkeyProxy.test.ts
```diff
@@ -0,0 +1,76 @@
...
+
+      const passkeyProxy = ExtrinsicHelper.executePassKeyProxy(fundedKeys, passkeyPayload);
+      assert.rejects(passkeyProxy.fundAndSendUnsigned(fundingSource));
+    });
+
+    it('should fail to transfer balance due to bad passkey signature', async function () {
+      const accountPKey = fundedKeys.publicKey;
+      const nonce = await getNonce(fundedKeys);
+      const transferCalls = ExtrinsicHelper.api.tx.balances.transferAllowDeath(receiverKeys.publicKey, 0n);
+      const { passKeyPrivateKey, passKeyPublicKey, passkeySignature } = createPassKeyAndSignAccount(accountPKey);
```
shannonwells, line 52: This confused me trying to understand how this process creates a bad passkeySignature, and if `passkeySignature` is bad, why this test fails since that value is apparently never reused in each of these tests.  The passkeySignature here is not what causes the failure it is the bad payload generated by the test function.

pallets/passkey/src/types.rs
aramikm: We might want to add a new Error if it doesn't exist. Something like  `PasskeyVerificationError::InvalidAuthenticatorData`

e2e/scaffolding/P256.ts
```diff
@@ -0,0 +1,76 @@
...
+  };
+
+  return passkeyCall;
+}
+
+export async function createPasskeyPayload(
+  passKeyPrivateKey: Uint8Array,
+  passKeyPublicKey: Uint8Array,
+  passkeyCallPayload: any = {},
+  bad: boolean = false
```
shannonwells, line 39: This was not the easiest to understand when reading through the tests, since you are just passing a boolean to get an invalid payload back.  It would be clearer to split these up or else pass named option(s)

shannonwells (approved):
The comments I have I don't think are sufficient to block merging, but I would prefer the test functions be a bit clearer, and also the one test title could maybe be reworded.

e2e/passkey/passkeyProxy.test.ts
saraswatpuneet: There are two passKeySignatures, one above `passkeySignature` is passkey signing the account and account signing the passkey, this resembles a registration process on provider side

When we submit the passkey we sign the payload, this test is testing bad passkey signature on transaction time

e2e/passkey/passkeyProxy.test.ts
saraswatpuneet: No specific reason but changed it to `transferKeepAlive`

e2e/scaffolding/P256.ts
saraswatpuneet: Gave is a meaningful name

e2e/passkey/passkeyProxy.test.ts
saraswatpuneet: Add checked for nonce

PR #2071, opened 2024-07-11
passkey: reorder validation and some docs

aramikm:
# Goal
The goal of this PR is to re order the validations to be as close as possible to the signed extension order considering the customizations that we did such as not having post_dispatch implemented.

Related to #2032

PR #2076, opened 2024-07-12
passkey:move to last

aramikm:
# Goal
There was a misunderstanding before that we thought if the fee was deducted from account in pre_dispatch and after that any failure happens inside pre_dispatch, only post_dispatch would be able to refund that amount but that was not a correct assumption and the caller of pre_dispatch is going to reverse all storage changes in it does not return with Ok. based on 
`/// Changes made to storage *WILL* be persisted if the call returns `Ok`.`

Knowing this, we could put the signature check as the final validation since it is the heaviest one.

Related to #2071


pallets/passkey/src/tests.rs
```diff
@@ -319,31 +317,6 @@ fn fee_withdrawn_for_failed_call() {
 	});
 }
 
-#[test]
-fn fee_not_removed_on_failed_validation() {
```
aramikm, line 323: We don't need this test because the fees would be deducted, but it is the responsibility of the caller to reverse the changes in storage based on docs.

mattheworris (approved):
lgtm

enddynayn (approved):
👍 awesome!

pallets/passkey/src/tests.rs
shannonwells: Would you mind linking to the documentation you're referring to?

saraswatpuneet (approved):

pallets/passkey/src/tests.rs
saraswatpuneet: So will it be a batch call ?

aramikm: @shannonwells https://paritytech.github.io/polkadot-sdk/master/polkadot_service/runtime_traits/trait.ValidateUnsigned.html#method.pre_dispatch

aramikm: I'm not sure if I understand what you mean by batch call. So my understanding is that the pre_dispatch caller which is inside executor or frame (I don't remember which one) checks for the result of pre_dispatch and only persists the storage changes if the result was OK. More details how it does is probably since all the writes go into the caching layer. it probably cleans the caching layer in case of any error so it does not get persisted at the end of block

saraswatpuneet: The link above solved my query, thanks

PR #2094, opened 2024-07-24
chore(Cargo): update to use crates

enddynayn:
Update crate dependencies to use
crates instead of Polkadot branches.


wilwade:
Closing for now

PR #2097, opened 2024-07-24
feat(runtime): add metadata signed-extension

enddynayn:
Add a new signed extension that enables the metadata hash verification feature approved under [RFC 0078](https://polkadot-fellows.github.io/RFCs/approved/0078-merkleized-metadata.html#rfc-0078-merkleized-metadata). This enhancement will support the new generic ledger hardware wallet app and future hardware wallets within the Polkadot ecosystem.

[Reference implementation](https://github.com/paritytech/polkadot-sdk/pull/4580)

### DIFF
--- a/runtime/frequency/build.rs
+++ b/runtime/frequency/build.rs
@@ -1,12 +1,16 @@
-use substrate_wasm_builder::WasmBuilder;
-
+#[cfg(all(feature = "std", not(feature = "metadata-hash")))]
 fn main() {
-	// VSCode Users: Uncomment the following line to disable the ANSI color codes.
-	// The OUTPUT pane does not understand ANSI color codes and will show garbage without this.
-	// std::env::set_var("WASM_BUILD_NO_COLOR", "1");
-	WasmBuilder::new()
-		.with_current_project()
-		.export_heap_base()
-		.import_memory()
+	substrate_wasm_builder::WasmBuilder::build_using_defaults()
+}
+
+#[cfg(all(feature = "std", feature = "metadata-hash"))]
+fn main() {
+	substrate_wasm_builder::WasmBuilder::init_with_defaults()
+		.enable_metadata_hash("FRQCY", 8)
 		.build()
 }
+
+/// The wasm builder is deactivated when compiling
+/// this crate for wasm to speed up the compilation.
+#[cfg(not(feature = "std"))]
+fn main() {}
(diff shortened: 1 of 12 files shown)


runtime/frequency/src/lib.rs
```diff
@@ -304,6 +304,7 @@ pub type SignedExtra = (
 	pallet_frequency_tx_payment::ChargeFrqTransactionPayment<Runtime>,
 	pallet_msa::CheckFreeExtrinsicUse<Runtime>,
 	pallet_handles::handles_signed_extension::HandlesSignedExtension<Runtime>,
+	frame_metadata_hash_extension::CheckMetadataHash<Runtime>,
```
saraswatpuneet, line 307: Should we do this for passkey unsigned ext? Maybe yes

saraswatpuneet (approved):
Lgtm , maybe we can add a story for passkey pallet to also add/support hash verification

JoeCap08055 (approved):

claireolmstead (approved):
Looks great!

runtime/frequency/src/lib.rs
aramikm: This is to ensure that the offline wallets are signing the expected data structure (metadata). Since we are doing the signing via passkey adding this does not bring any additional value at this point since the metadata is a totally foreign concept from passkey's perspective.

We might be able to utilize this check in future by adding this metadata hash inside `PasskeyCall<T: Config>` type and that would allow us to ensure the passkey signature was done on a correct metadata.

aramikm (approved):
Looks good overall! Added some questions

runtime/frequency/src/lib.rs
enddynayn: Yes, I’ve been considering it. Since this is an unsigned transaction, it might not be the best fit. However, I believe we are still signing the call function within the outer call. I wonder if we can verify that the hash of the metadata used at the time of signing matches the root.

runtime/frequency/build.rs
```diff
@@ -1,12 +1,16 @@
...
 fn main() {
-	// VSCode Users: Uncomment the following line to disable the ANSI color codes.
-	// The OUTPUT pane does not understand ANSI color codes and will show garbage without this.
-	// std::env::set_var("WASM_BUILD_NO_COLOR", "1");
-	WasmBuilder::new()
-		.with_current_project()
-		.export_heap_base()
-		.import_memory()
+	substrate_wasm_builder::WasmBuilder::init_with_defaults()
+		.enable_metadata_hash("FRQCY", 8)
```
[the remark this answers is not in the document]
enddynayn: Thanks for asking about the number of decimals for the token. It prompted me to check, and I confirmed that both Frequency and Paseo have the same number of decimals.

PR #2099, opened 2024-07-25
chore(upgrade): v1.8.0 to v1.9.0

enddynayn:
- Upgrade Polkadot-sdk to v.1.9.0.
- Update weights to reflect the new version.

Notable Changes:
- [System Callbacks](https://github.com/paritytech/polkadot-sdk/pull/1781)
- [Remove of pallet pallet::getter](https://github.com/paritytech/polkadot-sdk/pull/3456)
- [Add storage_proof_size host function](https://github.com/paritytech/polkadot-sdk/pull/3002)
- [Rename of storage version function](https://github.com/paritytech/polkadot-sdk/pull/1554/files#diff-01dc4f43df9baa537f30c6b369525715d596a3068944f38458e9f160d5412d58R306)

For more details, please refer to:

[Release Notes](https://github.com/paritytech/polkadot-sdk/releases/tag/polkadot-v1.9.0)

### DIFF
--- a/pallets/messages/src/tests/other_tests.rs
+++ b/pallets/messages/src/tests/other_tests.rs
@@ -668,7 +668,7 @@ fn migration_to_v2_should_work_as_expected() {
 
 		let old_count = v2::old::Messages::<Test>::iter().count();
 		let new_count = MessagesV2::<Test>::iter().count();
-		let current_version = MessagesPallet::current_storage_version();
+		let current_version = MessagesPallet::in_code_storage_version();
 
 		assert_eq!(old_count, 0);
 		assert_eq!(new_count, message_per_block.iter().sum::<usize>());
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -239,6 +239,20 @@ dependencies = [
 
 [[package]]
 name = "aquamarine"
+version = "0.3.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d1da02abba9f9063d786eab1509833ebb2fac0f966862ca59439c76b9c566760"
+dependencies = [
+ "include_dir",
+ "itertools 0.10.5",
+ "proc-macro-error",
+ "proc-macro2",
+ "quote",
+ "syn 1.0.109",
+]
+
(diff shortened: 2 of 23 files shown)


claireolmstead (approved):
🚢 LGTM!

pallets/messages/src/tests/other_tests.rs
```diff
@@ -668,7 +668,7 @@ fn migration_to_v2_should_work_as_expected() {
 
 		let old_count = v2::old::Messages::<Test>::iter().count();
 		let new_count = MessagesV2::<Test>::iter().count();
-		let current_version = MessagesPallet::current_storage_version();
+		let current_version = MessagesPallet::in_code_storage_version();
```
aramikm, line 671: We need to ensure these are exactly the same values. If not it would brick the chain.

aramikm (approved):
Overall looks good. Added 1 comment.

- To be able to use Multi Block Migrations we might need to look into using poll instead of on_initialize
- Added host function to clawback POV requires node upgrades before enabling from runtime

JoeCap08055 (approved):
- Pulled/built branch
- Ran unit tests
- Ran e2e tests
Looks good!

pallets/messages/src/tests/other_tests.rs
enddynayn: Thanks for pointing this out. I forgot to include it in the notable changes section of the PR. It appears to be a rename, and I’ll make sure to add it to the notable changes.

https://github.com/paritytech/polkadot-sdk/pull/1554/files#diff-01dc4f43df9baa537f30c6b369525715d596a3068944f38458e9f160d5412d58R306

enddynayn:
> Overall looks good. Added 1 comment.
> 
> * To be able to use Multi Block Migrations we might need to look into using poll instead of on_initialize
> * Added host function to clawback POV requires node upgrades before enabling from runtime


You are right. I have created a ticket so we can allocate time to investigate where we can move the on_initialize functions to poll.

https://github.com/frequency-chain/frequency/issues/2098

PR #2102, opened 2024-07-29
passkey: whitelisted capacity

### LINKED ISSUE #2032: [Feature]: Passkey - Finishing checks and cleanup and documentation

aramikm:
# Goal
The goal of this PR is whitelist Capacity pallet transactions for passkey.

Closes  #2032

# Discussion
- I checked `TransactionPayment` pallet and there were no extrinsics on that pallet so there were no reason to whitelist it
- I checked `FrequencyTxPayment` pallet and there were no token based extrinsics as far as I checked so it was also not included.

# Checklist

### DIFF
--- a/runtime/frequency/src/lib.rs
+++ b/runtime/frequency/src/lib.rs
@@ -286,7 +286,8 @@ impl Contains<RuntimeCall> for PasskeyCallFilter {
 
 			RuntimeCall::Balances(pallet_balances::Call::transfer_keep_alive { .. }) |
 			RuntimeCall::Balances(pallet_balances::Call::transfer_allow_death { .. }) |
-			RuntimeCall::Balances(pallet_balances::Call::transfer_all { .. }) => true,
+			RuntimeCall::Balances(pallet_balances::Call::transfer_all { .. }) |
+			RuntimeCall::Capacity(_) => true,
 			_ => false,
 		}
 	}
@@ -363,7 +364,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	spec_name: create_runtime_str!("frequency"),
 	impl_name: create_runtime_str!("frequency"),
 	authoring_version: 1,
-	spec_version: 99,
+	spec_version: 100,
 	impl_version: 0,
 	apis: RUNTIME_API_VERSIONS,
 	transaction_version: 1,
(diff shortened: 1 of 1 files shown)


claireolmstead (approved):
Simple enough! 🚢

enddynayn (approved):
👍 superb!

runtime/frequency/src/lib.rs
```diff
@@ -286,7 +286,8 @@ impl Contains<RuntimeCall> for PasskeyCallFilter {
 
 			RuntimeCall::Balances(pallet_balances::Call::transfer_keep_alive { .. }) |
 			RuntimeCall::Balances(pallet_balances::Call::transfer_allow_death { .. }) |
-			RuntimeCall::Balances(pallet_balances::Call::transfer_all { .. }) => true,
+			RuntimeCall::Balances(pallet_balances::Call::transfer_all { .. }) |
```
wilwade, line 289: @aramikm shouldn't this be just this to allow for any type of token transfer?

```suggestion
			RuntimeCall::Balances(_)
```

aramikm: It is pretty much the same in practice since the rest of the extrinsics inside that pallet all requires root access. `ensure_root` 

I can modify it if you think that might be simpler.

wilwade: image
There are at least these other two that should be allowed right?

I think allowing all is simplest way

aramikm: But those are allowed if you look at the allowed list but I'll modify it to just include all.

```
			RuntimeCall::Balances(pallet_balances::Call::transfer_keep_alive { .. }) |
			RuntimeCall::Balances(pallet_balances::Call::transfer_allow_death { .. }) |
			RuntimeCall::Balances(pallet_balances::Call::transfer_all { .. }) |
			RuntimeCall::Capacity(_) => true,
```

wilwade: Ah I completely missed those up above. Sorry about that. Thanks for the simplification PR.

PR #2108, opened 2024-08-01
provider boosting: some optimizations

aramikm:
# Goal
The goal of this PR is apply some optimizations in provider boosting

### DIFF
--- a/pallets/capacity/src/lib.rs
+++ b/pallets/capacity/src/lib.rs
@@ -26,7 +26,7 @@
 	missing_docs
 )]
 
-use sp_std::ops::Mul;
+use sp_std::{collections::btree_map::BTreeMap, ops::Mul};
 
 use frame_support::{
 	ensure,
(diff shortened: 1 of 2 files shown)


pallets/capacity/src/lib.rs
```diff
@@ -1076,39 +1080,41 @@ impl<T: Config> Pallet<T> {
...
+			staking_history.get_amount_staked_for_era(&(start_era.saturating_sub(1u32.into())));
 
 		let mut unclaimed_rewards: BoundedVec<
 			UnclaimedRewardInfo<BalanceOf<T>, BlockNumberFor<T>>,
 			T::ProviderBoostHistoryLimit,
 		> = BoundedVec::new();
-		while reward_era.le(&end_era) {
+
+		let mut chunk_cache = BTreeMap::new();
+
```
saraswatpuneet, line 1106: Any advantage of using of Btree of hashmap for temporary cache ?

enddynayn: Good question! I had to double-check this myself. The main reason for using BTreeMap instead of HashMap is that HashMap can’t be used in a no_std environment. [Here’s a detailed explanation](https://stackoverflow.com/questions/65469712/what-do-i-replace-vec-and-hashset-with-in-a-no-std-environmement#:~:text=You%20can't%20use%20HashMap,this%20security%20invariant%20is%20important) on why this is the case.

saraswatpuneet: Aah yes the wasm

pallets/capacity/src/lib.rs
```diff
@@ -1045,26 +1045,30 @@ impl<T: Config> Pallet<T> {
...
+				Self::has_unclaimed_rewards_with_params(current_era, &provider_boost_history),
 			None => false,
 		} // 1r
 	}
 
+	pub(crate) fn has_unclaimed_rewards_with_params(
+		current_era_index: u32,
+		boost_history: &ProviderBoostHistory<T>,
+	) -> bool {
+		match boost_history.count() {
```
enddynayn, line 1058: I creating a variable for naming:
let previous_era =  &current_era_index.saturating_sub(1u32.into())

saraswatpuneet (commented):
Nice optimization , looks good so far

aramikm:
Closing this since the benchmarks are considering the cached DB reads

PR #2109, opened 2024-08-02
Fix cargo deny

wilwade:
# Goal
The goal of this PR is to fix cargo deny config for the changes in v0.16

- Updated two yanked crates
- Updated config patterns, but not altered the results (new defaults and config setup)

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -11,6 +11,9 @@
 
 # Root options
 
+# The graph table configures how the dependency graph is constructed and thus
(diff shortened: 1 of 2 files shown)


deny.toml
wilwade, line 1: All these changes result in the same, but updated via re-running `cargo deny init` and editing.

enddynayn (approved):
👍 fantastic!

saraswatpuneet (approved):

Issue #2105, opened 2024-07-31
[Chore]: Remove pallet::getter Usage from All Pallets


enddynayn:
**Description**
We currently have several pallets that use the pallet::getter macro. Since this macro is going to be deprecated soon, it’s a good idea to remove it now and simplify our codebase. The preferred syntax moving forward is to use Key::<T>::get() instead.

**Acceptance Criteria**
- No more pallet::getter usage in the codebase.

[Polkadot Reference](https://github.com/paritytech/polkadot-sdk/issues/223)

PR #2112, opened 2024-08-05
Boosting: removed added getters

aramikm:
# Goal
The goal of this PR is to removed added getters

Related to #2105


wilwade (approved):
Nice

PR #2133, opened 2024-08-20
signed extension: check stateful storage hash

### LINKED ISSUE #1860: `target_hash` RPC Level validation
As someone transacting using stateful storage, I want to have my transaction stopped at the node level if I submit something with a bad `target_hash` so that I don't waste capacity on a transaction that will fail.

## Scenario
- User has two providers, A and B
- User is actively using both A and B
- User makes a change on A and a change on B that both effect the same stateful store
- Provider A makes the change
- Provider B uses the cached value from before the change by A
- The changes are in a race condition
- Provider B would like to safely use the cached value without fear of an invalid hash using capacity

## Checklist
- [ ] Make sure the pallet benchmarks are correct
- [ ] Make sure the overhead benchmarks are correct

## Questions
- Should this validation run on `pre_dispatch` as well as `validate`?
    - **A**: We want to do both 
    - If `pre_dispatch` and `validate`, we should remove the check in the extrinsic
    - If just `validate`, we cannot remove the extrinsic check, but also no weights etc...
- Should other checks also be moved into the signed extrinsics?
    - Just the one check for the hash

Draft PR: https://github.com/LibertyDSNP/frequency/pull/1288

aramikm:
# Goal
The goal of this PR is 

Closes #1860

# Discussion

- 

# Checklist
- [ ] Updated Pallet Readme?
- [ ] Updated js/api-augment for Custom RPC OR Runtime API?
- [ ] Design doc(s) updated?
- [ ] Unit Tests added?
- [ ] e2e Tests added?
- [ ] Benchmarks added?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3707,6 +3707,7 @@ dependencies = [
  "pallet-frequency-tx-payment",
  "pallet-handles",
  "pallet-msa",
+ "pallet-stateful-storage",
  "pallet-transaction-payment",
  "polkadot-cli",
  "polkadot-parachain-primitives",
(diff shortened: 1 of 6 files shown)


aramikm:
moved to https://github.com/frequency-chain/frequency/pull/2137

PR #2137, opened 2024-08-21
stateful-storage: check hash in signed extension

### LINKED ISSUE #1860: `target_hash` RPC Level validation

aramikm:
# Goal
The goal of this PR is to check the hash for stateful storage transactions and fail early.

Closes #1860

# Checklist
- [x] e2e Tests added?
- [X] Benchmarks added?


pallets/stateful-storage/src/signed_extension.rs
```diff
@@ -0,0 +1,200 @@
...
+use scale_info::TypeInfo;
+use sp_runtime::{
+	traits::{DispatchInfoOf, Dispatchable, SignedExtension},
+	transaction_validity::{InvalidTransaction, TransactionValidity, TransactionValidityError},
+	DispatchError,
+};
+
+/// The SignedExtension trait is implemented on CheckFreeExtrinsicUse to validate the request. The
+/// purpose of this is to ensure that the target_hash is verified in transaction pool before getting
+/// into block. This is to reduce the chance that capacity consumption due to stale hash
```
shannonwells, line 21: nit. only if you are already going to change something in this file
```suggestion
/// into block. This is to reduce the chance of capacity consumption due to stale hash
```

pallets/stateful-storage/src/signed_extension.rs
```diff
@@ -0,0 +1,200 @@
...
+		_call: &Self::Call,
+		_info: &DispatchInfoOf<Self::Call>,
+		_len: usize,
+	) -> Result<Self::Pre, TransactionValidityError> {
+		// Since we already check the hash in stateful-storage-pallet extrinsics we do not need to
+		// check the hash before dispatching
+		Ok(())
+	}
+
+	/// Filters stateful extrinsics and verifies the hashes to ensure we are proceeding
```
shannonwells, line 131: not?
```suggestion
	/// Filters stateful extrinsics and verifies the hashes to ensure we are not proceeding
```

shannonwells (approved):
Only a couple of little things for comments, not worth doing a CI round trip unless you are already going to change those files.

pallets/stateful-storage/src/signed_extension.rs
```diff
@@ -0,0 +1,200 @@
...
+				Self::verify_hash_paginated(
+					&state_owner_msa_id,
+					&payload.schema_id,
+					&payload.page_id,
+					&payload.target_hash,
+				)
+			},
+			_ => Ok(Default::default()),
+		}
+	}
```
saraswatpuneet, line 199: nit: can this match block be further refactored?

saraswatpuneet (dismissed):
Read through, looks good

Do we need to update any weights  (overhead) ?

pallets/stateful-storage/src/tests/signed_extension_tests.rs
```diff
@@ -0,0 +1,191 @@
...
+			msa_id,
+			paginated_schema_id,
+			0,
+			prev_content_hash,
+			BoundedVec::try_from(payload.clone()).unwrap(),
+		));
+		let calls: Vec<<Test as frame_system::Config>::RuntimeCall> = vec![
+			RuntimeCall::StatefulStoragePallet(Call::apply_item_actions {
+				actions: BoundedVec::try_from(actions.clone()).unwrap(),
+				target_hash: NONEXISTENT_PAGE_HASH,
```
JoeCap08055, line 53: maybe a nit, but I feel like using `NONEXISTENT_PAGE_HASH` (== 0) in the tests is a bit too simple; it feels like it has the potential to hit a special case in the code under test (I'm pretty sure it doesn't, but it feels weird to me anyway). Maybe better to use a random number? (Or test both cases: 0 and <random number>?)

pallets/stateful-storage/src/signed_extension.rs
```diff
@@ -0,0 +1,200 @@
...
+		if let Ok(Some(page)) = Pallet::<T>::get_itemized_page_for(*msa_id, *schema_id) {
+			let current_hash: PageHash = page.get_hash();
+
+			ensure!(
+				&current_hash == target_hash,
+				map_dispatch_error(Error::<T>::StalePageState.into())
+			);
+
+			return ValidTransaction::with_tag_prefix(TAG_PREFIX)
+				.and_provides((msa_id, schema_id))
```
enddynayn, line 49: 👍

pallets/stateful-storage/src/signed_extension.rs
```diff
@@ -0,0 +1,200 @@
...
+		if let Ok(Some(page)) = Pallet::<T>::get_paginated_page_for(*msa_id, *schema_id, *page_id) {
+			let current_hash: PageHash = page.get_hash();
+
+			ensure!(
+				&current_hash == target_hash,
+				map_dispatch_error(Error::<T>::StalePageState.into())
+			);
+
+			return ValidTransaction::with_tag_prefix(TAG_PREFIX)
+				.and_provides((msa_id, schema_id, page_id))
```
enddynayn, line 72: 👍

pallets/stateful-storage/src/signed_extension.rs
```diff
@@ -0,0 +1,200 @@
...
+		write!(f, "StatefulStorageSignedExtension<{:?}>", self.0)
+	}
+	#[cfg(not(feature = "std"))]
+	fn fmt(&self, _: &mut sp_std::fmt::Formatter) -> sp_std::fmt::Result {
+		Ok(())
+	}
+}
+
+/// Map a module DispatchError to an InvalidTransaction::Custom error
+pub fn map_dispatch_error(err: DispatchError) -> InvalidTransaction {
```
enddynayn, line 92: fancy

pallets/stateful-storage/src/signed_extension.rs
```diff
@@ -0,0 +1,200 @@
...
+	InvalidTransaction::Custom(match err {
+		DispatchError::Module(module_err) =>
+			<u32 as Decode>::decode(&mut module_err.error.as_slice())
+				.unwrap_or_default()
+				.try_into()
+				.unwrap_or_default(),
+		_ => 255u8,
```
enddynayn, line 99: It looks like we might end up with an error code that isn’t entirely accurate. Same in matching case - zero.

enddynayn (approved):
👍 great!

pallets/stateful-storage/src/signed_extension.rs
JoeCap08055: Also, just a note: the "Custom Error" code that we get as part of the RPC "1010" error is entirely arbitrary, and does not indicate the pallet. So a code of `9` (StalePageHash) is not meaningful to the end user, because they don't know what pallet enum that comes from. I do think somewhere there is an enum of custom RPC error codes for Frequency; perhaps we should map to that instead? Sadly Polkadot/Substrate does not give us a way to propagate anything more than an integer value for custom RPC errors.

pallets/stateful-storage/src/tests/signed_extension_tests.rs
aramikm: Changing to a random value on any run would create a flaky test with very low priority of failure but still using a deterministic stale value would be better. I'll use a different hash variable name to make it less confusing.

aramikm:
> Read through, looks good
> 
> Do we need to update any weights (overhead) ?

I calculated the overhead benchmark and it didn't change. My guess is that since we don't have anything inside `pre_dispatch` it didn't add to overhead.

pallets/stateful-storage/src/signed_extension.rs
```diff
@@ -0,0 +1,200 @@
...
+			Some(Call::delete_page {
+				state_owner_msa_id, schema_id, target_hash, page_id, ..
+			}) => Self::verify_hash_paginated(state_owner_msa_id, schema_id, page_id, target_hash),
+			Some(Call::apply_item_actions_with_signature { payload, .. }) =>
+				Self::verify_hash_itemized(
+					&payload.msa_id,
+					&payload.schema_id,
+					&payload.target_hash,
+				),
+			Some(Call::upsert_page_with_signature { payload, .. }) => Self::verify_hash_paginated(
```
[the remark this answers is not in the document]
enddynayn: That helps. I suggest doing it only if you need to run the build again,.

pallets/stateful-storage/src/signed_extension.rs
enddynayn: Interesting approach but works.

JoeCap08055: Probably a wider issue than this PR, but the universal enum idea might not be so bad; we wouldn't need to map every error from every pallet; only the ones that can happen in a SignedExtension check. There shouldn't be that many.

Anyway, this approach works for now; maybe it's worth a backlog issue to revisit how we do this across all of our signed extensions, or maybe not.

JoeCap08055 (approved):

runtime/frequency/src/lib.rs
```diff
@@ -1235,6 +1249,279 @@ construct_runtime!(
...
+	fn fmt(&self, f: &mut sp_std::fmt::Formatter) -> sp_std::fmt::Result {
+		write!(f, "StaleHashCheckExtension")
+	}
+	#[cfg(not(feature = "std"))]
+	fn fmt(&self, _: &mut sp_std::fmt::Formatter) -> sp_std::fmt::Result {
+		Ok(())
+	}
+}
+
+impl SignedExtension for StaleHashCheckExtension {
```
aramikm, line 1296: had to bring it out of pallet and inside runtime to be able to support multiple pallets and instead of using generics which was causing issues when used for multiple pallets it's directly using the concrete type used in the runtime

enddynayn: Could we have used a similar approach to how we inject CapacityEligibleCalls into the frequency-tx-payment pallet and injected a struct that includes the extract_hash_data function?

runtime/frequency/src/lib.rs
```diff
@@ -1235,6 +1249,279 @@ construct_runtime!(
...
+				)],
+				_ => vec![],
+			},
+			RuntimeCall::FrequencyTxPayment(FrequencyPaymentCall::pay_with_capacity {
+				call,
+				..
+			}) => Self::extract_hash_data(call),
+			RuntimeCall::FrequencyTxPayment(
+				FrequencyPaymentCall::pay_with_capacity_batch_all { calls, .. },
+			) => calls.iter().flat_map(|c| Self::extract_hash_data(c)).collect(),
```
enddynayn, line 1452: 👍

enddynayn (approved):
Great! 👍  I also suggest adding some unit tests to the signed extension.

runtime/frequency/src/lib.rs
JoeCap08055: Nice! I like this refactor.

JoeCap08055 (approved):
Nice rework.
- Pulled & built branch
- Ran unit tests
- Ran e2e tests

Looks good!

shannonwells (approved):
Looks good to me.  I think it's good to go and if it can be done the same way as CapacityEligibleCalls I think it can be revisited later.

runtime/frequency/src/lib.rs
aramikm: I think I can make that work. I'll close and merge this and add an issue for that

aramikm:
Refactor issue is created https://github.com/frequency-chain/frequency/issues/2145

PR #2144, opened 2024-08-29
docker: add health check

### LINKED ISSUE #1390: Add Heathcheck to Docker files
Docker supports a heathcheck https://docs.docker.com/engine/reference/builder/#healthcheck

We should use it with the rpc `/health` to know that the server is at least up.

aramikm:
# Goal
The goal of this PR is to add health check to docker images

Closes #1390
# Discussion
- Using the already existed script written by Puneet
- another alternative is to use something like the following
```bash
HEALTHCHECK --interval=300s --timeout=75s --start-period=30s --retries=3 \
  CMD curl -f http://localhost:9944/health || exit 1
```

### DIFF
--- a/docker/parachain-node.dockerfile
+++ b/docker/parachain-node.dockerfile
@@ -37,5 +37,9 @@ VOLUME ["/chain-data"]
 
 ENTRYPOINT ["/frequency/frequency"]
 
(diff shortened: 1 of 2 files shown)


wilwade:
Closing for now.

PR #2152, opened 2024-09-05
Chore/mv development genesis to runtime issue 2149

enddynayn:
# Goal
The goal of this PR is 

Closes 

# Discussion

- 

# Checklist
- [ ] Updated Pallet Readme?
- [ ] Updated js/api-augment for Custom RPC OR Runtime API?
- [ ] Design doc(s) updated?
- [ ] Unit Tests added?
- [ ] e2e Tests added?
- [ ] Benchmarks added?


runtime/frequency/src/development_genesis.rs
```diff
@@ -0,0 +1,154 @@
...
+
+pub fn development_technical_committee_members() -> Vec<AccountId> {
+	vec![
+		get_account_id_from_seed::<sr25519::Public>("Bob"),
+		get_account_id_from_seed::<sr25519::Public>("Dave"),
+	]
+}
+
+/// Helper function to generate a crypto pair from seed
+#[allow(clippy::expect_used)]
```
enddynayn, line 126: add comment // The panic from expect will not occur here because these input values are hardcoded.

runtime/frequency/src/development_genesis.rs
```diff
@@ -0,0 +1,154 @@
...
+/// Generate collator keys from seed.
+///
+/// This function's return type must always match the session keys of the chain in tuple format.
+pub fn get_collator_keys_from_seed(seed: &str) -> AuraId {
+	get_public_from_seed::<AuraId>(seed)
+}
+
+/// Helper function to generate a crypto pair from seed
+#[allow(clippy::expect_used)]
+pub fn get_public_from_seed<TPublic: Public>(seed: &str) -> <TPublic::Pair as Pair>::Public {
```
enddynayn, line 150: add comment // The panic from expect will not occur here because these input values are hardcoded.

runtime/frequency/src/development_genesis.rs
```diff
@@ -0,0 +1,154 @@
...
+// Generate the session keys from individual elements.
+//
+// The input must be a tuple of individual keys (a single arg for now since we have just one key).
+fn template_session_keys(keys: AuraId) -> SessionKeys {
+	SessionKeys { aura: keys }
+}
+
+#[allow(clippy::unwrap_used)]
+fn load_genesis_schemas() -> Vec<crate::pallet_schemas::GenesisSchema> {
+	serde_json::from_slice(include_bytes!("../../../resources/genesis-schemas.json")).unwrap()
```
enddynayn, line 19: Does anyone have any opinion on moving this file out of resources into a new location?

JoeCap08055: seems fine where it is to me

JoeCap08055 (approved):
- Reviewed changes
- Built & ran node, verified genesis config
- Ran unit & e2e tests, passing

runtime/frequency/src/lib.rs
```diff
@@ -26,6 +26,9 @@ use sp_runtime::{
 	DispatchError,
 };
 
+pub mod development_genesis;
+pub mod genesis_config_presets;
```
wilwade, line 30: Should these be feature flagged to `frequency-local` and `frequency-local`? They shouldn't be in mainnet or testnet right?

wilwade (changes requested):
Haven't gotten all the way through, but wanted to call out one fix. The rest looks good so far. Excited to see this happening!

runtime/frequency/src/lib.rs
enddynayn: I’ve been going back and forth on this. The API suggests that you can retrieve a list of all presets, which implies we should have more than just one. I also didn’t notice their example using flags. In any case, I don’t see it as a big deal to keep them. 

How do you feel about feature flagging them once we get all of them in as a final task?

runtime/frequency/src/apis.rs
```diff
@@ -149,11 +152,13 @@ impl_runtime_apis! {
...
 
 		fn get_preset(id: &Option<sp_genesis_builder::PresetId>) -> Option<Vec<u8>> {
-			get_preset::<RuntimeGenesisConfig>(id, |_| None)
+			get_preset::<RuntimeGenesisConfig>(id,  &genesis_config_presets::get_preset)
 		}
 
 		fn preset_names() -> Vec<sp_genesis_builder::PresetId> {
-			Default::default()
+			vec![
+				sp_genesis_builder::PresetId::from("development"),
```
aramikm, line 160: my guess is that you are currently hardcoding `development` in here but it should support mainnet and testnet in future PR also?

aramikm (approved):
Overall looks good one concern is bloating the size of runtime wasm. We already know wasm has a size limit on the polkadot and if it's bigger than a certain size the upgrade won't work.

runtime/frequency/src/apis.rs
enddynayn: That’s correct. To make it easier to review, I’m breaking each one into its own PR.

runtime/frequency/src/lib.rs
wilwade: Hmm... I don't like the idea of the development code being available in our mainnet wasm. That breaks with our rule of separation.

I'm open to being convinced, but I think we should go ahead and add the feature flags before merging.

I would be fine with merging into a feature development branch then a PR to do the flagging and a final PR to merge it all to main.

enddynayn: >Hmm... I don't like the idea of the development code being available in our mainnet wasm. That breaks with our rule of separation.

I agree with that. Less code means less issues. I am curious on the reasoning behind Polkadot not having the same concern.

>I would be fine with merging into a feature development branch then a PR to do the flagging and a final PR to merge it all to main.

That sounds good too me. I'll add to the acceptance criteria of the parent ticket feature flagging the presets.

mattheworris (approved):
- Read through changes

🚢 it!

deny.toml
```diff
@@ -74,6 +74,7 @@ ignore = [
   { id = "RUSTSEC-2024-0336", reason = "Only use of rustls@v0.20.9 is in futures-rustls which does not use the effected code" },
   { id = "RUSTSEC-2024-0344", reason = "We are only able to remove this once parity updates its dependencies. Older versions of curve25519-dalek should get replaces with >= 4.1.3" },
   { id = "RUSTSEC-2022-0093", reason = "The vulnerable code is not exploitable in Frequency because the signing function is not exposed in a way that allows the use of arbitrary public keys, ensuring protection against the described vulnerability." },
+  { id = "RUSTSEC-2024-0370", reason = "Although proc-macro-error is unmaintained, the code remains stable and is not expected to undergo significant changes. Its functionality is sufficient for current requirements, and immediate action is not necessary. We will evaluate alternatives if future updates or maintenance are needed." },
```
claireolmstead, line 77:
```suggestion
  { id = "RUSTSEC-2024-0370", reason = "Although proc-macro-error is unmaintained, the code remains stable and is not expected to undergo significant changes. Its functionality is sufficient for current requirements and immediate action is not necessary. We will evaluate alternatives if future updates or maintenance is needed." },
```

claireolmstead (approved):

runtime/frequency/src/lib.rs
enddynayn: Here’s the link to the new branch that includes moving the Frequency and Paseo Genesis: https://github.com/frequency-chain/frequency/pull/2155

node/service/src/chain_spec/frequency_paseo.rs
```diff
@@ -67,108 +54,6 @@ pub fn local_paseo_testnet_config() -> ChainSpec {
...
-		],
-		// Technical Committee members
-		vec![
-			get_account_id_from_seed::<sr25519::Public>("Bob"),
-			get_account_id_from_seed::<sr25519::Public>("Dave"),
-		],
-		// ParaId
-		2000.into(),
-	))
+	.with_genesis_config_preset_name("frequency-local")
```
aramikm, line 57: question: the filename is `frequency_paseo.rs` but the config name in here is `frequency-local `. Just wanted to make sure we are doing this deliberately.

runtime/frequency/src/development_genesis.rs
```diff
@@ -0,0 +1,156 @@
...
+						template_session_keys(aura), // session keys
+					)
+				})
+				.collect(),
+		},
+		// no need to pass anything to aura, in fact it will panic if we do. Session will take care
+		// of this.
+		aura: Default::default(),
+		aura_ext: Default::default(),
+		#[cfg(any(not(feature = "frequency-no-relay"), feature = "frequency-lint-check"))]
```
aramikm, line 58: having `any` before `not` is a bit weird. Should it be the other way around?

runtime/frequency/src/development_genesis.rs
```diff
@@ -0,0 +1,156 @@
...
+		parachain_system: Default::default(),
+		#[cfg(any(not(feature = "frequency"), feature = "frequency-lint-check"))]
```
aramikm, line 60: same if here? This is hard to evaluate.

aramikm (commented):
Added some questions and nits.

runtime/frequency/src/development_genesis.rs
enddynayn: I think I have a bad merge. I switched the PR to draft in order to fix it.

node/service/src/chain_spec/frequency_paseo.rs
enddynayn: Thank you for pointing that out! I see what you mean, and you’re right — the naming might need a bit of updating in a future PR. For now, though, this file is specifically for Paseo local, which is referred to as frequency-local.

runtime/frequency/src/apis.rs
```diff
@@ -149,11 +149,32 @@ impl_runtime_apis! {
...
+			#[cfg(any(
+				feature = "frequency-no-relay",
+				feature = "frequency-local",
+				feature = "frequency-lint-check"
+			))]
+			presets.extend(
+			vec![
+				sp_genesis_builder::PresetId::from("development"),
+				sp_genesis_builder::PresetId::from("frequency-local"),
+				sp_genesis_builder::PresetId::from("frequency"),
```
aramikm, line 167: question: why do you need `frequency` for local features?

aramikm (approved):
Reviewed the code and looked good.

mattheworris (approved):
- Read through changes

claireolmstead (commented):
Read through this PR. Was mainly focused on code organization and naming since I'm not a pro in all of the rust semantics. That being said, looks good to me!

runtime/frequency/src/apis.rs
enddynayn: Good question! This is something I mentioned during our stand-up, and I believe we agreed that having the frequency feature enabled by default was ok.

JoeCap08055 (approved):
Looks good!
- Ran unit tests
- Ran e2e tests
- Inspected build with `chain-spec-builder`

PR #2157, opened 2024-09-17
Display handles are whitespace trimed and collapsed

### LINKED ISSUE #1673: User handles doesn't strip leading/trailing whitespace
While leading and trailing whitespace are allowed (they are stripped before canonicalization), the length requirement is being broken by only checking before canonicalization instead of after.

image
image

### LINKED ISSUE #2148: Trim Display Handle Whitespace
While it doesn't effect lookup due to the canonical handle, Display handles should be cleaner.

- Remove all whitespace from the leading and trailing end of a display handle
     - ` this is a display handle              ` -> `this is a display handle`
- Collapse all whitespace to just one ` ` if duplicated
     - `this    is     a display        handle`   -> `this is a display handle`
     
## Notes

- Do not modify existing handles. They can remain
- Watch out for loop cost. Might be worth adding in with another loop such as ban loop for `@#`etc..

wilwade:
# Goal
The goal of this PR is to ensure that all new base handles (and thus the resulting display handle) have whitespace trimmed and concatenated

Closes #2148 

# Discussion

Fairly straightforward. I did benchmark against the collapse crate, and tested some with the versions here: https://stackoverflow.com/questions/71864137/whats-the-ideal-way-to-trim-extra-spaces-from-a-string

In the end this one benchmarked the fastest using a version of this:

```rust
#[cfg(test)]
mod tests {
    use super::*;
    use rand::distributions::Alphanumeric;
    use rand::distributions::DistString;
    use test::Bencher;

    #[bench]
    fn bench_one(b: &mut Bencher) {
        let mut rng = rand::thread_rng();
        b.iter(|| {
            trim_and_collapse_whitespace(
                Alphanumeric
                    .sample_string(&mut rng, 20)
                    .to_lowercase()
                    .replace("a", " ")
                    .replace("b", " ")
                    .as_str(),
            );
        });
    }
}
```

# Checklist
- [x] Updated Pallet Readme?
- [x] Design doc(s) updated?
- [x] Unit Tests added?

### DIFF
--- a/designdocs/user-handles.md
+++ b/designdocs/user-handles.md
@@ -36,7 +36,7 @@ The high level requirements for user handles are:
 * **Suffix**: A suffix is a numeric value that is appended to a base handle to make it unique.  The suffix is randomly constrained and selected within a range configured on chain.
 * **Delimiter**: A delimiter is a character that is used to separate the base handle and suffix.  The delimiter is a period (.) by default.
 * **Display Handle**: A display handle is formed by concatenating the base handle and suffix with the delimiter.  The display handle is the handle that is displayed to the user.
-  
+
 ### Handle Guidelines
 
 * User handles must be unique and each MSA can only be mapped to one handle.
@@ -48,6 +48,7 @@ The high level requirements for user handles are:
 * Suffix will be randomly constrained.
 * Homoglyph versions of handles should still resolve to the same ```msa_id``` (e.g. ```user.1234``` and ```u$er.1234``` should resolve to the same ```msa_id```).
 * `Display Handle` is the handle that is displayed to the user and must be unique.
+* `Base Handle` should have UTF-8 whitespace trimmed and concatenated (e.g. ```   u  s \t\t  e\tr ``` will be trimmed to ```u s e r```).
 
 ## Proposal
 
@@ -66,7 +67,7 @@ Handles```registry``` on Frequency chain.
 
 ### Handle replay attack
 
-To prevent replay attacks, the chain will require a mortality period for handles. This can be included in signed payload from user as a `expiration` field. The chain will check if the `expiration` is greater than the current block number. If not, the transaction will fail.  
+To prevent replay attacks, the chain will require a mortality period for handles. This can be included in signed payload from user as a `expiration` field. The chain will check if the `expiration` is greater than the current block number. If not, the transaction will fail.
 
 ### Handling Race Conditions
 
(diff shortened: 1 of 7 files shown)


pallets/handles/src/lib.rs
```diff
@@ -711,8 +713,9 @@ pub mod pallet {
 			base_handle: &str,
 			suffix: HandleSuffix,
 		) -> Result<DisplayHandle, DispatchError> {
+			let base_handle_trimmed = trim_and_collapse_whitespace(base_handle);
```
wilwade, line 716: Believe it or not, this appears to be the place that builds the display handle that is saved.

pallets/handles/src/handles-utils/src/converter.rs
```diff
@@ -84,7 +84,7 @@ pub fn split_display_name(display_name_str: &str) -> Option<(String, HandleSuffi
 	let parts: Vec<&str> = display_name_str.split(HANDLE_DELIMITER).collect();
 	let base_handle_str = parts[0].to_string();
 	if parts.len() != 2 {
-		return None
+		return None;
```
wilwade, line 87: My formatter really wants to have `;` at the end of the line when it is `return`

saraswatpuneet (approved):
clean! lgtm

designdocs/user-handles.md
```diff
@@ -66,7 +67,7 @@ Handles```registry``` on Frequency chain.
 
 ### Handle replay attack
 
-To prevent replay attacks, the chain will require a mortality period for handles. This can be included in signed payload from user as a `expiration` field. The chain will check if the `expiration` is greater than the current block number. If not, the transaction will fail.  
+To prevent replay attacks, the chain will require a mortality period for handles. This can be included in signed payload from user as a `expiration` field. The chain will check if the `expiration` is greater than the current block number. If not, the transaction will fail.
```
mattheworris, line 70: nit: correct a to an, and active voice
```suggestion
To prevent replay attacks, the chain will require a mortality period for handles. This can be included in the user's signed payload as an `expiration` field. The chain will check if the `expiration` is greater than the current block number. If not, the transaction will fail.
```

designdocs/user-handles.md
```diff
@@ -135,7 +136,7 @@ ClaimHandlePayload {
 
 ### Claim handle
 
- As a network, Frequency should allow users to choose their own handle, while the chain will generate a random numeric suffix within the range of suffixes allowed. The display handle will be the base handle with the suffix.
+ As a network, Frequency should allow users to choose their own handle, while the chain will generate a random numeric suffix within the range of suffixes allowed. The display handle will be the base handle with whitespace trimmed and concatenated with the suffix.
```
JoeCap08055, line 139:
```suggestion
 As a network, Frequency should allow users to choose their own handle, while the chain will generate a random numeric suffix within the range of suffixes allowed. The display handle will be the base handle with whitespace trimmed and consolidated, concatenated with the suffix.
```

Issue #2011, opened 2024-06-05
Feature: Support New Ledger Wallet App


wilwade:
We want to support the upcoming Generic Polkadot Ledger Application on Frequency.

How to: https://forum.polkadot.network/t/new-polkadot-ledger-app/8817

Known Tasks
- [x] Add the `CheckMetadataHash` Signed extension See: https://paritytech.github.io/polkadot-sdk/master/polkadot_sdk_docs/guides/enable_metadata_hash/index.html

Waiting on...
- [x] https://github.com/polkadot-fellows/runtimes/pull/337
- [x] #1894 Upgrade to v1.8.0+
- [x] Ledger App Live [Tracking Forum Post](https://forum.polkadot.network/t/polkadot-generic-ledger-app/4295/)
- [x] Extension support for the new Ledger App
- [x] Test on Paseo Testnet: Difficult to test

Post Chain Updates
- [x] Check to make sure that https://github.com/TalismanSociety/chaindata correctly identifies `hasCheckMetadataHash = true`
- [x] PR of https://github.com/frequency-chain/polkadot-js-common/tree/add-frequency-to-generic-ledger-apps to https://github.com/polkadot-js/common

wilwade:
All work completed.

PR #2159, opened 2024-09-20
Fix token symbol testnet

wilwade:
# Goal
The goal of this PR is to fix an issue with the token symbols for testnet found while testing CheckMetadataHash extension.

Part of #2011 

# Discussion

- Corrected the symbols used for `node/service/src/chain_spec/frequency_paseo.rs`
- Corrected the symbols used in the `substrate_wasm_builder::WasmBuilder::init_with_defaults().enable_metadata_hash`
- Added a skipped e2e test that can be run to test transacting using the `metadata_hash`
- Required: Update cargo deny exclusions
- Adjacent: Updated the API Augment packages
- Adjacent: Updated the API Augment to have all the no-op signed extensions and remove warnings
- Adjacent: Updated the API Augment to list all the old v1 rpcs remove warnings

# Checklist
- [x] Updated js/api-augment for Custom RPC OR Runtime API?
- [x] e2e Tests added?

# How to Test

With the e2e test:
- Build Frequency using the feature `metadata-hash` (EZ mode: edit line 109 of `init.sh` to have it as one of the features)
- Set the `skip` in `e2e/signed-extensions/checkMetadataHash.test.ts` to be only
- Run `make js`
- `cd e2e`
- `npm i ../js/api-augment/dist/frequency-chain-api-augment-0.0.0.tgz`
- `npm run test:serial`

1. See that the test passes
2. See that there are no warnings about signed extensions
3. See that there are no warnings about rpcs

Note: You will see an warning about `PORTABLEREGISTRY: Unable to determine runtime Call type, cannot inspect sp_runtime::generic::unchecked_extrinsic::UncheckedExtrinsic` However that appears to be due to an issue with the latest version of Polkadotjs/api and not related to these changes nor does it impact the running of the api.


e2e/signed-extensions/checkMetadataHash.test.ts
wilwade, line 1: Here's the new test, but it is skipped by default

js/api-augment/index.ts
```diff
@@ -38,6 +62,14 @@ export const signedExtensions: ExtDef = {
...
     payload: {},
   },
+  StaleHashCheckExtension: {
+    extrinsic: {},
+    payload: {},
+  },
+  StorageWeightReclaim: {
+    extrinsic: {},
+    payload: {},
+  },
```
wilwade, line 72: These are the additional signed extensions, but they don't need anything extra. This removes the warning.

node/service/src/chain_spec/frequency_paseo.rs
```diff
@@ -54,7 +54,7 @@ fn template_session_keys(keys: AuraId) -> frequency_runtime::SessionKeys {
 pub fn local_paseo_testnet_config() -> ChainSpec {
 	// Give your base currency a unit name and decimal places
 	let properties =
-		get_properties(FREQUENCY_LOCAL_TOKEN, TOKEN_DECIMALS as u32, Ss58Prefix::get().into());
+		get_properties(FREQUENCY_TESTNET_TOKEN, TOKEN_DECIMALS as u32, Ss58Prefix::get().into());
```
wilwade, line 57: I set paseo-local to use `XRQCY`, so that it better matches with paseo live, but willing to move it back. However this aligns it for the metadata on all "paseo" chains.

runtime/common/src/constants.rs
```diff
@@ -10,6 +10,7 @@ use frame_support::{
 	PalletId,
 };
 
+// Duplicated in runtime/frequency/build.rs to keep build dependencies low
```
wilwade, line 13: I did try importing it, but having runtime common as a build dependency was having issues and it was way too much for a few constants.

runtime/common/src/constants.rs
```diff
@@ -10,6 +10,7 @@ use frame_support::{
...
 pub const FREQUENCY_TESTNET_TOKEN: &str = "XRQCY";
 pub const FREQUENCY_LOCAL_TOKEN: &str = "UNIT";
```
JoeCap08055, line 15: Do we need this constant any more, then? 'UNIT'?

node/service/src/chain_spec/frequency_paseo.rs
JoeCap08055: The variety of configs gets confusing sometimes--under what scenarios would we still use `FREQUENCY_LOCAL_TOKEN`, then?

JoeCap08055: Ah, nvm. I see `paseo-local` !== `dev`

JoeCap08055 (approved):
- Read through changes
Looks good.

js/api-augment/substrate_v1_rpcs.ts
```diff
@@ -0,0 +1,75 @@
+import { DefinitionRpc } from '@polkadot/types/types';
+
+export const v1SubstrateRpcs: Record<string, Record<string, DefinitionRpc>> = {
```
aramikm, line 3: Is this generated? If not what is the source and how it might evolve in future?

runtime/frequency/build.rs
```diff
@@ -1,12 +1,40 @@
...
+fn main() {
+	substrate_wasm_builder::WasmBuilder::init_with_defaults()
+		.enable_metadata_hash(FREQUENCY_TOKEN, TOKEN_DECIMALS)
+		.build()
+}
+
+#[cfg(all(feature = "std", feature = "metadata-hash", feature = "frequency-testnet"))]
+fn main() {
+	substrate_wasm_builder::WasmBuilder::init_with_defaults()
+		.enable_metadata_hash(FREQUENCY_TOKEN, TOKEN_DECIMALS)
```
aramikm, line 26:
```suggestion
		.enable_metadata_hash(FREQUENCY_TESTNET_TOKEN, TOKEN_DECIMALS)
```

js/api-augment/substrate_v1_rpcs.ts
wilwade: It is not. I generated it via the list of "unknown" rpcs. I went looking for being able to pull it from somewhere, but nothing had it from what I could find on GitHub. :(

runtime/frequency/build.rs
wilwade: Thank you. The one thing I am trying to fix and my copy paste for the final setup broke it!

mattheworris (approved):
- Read through the changes
- Confirmed the testing results
🚢 it!

aramikm (approved):
Read the code. Looks good!

PR #2161, opened 2024-09-24
Reserved word check should be post canonical conversion

### LINKED ISSUE #2147: Reserved handle word check should be post-canonical
## Summary
A user can claim a handle very similar to those in the `RESERVED_WORDS` list by using Unicode characters.

## Issue details
During the execution of the [`claim_handle`](https://github.com/frequency-chain/frequency/blob/9ef58185f2230aaf92dc166fb8eade56a78e4cf3/pallets/handles/src/lib.rs#L371) extrinsic, the base handle, which is passed by the user, is validated by the `validate_base_handle` function. One of the checks in this function ensures that the handle is not part of a list of reserved words:

```rust
ensure!(!is_reserved_handle(&base_handle_str), Error::<T>::HandleIsNotAllowed);
```

The `is_reserved_handle` function does not check the canonical base representation of the `base_handle_str`. As a result, a user can create a handle which looks very similar to a reserved one.

As an example, the following test claims a handle with "here" as prefix, by using Unicode character "\u0435" instead of "e" for the first 'e':

```rust
fn claim_handle_unicode() {
		let alice = sr25519::Pair::from_seed(&[0; 32]);
		let expiry = 100;
		let (payload, proof) =
			get_signed_claims_payload(&alice, "hеre".as_bytes().to_vec(), expiry);
		assert_ok!(Handles::claim_handle(
			RuntimeOrigin::signed(alice.public().into()),
			alice.public().into(),
			proof,
			payload
		));

	});
}
```

The code for the final handle creation correctly uses the `base_handle_str` with a suffix that does take the canonical representation into account. As a result, there will never be two handles that only differ by a similar Unicode character. 

However, to prevent any kind of confusion, the check for reserved handles in the `RESERVED_WORDS` list should  use the canonical representation.

## Mitigation
A simple mitigation would be to adjust the `is_reserved_handle` function to check the canonicalized representations of the strings:
```rust
pub fn is_reserved_handle(input_str: &str) -> bool {
  RESERVED_WORDS.map(|w| convert_to_canonical(w)).contains(&convert_to_canonical(&input_str))
}
```

However for efficiency, reserved word map should be converted for the constant and test moved to after the canonical conversion of the input.

wilwade:
# Goal
The goal of this PR is to make sure that reserved words cannot be used in different forms.

Closes #2147

# Discussion

- Used `to_lowercase` instead of `to_ascii_lowercase` for better handling
- Move the three canonical validations together

# Checklist
- [x] Unit Tests added?


pallets/handles/src/handles-utils/src/converter.rs
```diff
@@ -35,7 +35,7 @@ pub fn convert_to_canonical(input_str: &str) -> alloc::string::String {
 	let white_space_stripped = strip_unicode_whitespace(input_str);
 	let diacriticals_stripped = strip_diacriticals(&white_space_stripped);
 	let confusables_removed = replace_confusables(&diacriticals_stripped);
-	confusables_removed.to_ascii_lowercase()
+	confusables_removed.to_lowercase()
```
wilwade, line 38: Not sure why we didn't use this before, but we should have.

pallets/handles/src/handles-utils/src/validator.rs
```diff
@@ -10,7 +10,21 @@ use constants::ALLOWED_UNICODE_CHARACTER_RANGES;
 
 /// Reserved words that cannot be used as the handle.
 const RESERVED_WORDS: [&str; 8] =
-	["admin", "everyone", "all", "administrator", "mod", "moderator", "here", "channel"];
+	["adm1n", "every0ne", "a11", "adm1n1strat0r", "m0d", "m0derat0r", "here", "channe1"];
```
wilwade, line 13: Oddly enough, the best way to do this was to do the conversion statically.

Note that `convert_to_canonical` cannot be used here as it cannot be converted into a const fn

pallets/handles/src/lib.rs
```diff
@@ -668,33 +664,36 @@ pub mod pallet {
 				Error::<T>::InvalidHandleCharacterLength
 			);
 
-			ensure!(!is_reserved_handle(&base_handle_str), Error::<T>::HandleIsNotAllowed);
```
wilwade, line 671: This is no longer a check done on the base_handle pre-canonicalization

pallets/handles/src/lib.rs
```diff
@@ -765,10 +764,7 @@ pub mod pallet {
 					let (canonical_handle_str, _) =
 						Self::get_canonical_string_vec_from_base_handle(&base_handle_str);
 
-					return Self::validate_canonical_handle_contains_characters_in_supported_ranges(
-						&canonical_handle_str,
-					)
-					.is_ok();
+					return Self::validate_canonical_handle(&canonical_handle_str).is_ok();
```
wilwade, line 767: Not sure why this wasn't doing all the other canonical checks, but it should have been.

mattheworris (approved):
- Read through changes
lgtm

pallets/handles/src/handles-utils/src/validator.rs
```diff
@@ -10,7 +10,21 @@ use constants::ALLOWED_UNICODE_CHARACTER_RANGES;
...
+fn ensure_reserved_words_canonical() {
+	for &word in &RESERVED_WORDS {
+		let canonical = crate::convert_to_canonical(word);
+		assert_eq!(
+			word, canonical,
+			"The reserved word '{}' MUST match canonical form: '{}'",
+			word, canonical
+		);
+	}
+}
```
JoeCap08055, line 27: not sure if this would be a better test or not; otherwise technically the test passes if even if `convert_to_canonical` is a no-op
```suggestion

// We MUST have the RESERVED_WORDS constant as canonical.
// Cannot easily be canonicalized at compile time currently
#[test]
fn ensure_reserved_words_canonical() {
    const TEST_WORDS: [&str; 8] =
	["admin", "everyone", "all", "administrator", "mod", "moderator", "here", "channel"];

	for (i, &word) in TEST_WORDS.iter().enumerate() {
		let canonical = crate::convert_to_canonical(word);
		assert_eq!(
			RESERVED_WORDS[i], canonical,
			"The reserved word '{}' MUST match canonical form: '{}'",
			word, canonical
		);
	}
}
```

pallets/handles/src/handles-utils/src/validator.rs
aramikm: This is a bit confusing since we are using these keywords with like the following pub fn 
```rust
is_reserved_canonical_handle(input_str: &str) -> bool {
	RESERVED_WORDS.contains(&input_str)
}
```
this would only work if we convert all 
o -> 0
l -> 1 
and ...
which doesn't look like the canonical form.

aramikm: Well it looks like that's what we do which for a canonical form seems confusing.

pallets/handles/src/handles-utils/src/validator.rs
wilwade: True. This is just a check to make sure that `convert_to_canonical` does NOT change the output of `RESERVED_WORDS`. Other test (`test_is_reserved_canonical_handle_happy_path) make sure that reserved words are rejected.

aramikm (approved):
Read the code. Looked good to me

pallets/handles/src/handles-utils/src/validator.rs
wilwade: In retrospect canonical might have been better named something else, but yes, canonical is the trimmed, stripped, converted, etc... version

shannonwells (approved):
LGTM

PR #2163, opened 2024-09-24
Feat/support more languages in handles

### LINKED ISSUE #2162: [Feature]: Handles updates: disallow '/', support more languages
## Details
* Support more languages in handles o support the goal of "a social network platform for everybody"
* Also disallow '/' to help prevent people easily registering URIs in as handles

- [x] Added/updated tests

shannonwells:
# Goal
The goal of this PR is to:
1. support more languages in handles, 
2. close some gaps in language support for existing ones
3. disallow a '/' in handles.

Closes #2162 

# Discussion
In adding test cases, I discovered that there were some valid language characters in existing language support that would fail the handle check, such as some diacriticals and space indicators.

I also trimmed off the first 19 Latin characters since they wouldn't be printable, or they represent a backspace or a control key.

# Checklist
- [x] Unit Tests for all new languages


pallets/handles/src/handles-utils/constants.rs
```diff
@@ -2,27 +2,88 @@
...
+    0x1A00..=0x1A1F, // Buginese
+    0x1A20..=0x1AAD, // TaiTham
+    0x1B00..=0x1B7C, // Balinese
+    0x1B80..=0x1BB9, // Sundanese
+    0x1BC0..=0x1BFF, // Batak
+    0x1C00..=0x1C4F, // Lepcha
+    0x1C50..=0x1C7F, // Ol Chiki
+    0x1E00..=0x1EFF, // Latin Extended Additional
+    0x1F00..=0x1FFF, // Greek Extended
+    0x200C..=0x206F, // General punctuation, used in some languages to indicate syllables such as glottal stops
```
shannonwells, line 67: Not including this was causing some languages to fail even though they used things like Latin Extended-A otherwise

pallets/handles/src/handles-utils/constants.rs
```diff
@@ -2,27 +2,88 @@
...
-	0xF900..=0xFAFF, // CJK Compatibility Ideographs
-	0xFB50..=0xFDFF, // Arabic Presentation Forms-A
+/// Characters that are allowed.
+#[rustfmt::skip]
+pub const ALLOWED_UNICODE_CHARACTER_RANGES: [RangeInclusive<u16>; 81] = [
+    0x0020..=0x007A, // BasicLatin
+    0x0080..=0x00FF, // Latin-1 Supplement
+    0x0100..=0x017F, // Latin Extended-A
+    0x180..=0x24F,   // Latin Extended-B
+    0x02B0..=0x02FF, // Spacing Modifier Letters
```
wilwade, line 12: This is an odd one to include if it only includes "characters for the IPA, UPA, and other phonetic transcriptions"

https://en.wikipedia.org/wiki/Spacing_Modifier_Letters

pallets/handles/src/handles-utils/constants.rs
```diff
@@ -2,27 +2,88 @@
...
-	0x3040..=0x309F, // Hiragana
-	0x30A0..=0x30FF, // Katakana
-	0x3400..=0x4DBF, // CJK Unified Ideographs Extension A
-	0x4E00..=0x9FFF, // CJK Unified Ideographs
-	0xAC00..=0xD7AF, // Hangul Syllables
-	0xF900..=0xFAFF, // CJK Compatibility Ideographs
-	0xFB50..=0xFDFF, // Arabic Presentation Forms-A
+/// Characters that are allowed.
+#[rustfmt::skip]
+pub const ALLOWED_UNICODE_CHARACTER_RANGES: [RangeInclusive<u16>; 81] = [
```
wilwade, line 7: We didn't do this before due to the short list, but do we need to have a `const fn` that converts this at build time to a compressed list of ranges? Aka (in decimal) [1-5, 5-7, 9-10]` -> `[1-7, 9-10]`?

This does turn into `O(<range count>*<handle char length>)` so likely worth it with the increased size and with many abutting ranges.

Could also just do it manually (which might be easier).

pallets/handles/src/handles-utils/src/validator.rs
```diff
@@ -27,8 +27,8 @@ fn ensure_reserved_words_canonical() {
 }
 
 /// Characters that cannot be used in the handle.
-const BLOCKED_CHARACTERS: [char; 16] =
-	['"', '#', '%', '(', ')', ',', '.', ':', ';', '<', '>', '@', '\\', '`', '{', '}'];
+const BLOCKED_CHARACTERS: [char; 17] =
+	['"', '#', '%', '(', ')', ',', '.', ':', ';', '<', '>', '@', '\\', '`', '{', '}', '/'];
```
wilwade, line 31: The test error is that this has to sorted. (see test right below this).
```suggestion
	['"', '#', '%', '(', ')', ',', '.', '/', ':', ';', '<', '>', '@', '\\', '`', '{', '}'];
```

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=a0fbf3b89116787e84db490a012234a0

wilwade (commented):
Few notes, but I think we can get this in.

pallets/handles/src/handles-utils/constants.rs
shannonwells: it's included because of failing checks for some words in certain languages.

pallets/handles/src/handles-utils/constants.rs
shannonwells: Added a test that generates the compacted list, which is now down to 54 items.

pallets/handles/src/handles-utils/src/tests/validator_tests.rs
```diff
@@ -39,37 +39,110 @@ fn test_contains_blocked_characters_negative() {
...
-		"דָּנִיֵּאלאבּבגּגדּדהווּוֹזחטי ִיכּךּכךלמםנןסעפּףּפףצץקרשׁשׂתּת",            // Hewbrew
 		"AaĄąBbCcĆćDdEeĘęFfGgHhIiJjKkLlŁłMmNnŃńOoÓóRrSsŚśYyZzŹźŻż", // Polish
 		"ÄäÖöÜüẞß",                                                 // German
 		"AÁBCČDĎEÉĚFGHChIÍJKLMNŇOÓPQRŘSŠTŤUÚŮVWXYÝZŽaábcčdďeéěfghchiíjklmnňoópqrřsštťuúůvwxyýzž", // Czech
 		"αιαιαιᾳειειηιῃοιοιυιυιωιῳαυαυᾹυᾱυευευηυηυουουωυωυγγγγγκγκγξγξγχγχμπμπντντΖζΤΖτζ", // Greek
 		"ÅåÄäÖö",                                                   // Swedish
 		"ÅåÄäÖöŠšŽž",                                               // Finnish
 		"ÆæØøÅå",                                                   // Danish
+		"Александр",                                                // Cyrillic
+		"Կրնամ ապակի ուտել և ինծի անհանգիստ չըներ։",             // Armenian
```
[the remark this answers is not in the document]
shannonwells: I really struggled with finding example text in all these languages.  But I found the "I can eat glass" on the Kermit website (which is a whole other rabbit hole) and you can read why they picked that there:  https://www.kermitproject.org/utf8.html

pallets/handles/src/handles-utils/src/tests/constants_tests.rs
```diff
@@ -0,0 +1,17 @@
+#[path = "../../constants.rs"]
+mod constants;
+use constants::*;
+
+// Use this to regenerate the compacted ALLOWED_UNICODE_CHARACTER_RANGES.
```
shannonwells, line 5: This is a tedious, semi-manual way of doing it, but I decided it was the best balance between 100% manual which was resulting in lots of mistakes, and spending a bunch of time writing a fully automated file generation script for something we will change only rarely.

claireolmstead (approved):
Not super well versed on how this all works, but at first glance, all of your changes look good!

JoeCap08055 (approved):
nice work--good compromise approach

pallets/handles/src/handles-utils/src/validator.rs
```diff
@@ -27,8 +27,8 @@ fn ensure_reserved_words_canonical() {
 }
 
 /// Characters that cannot be used in the handle.
-const BLOCKED_CHARACTERS: [char; 16] =
-	['"', '#', '%', '(', ')', ',', '.', ':', ';', '<', '>', '@', '\\', '`', '{', '}'];
+const BLOCKED_CHARACTERS: [char; 17] =
+	['"', '#', '%', '(', ')', ',', '.', '/', ':', ';', '<', '>', '@', '\\', '`', '{', '}'];
```
enddynayn, line 31: 👍

enddynayn (approved):
👍 nice!

Issue #2134, opened 2024-08-20
Signature Verification: Support Unwrapped Signatures in addtion to <Bytes> Wrapped


wilwade:
## Summary

The [account_ownership_proof](https://github.com/frequency-chain/frequency/blob/9ef58185f2230aaf92dc166fb8eade56a78e4cf3/pallets/passkey/src/types.rs#L66) parameter in `PasskeyCall` signifies the signature of `<Bytes>PASSKEY_PUBLIC_KEY</Bytes>`. This means that the wallet client or browser must use this format during registration.

## Issue details

Currently, the polkadot-js-extension wraps all raw signed bytes inside the [<Bytes\>](https://github.com/polkadot-js/extension/pull/743) tag, and the Frequency protocol follows the same approach during passkey registration. Specifically, it signs the payload `<Bytes>PASSKEY_PUBLIC_KEY</Bytes>` using the private key of the user's Substrate Account. During the transaction process, it [wraps](https://github.com/frequency-chain/frequency/blob/9ef58185f2230aaf92dc166fb8eade56a78e4cf3/pallets/passkey/src/lib.rs#L318) the `passkey_public_key` inside `<Bytes>`.

However, since any entity can develop an infrastructure for passkey registration and transaction submission, there is a risk that others might not wrap the `PASSKEY_PUBLIC_KEY` in `<Bytes>`, potentially causing confusion.

## Risk

The current method could lead to interoperability issues if different implementations do not use the `<Bytes>` wrapping for the `PASSKEY_PUBLIC_KEY`. This inconsistency could result in failed signature verification and transaction errors.

## Mitigation

The [check_account_signature](https://github.com/frequency-chain/frequency/blob/9ef58185f2230aaf92dc166fb8eade56a78e4cf3/pallets/passkey/src/lib.rs#L312) function should verify signatures for both the raw `signed_data` and `<Bytes>signed_data</Bytes>`. This dual-check approach, as implemented in [pallet-nfts](https://github.com/paritytech/polkadot-sdk/blob/ebcbca3ff606b22b5eb81bcbfaa9309752d64dde/substrate/frame/nfts/src/common_functions.rs#L41), would ensure compatibility across different implementations and reduce the risk of verification failures.

## Suggestion

This should be fixed everywhere

- [x] Fix for passkey
- [x] Fix for handles
- [x] Fix for MSA Pallet: Delegation, Add Key, etc...
- [x] Fix for Stateful Storage: Signature Requried

[...]

PR #2169, opened 2024-09-30
feat(msa): add support for unwrapped bytes

enddynayn:
Add support to verify signatures against unwrapped bytes. 
This allows for verification of signatures not using PolkadotUi.

#2134


pallets/msa/src/lib.rs
```diff
@@ -1097,6 +1097,11 @@ impl<T: Config> Pallet<T> {
 		payload: Vec<u8>,
 	) -> DispatchResult {
 		let key = T::ConvertIntoAccountId32::convert((*signer).clone());
+
+		if signature.verify(&payload[..], &key) {
+			return Ok(())
+		}
+
 		let wrapped_payload = wrap_binary_data(payload);
```
wilwade, line 1105: Hmm... Not sure how heavy the verification step is, but do we want to assume wrapped or unwrapped as more common? (aka which one to try first?)

pallets/msa/src/tests/creation_tests.rs
```diff
@@ -276,3 +276,45 @@ fn it_does_not_allow_duplicate_keys() {
...
+		assert_ok!(Msa::verify_signature(
+			&signature,
+			&key_pair_delegator.public().into(),
+			add_provider_payload.encode()
+		));
+	});
+}
+
+#[test]
+fn verify_signature_without_wrapped_bytes() {
```
wilwade, line 302: Nice pair of tests. Should be easy to duplicate across the other signature verification places.

pallets/msa/src/lib.rs
enddynayn: Good point! I’m thinking the optimization might not matter much here since we base our weights on the worst-case scenario. This flow seems like it would be the worst case, so it might make more sense to focus on handling that rather than optimizing for the less common cases.

wilwade (approved):
Looks good to me.

- Reviewed code
- Pulled and tested locally
- Verified no additional `.verify` calls (outside of VerifiablePasskeySignature)

runtime/common/src/signature.rs
```diff
@@ -0,0 +1,53 @@
...
+
+pub fn check_signature(signature: &MultiSignature, signer: AccountId32, payload: Vec<u8>) -> bool {
+	let verify_signature = |payload: &[u8]| signature.verify(payload, &signer.clone().into());
+
+	if verify_signature(&payload) {
+		return true;
+	}
+
+	let wrapped_payload = wrap_binary_data(payload);
+	verify_signature(&wrapped_payload)
```
aramikm, line 13: nit: I would prefer if we check the wrapped one first as a minor optimization due to most of our signatures are being wrapped.

aramikm (approved):

runtime/common/src/signature.rs
enddynayn: > nit: I would prefer if we check the wrapped one first as a minor optimization due to most of our signatures are being wrapped.

https://github.com/frequency-chain/frequency/pull/2169#discussion_r1784764586

runtime/common/src/signature.rs
```diff
@@ -0,0 +1,53 @@
+use common_primitives::utils::wrap_binary_data;
+use sp_runtime::{traits::Verify, AccountId32, MultiSignature};
+use sp_std::vec::Vec;
+
+pub fn check_signature(signature: &MultiSignature, signer: AccountId32, payload: Vec<u8>) -> bool {
```
JoeCap08055, line 5: I'm curious, why have this return `bool` instead of following the original pattern of `verify_signature` and return `DispatchResult`? Doing it this way requires more context every place it's called (each call to `ensure!(check_signature(...), Error::?)` requires passing in the error to be thrown, but since it's always `InvalidSignature`, why not just throw that from here and call as `verify_signature(...)?` in the pallet methods?

enddynayn: Good question! My reasoning behind returning bool is to give each pallet the flexibility to be more precise with the error handling. For instance, in the MSA pallet, there are multiple signature checks, and it’s useful to provide more specific error messages depending on which check fails. This approach allows us to give the user more detailed feedback, rather than defaulting to a generic InvalidSignature error.

pallets/stateful-storage/src/lib.rs
```diff
@@ -687,10 +686,10 @@ impl<T: Config> Pallet<T> {
 		signer: &T::AccountId,
 		payload: Vec<u8>,
 	) -> DispatchResult {
-		let key = T::ConvertIntoAccountId32::convert((*signer).clone());
-		let wrapped_payload = wrap_binary_data(payload);
+		let key = T::ConvertIntoAccountId32::convert(signer.clone());
+		use common_runtime::signature::check_signature;
```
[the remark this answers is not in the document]
enddynayn: Ah, I included it to avoid overriding the existing function due to a naming collision.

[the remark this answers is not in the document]
JoeCap08055: nit: maybe a pallet-level import, but `use common_runtime::signature::check_signature as <some other name>` would be more consistent?

Or, do we even really need the `use` statement there, or can we just call `common_runtime::signature::check_signature()` directly, if we're only doing it in that one spot?

non-blocking.

JoeCap08055 (approved):
One additional (non-blocking) comment.
Looks great!

Issue #2141, opened 2024-08-29
Review and update Capacity epoch times


enddynayn:

enddynayn:
Part of https://github.com/frequency-chain/frequency/pull/2175

wilwade:
Merged with https://github.com/frequency-chain/frequency/pull/2175

Issue #1920, opened 2024-03-13
Move to 6s Blocktimes


wilwade:
When async backing is supported on Polkadot, Frequency can move to a default of 6s block times.

- [x] Polkadot enabled 6s block times

## Tasks
- [x] Upgrade Polkadot-SDK to v1.8.0+
- [x] Update our blocktime from 12 to 6 via `MILLISECS_PER_BLOCK`
- [x] Review and update Governance times via migration
- [x] #2141
- [x] Review and possibly update message pagination
- [x] Change the default interval sealing time

## Helpful resources
- https://polkadot.network/blog/the-way-to-a-10x-throughput-lift-on-parachains

enddynayn:
Should we also update the mortality window for signature verification?

enddynayn:
Hi @wilwade, @aramikm  and I took a look at the “Review and possibly update message pagination” check item, but we couldn’t find anything that seemed like it needed updating. Could you clarify what changes you’re thinking of? Just want to make sure we get those updates in. Thanks!

enddynayn:
Do we know the intended duration for the following constant?

```
pub type StatefulMortalityWindowSize = ConstU32<14400>;
```
[link to code](https://github.com/frequency-chain/frequency/blob/ba31a73a5e4288099fd32fe2a72f903184a714c2/runtime/common/src/constants.rs#L353)

The comment mentions that the mortality window is meant to be 24 hours, but with a 12-second block time, it actually covers two days. Once we update to 6-second block times, it would match the 1-day duration.

Could you clarify the correct intention here — should the mortality window be 1 day or 2 days?

wilwade:
> Should we also update the mortality window for signature verification?

Yes!

The other ones @aramikm is better to answer than I.

enddynayn:
@wilwade For the “Review and update Governance times via migration” task, I went through all the constants to check if any needed updating but couldn't find any that required changes. I also looked into whether any migrations were necessary and didn't find anything there either. Additionally, I checked the Polkadot releases for references to any related migrations but found nothing. I also checked with @shannonwells  to see if there was anything she was aware of, but nothing came up.

Was there a specific migration you had in mind?

wilwade:
@enddynayn I do not know of any other particular ones.

[...]

PR #2175, opened 2024-10-04
feat(*): upgrade Frequency Mainnet to 6-Second Block Time

enddynayn:
This update reduces block time to 6 seconds. As a result, the Capacity Epoch length has been adjusted to account for the reduced latency. Additionally, the default interval sealing time has been changed to 6 seconds.

A migration has also been added to update the StorageVersion of the CollatorSelection pallet.

Commands to test the runtime migration:

Paseo
```
make try-runtime-create-snapshot-paseo-testnet
make try-runtime-use-snapshot-paseo-testnet
```

Mainnet
```
cargo build --features frequency,try-runtime&& \
try-runtime --runtime ./target/release/wbuild/frequency-runtime/frequency_runtime.wasm on-runtime-upgrade live --uri wss://1.rpc.frequency.xyz:443 --pallet Capacity --pallet System
```

Or

Paseo
```
cargo build --features frequency,try-runtime && \
try-runtime --runtime ./target/release/wbuild/frequency-runtime/frequency_runtime.wasm on-runtime-upgrade live --uri wss://1.rpc.frequency.xyz:443 --pallet Capacity --pallet System
```

```
cargo build --features frequency-testnet,try-runtime && \
try-runtime --runtime ./target/debug/wbuild/frequency-runtime/frequency_runtime.wasm on-runtime-upgrade live --uri wss://rpc.testnet.amplica.io:443 -pallet  Capacity --pallet
```

Screenshot 2024-10-07 at 3 13 25 PM

#1920 
#2141


aramikm (approved):
I've only read the code but didn't execute the try runtimes.
It looked good to me!

runtime/common/src/constants.rs
```diff
@@ -350,7 +338,7 @@ pub type MaxPaginatedPageId = ConstU16<32>;
 /// The maximum number of actions in itemized actions
 pub type MaxItemizedActionsCount = ConstU32<5>;
 /// The number of blocks for Stateful mortality is 24 hours
```
mattheworris, line 340: nit: This comment should be 48 hours to match the value below.

shannonwells (commented):
The CapacityEpoch should still be one day?

runtime/common/src/constants.rs
```diff
@@ -397,7 +385,7 @@ impl sp_std::fmt::Debug for MaxItemizedBlobSizeBytes {
 pub type CapacityMinimumStakingAmount = ConstU128<{ currency::EXISTENTIAL_DEPOSIT }>;
 pub type CapacityMinimumTokenBalance = ConstU128<{ currency::DOLLARS }>;
 pub type CapacityMaxUnlockingChunks = ConstU32<4>;
-pub type CapacityMaxEpochLength = ConstU32<7_200>; // one day, assuming 12 second blocks.
+pub type CapacityMaxEpochLength = ConstU32<{ 2 * DAYS }>; // Two days, assuming 6 second blocks.
```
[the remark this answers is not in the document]
enddynayn: Great question! The move from a 1-day to a 2-day epoch is mainly for future-proofing. Just to clarify, this 2-day setting represents the maximum, not the actual epoch length. The actual epoch is still set to 1 day, as shown in the migration. By setting the max to 2 days, we allow for future increases via governance without needing code changes.

runtime/common/src/fee.rs
```diff
@@ -59,7 +59,7 @@ mod test {
 		let full_block = WeightToFee::weight_to_fee(&MAXIMUM_BLOCK_WEIGHT);
 		// A bounded assertion to consider changes in generated extrinsic base weight.
 		assert!(full_block >= 2 * 150 * CENTS);
-		assert!(full_block <= 10 * DOLLARS);
+		assert!(full_block <= 80 * DOLLARS);
```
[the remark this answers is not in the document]
enddynayn: The throughput has been increased by 8x, which means the cost to purchase a whole block has also gone up accordingly. As a result, the test failed correctly, so I’ve updated it to reflect the new approximate cost for purchasing a block.

runtime/common/src/constants.rs
```diff
@@ -92,7 +85,7 @@ pub type MsaMaxPublicKeysPerMsa = ConstU8<25>;
 /// The maximum size of the provider name (in bytes)
 pub type MsaMaxProviderNameSize = ConstU32<16>;
 /// The number of blocks per virtual bucket
-pub type MSAMortalityWindowSize = ConstU32<100>;
+pub type MSAMortalityWindowSize = ConstU32<{ 20 * MINUTES }>;
```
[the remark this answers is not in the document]
enddynayn: They should be equivalent, but measured in terms of blocks over time. For example:
```
100 * 12 sec/ 60
```
This calculates to 20 minutes, with the measurement based on the number of blocks.

scripts/init.sh
```diff
@@ -136,7 +136,7 @@ start-frequency-instant)
   ;;
 
 start-frequency-interval)
-  defaultInterval=12
+  defaultInterval=6
```
wilwade, line 139: Do we want to also change the default for the `sealing_interval` parameter? (doesn't have to be this PR)

pallets/capacity/src/lib.rs
```diff
@@ -88,7 +88,7 @@ pub mod pallet {
 	}
 
 	/// the storage version for this pallet
-	pub const STORAGE_VERSION: StorageVersion = StorageVersion::new(3);
+	pub const STORAGE_VERSION: StorageVersion = StorageVersion::new(4);
```
wilwade, line 91: ❤️

wilwade (approved):
One optional change, but otherwise looks great!

- Reviewed code
- Pulled and searched for any remaining usages of `12` or `7200`

runtime/common/src/fee.rs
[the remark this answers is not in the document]
aramikm: 8x does not make sense. It should be 4x. The reason is that we increase the block processing capacity from 0.5 seconds to 2 seconds. This means each block can process 4 times more. So this should really be 4 times. The throughput is 8x becasue we generate 2 blocks in the same timestamp but here we are testing just 1 block as far as I can tell.

scripts/init.sh
enddynayn: Good idea! I update the default to 6 seconds.
```
#[clap(long, help = "The interval in seconds", default_value = "6", value_name = "SECONDS")]
```

runtime/common/src/constants.rs
enddynayn: Good point!

mattheworris (approved):
- Read through the changes
- Non-blocking comments
🚢 it!

saraswatpuneet (approved):
Read through the changes and comments
Looks all good. Would try this new image soon 🚢

PR #2184, opened 2024-10-11
revert: stalehash extension

### LINKED ISSUE #2183: [Bug]: Stale Hash potential issues
# Summary
It appears that the stale hash extension might be causing some issues with graph updates

# Context

One of our providers has an issue with the graph updates. The exact reason is not currently known but it is highly likely the result of the new stale hash check extension added to ensure non valid graph transactions are going to get ignored early in the pool. 

# Acceptence Criterea

- [ ] Revert the changes related to stale hash check

### Relevant log output

```shell
[Nest] 35  - 10/11/2024, 7:31:18 PM   DEBUG [NonceService] nextNonce 2990897
2024-10-11 19:31:19        RPC-CORE: submitAndWatchExtrinsic(extrinsic: Extrinsic): ExtrinsicStatus:: 1014: Priority is too low: (757 vs 757): The transaction has too low priority to replace another transaction already in the pool.
[Nest] 35  - 10/11/2024, 7:31:19 PM   ERROR [ReconnectionGraphService] Error processing batch: RpcError: 1014: Priority is too low: (757 vs 757): The transaction has too low priority to replace another transaction already in the pool.
2024-10-11 19:31:19        RPC-CORE: submitAndWatchExtrinsic(extrinsic: Extrinsic): ExtrinsicStatus:: 1014: Priority is too low: (757 vs 757): The transaction has too low priority to replace another transaction already in the pool.
[Nest] 35  - 10/11/2024, 7:31:19 PM   ERROR [ReconnectionGraphService] Error updating graph for user 147584, provider 2: RpcError: 1014: Priority is too low: (757 vs 757): The transaction has too low priority to replace another transaction already in the pool.
    at checkError (/app/node_modules/@polkadot/rpc-provider/cjs/coder/index.js:23:15)
```

aramikm:
# Goal
The goal of this PR is 

Closes #2183 

Reverting #2137 

# Checklist
- [x] e2e Tests updated
- [X] Benchmarks updated


runtime/frequency/src/lib.rs
```diff
@@ -1268,279 +1254,6 @@ construct_runtime!(
...
-			DispatchError::Module(module_err) =>
-				<u32 as Decode>::decode(&mut module_err.error.as_slice())
-					.unwrap_or_default()
-					.try_into()
-					.unwrap_or_default(),
-			_ => 255u8,
-		})
-	}
-}
-
```
saraswatpuneet, line 1543: nit: assuming there is still stalehash check in extrinsics?

saraswatpuneet (approved):
one comment, lgtm

runtime/frequency/src/lib.rs
aramikm: That is correct, we didn't remove that check.

wilwade (approved):
Not sure it's the issue, but we'll revert just in case

mattheworris (approved):

e2e/stateful-pallet-storage/handlePaginated.test.ts
```diff
@@ -232,22 +232,6 @@ describe('📗 Stateful Pallet Storage', function () {
         section: 'statefulStorage',
       });
     });
-
-    it('🛑 should fail call to remove page for un-delegated attempts', async function () {
```
aramikm, line 236: These tests are not correct and I had to remove since we are using a `badMsaId` which it's nonce is lower than current nonce for providerKeys it is throwing following error. The only reason this was not happening before was because it was erroring out on the stalehash check before getting to this error.

`1010: Invalid Transaction: Transaction is outdated` which might be a nonce check race condition issue which is not related to this specific pallet or extrinsic

aramikm: On a second thought it is weird why this gets triggered

The following is the only place we are returning `InvalidTransaction::Stale`.
```
		// The default account (no account) has a nonce of 0.
		// If account nonce is not equal to the tx nonce (self.0), the tx is invalid.  Therefore, check if it is a stale or future tx.
		if self.0 != account.nonce {
			return Err(if self.0 < account.nonce {
				InvalidTransaction::Stale
			} else {
				InvalidTransaction::Future
			}
			.into())
		}
```

aramikm: I don't think it's a blocker for this PR merge but we should take a deeper look

PR #2186, opened 2024-10-14
Disable `report_invalid` in the vendor'd transaction pool

wilwade:
# Goal
The goal of this PR is to separate into a separate PR the only change happening to the vender'd transaction pool

### DIFF
--- a/vendor/transaction-pool/src/graph/ready.rs
+++ b/vendor/transaction-pool/src/graph/ready.rs
@@ -508,7 +508,7 @@ impl<Hash: hash::Hash + Member, Ex> BestIterator<Hash, Ex> {
 impl<Hash: hash::Hash + Member, Ex> sc_transaction_pool_api::ReadyTransactions
 	for BestIterator<Hash, Ex>
 {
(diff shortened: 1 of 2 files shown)


vendor/transaction-pool/src/graph/ready.rs
```diff
@@ -508,8 +508,12 @@ impl<Hash: hash::Hash + Member, Ex> BestIterator<Hash, Ex> {
 impl<Hash: hash::Hash + Member, Ex> sc_transaction_pool_api::ReadyTransactions
 	for BestIterator<Hash, Ex>
 {
-	fn report_invalid(&mut self, tx: &Self::Item) {
-		BestIterator::report_invalid(self, tx)
+	fn report_invalid(&mut self, _tx: &Self::Item) {
```
aramikm, line 511: I'm worried about side effects. Maybe we can do some tests using https://github.com/paritytech/polkadot-stps?tab=readme-ov-file?

wilwade: Our e2e tests on testnet get close to that (and they were showing the issue) I think we could setup something like that as well for it.

Issue #2154, opened 2024-09-10
Testnet e2e Tests Failing


wilwade:
With the latest release v1.13.0 live on testnet, the e2e tests are failing: https://github.com/frequency-chain/frequency/actions/runs/10797691465

## Acceptance Criteria

- e2e tests pass

PR #2191, opened 2024-10-21
API Augment update to use metadata v15

wilwade:
# Goal
The goal of this PR is to update API Augment to use the newer metadata v15 provided with Polkadotjs/api v 14.1.1

Part of #2154 

# Discussion

- Updated the Export Metadata Cmd to use Metadata v15
- Update the fetch:local to use Metadata v15
- Fixed a test that doesn't work under all situations (description is not always returned on all networks)
- Setup e2e testnet workflow to run on the provided branch instead of against the release branch so that other test setups can test against testnet

## Testing

- `make js`
- `/target/debug/frequency export-metadata --tmp ./js/api-augment/metadata.json && cd js/api-augment && npm run build`


node/cli/src/export_metadata_cmd.rs
```diff
@@ -52,7 +52,7 @@ impl ExportMetadataCmd {
 		let maybe_hash = client.hash(block_number.into())?;
 		let block_hash = maybe_hash.ok_or_else(|| Error::from("Block not found"))?;
 
-		let metadata: Bytes = api.metadata(block_hash).unwrap().into();
+		let metadata: Bytes = api.metadata_at_version(block_hash, 15u32).unwrap().unwrap().into();
```
wilwade, line 55: metadata_at_version returns an option, thus the second unwrap after unwrapping the result

enddynayn: I’m not sure if I’m overlooking something, but it seems like metadata_at_version only accepts one argument. Could you please confirm?
https://github.com/frequency-chain/frequency/blob/9fd47cf90b23b3030749790aceffbc726bd3dd40/runtime/frequency/src/apis.rs#L88

enddynayn (approved):
👍 great!

node/cli/src/export_metadata_cmd.rs
enddynayn: Ahh its wrapped around `impl_runtime_apis`.

saraswatpuneet (approved):

PR #2218, opened 2024-11-13
ethereum: added e2e tests

### LINKED ISSUE #2203: Verify Support for Address20
Frequency supports `secp256k1` signatures in some places, but doesn't support `Address20`

## AC: `secp256k1` signatures and `Address20` should be supported everywhere a `sr25519` is for extrinsics

- [x] Passkey roots
- [ ] MSA Control Keys
- [ ] Balance Transfers
- [ ] Signatures for Stateful Storage
- [ ] Signatures for Delegation actions
- [ ] Others?

### Testing

Should be able to configure all e2e tests to use `Address20` and still pass

aramikm:
# Goal
The goal of this PR is 

Closes #2203 

# Discussion
- added e2e tests for more usecase for ethereum keys
- Some clean up and refactoring
- Added eslint rules to enforce correct usage

# Checklist
- [X] e2e Tests added?


e2e/eslint.config.mjs
```diff
@@ -51,6 +51,17 @@ export default tseslint.config(
         },
       ],
       'allow-namespace': 'off',
+	  'no-restricted-syntax': [
```
aramikm, line 54: adding new rules so going forward we have to use correct functions

e2e/scaffolding/ethereum.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable no-restricted-syntax */
```
aramikm, line 1: Only ignoring the custom rule in here

e2e/eslint.config.mjs
```diff
@@ -51,6 +51,19 @@ export default tseslint.config(
         },
       ],
       'allow-namespace': 'off',
+      'no-restricted-syntax': [
```
aramikm, line 54: adding new rules so going forward we have to use correct functions

saraswatpuneet (approved):
read through, lgtm :100:

e2e/scenarios/grantDelegation.ethereum.test.ts
```diff
@@ -0,0 +1,183 @@
...
+
+    schemaId2 = await ExtrinsicHelper.getOrCreateSchemaV3(
+      keys,
+      schema,
+      'AvroBinary',
+      'OnChain',
+      [],
+      'test.grantDelegationSecond'
+    );
+  });
```
JoeCap08055, line 93: nit: probably don't need to create schemas any more, unless actually testing the schemas pallet, as mainnet schemas should be available on all chains now

JoeCap08055 (commented):
One general comment: it seems you've cloned some of the tests solely to validate that they work as well with Ethereum-style keys. Might it be simpler to just convert the existing tests to table tests that run the same test using different keys?

aramikm:
@JoeCap08055 There are some minor other differences between ethereum ones and other ones besides the keys used. One of the benefits of having them in separate files is that I can just change the test selector to `*.ethereum.test.ts` and it would only run ethereum related tests but open to other suggestions

JoeCap08055:
> @JoeCap08055 There are some minor other differences between ethereum ones and other ones besides the keys used. One of the benefits of having them in separate files is that I can just change the test selector to `*.ethereum.test.ts` and it would only run ethereum related tests but open to other suggestions

If there's a good reason, then I'm fine with it.

e2e/scaffolding/helpers.ts
```diff
@@ -79,6 +81,19 @@ export function signPayloadSr25519(keys: KeyringPair, data: Codec): Sr25519Signa
   return { Sr25519: u8aToHex(keys.sign(u8aWrapBytes(data.toU8a()))) };
 }
 
+export function signPayload(keys: KeyringPair, data: Codec): MultiSignatureType {
+  switch (keys.type) {
+    case 'ecdsa':
+      throw new Error('Ecdsa key type is not supported!');
```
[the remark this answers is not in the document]
aramikm: I agree that it is confusing. I will modify the error message but the issue is that in the PolkadotJs they have 4 types `ecdsa`, `sr25519`, `ed25519 `, `ethereum` but on the chain there are only 3 types supported `ecdsa `, `sr25519`, `ed25519 `. 

The difference between `ecdsa` and `ethereum` types are the hashing function they use on payload before signing. `ecdsa` uses `blake-256` but `ethereum` uses `keccak-256`. So from chain perspective if we are getting an `ecdsa` signature we can only support one of them which is `keccak-256` going forward. due to ethereum compatibility

e2e/miscellaneous/balance.ethereum.test.ts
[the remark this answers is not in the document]
aramikm: Added checks to ensure the amount is added to the receiving account

e2e/scenarios/grantDelegation.ethereum.test.ts
aramikm: yeah this is getting the schemaId if it already exists but I agree that we can refactor all these schema creation everywhere and simplify it

mattheworris (approved):
- Read through changes

🚢 it!

PR #2225, opened 2024-11-19
init: static schemas package

### LINKED ISSUE #2176: [Feature]: Static Schemas package
## Details

Always fetching the schema ids from the chain works, but usually it is not needed as the values are static.

Let's publish a package (suggested at `@frequency-chain/schemas`) that has the following:

- A way to get the schema id from the schema namespace, name, and version. (see below for example)
- A way to get the schema namespace, name, and version from the id
- Have zero dependencies
- Placed in the `js/schemas` folder
- Releases should be tagged only when the contents change, but this should still work to release at the same time as the node image with the mainnet schema included.
- Perhaps a generator to generate from the mainet schema data

```typescript
// This has already applied the transformation from the v1.13.0 migration
export const SCHEMA_NAME_TO_ID = new Map<string, number>([
  ['dsnp.broadcast@v1', 2],
  ['dsnp.broadcast@v2', 17],
  ['dsnp.dsnp-content-attribute-set@v1', 12],
  ['dsnp.ext-content-attribute-set@v1', 13],
  ['dsnp.private-connections@v1', 10],
  ['dsnp.private-follows@v1', 9],
  ['dsnp.profile-resources@v1', 15],
  ['dsnp.profile@v1', 6],
  ['dsnp.public-follows@v1', 8],
  ['dsnp.public-key-assertion-method@v1', 14],
  ['dsnp.public-key-key-agreement@v1', 7],
  ['dsnp.reaction@v1', 4],
  ['dsnp.reply@v1', 3],
  ['dsnp.reply@v2', 18],
  ['dsnp.tombstone@v1', 1],
  ['dsnp.tombstone@v2', 16],
  ['dsnp.update@v1', 5],
  ['dsnp.update@v2', 19],
  ['dsnp.user-attribute-set@v1', 11],
  ['dsnp.user-attribute-set@v2', 20],
  ['frequency.default-token-address@v1', 21],
]);

```

aramikm:
# Goal
The goal of this PR is create a new js package for schema name and ID mappings

Closes #2176 

# Discussion
- This only generates data for `dsnp` schemas from chain. The issue is that currently there is no easy way to get all schema names without providing namespaces. We can get all using raw state but then we would need to scale decode it.


js/schemas/scripts/package-lock.json
wilwade, line 1: Extra file?

js/schemas/index.ts
```diff
@@ -0,0 +1,2 @@
+import './data';
+import './schemas';
```
wilwade, line 2: A. If you just want to export without a default aka to use `import { SCHEMA_INFOS } from '@frequency-chain/schemas'`:
```suggestion
export * from './data';
export * from './schemas';
```

B. If you want to export nested aka to use `import { data } from '@frequency-chain/schemas'`:

```suggestion
export * as data from './data';
export * as schemas from './schemas';
```

C. If you want to export with a default with flat aka to use `import schemas from '@frequency-chain/schemas'`:

```suggestion
import * as data from './data';
import * as schemas from './schemas';

export default { ...data, ...schemas };
```


D. (Suggested?) Export with a default with flat and each function (aka to use A and C)


CJS Tested using:

`npm run build && cd dist && npm pack && cd ..`

And then in a new `npm init` with an `npm i ../../<etc...>`

index.js
```javascript
const schemas = require("@frequency-chain/schemas");

console.log(schemas);
```

wilwade (changes requested):
Some fixes for the exports and a few other notes, but looking good.

enddynayn (approved):
👍 LGTM

js/schemas/index.ts
aramikm: Thanks for great suggestions. So combining A + C together. One thing I noticed after doing suggested A+C. I have to use `schemas.default.FULL_NAME_TO_ID` to be able to access it. It's wrapped inside the default and I can not access it directly like `schemas.FULL_NAME_TO_ID`

wilwade: Depends on how you do the import at that point. Currently these all work for me (which is ✅ ):

```typescript
import { SCHEMA_INFOS } from "@frequency-chain/schemas";

console.log(SCHEMA_INFOS);
```


```typescript
import * as schemas from "@frequency-chain/schemas";

console.log(schemas.SCHEMA_INFOS);
```


```typescript
import schemas from "@frequency-chain/schemas";

console.log(schemas.SCHEMA_INFOS);
```

```javascript
const { SCHEMA_INFOS } = require("@frequency-chain/schemas");
console.log(SCHEMA_INFOS);
```

```javascript
const schemas = require("@frequency-chain/schemas");
console.log(schemas.SCHEMA_INFOS);
```

wilwade (approved):
🥅 

- Tested build locally
- Imported in typescript ✅ 
- Imported in cjs ✅

js/schemas/README.md
```diff
@@ -0,0 +1,89 @@
...
+    payloadLocation: 'Itemized',
+    appendOnly: true,
+    signatureRequired: true,
+  },
+```
+
+## Upgrades and Matching Versions
+
+Assuming you are using no deprecated methods, any release version should work against a release version of `@frequency-chain/schemas`.
+If you are working against a development version it is suggested that you match against the commit hash using `v0.0.0-[First 6 of the commit hash]`.
```
JoeCap08055, line 56: This is a bit confusing. Perhaps:
```suggestion
Assuming you are using no deprecated methods, any release version of `@frequency-chain/schemas` should work against a release version of a Frequency node.
If you are working against a development node it is suggested that you match the package version against the commit hash using `v0.0.0-[First 6 of the commit hash]`.
```

js/schemas/scripts/fetchSchemaData.mjs
```diff
@@ -0,0 +1,135 @@
+// Only way to silence PolkadotJS API warnings we don't want
+console.warn = () => {};
```
JoeCap08055, line 2: 💯

JoeCap08055 (commented):
Looks great--however I think we're missing a couple of key functions:

1. Should have a method that can retrieve all versions of a given schema name
2. Should be able to query `<schema.fullname>@latest` to get the latest version

Granted, these are easy enough for anyone to do with access to the underlying array, but I think this package should provide those lookups as a convenience.

JoeCap08055 (changes requested):
Please respond to:
https://github.com/frequency-chain/frequency/pull/2225#pullrequestreview-2449468169

mattheworris (approved):
- Read through changes
🚢 it!

PR #2233, opened 2024-12-05
Switch the sourcing of the funds from strings to test file names

wilwade:
Note: This will merge into the feature branch `e2e-improvements-development`

# Goal

Simple swap from string to file name and path based account funding for tests

## Checking

- Tests all still run!
- Stopping the tests means that some accounts don't have to be topped off in the re-run


saraswatpuneet (approved):
Neat 🚢

Issue #2227, opened 2024-11-20
[Bug]: Fix the key issues


aramikm:
### What happened?

Currently some of the keys imported into the chain don't have the correct Avro prefix. We need to fix these keys.

The fix is prefixing all stored itemized storage items on schema_id `7` with `0x40`

# Affected Msa Accounts

## testnet 
```
8004, 8009, 8816, 8817, 8818, 8819, 8820, 8822, 8823, 8824, 8825, 8826, 9384, 9753, 9919,
9992, 9994, 9996, 9997, 10009, 10010, 10012, 10013, 10014, 10015, 10019, 10020, 10021,
10022, 10023, 10024, 10025, 10026, 10027, 10028, 10029, 10030, 10031, 10032, 10033, 10034,
10035, 10036, 10037, 10038, 10039, 10040, 10041, 10042, 10043, 10044, 10045, 10046, 10047,
10048, 10049, 10050, 10051, 10052, 10053, 10054, 10055, 10056, 10057, 10058, 10059, 10061,
10062, 10064, 10067, 10068, 10069, 10070, 10071, 10072, 10075, 10076, 10077, 10078, 10079,
10138, 10139, 10140, 10206, 10207, 10209, 10212, 10218, 10219, 10220, 10221, 10222, 10223,
10224, 10231, 10232, 10233, 10234, 10235, 10236, 10237, 10238, 10239, 10240, 10241, 10242,
10243, 10247, 10248, 10251, 10253, 10254, 10255, 10256, 10257, 10258, 10259, 10260, 10261,
10262, 10263, 10264, 10265, 10266, 10267, 10268, 10269, 10270, 10271, 10272, 10273, 10274,
10275, 10287, 10288, 10289, 10290, 10291, 10292, 10293, 10294, 10295, 10296, 10297, 10298,
10299, 10300, 10301, 10302, 10303, 10304, 10305, 10306, 10307, 10308, 10309, 10311, 10312,
10313, 10314, 10315, 10316, 10317, 10318, 10319, 10320, 10321, 10322, 10323, 10324, 10325,
10326, 10327, 10328, 10329,
```
- [x] Done
## mainnet

```
227, 542, 1249820, 1287729, 1288925, 1309067, 1309241, 1309258, 1309367, 1309397,
1329112, 1329535, 1330067,
```
- [x] Done

wilwade:
@aramikm Are the mainnet ones going to be fixed via migration or "user" (Frequency Access in this case) action?

[...]

PR #2236, opened 2024-12-09
cleanup: storage migration

aramikm:
# Goal
The goal of this PR is to remove extra migrations

Related to #2227 

# Discussion
- keeping the `MigrationPageIndex` storage around for future use. Removing it requires another migration which is not justified currently.

# Checklist


wilwade (approved):
Looks good to me. Any other migrations we need to remove while we are at it?

saraswatpuneet (approved):

mattheworris (approved):

shannonwells (approved):
1300+ lines removed, I like that.

PR #2239, opened 2024-12-11
Update handle allowed set and add new handles RPC 

### LINKED ISSUE #2237: Remove "General Punctuation" from handle support
Currently the Unicode General Punctuation Range: 2000–206F is mostly allowed (`0x200C..=0x206F`), but several of the characters are not condensing correctly nor does it appear any of them are needed for multilingual handles.

https://www.unicode.org/charts/PDF/U2000.pdf

Action: Remove the range from the allowed list and double check that nothing needs it.

Review result: We need to keep the zero-width joiner and zero-width non-joiner

wilwade:
# Goal
The goal of this PR is to remove some undesirable characters from the handle allowed list and create new RPCs to make testing handles (and the edge cases) easier

Closes #2237

# Discussion

- Removing some punctuation only characters from the allowed list
- Added a new `check_handle` RPC so that it is easier to check the canonical value along with validity and if there are suffixes remaining

# Checklist
- [x] Unit Tests added?
- [x] e2e Tests added?


common/primitives/src/handles.rs
```diff
@@ -94,3 +94,21 @@ pub struct PresumptiveSuffixesResponse {
 	/// The suffixes
 	pub suffixes: Vec<HandleSuffix>,
 }
+
+/// Output response for retrieving the next suffixes for a given handle
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+#[derive(Clone, Encode, Decode, PartialEq, Debug, TypeInfo, Eq)]
+pub struct CheckHandleResponse {
```
wilwade, line 101: I cannot think of anything else we might want to be returned, but if you can, let me know and I'll add it

e2e/handles/handles.test.ts
```diff
@@ -166,4 +170,59 @@ describe('🤝 Handles', function () {
...
+        canonical_base: 'नहनन।',
+        suffix_index: '0',
+        suffixes_available: true,
+        valid: true,
+      });
+
+      // Ensure these are the same canonically
+      const withoutWhitespace = await ExtrinsicHelper.apiPromise.rpc.handles.checkHandle('नी हुन्न् ।');
+      assert(!withoutWhitespace.isEmpty, 'Expected a response');
+      assert.equal(withoutWhitespace.toHuman().canonical_base, res.toHuman().canonical_base);
```
wilwade, line 225: It was hard to manually test this issue, so here's an example of why the RPC exists.

pallets/capacity/src/runtime-api/src/lib.rs
```diff
@@ -27,7 +27,6 @@ use sp_std::vec::Vec;
 sp_api::decl_runtime_apis! {
 	/// Runtime Version for Capacity
 	/// - MUST be incremented if anything changes
-	/// - Also update in js/api-augment
```
wilwade, line 30: This is no longer true, so removing it everywhere

pallets/handles/src/handles-utils/constants.rs
```diff
@@ -3,6 +3,7 @@
 use core::ops::RangeInclusive;
 
 #[cfg(test)]
+#[allow(dead_code)]
```
wilwade, line 6: No more warning about the dead code

pallets/handles/src/handles-utils/constants.rs
```diff
@@ -70,7 +71,7 @@ pub const ALLOWED_UNICODE_CHARACTER_RANGES: [RangeInclusive<u16>; 54] = [
 0x1B80..=0x1BB9,
 0x1BC0..=0x1C7F,
 0x1E00..=0x1FFF,
-0x200C..=0x206F,
+0x200C..=0x200D,
```
wilwade, line 74: HERE'S THE CHANGE for which characters are allowed

pallets/handles/src/handles-utils/src/converter.rs
```diff
@@ -105,7 +105,11 @@ pub fn split_display_name(display_name_str: &str) -> Option<(String, HandleSuffi
 pub fn strip_unicode_whitespace(input_str: &str) -> String {
 	input_str
 		.chars()
-		.filter(|character| !character.is_whitespace())
+		// U+200C is a zero-width Non-joiner needed for some writing systems
+		// U+200D is a zero-width joiner needed for some writing systems
+		.filter(|character| {
+			!character.is_whitespace() && character.ne(&'\u{200C}') && character.ne(&'\u{200D}')
```
wilwade, line 111: Not super happy with this fix, but no better ones that I currently am aware of

e2e/handles/handles.test.ts
```diff
@@ -119,6 +119,10 @@ describe('🤝 Handles', function () {
         assert(msaOption.isSome, 'msaOption should be Some');
         const msaFromHandle = msaOption.unwrap();
         assert.equal(msaFromHandle.toString(), msa_id.toString(), 'msaFromHandle should be equal to msa_id');
+
+        // Check that the rpc returns the index as > 0
+        const apiCheck = await ExtrinsicHelper.apiPromise.rpc.handles.checkHandle(handle);
+        assert(apiCheck.suffix_index > 0);
```
JoeCap08055, line 125: Since the returned value is a `u16`, does the `>` operator work correctly, or should you `.toNumber()` the result first?

wilwade: I tested and it was valid. JS does the coercion.

common/primitives/src/handles.rs
JoeCap08055: Any value in returning the actual suffix (or next N suffixes) instead of/in addition to the suffix index, to eliminate potentially having to make a separate call to the `get_next_available_suffixes` RPC?

pallets/handles/src/rpc/src/lib.rs
```diff
@@ -133,4 +137,15 @@ where
...
+	fn check_handle(&self, base_handle: String) -> RpcResult<CheckHandleResponse> {
+		let api = self.client.runtime_api();
+		let at = self.client.info().best_hash;
+		let base_handle: BaseHandle = base_handle
+			.into_bytes()
+			.try_into()
+			.map_err(|_| HandlesRpcError::InvalidHandle)?;
+		let result = api.check_handle(at, base_handle);
+		map_rpc_result(result)
+	}
```
JoeCap08055, line 150: I'm not clear on the need to have both an RPC and a custom runtime call for this--also, per the latest guidance from Parity, aren't we supposed to *not* be creating custom RPCs any more?

wilwade: I did both as we have a lot of these and I wanted to continue the pattern for now.

I'm not sure what we want to do, but I'll add it to a conversation point for the Community Call: https://github.com/frequency-chain/frequency/issues/2240

e2e/handles/handles.test.ts
JoeCap08055: meh. Stylistically, I prefer explicit unwrapping/conversion of SCALE to native JS types, as it makes it clear when you read the code that's what's happening. Otherwise a casual reader might assume that `suffix_index` is a regular JS `number` when it actually isn't.

But, this is my non-blocking preference, as this is a test & not deployed code.

enddynayn (approved):
👍 superb!

shannonwells (changes requested):
I think the custom RPC should be removed given the general transition away from them and the problems with them.  See my comment in the code.

e2e/handles/handles.test.ts
JoeCap08055: nvm... apparently, RPC calls return native JS; it's only runtime calls that return SCALE.

pallets/handles/src/rpc/src/lib.rs
JoeCap08055: Given that a larger review/revamp of all RPCs seems to be around the corner anyway, I don't feel strongly either way, so non-blocking for me.

JoeCap08055 (approved):
My questions addressed to my satisfaction. Approved.

pallets/handles/src/lib.rs
```diff
@@ -567,6 +567,56 @@ pub mod pallet {
...
+		/// * `handle` - The handle to check.
+		///
+		/// # Returns
+		///
+		/// * `CheckHandleResponse`
+		///
+		pub fn check_handle(base_handle: BaseHandle) -> CheckHandleResponse {
+			let valid = Self::validate_handle(base_handle.to_vec());
+
+			if !valid {
```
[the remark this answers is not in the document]
wilwade: Added a Default impl

shannonwells (approved):
Thanks, looks good to go.

PR #2242, opened 2024-12-13
Feature Passkey Proxy v2, Simple Param Shift

### LINKED ISSUE #2241: Feature: Passkey Proxy shift `accountOwnershipProof` parameter
Users can have multiple passkeys for the same domain, and a website cannot know which passkey signed the payload until after the payload has been signed.

This causes an issue as currently the `PasskeyPayload` requires that the `accountOwnershipProof` be part of the `PasskeyCall`.

As we can move the `accountOwnershipProof` out of the `PasskeyCall` without a security issue, moving it one level up will remove the edge case where a user has multiple passkeys and allow clients to correctly attach the correct `accountOwnershipProof` to the `PasskeyPayload`.

Example V2 structs:

```rust
pub struct PasskeyPayloadV2<T: Config> {
	/// passkey public key
	pub passkey_public_key: PasskeyPublicKey,
	/// a self-contained verifiable passkey signature with all required metadata
	pub verifiable_passkey_signature: VerifiablePasskeySignature,
	/// MOVED passkey_public_key signed by account_id's private key
	pub account_ownership_proof: MultiSignature,
	/// PassKey Call
	pub passkey_call: PasskeyCall<T>,
}

pub struct PasskeyCallV2<T: Config> {
	/// account id which is the origin of this call
	pub account_id: T::AccountId,
	/// account nonce
	pub account_nonce: T::Nonce,
	/// Extrinsic call
	pub call: Box<<T as Config>::RuntimeCall>,
}
```

Expected result is having a v2 call: `passkey.proxy_v2` that uses the new data structure. The v1 of the call can be left for now for future deprecation.

wilwade:
# Goal
The goal of this PR is to shift the `accountOwnershipProof` out of the signature payload. For reasons why, see #2241  

Closes #2241 

With @aramikm 

# Discussion

- Added `proxy_v2`
- Deprecated `proxy`
- Duplicated tests, as eventually we'll remove v1

# Checklist
- [x] Updated Pallet Readme?
- [x] Unit Tests added?
- [x] e2e Tests added?
- [x] Benchmarks added?


aramikm (approved):
Looks great!

JoeCap08055 (approved):

saraswatpuneet (approved):
Would we need to update the iframe ? Or it's not being used

PR #2244, opened 2024-12-16
update image in readme to replace rococo with paseo

demisx:
# Goal
The goal of this PR is to update reference from Rococo to Paseo in the readme image.


wilwade (approved):
👍

PR #2246, opened 2024-12-30
fix: make e2e tests for passkey more rigorous in their success checks

JoeCap08055:
# Goal
The goal of this PR is to make the e2e tests for the passkey pallet a bit more discerning, and more clear to read.
Previously, the tests were determining a successful token transfer by the absence of an exception and an increased nonce. While this is technically correct, according to the logic by which the extrinsic call would complete with no exceptions, it is a bit opaque.

This PR enhances the test conditions to check that:
* The appropriate events are generated in the block, and the event contents are correct
* The correct reciever's balance is incremented by the transfer amount
* The correct sender's balance is decremented by the transfer amount + fee
* Failure tests are enhance to check the specific error that caused the failure

Test conditions for the success cases are also enhanced so that successful tests are more verbose, as each check is encapsulated in a separate test block.

Also, the function `Extrinsic.sendUnsigned()` was fixed to properly parse and return the event stream from the blockchain, in the same manner that `Extrinsic.signAndSend()` does.

# Checklist
- [ ] Updated Pallet Readme?
- [ ] Design doc(s) updated?
- [ ] Unit Tests added?
- [ ] e2e Tests added?
- [ ] Benchmarks added?

### DIFF
--- a/e2e/passkey/passkeyProxyV2.ethereum.test.ts
+++ b/e2e/passkey/passkeyProxyV2.ethereum.test.ts
@@ -2,11 +2,13 @@ import '@frequency-chain/api-augment';
 import assert from 'assert';
 import { createAndFundKeypair, EcdsaSignature, getNonce, Sr25519Signature } from '../scaffolding/helpers';
 import { KeyringPair } from '@polkadot/keyring/types';
-import { ExtrinsicHelper } from '../scaffolding/extrinsicHelpers';
+import { Extrinsic, ExtrinsicHelper } from '../scaffolding/extrinsicHelpers';
 import { getFundingSource } from '../scaffolding/funding';
 import { getUnifiedPublicKey, getUnifiedAddress } from '../scaffolding/ethereum';
 import { createPassKeyAndSignAccount, createPassKeyCallV2, createPasskeyPayloadV2 } from '../scaffolding/P256';
 import { u8aToHex, u8aWrapBytes } from '@polkadot/util';
+import { AccountId32 } from '@polkadot/types/interfaces';
+import { ISubmittableResult } from '@polkadot/types/types';
 const fundingSource = getFundingSource(import.meta.url);
 
 describe('Passkey Pallet Proxy V2 Ethereum Tests', function () {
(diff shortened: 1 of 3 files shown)


e2e/passkey/passkeyProxyV2.ethereum.test.ts
```diff
@@ -46,21 +42,56 @@ describe('Passkey Pallet Proxy V2 Ethereum Tests', function () {
         false
       );
       const passkeyProxy = ExtrinsicHelper.executePassKeyProxyV2(fundedSr25519Keys, passkeyPayload);
-      await assert.doesNotReject(passkeyProxy.fundAndSendUnsigned(fundingSource));
-      await ExtrinsicHelper.waitForFinalization((await getBlockNumber()) + 2);
-      // adding some delay before fetching the nonce to ensure it is updated
-      await new Promise((resolve) => setTimeout(resolve, 1000));
```
JoeCap08055, line 52: We don't need to await finalization now that `sendUnsigned` has been fixed to not resolve until the extrinsic is included in a finalized block.

e2e/scaffolding/extrinsicHelpers.ts
```diff
@@ -264,6 +264,10 @@ export class Extrinsic<N = unknown, T extends ISubmittableResult = ISubmittableR
   public async fundAndSendUnsigned(source: KeyringPair) {
     await this.fundOperation(source);
     log('Fund and Send', `Fund Source: ${getUnifiedAddress(source)}`);
+    return this.sendUnsigned();
```
JoeCap08055, line 267: Deleted code block; now that `sendUnsigned` is doing what this block used to do.

e2e/passkey/passkeyProxyV2.ethereum.test.ts
```diff
@@ -28,11 +22,13 @@ describe('Passkey Pallet Proxy V2 Ethereum Tests', function () {
     });
 
     it('should transfer via passkeys with root sr25519 key into an ethereum style account', async function () {
+      const startingBalance = (await ExtrinsicHelper.getAccountInfo(receiverKeys)).data.free.toBigInt();
       const accountPKey = getUnifiedPublicKey(fundedSr25519Keys);
       const nonce = await getNonce(fundedSr25519Keys);
+      const transferAmount = 55_000_000n;
```
enddynayn, line 28: Do we have constants for these like DOLLARS and CENTS?

enddynayn (approved):
👍 great!

saraswatpuneet (approved):
Lgtm

pfrank13 (approved):

e2e/passkey/passkeyProxyV2.ethereum.test.ts
JoeCap08055: Not sure, but the amount isn't important anyway; we just want to be sure to check that whatever amount is transferred, the balance changes by that amount.

JoeCap08055: But that does remind me, it would be good to also check that the correct sender's balance decreases by the appropriate amount.

PR #2250, opened 2025-01-06
Fix Release Metadata Comparison

wilwade:
# Goal
The goal of this PR is to fix metadata comparison in the release process.

# Discussion

- Had to tweak some of the values to get it all to work. Working here:
  - https://github.com/frequency-chain/frequency/actions/runs/12639367825
  - https://github.com/frequency-chain/frequency/releases/tag/v0.0.1-rc6

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -533,7 +533,7 @@ jobs:
         id: get-previous-full-release-version
         uses: ./.github/workflows/common/get-previous-full-release-version
         with:
-          full-release-version-tag: v1.15.0
+          full-release-version-tag: ${{env.LATEST_FULL_RELEASE_TAG}}
       - name: Fetch Reference Binary
         uses: robinraju/release-downloader@a96f54c1b5f5e09e47d9504526e96febd949d4c2
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -544,7 +544,7 @@ jobs:
         id: download-binaries
         uses: actions/download-artifact@v4
         with:
-          pattern: artifacts-binary-${{env.NETWORK}}-${{github.run_id}}
+          pattern: artifacts-binary-local-${{github.run_id}}
```
wilwade, line 547: This was the main fix, but also swapped from dev to local so that we could also tag any parachain related pallet changes.

enddynayn (approved):
👍 great!

.github/workflows/release.yml
```diff
@@ -570,7 +572,6 @@ jobs:
             --name test-node \
             ubuntu:24.04 \
               /bin/sh -c "chmod +x /app/${{env.TEST_BIN_FILENAME}} && /app/${{env.TEST_BIN_FILENAME}} \
-              --chain=dev \
```
JoeCap08055, line 573: Assuming this node we're launching doesn't need the funded dev accounts? (same below as well?)

wilwade: Correct. It can just start with basic genesis

JoeCap08055 (approved):

shannonwells (approved):
let 'er rip

PR #2272, opened 2025-03-12
feat(TimeRelease): Add support for scheduling and executing time-release transfers 

### LINKED ISSUE #2261: Time Release Pallet Cancellable Future Transfers
###  Description
As a user of the Time Release Pallet, I want to be able to schedule a transfer in the future that can still be cancelled by the initiator.

### Acceptance Criteria
- [x] I can create a transfer that will happen at a future block
- [ ] #2275
- [x] Once transferred, it works as it does now.

**Scheduling a Transfer**
- [x] A user can schedule a transfer to execute at a specific future block based on frequency block time.
- [x] The scheduled transfer reserves (holds) the sender’s total scheduled amount until execution or cancellation.
- [x]  The Time Release Pallet schedules the transfer by augmenting the ScheduleOrigin config of the Scheduler Pallet to include EnsureTimeReleaseOrigin.

**Executing a Scheduled Transfer**
- [x] When the scheduled block arrives, the Scheduler automatically executes the transfer by calling execute_scheduled_transfer.

execute_scheduled_transfer must:
- [x] Deduct the transfer amount from the sender’s held (reserved) balance, not their free balance.
- [x] Transfer the amount to the receiver and freeze the receiver’s balance upon receipt, similar to an immediate transfer.

**Access Control & Security**
- [x] Only the Time Release Origin can execute the execute_scheduled_transfer extrinsic.
- [x] No other user or external actor can execute the transfer manually.
- [x] The Scheduler is the only entity authorized to trigger the execution at the designated block time.

### Ideas/Notes
- One option is using the scheduler pallet
- Another option is having a lead in period on the transfer where it can be cancelled still

enddynayn:
Add ability to schedule a time release transfer in the future.

- Users can now schedule transfers to execute at a future block based on frequency block time.  
- Added a time release schedule by augmenting `ScheduleOrigin` in the Scheduler Pallet to include `EnsureTimeReleaseOrigin`, allowing the Time Release Pallet to manage scheduled calls.  
- Scheduled transfers reserve the sender’s total scheduled amount until execution.  
- Implemented `execute_scheduled_transfer`, which transfers funds from the sender’s reserved balance instead of their free balance.  
- Ensured that only the Time Release Origin can execute scheduled transfers, restricting manual execution.  
- Allowed initiators to cancel scheduled transfers before execution.  
- Improved failure handling to ensure transactions fail gracefully if execution conditions are not met.  

Closes #2261 

scheduled-transfer-sequenced-diagram


pallets/time-release/src/lib.rs
```diff
@@ -90,20 +113,37 @@ pub mod module {
...
 
 		/// Required origin for time-release transfer.
-		type TransferOrigin: EnsureOrigin<Self::RuntimeOrigin, Success = Self::AccountId>;
+		type TransferOrigin: EnsureOrigin<
+			<Self as frame_system::Config>::RuntimeOrigin,
+			Success = Self::AccountId,
+		>;
+
+		/// Timerelase origin
+		type TimeReleaseOrigin: EnsureOrigin<
```
enddynayn, line 143: rename: ScheduledTransferOrigin

pallets/time-release/src/tests.rs
```diff
@@ -415,6 +415,230 @@ fn exceeding_maximum_schedules_should_fail() {
...
+			schedule.clone(),
+			when
+		));
+
+		assert_ok!(TimeRelease::execute_scheduled_transfer(
+			Origin::<Test>::TimeRelease(ALICE).into(),
+			BOB,
+			schedule
+		));
+		assert_eq!(PalletBalances::free_balance(BOB), 20u64);
```
shannonwells, line 586: These tests could use comments reminding a reader how much hold/freeze balance these accounts have.  

Also it seems like if there are 2 periods I would expect 2 scheduled transfers. But I haven't looked at this pallet in a really long time.

pallets/time-release/src/lib.rs
```diff
@@ -79,9 +86,25 @@ pub mod module {
 		TimeReleaseVesting,
 	}
 
+	/// A reason for holding funds
+	/// Create a hold reason for this pallet that is aggregated by 'contract_runtime'.
+	#[pallet::composite_enum]
+	pub enum HoldReason {
+		/// Funds are frozen until
```
shannonwells, line 93: nit:  finish comment, until when?

pallets/time-release/src/lib.rs
```diff
@@ -113,6 +153,21 @@ pub mod module {
 
 		/// The block-number provider.
 		type BlockNumberProvider: BlockNumberProvider<BlockNumber = BlockNumberFor<Self>>;
+
+		/// The aggregate call type
+		type RuntimeCall: Parameter
+			+ Dispatchable<
+				RuntimeOrigin = <Self as frame_system::Config>::RuntimeOrigin,
+				PostInfo = PostDispatchInfo,
+			> + From<Call<Self>>
```
shannonwells, line 162: nit: indentation is confusing here

pallets/time-release/src/lib.rs
```diff
@@ -450,6 +587,122 @@ impl<T: Config> Pallet<T> {
...
+		T::Currency::balance_on_hold(&HoldReason::TimeReleaseScheduledVesting.into(), &who)
+	}
+
+	fn ensure_sufficient_hold_balance(from: &T::AccountId, amount: BalanceOf<T>) -> DispatchResult {
+		ensure!(Self::hold_balance_for(&from) >= amount, Error::<T>::InsufficientBalanceToFreeze);
+
+		Ok(())
+	}
+
+	fn validate_and_get_schedule_amount(
```
shannonwells, line 601: nit: seems like this could be inlined

pallets/time-release/src/lib.rs
```diff
@@ -450,6 +587,122 @@ impl<T: Config> Pallet<T> {
...
+
+		transfer_fn(from, to, schedule_amount)?;
+		Self::update_freeze(to, total_amount)?;
+
+		Self::append_release_schedule(to, schedule)?;
+
+		Ok(())
+	}
+
+	fn finalize_vesting_transfer_from_free_balance(
```
shannonwells, line 691: nit: These last two abstractions seem unnecessary since they are each used in only one place and the single line in each are clear what they're doing.

pallets/time-release/src/lib.rs
```diff
@@ -268,15 +323,10 @@ pub mod module {
...
 
-			if to == from {
-				ensure!(
-					T::Currency::balance(&from) >=
-						schedule.total_amount().ok_or(ArithmeticError::Overflow)?,
-					Error::<T>::InsufficientBalanceToFreeze,
-				);
-			}
+			let total_amount = schedule.total_amount().ok_or(ArithmeticError::Overflow)?;
+			Self::ensure_suffient_free_balance(&from, &to, total_amount)?;
```
shannonwells, line 327: spelling for a used function
```suggestion
			Self::ensure_sufficient_free_balance(&from, &to, total_amount)?;
```

pallets/time-release/src/lib.rs
```diff
@@ -450,6 +587,122 @@ impl<T: Config> Pallet<T> {
...
+			schedule_amount,
+			Exact,
+			Free,
+			Polite,
+		)?;
+
+		Ok(())
+	}
+
+	fn ensure_suffient_free_balance(
```
shannonwells, line 659: spelling for functions should be correct
```suggestion
	fn ensure_sufficient_free_balance(
```

shannonwells (changes requested):
The misspelled function needs to be corrected; everything else is non-blocking, but I do have a strong preference against one line abstractions.  I think it makes the code harder to follow.

pallets/time-release/src/tests.rs
enddynayn: There is only 1 transfer here.

pallets/time-release/src/lib.rs
enddynayn: Yeah, I agree, the linter is doing that.

pallets/time-release/src/lib.rs
enddynayn: I agree `make format` is doing this.

pallets/time-release/src/lib.rs
enddynayn: I thought this made the code a little easier to read. Let me know if you want me to change it.

pallets/time-release/src/lib.rs
shannonwells: well poop on that linter

pallets/time-release/src/lib.rs
```diff
@@ -450,6 +587,122 @@ impl<T: Config> Pallet<T> {
...
+	/// * `ArithmeticError::Overflow` - If the transfer amount overflows.
+	/// * `Error::InsufficientBalanceToFreeze` - If the hold balance is insufficient.
+	fn transfer_from_hold_balance(
+		from: &T::AccountId,
+		to: &T::AccountId,
+		schedule_amount: BalanceOf<T>,
+	) -> DispatchResult {
+		// Exact - Will fail if the exact amount cannot be transfered
+		// Free - deposit into the free balance but will locked following
+		// Polite - Execute with regular previlages it consideres existential
```
JoeCap08055, line 645: I'm a little confused on these comments; perhaps a little more clarification?

pallets/time-release/src/lib.rs
```diff
@@ -326,16 +376,95 @@ pub mod module {
...
+			let total_amount = Self::validate_and_get_schedule_amount(&schedule)?;
+			Self::ensure_sufficient_hold_balance(&from, total_amount)?;
+
+			Self::finalize_vesting_transfer_from_hold(&from, &to, schedule.clone())?;
+
+			Self::deposit_event(Event::ReleaseScheduleAdded {
+				from,
+				to,
+				release_schedule: schedule,
+			});
```
JoeCap08055, line 459: I'm confused on why we're emitting this event; didn't we emit this event when the schedule was initially added, and now we're executing a previously-added schedule? Or maybe we just have a naming problem...?

runtime/common/src/constants.rs
```diff
@@ -121,8 +121,7 @@ pub type HandleSuffixMax = ConstU16<99>;
 // -end- Handles Pallet
 
 // --- TimeRelease Pallet ---
-// Update
-pub type MinReleaseTransfer = ConstU128<0>;
+pub type MinReleaseTransfer = ConstU128<{ currency::EXISTENTIAL_DEPOSIT }>;
```
JoeCap08055, line 124: Hmm, could this be problematic?
I can understand that we might not want to schedule a transfer that might result in an underfunded account that will just get reaped, but won't this also prevent scheduling potentially legitimate transfers to accounts that already hold enough tokens to not be reaped?

Or is this sort of a "lesser of two evils" situation, ie yes, it disallows some minimal (and therefore unlikely?) scheduled transfer amounts, but guarantees that (at least at the time of scheduling) we won't schedule a transfer that will end up getting reaped?

Or, am I misunderstanding the purpose of this value?

enddynayn: This amount is very small, just one cent. We can reasonably expect users to send transfers larger than that. The main purpose of this limit is to help prevent storage spam.

pallets/time-release/src/lib.rs
enddynayn: Resolution to add documentation.

pallets/time-release/src/lib.rs
enddynayn: Oops, left over clean up work.

JoeCap08055: Suggest either:
* Replace comment with a reference to https://github.com/paritytech/polkadot-sdk/blob/c8d33396345237c1864dfc0a9b2172b7dfe7ac8f/substrate/frame/support/src/traits/tokens/fungible/hold.rs#L346 for the meaning of these parameters
* Replace comment with copied text of above reference

JoeCap08055 (approved):
one comment nit, otherwise looks great!

pallets/time-release/src/lib.rs
```diff
@@ -326,16 +383,100 @@ pub mod module {
...
+		///
+		/// * [`Event::ReleaseScheduleAdded`] - Indicates that a new release schedule was added.
+		#[pallet::call_index(5)]
+		#[pallet::weight(T::WeightInfo::execute_scheduled_transfer())]
+		pub fn execute_scheduled_transfer(
+			origin: OriginFor<T>,
+			dest: <T::Lookup as StaticLookup>::Source,
+			schedule: ReleaseScheduleOf<T>,
+		) -> DispatchResult {
+			let from = T::TimeReleaseOrigin::ensure_origin(origin)?;
```
aramikm, line 454: Q: I'm a little confused about the enforcement of the origin. How hard is it to just wrap the origin with `Origin::<T>::TimeRelease` and just call the extrinsic or that is not possible?

pallets/time-release/src/lib.rs
```diff
@@ -376,23 +517,23 @@ impl<T: Config> Pallet<T> {
...
-		Self::update_freeze(&to, total_amount)?;
+		let schedule_call =
+			<T as self::Config>::RuntimeCall::from(Call::<T>::execute_scheduled_transfer {
+				dest,
+				schedule,
+			});
 
-		<ReleaseSchedules<T>>::try_append(to, schedule)
-			.map_err(|_| Error::<T>::MaxReleaseSchedulesExceeded)?;
+		let _ = T::SchedulerProvider::schedule(
```
aramikm, line 532: It returns () so no need to ignore the result.
```suggestion
		T::SchedulerProvider::schedule(
```

runtime/frequency/src/lib.rs
```diff
@@ -121,7 +123,20 @@ use frame_support::traits::{TryStateSelect, UpgradeCheckSelect};
...
-/// Interface to collective pallet to propose a proposal.
+pub struct SchedulerProvider;
+
+impl SchedulerProviderTrait<RuntimeOrigin, BlockNumber, RuntimeCall> for SchedulerProvider {
+	fn schedule(
+		origin: RuntimeOrigin,
+		when: BlockNumber,
+		call: Box<RuntimeCall>,
+	) -> Result<(), DispatchError> {
+		let _ = Scheduler::schedule(origin, when, None, LOWEST_PRIORITY, call)?;
```
aramikm, line 134:
```suggestion
		Scheduler::schedule(origin, when, None, LOWEST_PRIORITY, call)?;
```

pallets/time-release/src/tests.rs
```diff
@@ -415,6 +415,230 @@ fn exceeding_maximum_schedules_should_fail() {
 	});
 }
 
+#[test]
+fn schedule_transfer_success() {
```
aramikm, line 419: if you already didn't have a test it might be useful to a have test that would cause the account to get reaped or any one the edge cases after transfer.

aramikm (commented):
Great effort but there are some fixes and questions and nits.

I also see the following warnings when I run the node which are not great and not sure if problematic.

```
warning: patch for `bandersnatch_vrfs` uses the features mechanism. default-features and features will not take effect because the patch dependency does not support this mechanism
warning: Patch `bandersnatch_vrfs v0.0.4 (https://github.com/davxy/ring-vrf.git?rev=48b642a#48b642a0)` was not used in the crate graph.
Perhaps you misspelled the source URL being patched.
Possible URLs for `[patch.<URL>]`:
    https://github.com/w3f/ring-vrf
warning: Patch `fflonk v0.1.1 (https://www.github.com/w3f/fflonk?rev=be95d4c971b1d15b5badfc06ff13f5c07987d484#be95d4c9)` was not used in the crate graph.
Check that the patched package version and available features are compatible

```

pallets/time-release/src/lib.rs
```diff
@@ -326,16 +383,100 @@ pub mod module {
...
+			let from = T::TransferOrigin::ensure_origin(origin)?;
+			let to = T::Lookup::lookup(dest.clone())?;
+
+			let total_amount = Self::validate_and_get_schedule_amount(&schedule)?;
+			Self::ensure_sufficient_free_balance(&from, &to, total_amount)?;
+
+			let _ = Self::update_schedule_vesting_hold(
+				&from,
+				schedule.total_amount().ok_or(ArithmeticError::Overflow)?,
+			);
```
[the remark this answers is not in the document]
aramikm: This one is important since it is eating the error

Cargo.toml
```diff
@@ -162,6 +162,12 @@ substrate-frame-rpc-system = { git = "https://github.com/paritytech/polkadot-sdk
 substrate-prometheus-endpoint = { git = "https://github.com/paritytech/polkadot-sdk", branch = "release-polkadot-v1.13.0" }
 sp-debug-derive = { git = "https://github.com/paritytech/polkadot-sdk", branch = "release-polkadot-v1.13.0", default-features = false }
 
+# hotfix for https://github.com/paritytech/polkadot-sdk/issues/7653
+[patch.'https://github.com/w3f/fflonk']
+fflonk = { git = "https://www.github.com/w3f/fflonk", rev = "be95d4c971b1d15b5badfc06ff13f5c07987d484" }
+bandersnatch_vrfs = { git = "https://github.com/davxy/ring-vrf.git", rev = "48b642a", default-features = false }
```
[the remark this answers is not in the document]
enddynayn: This is currently an issue on main.

pallets/time-release/src/lib.rs
```diff
@@ -450,6 +599,106 @@ impl<T: Config> Pallet<T> {
...
+		Ok(())
+	}
+
+	fn ensure_sufficient_free_balance(
+		from: &T::AccountId,
+		to: &T::AccountId,
+		amount: BalanceOf<T>,
+	) -> DispatchResult {
+		if to == from {
+			ensure!(T::Currency::balance(&from) >= amount, Error::<T>::InsufficientBalanceToFreeze,);
```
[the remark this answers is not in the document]
enddynayn: Perhaps, but this might be a bit out of scope. These errors weren’t created by me; they are part of the main setup. If we’d like to make changes to the current setup, I’d suggest creating a new ticket for it. Let me know what you think!

pallets/time-release/src/lib.rs
```diff
@@ -450,6 +599,106 @@ impl<T: Config> Pallet<T> {
...
+		Ok(())
+	}
+
+	fn validate_and_get_schedule_amount(
+		schedule: &ReleaseScheduleOf<T>,
+	) -> Result<BalanceOf<T>, DispatchError> {
+		ensure_valid_release_schedule::<T>(schedule)
+	}
+
+	fn append_release_schedule(
```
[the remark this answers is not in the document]
enddynayn: I think having these go into functions makes it easier for name changes and updates.

pallets/time-release/src/lib.rs
```diff
@@ -433,6 +574,14 @@ impl<T: Config> Pallet<T> {
 		Ok(())
 	}
 
+	fn update_schedule_vesting_hold(
```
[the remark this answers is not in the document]
enddynayn: I see your point, but I think isolating dependencies from the outside can be beneficial. Since names and structures might change over time, encapsulating them could make the code more maintainable and easier to work with in the long run. Let me know what you think!

pallets/time-release/src/lib.rs
enddynayn: Good question! I believe that’s not possible. The Origin serves as an access control mechanism. We ask the same question for any origin that requires sudo or governance, and if it were possible to bypass it, those mechanisms would break.

pallets/time-release/src/lib.rs
```diff
@@ -450,6 +599,106 @@ impl<T: Config> Pallet<T> {
...
+		T::Currency::balance_on_hold(&HoldReason::TimeReleaseScheduledVesting.into(), &who)
+	}
+
+	fn ensure_sufficient_hold_balance(from: &T::AccountId, amount: BalanceOf<T>) -> DispatchResult {
+		ensure!(Self::hold_balance_for(&from) >= amount, Error::<T>::InsufficientBalanceToFreeze);
+
+		Ok(())
+	}
+
+	fn validate_and_get_schedule_amount(
```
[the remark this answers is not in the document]
enddynayn: I wanted to avoid modifying the existing code since it seems to be outside the scope of this pallet. However, you make a great point, and we could create a ticket to further improve it. I also have a few changes in mind that could make it much easier to work with.

pallets/time-release/src/lib.rs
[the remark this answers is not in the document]
aramikm: This is not a big deal but since they recommend using `T::Currency::reducible_balance` instead of `T::Currency::balance` it might worth taking a look at it.

pallets/time-release/src/lib.rs
aramikm: I might be wrong but I think sudo or governance check it against a state (accounts) stored on-chain but this doesn't seems to do that.

enddynayn: That’s true. Let me clarify my understanding. The origin is set by the runtime/node and is typically one of the following:
```
enum RawOrigin {
    Signed,
    Root,
    None,
}
```

The origin type is determined based on who or what initiated the action:
- Signed -> A user-submitted transaction.
- Root -> A governance or privileged system call.
- None -> A system-initiated operation without a specific identity.

In this case, when a signed transaction comes our way we modify the origin and pass it to the scheduler. Only in this piece of code can you get that origin (unless we add it else where). 

A transaction **cannot modify its own origin**, as origin assignment is strictly handled internally by the runtime for security and access control reasons.

Let me know if you have any questions! This is my understanding, but I might have missed something. Happy to discuss further.

pallets/time-release/src/lib.rs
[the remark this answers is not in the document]
enddynayn: I think thats a good idea. I will create a new ticket for this.

Cargo.toml
```diff
@@ -162,6 +162,12 @@ substrate-frame-rpc-system = { git = "https://github.com/paritytech/polkadot-sdk
 substrate-prometheus-endpoint = { git = "https://github.com/paritytech/polkadot-sdk", branch = "release-polkadot-v1.13.0" }
 sp-debug-derive = { git = "https://github.com/paritytech/polkadot-sdk", branch = "release-polkadot-v1.13.0", default-features = false }
 
+# hotfix for https://github.com/paritytech/polkadot-sdk/issues/7653
```
enddynayn, line 165: Remove and comment out Cargo CI and create a ticket too fix. Issue should resolve upon upgrade to new Polkadot version.

pallets/time-release/src/lib.rs
```diff
@@ -326,16 +383,100 @@ pub mod module {
...
+
+		/// This function schedules a transfer by calling the `Scheduler` pallet's `schedule` function.
+		/// The transfer will be executed at the specified block number.
+		///
+		/// * [`Error::InsufficientBalanceToFreeze`] - Insufficient amount of balance to freeze.
+		/// * [`Error::MaxReleaseSchedulesExceeded`] - Failed because the maximum release schedules was exceeded-
+		/// * [`ArithmeticError::Overflow] - Failed because of an overflow.
+		///
+		#[pallet::call_index(4)]
+		#[pallet::weight(T::WeightInfo::schedule_transfer().saturating_add(T::WeightInfo::transfer()))]
```
enddynayn, line 400: execute_schedule_transfer

PR #2292, opened 2025-03-21
feat(TimeRelease): add ability to cancel scheduled transfers

### LINKED ISSUE #2275: I can cancel a future transfer

enddynayn:
Add ability cancel a scheduled transfer by providing a unique identifier.
When a transfer is canceled, any held funds are released. Scheduled transfers now
require a unique identifier at the time of creation, which is used to reference and manage each transfer.

Closes #2275


enddynayn:
Running benchmarks for TimeRelease pallet but action is not committing the updates https://github.com/frequency-chain/frequency/actions/runs/14002616793

enddynayn:
> Looks good, just one comment/question:
> 
> Was having a human-readable string identifier a requirement? Because it seems like a costly storage impact (32-byte storage key) (and associated reads/writes). Why not just use a monotonically-increasing `u32` and emit the allocated identifier in an event after the schedule is created (like we do with MSA IDs)?
> 
> Maybe we don't care about the storage cost due to the expected limited usage of this feature...? (Also, I suppose the storage is reclaimed once the schedule executes or is canceled, so there's that...)

Good question. We’re passing this name to the Scheduler pallet, which stores it as a 32-byte key. This should be fine, as the key is removed once the scheduled transaction is either executed or canceled.

JoeCap08055 (approved):
- Reviewed code
- Built branch
- Ran unit & e2e tests locally
- 🚢 it!

aramikm (approved):
Looks good! a few nits.

e2e/time-release/timeRelease.test.ts
```diff
@@ -50,4 +50,92 @@ describe('TimeRelease', function () {
       assert.notEqual(target, undefined, 'should have returned ReleaseScheduleAdded event');
     });
   });
+
+  describe('create schedule and cancel transfer', function () {
+    it('create a schedule transfer', async function () {
```
shannonwells, line 55: nit: copypasta -  has the same name as the test below
```suggestion
    it('creates and cancels named scheduled transfers', async function () {
```

e2e/time-release/timeRelease.test.ts
```diff
@@ -50,4 +50,81 @@ describe('TimeRelease', function () {
...
+        scheduleName1
+      );
+      const { target: cancelTarget1 } = await cancelScheduleTransferTx1.signAndSend();
+      assert.notEqual(cancelTarget1, undefined, 'should have returned scheduler Canceled event');
+
+      const cancelScheduleTransferTx2 = await ExtrinsicHelper.timeReleaseCancelScheduledNamedTransfer(
+        fundingSource,
+        scheduleName2
+      );
+      const { target: cancelTarget2 } = await cancelScheduleTransferTx2.signAndSend();
```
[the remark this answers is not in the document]
shannonwells: Agree, it would be nice to see in this test that the name applied to the one & not the other.

runtime/frequency/src/lib.rs
```diff
@@ -128,10 +128,17 @@ pub struct SchedulerProvider;
 impl SchedulerProviderTrait<RuntimeOrigin, BlockNumber, RuntimeCall> for SchedulerProvider {
 	fn schedule(
 		origin: RuntimeOrigin,
+		id: [u8; 32],
```
shannonwells, line 131: nit: it's a little confusing to call it id in some places and named in others.

pallets/time-release/src/lib.rs
```diff
@@ -572,14 +615,28 @@ impl<T: Config> Pallet<T> {
 		Ok(())
 	}
 
-	fn update_schedule_vesting_hold(
+	fn hold_funds_for_scheduled_vesting(
```
shannonwells, line 618: better

shannonwells (approved):
Changes look good! I have only the couple of nitpicks but some are nitpickier than others - I agree with Aramik's request to add another check to the 2nd canceled transfer though.

I reviewed the code, I looked at the tests for completeness.

Issue #2277, opened 2025-03-20
Classify Ethereum Keys and Accounts


p5150j:
### Description
Classify all existing keys in the system to determine their types and associated tokens. This classification will guide subsequent migration steps and help prioritize actions for different key categories.

Tasks

- Create scripts to scan and identify control keys with tokens
- Create scripts to scan and identify regular token keys
- Create scripts to scan and identify control-only keys (no tokens)
- Document the criteria used for classification

Acceptance Criteria

- [ ]  Complete inventory of Phase1 keys with their classification
- [ ]  Classification data stored in an accessible format for other tasks
- [ ]  Documentation explaining classification methodology

Engineering Notes
- These have to executed on paseo-testnet and Mainnet

aramikm:
Uploaded the list of the keys and their classifications on the PR but We will need to rerun these again exactly before the migration begins. Closing for now

PR #2294, opened 2025-03-25
key-migration: control key scripts

aramikm:
# Goal
The goal of this PR is to create scripts to be able to fetch control keys

Related to #2277 

# Details
- added a tool to be able to fetch all msa ids and their control keys from `publicKeyToMsaId` storage
- added a tool to be able to fetch all msa ids and their control keys from offchain indexer
- the output is in compatible so that after sorting we can compare them for any discrepancies
- added a tool to be able to fetch all token keys

# Control Keys with Tokens
The script should be executed close to the migration to get the latest state but for now we have these control keys with tokens on main-net.
```
"0xca94146d680192fbeff69aa6080e2f9daff8d8d238b9a76f136e35d5d3083d3c"
"0x748a616c3171d049c7f331df9918e010ff27625d22c016ce6a6d3c2f24004166"
"0x54fa640b48cbfdea066434189cecdd11f76c26b2499ece246fc1c69ea0e74761"
"0x94c602c107c5c11a5e9fed59f65140701bb48678a10e667c32c8a03914a0ac0c"
"0x5e1fb4d1e2faebbf2ffa2cb50718b57064cfd3c6b1445948f87fcef923108f4d"
"0x068cbd7799b593960797a2c698f7943dd60ec8720bb28ed4019f0387444a7862"
"0xaed8aa458599d23d83ba2bd13d5e9afb4aacb826e54d6876a41e01c65f72da4e"
"0xae139214802b320dff4f202d37554e6899d37c5913927b9af440f783f8ec5a38"
"0x58083af872b860bb71a359b92f1d2c67b9c5c970a64b5f264130e02533b0f372"
"0x861b30dae9083e3bebaf427d08ec20b4b71282a987342c226cfc3bfbdeabaf61"
```


wilwade (approved):
Only validated that it did not impact any other code.

tools/eth-migration/README.md
```diff
@@ -0,0 +1,20 @@
+# Eth migrations tools
+Some scripts and tools to facilitate the key migration
+
+## Get Onchain msa id and keys
+`npm run control-keys --silent`
```
shannonwells, line 5: nit: there is 1 line difference (the first one, a header) in output with/without `--silent`, and also, this command outputs the values but doesn't direct them to a file.

tools/eth-migration/fetch-control-keys.mjs
```diff
@@ -0,0 +1,98 @@
+// Only way to silence PolkadotJS API warnings we don't want
+console.warn = () => {};
+
+import { ApiPromise, WsProvider } from "@polkadot/api";
+
+const SOURCE_URL = "wss://1.rpc.frequency.xyz";
```
shannonwells, line 6: is it worth making this || with an environment value?

tools/eth-migration/fetch-offchain-index-keys.mjs
```diff
@@ -0,0 +1,85 @@
...
+	if (promises.length > 0) {
+		printRpcResults(await Promise.all(promises), batchNumber);
+	}
+}
+
+function printRpcResults(results, batchNumber) {
+	let index = 0;
+	for (const r of results) {
+		if (!r.isSome) {
+			console.error(`No keys for MsaId: ${(batchNumber * PROMISE_BATCH_SIZE + index)}!`);
```
shannonwells, line 62: I'm not sure how you intend to handle the error case when you're using the sorted output, but I wonder if outputting a valid row with 0x0 as the key, instead of outputting an error might be more helpful.  That or you could redirect errors to another file.

shannonwells (approved):
All worked for me, I read over the scripts and tried each one against testnet.

mattheworris (approved):
- Read through code.
- Ran test commands.

tools/eth-migration/fetch-control-keys.mjs
JoeCap08055, line 6:
```suggestion
const SOURCE_URL = process.env["FREQUENCY_URL"] || "wss://1.rpc.frequency.xyz";
```

tools/eth-migration/fetch-control-keys.mjs
```diff
@@ -0,0 +1,98 @@
...
+		},
+	},
+	types: {
+		MessageSourceId: 'u64',
+		KeyInfoResponse: {
+			msa_keys: 'Vec<AccountId>',
+			msa_id: 'MessageSourceId',
+		},
+	},
+};
```
JoeCap08055, line 31: nit: not worth the dependency to import `@frequency-chain/api-augment`?

wilwade: I did this on a few tools. It limits the dependency on the Polkadotjs version trees for a cleaner setup. (Any you don't need the TS support)

tools/eth-migration/fetch-control-keys.mjs
```diff
@@ -0,0 +1,98 @@
...
+export async function printPublicKeysWithMsaId(sourceApi, publicKeys) {
+	let promises = [];
+	for (const publicKey of publicKeys) {
+		promises.push(sourceApi.query.msa.publicKeyToMsaId(publicKey));
+	}
+
+	if (promises.length > 0) {
+		printResults(await Promise.all(promises), publicKeys);
+	}
+}
```
JoeCap08055, line 72: You already fetch the storage from `msa.publicKeyToMsaId`, but then you threw away the value, keeping the key, and making an extra storage query _per key_ to get its MSA; information you already fetched in the paginated request--also it really slams the node.

Suggested optimization:
```js
export async function fetchControlKeysFromState(sourceUrl) {
	// Connect to the state source
	const sourceProvider = new WsProvider(sourceUrl);
	const sourceApi = await ApiPromise.create({ provider: sourceProvider, ...options });

    let startKey;
    let result;
    do {
      result = await sourceApi.query.msa.publicKeyToMsaId.entriesPaged({ args: [], pageSize: BATCH_SIZE, startKey });

      result.forEach(([key, value]) => {
        if (!value.isSome) {
          console.error(`No MsaId for 0x${key.args[0].toHex()}`);
        } else {
          console.log(`${value.unwrap().toString()},0x${key.args[0].toHex()}`);
        }
	  });

      startKey = result[result.length - 1][0];
      } while (result.length > 0);
}
```

tools/eth-migration/fetch-token-keys.mjs
JoeCap08055, line 1: Here again, you're slamming the chain with individual queries for each key--redundantly, because we've already fetched the account value. And we can use the 'multi query' api to make the msa queries more efficient.

```js
export async function fetchTokenKeysFromState(sourceUrl) {
	// Connect to the state source
	const sourceProvider = new WsProvider(sourceUrl);
	const sourceApi = await ApiPromise.create({provider: sourceProvider, ...options});

	let result;
	do {
		result = await sourceApi.query.system.account.entriesPaged({ args: [], pageSize: BATCH_SIZE, startKey: result && result[result.length - 1][0] });
		let msaResult = await sourceApi.query.msa.publicKeyToMsaId.multi(result.map(([key, _]) => key.args[0]));
		result.forEach(([key, account], index) => {
			const obj = {
				publicKey: key.args[0].toString(),
				msaId: msaResult[index].isSome ? msaResult[index].unwrap().toNumber() : 0,
				values: account,
			}
		    console.log(`${JSON.stringify(obj)}`);
		});
	} while (result.length === BATCH_SIZE);
}
```

Granted, the load on the node is much less since there should be very few token accounts, but it's not bad to eliminate unnecessary RPC calls.

tools/eth-migration/README.md
aramikm: I always preferred to have the option to pipe it to file in case or see it in console so that's why I didn't add the pipe to file.  But open to change it in case you have strong opinions about it

tools/eth-migration/fetch-control-keys.mjs
aramikm: Sure

tools/eth-migration/fetch-control-keys.mjs
aramikm: It keeps it simpler

tools/eth-migration/fetch-control-keys.mjs
aramikm: Awesome suggestion, I was not aware of `entriesPaged`.

tools/eth-migration/fetch-token-keys.mjs
aramikm: Thanks, much simpler

tools/eth-migration/fetch-offchain-index-keys.mjs
aramikm: So by default you would see the errors in console but you can pipe it into another file if you want. I think it's a matter of taste how you would like to see it . Currently it shows up in the diff and it's not a big deal imo.

PR #2297, opened 2025-03-25
key-migration: compare DB with on-chain

### LINKED ISSUE #2280: Verify Legacy ETH Accounts and Control Keys
### Description
Check for the existence of legacy ETH accounts and verify them against Control keys in the Custodial wallet to determine the scope of migration needed.

Tasks

- Develop process to identify legacy ETH accounts
- Compare against Control keys in Custodial wallet
- Document findings and highlight any discrepancies

Acceptance Criteria

- [x]  Complete list of legacy ETH accounts
- [x]  Mapping between legacy accounts and Control keys in Custodial wallet
- [x]  Find all accounts in custodial-wallet that are not part of the chain

Engineering Notes
- We need to ensure there are no Legacy ETH accounts on mainnet (there might be some on testnet which should be fine). This is to ensure there are no 2 incompatible ETH address mappings at the same time

Blocked by #2277

aramikm:
# Goal
The goal of this PR is to get the script for list of accounts for
- on-chain control keys which are not in the DB
- DB account which are not on-chain

Closes #2280


tools/eth-migration/README.md
```diff
@@ -18,3 +18,7 @@ Some scripts and tools to facilitate the key migration
 
 ### Select token keys which are also control keys
 `jq -c 'select(.msaId != 0)' tokens-keys.txt`
+
+## Get the difference between on-chain control keys and Custodial Wallet
+1.  fill in the CW DB connection string and other details
+2. `npm run cw-compare`
```
wilwade, line 24: I think we should swap this to be "db" instead of "cw" as (correctly so) that's all it is actually doing.

```suggestion
## Get the difference between on-chain control keys and a database
1.  fill in the DB connection string and other details
2. `npm run db-compare`
```

wilwade (changes requested):
One change to swap to db language, but otherwise looks good to me

tools/eth-migration/cw-comparison.mjs
JoeCap08055, line 1: Would be nice if this script either took CLI arguments or env files for configurable parameters; having to edit the source for your input parameters seems clunky, and also not scriptable.

aramikm: Done

wilwade (approved):
Nice

JoeCap08055 (approved):

PR #2307, opened 2025-04-03
use fork aware txpool

aramikm:
# Goal
The goal of this PR is 

Closes 

# Discussion

- 

# Checklist
- [ ] Updated Pallet Readme?
- [ ] Design doc(s) updated?
- [ ] Unit Tests added?
- [ ] e2e Tests added?
- [ ] Benchmarks added?

### DIFF
--- a/scripts/run_collator.sh
+++ b/scripts/run_collator.sh
@@ -50,7 +50,7 @@ bootnode () {
     echo "$bootnode"
 }
 
-args+=( "--" "--wasm-execution=compiled" "--chain=${chain}" "--bootnodes=$(bootnode "$alice" "$alice_rpc_port")" "--bootnodes=$(bootnode "$bob" "$bob_rpc_port")" )
+args+=( "--" "--wasm-execution=compiled" "--pool-type=fork-aware" "--chain=${chain}" "--bootnodes=$(bootnode "$alice" "$alice_rpc_port")" "--bootnodes=$(bootnode "$bob" "$bob_rpc_port")" )
 
 set -x
 "$ctpc" "${args[@]}"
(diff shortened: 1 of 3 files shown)


shannonwells (approved):
Nice, thank you

scripts/run_collator.sh
```diff
@@ -50,7 +50,7 @@ bootnode () {
     echo "$bootnode"
 }
 
-args+=( "--" "--wasm-execution=compiled" "--chain=${chain}" "--bootnodes=$(bootnode "$alice" "$alice_rpc_port")" "--bootnodes=$(bootnode "$bob" "$bob_rpc_port")" )
+args+=( "--" "--wasm-execution=compiled" "--pool-type=fork-aware" "--chain=${chain}" "--bootnodes=$(bootnode "$alice" "$alice_rpc_port")" "--bootnodes=$(bootnode "$bob" "$bob_rpc_port")" )
```
wilwade, line 53: @aramikm is there a way to make the fork aware pool default instead of needing to set it?

aramikm: I think it should be possible. I'll look into it.

aramikm: Since the `Configuration` is part of the code we depend on, one thing we can do is just check for the  SealingMode. and default it to ForkAware pool 
`if SealingMode != Instant `

wilwade: That sounds good. I'm also ok with just forcing the pool one way in Instant sealing mode. It's a special mode, so its ok to ignore that setting (as there are no forks in instant mode anyway).

In the end, we'll get that all collators and such upgrade, they default to the fork aware pool and avoid having to ask people to enable it when we need it.

PR #2308, opened 2025-04-03
default fork-aware

aramikm:
# Issues
- We can not get a `mut` reference to override the inner passed config so this forces us to just use the default pool values which is not great.


node/service/src/service.rs
```diff
@@ -136,14 +140,38 @@ pub fn new_partial(
 		telemetry
 	});
 
-	// See https://github.com/paritytech/polkadot-sdk/pull/4639 for how to enable the fork-aware pool.
+	// default to Fork-Aware pool
+	// Warning: This overrides all the TransactionPoolParams with the default values (config.transaction_pool)
+	let pool_params = TransactionPoolParams {
```
aramikm, line 145: @wilwade This is one way of doing it but I don't like it since we have no other choice than use default params due to Rust limits.

aramikm:
Replaced with #2349

PR #2309, opened 2025-04-04
remove sp_std

aramikm:
# Goal
Replacing sp_std with core and alloc

https://forum.polkadot.network/t/polkadot-release-analysis-polkadot-stable2407/9452


aramikm:
replaced with https://github.com/frequency-chain/frequency/pull/2327

PR #2311, opened 2025-04-04
Design Doc: Provider Contexts

### LINKED ISSUE #2269: Provider Application Details Design Doc
As a social login provider, I want to be able to get trusted details about the provider and the applications the provider runs.

## AC
- Anyone should be able to fetch the name, domain, and logo of a Provider
- A provider should be able to create and update name, domain, and logo
- A provider should be able to create, update, and remove applications that have a unique (to that provider) string identifier, name, domain, and logo
- Anyone who has a provider id and application id should be able to fetch the details of that provider's application
- Anyone should be able to discover a list of a provider's application ids
- The data (name, domain, logo) should all be onchain
- The logo should be a 250x100 px png
- Able to replace: https://projectlibertylabs.github.io/siwf/v2/docs/DataStructures/All.html#application-context

## Open Questions
- How do we protect against a phishing provider (for mainnet)
    - Does Governance need to be a part of approving the changes to a provider?
    - Outside verification options?
- What, if anything, should this have for recovery providers
- Evaluate Recovery Provider storage requirements: should `RecoveryProviderApproved` storage be rolled into Provider Application Details storage?

## Notes
- Don't forget the updates to Provider Dashboard to support it

wilwade:
# Goal
The goal of this PR is to layout a design doc for storing Provider Contexts on chain. This would allow a provider to have one or more applications and provider user consumed logos and names (in multiple languages).

Closes #2269


designdocs/provider_contexts.md
```diff
@@ -0,0 +1,230 @@
+# Provider Contexts Design Doc
+
+## Context and Scope
+
+In Frequency, a Provider has been representative of both an Application and a Company.
```
jeanettedepatie, line 5: "On Frequency a Provider may either represent an application or a company.

designdocs/provider_contexts.md
```diff
@@ -0,0 +1,230 @@
...
+1. Verify the SIWF Signed Request.
+2. Lookup the Provider via the `publicKey` in the SIWF Signed Request.
+3. If any, fetch the Application Identifier from Frequency.
+4. Display the information from the ApplicationRegistryEntry (or default Provider one) to the user to help them know who they are authorizing.
+5. User is happy to continue with the login process.
+
+### **Provider Dashboard Steps** <a id='dashboard'></a>
+
+Provider Dashboard needs be able to:
+
```
jeanettedepatie, line 171: needs to be able

designdocs/provider_contexts.md
```diff
@@ -0,0 +1,230 @@
...
+- Create a provider without any application context other than the default provider context
+- Update the default provider context, logos, and translations
+- Add new Application Contexts with a unique to that provider id, logos, and translations
```
jeanettedepatie, line 174: a unique what?

designdocs/provider_contexts.md
```diff
@@ -0,0 +1,230 @@
...
+## Non-goals
+
+- Implementation of verified credentials for applications (Aka FooBar is approved by <Consumer Org>)
+- Providing for independent body to perform verification
+
+## Benefits and Risk
+
+### Benefit: User protection and Application diversity
+
+Users will have a larger trust when they are logging into an application that the application is represented honestly, even with the continuous risk of trusting any new application with user data.
```
jeanettedepatie, line 188: replace a larger trust with greater confidence

designdocs/provider_contexts.md
```diff
@@ -0,0 +1,230 @@
...
+Tricking a single member is much simpler than a more detailed vetting process; however, at this stage, the number of providers is small.
+
+This process should be re-evaluated as the number of Providers grows.
+
+### Risk: Logo Images on Chain
+
+This provides a direct way to place image content on chain, intended to be interpreted as image content.
+While this must still pass through the governance step to be used by others, there is always a risk of harmful images being proposed.
+This risk is mitigated by:
+- A registered provider must be taking the action
```
wesbiggs, line 201: This is not clear from the sequence diagram, at least for the initial provider creation flow; it looks like all application data (including image bytes) would be recorded on-chain as part of the provider proposal.

Wouldn't the mitigation be submitting the image hashes rather than the images, as discussed in the "Implementation Suggestion" section?

designdocs/provider_contexts.md
```diff
@@ -0,0 +1,230 @@
...
+### **Example of Wallet Usage via SIWF** <a id='siwf'></a>
+
+SIWF [Signed Request Payload](https://projectlibertylabs.github.io/siwf/v2/docs/DataStructures/All.html) can expand the optional `applicationContext` value with a new optional `id` field that is the text of the `ApplicationIdentifier` on chain.
+This would allow a smooth transition between `applicationContext.url` and `applicationContext.id` with both being optional.
+
+The Wallet would then:
+1. Verify the SIWF Signed Request.
+2. Lookup the Provider via the `publicKey` in the SIWF Signed Request.
+3. If any, fetch the Application Identifier from Frequency.
+4. Display the information from the ApplicationRegistryEntry (or default Provider one) to the user to help them know who they are authorizing.
```
wesbiggs, line 165: Can we better specify what is meant by the "default Provider one"?

designdocs/provider_contexts.md
wilwade:
```suggestion
On Frequency, a Provider may either represent an application or a company.
```

designdocs/provider_contexts.md
```diff
@@ -0,0 +1,230 @@
...
+Reference Document: [SIWF Specification](https://projectlibertylabs.github.io/siwf/v2/docs/)
+
+### Why Is This Needed?
+
+#### Concept of an Application Separate from a Provider
+
+While most companies have a single major product, companies already want to take advantage of Frequency's universal data to provide many different applications to serve different needs.
+While each application could be a separate provider, this causes two issues: The sharing of capacity and the user permissions.
+
+To have multiple providers requires separate Capacity staking.
```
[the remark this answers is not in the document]
wilwade:
```suggestion
Multiple providers require separate Capacity staking.
```

designdocs/provider_contexts.md
```diff
@@ -0,0 +1,230 @@
...
+As these applications would be from the same company, it is more reasonable to say that they should share Capacity.
```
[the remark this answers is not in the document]
wilwade:
```suggestion
As these applications would be from the same company, it is reasonable to say that they should share Capacity.
```

designdocs/provider_contexts.md
```diff
@@ -0,0 +1,230 @@
...
+To align with the data security model of company-to-user in the world today, permissions, and thus Provider-level delegations continue to reflect the reality best.
+
+Finally, this structure does not prevent the setup of a 1:1 relationship between Application and Provider.
+If a company wishes, they can create more than one Provider for the purposes of separation.
+This may be desirable if there is a subsidiary company or other complex structures.
+
+#### User Trust in Application Presentation
+
+Users need to know what application they are logging into and trust that the displayed information is not a phishing attack.
+Users have a reasonable expectation that the chain and wallet provides a level of protection against phishing attacks from Providers and Applications.
```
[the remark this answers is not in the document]
wilwade:
```suggestion
Users have a reasonable expectation that the chain and wallet provide a level of protection against phishing attacks from Providers and Applications.
```

designdocs/provider_contexts.md
```diff
@@ -0,0 +1,230 @@
...
+
+Currently, the `ProviderRegistryEntry` has only a `name` property, and no way to update said name.
+
+Applications have two different pieces of default and (potentially) internationalized data:
+- name
+- logo/image
+
+Limits
+1. Provide a space to internationalize up to 150 localizations (Windows and macOS each have < 150).
+2. Image should be `base64` encoded `png` that is `250x100` (support for future).
```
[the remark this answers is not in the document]
wilwade: Might have been broken as I switched (and failed here) to bytes over base64. Will fix.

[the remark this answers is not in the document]
wilwade:
```suggestion
2. Image MUST BE encoded (`png`)[https://www.w3.org/TR/png-3/] that is `250x100` (support for future sizes and light/dark should be considered).
```

designdocs/provider_contexts.md
wilwade:
```suggestion
4. Display the information from the `ApplicationRegistryEntry` (or the Provider Registry Entry `ProviderToRegistryEntry` if there is no application context identifier) to the user to help them know who they are authorizing.
```

designdocs/provider_contexts.md
```diff
@@ -0,0 +1,230 @@
...
+
+SIWF [Signed Request Payload](https://projectlibertylabs.github.io/siwf/v2/docs/DataStructures/All.html) can expand the optional `applicationContext` value with a new optional `id` field that is the text of the `ApplicationIdentifier` on chain.
+This would allow a smooth transition between `applicationContext.url` and `applicationContext.id` with both being optional.
+
+The Wallet would then:
+1. Verify the SIWF Signed Request.
+2. Lookup the Provider via the `publicKey` in the SIWF Signed Request.
+3. If any, fetch the Application Identifier from Frequency.
+4. Display the information from the ApplicationRegistryEntry (or default Provider one) to the user to help them know who they are authorizing.
+5. User is happy to continue with the login process.
```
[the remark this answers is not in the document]
wilwade:
```suggestion
5. Allow the user to continue the login process.
```

designdocs/provider_contexts.md
```diff
@@ -0,0 +1,230 @@
...
+
+### **Provider Dashboard Steps** <a id='dashboard'></a>
+
+Provider Dashboard needs be able to:
```
wilwade, line 170:
```suggestion
Provider Dashboard needs to be able:
```

designdocs/provider_contexts.md
wilwade:
```suggestion
- Add new Application Contexts with a identifier (unique to that provider), logos, and translations
```

designdocs/provider_contexts.md
wilwade:
```suggestion
Users will have a greater confidence when they are logging into an application that the application is represented honestly, even with the continuous risk of trusting any new application with user data.
```

designdocs/provider_contexts.md
wilwade: Fixed

designdocs/provider_contexts.md
wilwade: Applied some notes to highlight this.

designdocs/provider_contexts.md
```diff
@@ -0,0 +1,231 @@
...
+	pub type ProviderToApplicationRegistryEntry<T: Config> = StorageDoubleMap<
+		_,
+		Twox64Concat,
+		ProviderId,
+		Twox64Concat,
+		ApplicationIdentifier<T>,
+		ApplicationRegistryEntry<T::MaxProviderNameSize, T::MaxProviderLogo250X100Size>,
+		OptionQuery,
+	>;
+	```
```
JoeCap08055, line 105: This looks confusing to me. Can a `StorageMap` reference another StorageMap? Seems like there's some naming mixups.

Also, I realize that `ApplicationIdentifier` is not in the critical path anywhere, but should we keep identifiers normalized (ie, numeric/integer)?

designdocs/provider_contexts.md
JoeCap08055, line 1: This appears to get rid of Provider info (ie, `ProviderRegistryEntry` completely, replaced by application-level info. (Though I suppose, a ProviderId being an MsaId, Providers could still register `dsnp.profile` data?)

I think it's important that Provider-level information be discoverable by the user as well as Application-level.

Also, perhaps out of scope, but this seems like an opportune moment to expand the information associated with a Provider or Application, perhaps a link to off-chain info _à la_ `dsnp.profile` or a DNS-like record?

designdocs/provider_contexts.md
```diff
@@ -0,0 +1,231 @@
...
+    }
+    ```
+3. The `ProviderToRegistryEntry` be updated to use `ProviderToApplicationRegistryEntry`
+    ```rust
+    #[pallet::storage]
+	pub type ProviderToRegistryEntry<T: Config> = StorageMap<
+		_,
+		Twox64Concat,
+		ProviderId,
+		ProviderToApplicationRegistryEntry<T::MaxProviderNameSize, T::MaxProviderLogo250X100Size>,
```
wilwade, line 86: @JoeCap08055 This was a typo. It should ALSO be a `ApplicationRegistryEntry` (or an alias'd version of it)

```suggestion
3. The `ProviderToRegistryEntry` be updated to use `ApplicationRegistryEntry` so that the Provider has a "default" `ApplicationRegistryEntry`.
    ```rust
    #[pallet::storage]
	pub type ProviderToRegistryEntry<T: Config> = StorageMap<
		_,
		Twox64Concat,
		ProviderId,
		ApplicationRegistryEntry<T::MaxProviderNameSize, T::MaxProviderLogo250X100Size>,
```

designdocs/provider_contexts.md
wilwade: Fixed. See other threads.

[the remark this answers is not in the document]
wilwade: Hmm... I'll have to think on this. It does feel like a future version of governance has a separate committee to handle this.

designdocs/provider_contexts.md
```diff
@@ -0,0 +1,231 @@
...
+	pub type ProviderToApplicationRegistryEntry<T: Config> = StorageDoubleMap<
+		_,
+		Twox64Concat,
+		ProviderId,
+		Twox64Concat,
+		ApplicationIdentifier<T>,
+		ApplicationRegistryEntry<T::MaxProviderNameSize, T::MaxProviderLogo250X100Size>,
+		OptionQuery,
+	>;
+	```
```
JoeCap08055, line 105: I appreciate the reuse of the `ApplicationRegistryEntry` structure, but I think they will end up diverging (either over time, or before this document is accepted...). `ProviderRegistryEntry` may end up containing more information regarding the Provider as a legal entity (similar to a DNS record), while `ApplicationRegistryEntry` would not necessarily evolve in such a way, being more or less a marketing distinction drawn by the Provider.

ex:
Provider `Meta` could/should reference information about the real-world ownership + "trustworthiness" of said entity.
Applications `Meta.Facebook`, `Meta.Instagram`, and `Meta.Threads`, being simply different product offerings of the owning entity, need have nothing other than name+icon (possible other descriptive fields, but entirely at the discretion of the Provider to describe).

wilwade: I think it is reasonable to at least alias or even duplicate the `ApplicationRegistryEntry` to a separate `ProviderRegistryEntry`. For clarity if nothing else.

I would sat that your namespacing is what `ProviderToApplicationRegistryEntry` does: `<ProviderId>.<ApplicationId>`

That said, I'm not sure the trustworthiness can remain at the Provider level. For your example, it would be important that `Meta.Threads` has the validated DNS: `threads.net` and that it not apply to other Meta applications.

I am trying (and perhaps this is part of the issue here) to avoid needing to have every Provider also complete the Application data. Given that I expect <50% need both. Perhaps a better way to handle this is flipping the naming around. Rename `ApplicationRegistryEntry` to `ProviderRegistryEntry`. Then an `ApplicationId` can "represent" themselves with a different `ProviderRegistryEntry` and it is a bit more clear that Application level is an override of the default for the Provider level.

designdocs/provider_contexts.md
```diff
@@ -0,0 +1,231 @@
...
+### Why Is This Needed?
+
+#### Concept of an Application Separate from a Provider
+
+Many companies have a single major product, but some companies have multiple applications that serve different needs.
+While each application could be a separate provider, this causes two issues: The sharing of capacity and the user permissions.
+
+Multiple providers require separate Capacity staking.
+As these applications would be from the same company, it is reasonable to say that they should share Capacity.
+Instead of building a Capacity sharing system, it is more reasonable to merely be the same Provider and provide a system of application identification.
```
wilwade, line 22:
```suggestion
```suggestion
Having separate providers would require a single company to have separate Capacity staking for each application.
As each these applications are still part of the same company, it is reasonable to say that those applications should share Capacity.
Instead of building a Capacity sharing system which has different trade-offs, it is reasonable for those applications to use the same Provider and provide a system of application identification that is separate from provider identification for these cases..
```
```

File: designdocs/provider_contexts.md [hunk 17]
```diff
@@ -0,0 +1,232 @@
...
+Reference Document: [SIWF Specification](https://projectlibertylabs.github.io/siwf/v2/docs/)
+
+### Why Is This Needed?
+
+#### Concept of an Application Separate from a Provider
+
+Many companies have a single major product, but some companies have multiple applications that serve different needs.
+While each application could be a separate provider, this causes two issues: The sharing of capacity and the user permissions.
+
+```suggestion
```
wilwade, line 20:
```suggestion
```

designdocs/provider_contexts.md
```diff
@@ -0,0 +1,231 @@
...
+
+### Why Is This Needed?
+
+#### Concept of an Application Separate from a Provider
+
+Many companies have a single major product, but some companies have multiple applications that serve different needs.
+While each application could be a separate provider, this causes two issues: The sharing of capacity and the user permissions.
+
+Having separate providers would require a single company to have separate Capacity staking for each application.
+As each these applications are still part of the same company, it is reasonable to say that those applications should share Capacity.
```
wilwade, line 21:
```suggestion
As each these applications are part of the same company, it is reasonable to say that those applications should share Capacity.
```

designdocs/provider_contexts.md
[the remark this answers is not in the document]
JoeCap08055: I do mention in my doc, that this kind of brings up the issue of whether we should "take the plunge" on at least a partial ENS implementation, which would handle the governance and ownership aspects of names/namespaces across all pallets/contracts.

designdocs/provider_contexts.md
JoeCap08055: If we anticipate that an Application may require a good deal of the same meta-info as a Provider for purposes of trustworthiness, then I think it's reasonable to clone/alias the info types for now.

designdocs/provider_contexts.md
```diff
@@ -0,0 +1,231 @@
...
+### Why Is This Needed?
+
+#### Concept of an Application Separate from a Provider
+
+Many companies have a single major product, but some companies have multiple applications that serve different needs.
+While each application could be a separate provider, this causes two issues: The sharing of capacity and the user permissions.
+
+Having separate providers would require a single company to have separate Capacity staking for each application.
+As each these applications are part of the same company, it is reasonable to say that those applications should share Capacity as a single legal entity.
+Instead of building a Capacity sharing system which has different trade-offs, it is reasonable for those applications to use the same Provider and provide a system of application identification that is separate from provider identification for these cases..
```
wilwade, line 22:
```suggestion
```

enddynayn (approved):
👍 Great! Thanks!

designdocs/provider_contexts.md
```diff
@@ -0,0 +1,232 @@
...
+1. `ProviderRegistryEntry` struct be renamed `ApplicationRegistryEntry`
+2. `ApplicationRegistryEntry` have the following properties:
+    ```rust
+    pub struct ApplicationRegistryEntry<T, U>
+    where
+        T: Get<u32>,
+        U: Get<u32>,
+    {
+        pub default_name: BoundedVec<u8, T>,
+        pub localized_names: BTreeMap<Vec<u8>, BoundedVec<u8, T>>,
```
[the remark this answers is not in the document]
wilwade: Yeah. I think that was the idea around the BTreeMap, but I didn't put any details in there.

designdocs/provider_contexts.md
```diff
@@ -0,0 +1,232 @@
...
+Provider Dashboard needs to be able to:
+
+- Create a provider without any application context other than the default provider context
+- Update the default provider context, logos, and translations
+- Add new Application Contexts with a identifier (unique to that provider), logos, and translations
+- Remove an existing Application Context
+- Update an Application Contexts with new logos, and translations
+
+
+## Non-goals
```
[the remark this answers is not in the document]
wilwade: Since Applications are still under a Provider, the capacity sharing is the default (and only) setup out of the box.

designdocs/provider_contexts.md
```diff
@@ -0,0 +1,276 @@
...
+
+### Why Is This Needed?
+
+#### Concept of an Application Separate from a Provider
+
+Many companies have a single major product, but some companies have multiple applications that serve different needs.
+While each application could be a separate provider, this causes two issues: The sharing of capacity and the user permissions.
+
+Having separate providers would require a single company to have separate Capacity staking for each application.
+As each these applications are part of the same company, those applications should share Capacity as a single legal entity.
```
enddynayn, line 21: As each "of" these...

designdocs/provider_contexts.md
```diff
@@ -0,0 +1,272 @@
...
+
+Separate providers per application would NOT align with the data security model of user-to-company that exists in the world today.
+Provider-level delegations are better than introducing application-level delegations.
```
enddynayn, line 24: Can you please help me understand why Provider-level delegations are better than application-level delegations?

designdocs/provider_contexts.md
```diff
@@ -0,0 +1,323 @@
...
+            pub localized_logo_250_100_png_bytes: BTreeMap<Vec<T::MaxLanguageCodeSize>, BoundedVec<u8, T::MaxProviderLogo250X100Size>>,
+        }
+    ```
+
+3. New `ProviderToApplicationRegistryEntry` storage be initialized:
+
+    ```rust
+    // Alias for clarity
+    type ApplicationIdentifier<T: Config> = BoundedVec<u8, T::MaxProviderNameSize>
+
```
saraswatpuneet, line 91: @wilwade do we need a special application identifiers? These are per provider basis so we can just use incremental number 0-Nth app? Or is there a utility tied to it

designdocs/provider_contexts.md
```diff
@@ -0,0 +1,272 @@
...
+SIWF [Signed Request Payload](https://projectlibertylabs.github.io/siwf/v2/docs/DataStructures/All.html) can expand the optional `applicationContext` value with a new optional `id` field that is the text of the `ApplicationIdentifier` on chain.
+This would allow a smooth transition between `applicationContext.url` and `applicationContext.id` with both being optional.
+
+The Wallet would then:
+
+1. Verify the SIWF Signed Request.
+2. Lookup the Provider via the `publicKey` in the SIWF Signed Request.
+3. If any, fetch the Application Identifier from Frequency.
+4. Display the information from the `ApplicationRegistryEntry` (or the Provider Registry Entry `ProviderToRegistryEntry` if there is no application context identifier) to the user to help them know who they are authorizing.
```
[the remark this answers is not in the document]
saraswatpuneet: Thinking through this it would make sense for user be aware to the suite of applications they maybe granting delegation to, reminding me of how Disney used ToS of their app against a case incident in their park, but that would be a ux nightmare to show all apps to users

Also users might find it spooky that they are on an app saying "My family friendly app" but they now see "My adult app" asking for same permissions

JoeCap08055:
I think I've come around to the idea of keeping Delegations at the Provider level:

* This mirrors the real-world trust model of users trusting corporate entities, not apps, with their data
* Delegations allow that corporate entity to take action on the user's behalf using their Capacity; such actions (currently) contain no indication of the Application used as the conduit, so it doesn't matter which application actually processed the action
* Separating the Delegatee from the Capacity User involves more complexity than I first realized, which, given the above points, doesn't seem worth it given it's not really needed

There's still a small tickle in the back of my mind that a use case could arise which would make the indication of _which Application_ performed an action significant. However, as such a use case would necessitate an evolution of both Frequency and DSNP, I'm satisfied that it can be left to a future discussion if necessary.

designdocs/provider_contexts.md
```diff
@@ -0,0 +1,339 @@
...
+### **Changes and additions in extrinsics** <a id='extrinsics'></a>
+
+1. The `propose_to_be_provider` extrinsic will now accept an optional list of hashes for images/logos to be approved by governance.
+
+    ```rust
+        #[pallet::call_index(0)]
+        pub fn propose_to_be_provider(
+            origin: OriginFor<T>,
+            provider_name: Vec<u8>,
+            logo_hashes: Vec<[u8; 32]>, // blake2_256 hashes of the logos
```
saraswatpuneet, line 135: Question: do we also take a list of localized names to be approved by goverance? or create a typed payload like `ProviderRegisterEntry` but with hashes instead of image bytes?

designdocs/provider_contexts.md
```diff
@@ -0,0 +1,354 @@
...
+            /// A list of pre-approved logo image hashes (blake2_256).
+            pub logo_hashes: BoundedVec<[u8; 32], T::MaxApprovedLogoCount>,
+
+            /// Localized translations of the name, keyed by BCP 47 language code.
+            pub translations: BoundedBTreeMap<
+                BoundedVec<u8, T::MaxLanguageCodeSize>, 
+                BoundedVec<u8, T::MaxProviderNameSize>, 
+                T::MaxTranslations,
+            >,
+        }
```
saraswatpuneet, line 146: Proposed payload to register provider/application, this is minimal to get a bulk approved and update later

designdocs/provider_contexts.md
```diff
@@ -0,0 +1,354 @@
...
+
+    Each registry entry supports:
+    - A default name and logo.
+    - Localized names and logos (up to 150 locales).
+    - PNG logo support (250x100, up to 128 KiB).
+    - BCP 47-compliant language codes (e.g., en-US, fr-FR).
+2. Updated `ProviderRegistryEntry` struct have the following properties:
+
+    ```rust
+        pub struct ProviderRegistryEntry<T::Config> {
```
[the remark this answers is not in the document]
saraswatpuneet: Did we consider setting `url` but I guess that does not necessarily mean ownership of brand and logo things

designdocs/provider_contexts.md
```diff
@@ -0,0 +1,331 @@
...
+            pub default_logo_250_100_png_bytes: BoundedVec<u8, T::MaxProviderLogo250X100Size>,
+            pub localized_logo_250_100_png_bytes: BTreeMap<Vec<T::MaxLanguageCodeSize>, BoundedVec<u8, T::MaxProviderLogo250X100Size>>,
+        }
+    ```
+
+3. New `ProviderToApplicationRegistryEntry` storage be initialized:
+
+    ```rust
+    // Alias for clarity
+    type ApplicationIdentifier<T: Config> = BoundedVec<u8, T::MaxProviderNameSize>
```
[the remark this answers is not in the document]
saraswatpuneet: We can duly do both? A metric worth collecting, total apps and give 0..N per provider. 

Not sure if there is any utility for providers to know their first app is some rth app on frequency

designdocs/provider_contexts.md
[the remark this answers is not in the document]
saraswatpuneet: Was the support for 30 locales discussed somewhere? Not sure if this number was introduced by me

designdocs/provider_contexts.md
```diff
@@ -0,0 +1,353 @@
...
+
+SIWF [Signed Request Payload](https://projectlibertylabs.github.io/siwf/v2/docs/DataStructures/All.html) can expand the optional `applicationContext` value with a new optional `id` field that is the text of the `ApplicationIndex` on chain.
+This would allow a smooth transition between `applicationContext.url` and `applicationContext.id` with both being optional.
+
+The Wallet would then:
+
+1. Verify the SIWF Signed Request.
+2. Lookup the Provider via the `publicKey` in the SIWF Signed Request.
+3. If any, fetch the `ApplicationIndex` from Frequency.
+4. If present, fetch the `ProviderToApplicationRegistryEntry` for that (ProviderId, ApplicationId) pair. Otherwise, fetch the default `ProviderToRegistryEntry` for the ProviderId
```
saraswatpuneet, line 285: Do we just keep provider side registry to contain just the name and empty for others and instead create a `default` application with id:0 when a provider is registered?

 that way there is atleast one application is   for providers and mutations can be limited to applications root and keep `ProviderToRegistry` immutable

designdocs/provider_contexts.md
```diff
@@ -0,0 +1,401 @@
...
+            localized_names: BTreeMap::new(), // Initially empty, can be updated later
+            default_logo_250_100_png_bytes: BoundedVec::default(), // Initially empty, can be updated later
+            localized_logo_250_100_png_bytes: BTreeMap::new(), // Initially empty, can be updated later
+        };
+
+        ProviderToApplicationRegistryEntry::<T>::insert(provider_id, application_identifier, application_entry);
+    ```
+
+5. `propose_to_add_application` will also insert the logo hashes into the `ApprovedLogoHashes` storage map.
+6. Introduce a new extrinsic to `update_name` for updating an existing application context (post goveranance registration) provided with new name  `ProviderRegistryEntry` with the `ApplicationIndex`.
```
saraswatpuneet: For application updates post governance, we introduce 3 extrinsics to update `logo` and `localized logos` , for name and local updates one has to go via `governance` like before

designdocs/provider_contexts.md
wilwade:
```suggestion
For example, imagine if Alice is delegating to an application, Bar. Bar application is owned by the Foo Company which also owns and controls Baz. The Foo Company is able to collect all data shared with both the Bar and the Baz. A delegation to just Bar or just Baz doesn't limit the information sharing between those Applications because the data agreement is with Foo Company.
Additionally, to limit the impact on the network, the Bar and Baz Applications are able to share capacity and perform batching together.
This leads to the conclusion that Provider-level delegations are better than introducing new Application-level delegations.
```

wilwade: @enddynayn how does that work?

designdocs/provider_contexts.md
```diff
@@ -0,0 +1,354 @@
...
+        ProviderRegistryEntry<T>,
+        OptionQuery,
+    >;
+    ```
+
+4. `MaxProviderNameSize` be increased to `256`.
+5. `MaxProviderLogo250X100Size` be created and the limit set to `131_072` (128 KiB).
+6. Introduce hash-based logo approval mechanism:
+    Logos are not embedded during proposal submission. Instead, a blake2_256 hash of the logo image is included in the proposal.
+    Governance must explicitly approve these logo hashes as part of the provider registration process.
```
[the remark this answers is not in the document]
wilwade: I would think if we didn't store on-chain, then it would be an IPFS CID.

designdocs/provider_contexts.md
wilwade: I was thinking just incremental ids. Names could change (Twitter -> 𝕏).

The application name doesn't have to be unique from a chain perspective, but I cannot see to many reasons for duplicates... Or well if you are trying out different branding you might create a second app context with the new logo and then you can roll it out slower.

designdocs/provider_contexts.md
wilwade: I do like the idea of an "application 0" being the default. `ProviderToRegistry` could be more minimal. hmm... I'm open to that solution. Aka all Providers have at least one application. Makes the data storage simpler for sure.

saraswatpuneet: using a hash table `ApprovedLogos` makes it easy now, since `ProviderRegitry` would not contain raw bytes of logos but hashes and a second call to logo tables can resolve hash to image for users, so I believe we dont need a application 0 if fallback is `ProviderRegistry`

designdocs/provider_contexts.md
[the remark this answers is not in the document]
saraswatpuneet: @wilwade do we need to shard with provider id, would just a hash table for logo/vs bytes would suffice? since ProviderRegistry would have the hash at time of provider/app registration. But if there is reason we would want to do ProviderId, Hash, Logo type doublemap?

designdocs/provider_contexts.md
enddynayn: Perfect! Thanks!

designdocs/provider_contexts.md
```diff
@@ -0,0 +1,379 @@
...
+
+    The approved hashes are recorded in a dedicated storage map:
+
+    ```rust
+        #[pallet::storage]
+        #[pallet::getter(fn approved_logo_hashes)]
+        /// Approved logo hashes for providers.
+        /// This map stores the blake2_256 hash of the logo as the key and an empty tuple as the value.
+        pub ApprovedLogoHashes: StorageMap<
```
wilwade, line 122: I don't think we need the `ApprovedLogoHashes` storage do we? Instead the updating of the `ApprovedLogos` requires the path for the lookup in the `ProviderRegistryEntry` right?

designdocs/provider_contexts.md
```diff
@@ -0,0 +1,379 @@
...
+
+    Each registry entry supports:
+    - A default name and logo (hash of the logo image).
+    - Localized names and logos (up to 150 locales).
+    - PNG logo support (250x100, up to 128 KiB). Note that the logo is not embedded in the proposal but rather a hash of the logo image.
+    - BCP 47-compliant language codes (e.g., en-US, fr-FR).
+2. Updated `ProviderRegistryEntry` struct have the following properties:
+
+    ```rust
+    type LogoHash = [u8; 32]; // blake2_256 hash of the logo
```
wilwade, line 82: What if LogoHash was instead the CID? Stay with me here, as I'm not sure this works:

1. LogoHash is the CID. This way when submitted to the chain for approval, there is a simple way for the approval process to check the image.
2. (Not sure if this works) Submission of the image to be added to `ApprovedLogos` would need to check that the image matches the given CID. (I think this would require making some CID assumptions or documenting the defaults?)

designdocs/provider_contexts.md
```diff
@@ -0,0 +1,379 @@
...
+        pub localized_names: BTreeMap<BoundedVec<u8, T::MaxLanguageCodeSize>, BoundedVec<u8, T::MaxProviderNameSize>>,
+        pub default_logo_250_100_png: BoundedVec<u8, T::MaxProviderLogo250X100Size>,
+        pub localized_logo_250_100_png: BTreeMap<BoundedVec<u8, T::MaxLanguageCodeSize>, BoundedVec<u8, T::MaxProviderLogo250X100Size>>,
+    }
+
+    // The rpc should fetch the data from `ProviderToApplicationRegistryEntry` and construct the ApplicationContext with logos from `ApprovedLogos`.
+    #[rpc]
+    pub fn get_provider_application_context(
+        provider_id: ProviderId,
+        application_index: ApplicationIndex,
```
wilwade, line 236: As this could get large , I wonder if we should have a `locale` or locale version as well. That way the output could be simplified and doesn't need to have potentially 150 different logos pulled just to get the one needed.

designdocs/provider_contexts.md
```diff
@@ -0,0 +1,379 @@
...
+    ) -> Option<ApplicationContext<T>>;
+    ```
+
+2. Optionally, introduce "get_provider" RPC method to include the new fields for localized names and logos and return a provider registry entry.
```
wilwade, line 240: I like the idea of one that pulls the info for the `application` and `locale` and one that just pulls all the data.

wilwade (commented):
Few final thoughts and questions, but I think we are good to merge, make issues, and figure out the rest of it along the way.

designdocs/provider_contexts.md
saraswatpuneet: Agreed, its easier to have it entered in ApprovedLogos and set incoming logos directly against a preset hash

designdocs/provider_contexts.md
saraswatpuneet: This is doable and we already have some basic CID validation in `messages` for IPFS, my only question here is that is it ok to?

1. Provider may pin a logo and get CID
2. Proposes application with CID as logo hash
3. Upon approval submits a logo with cid but it might be a different file than originally pinned to bypass some governance restrictions 

If we get a CID is it possible to have some offchain worker that creates an offchain index and pull the image from CID and compute a hash of actual image for verification but i guess that might be an overkill

designdocs/provider_contexts.md
saraswatpuneet: Yep added an optional locale for such case for both RPCs

designdocs/provider_contexts.md
wilwade: As long as we are able to validate the CID on chain, then the file must be the same. As the CID has a sha2-256 hash "inside" of it. Ok not quite inside of it and this is where the question sits. Since it isn't just the bytes being hashed: https://docs.ipfs.tech/concepts/content-addressing/#cids-are-not-file-hashes

saraswatpuneet: Re imagined LogoHash -> LogoCID,  atleast on the messages side we don't do a full CIDv1 validation, here we can recompute on file upload ( not sure if that would be possible) .

Following same rules around CIDv1 as in our message design.

designdocs/provider_contexts.md
```diff
@@ -0,0 +1,383 @@
...
+        }
+    ```
+
+2. `propose_to_be_provider` to will insert hashes into the `ApprovedLogos` storage map.
+3. Introduce a new extrinsic `propose_to_add_application` which work in similar way to `propose_to_be_provider` but will be used for adding or updating application contexts. It will accept a `ProviderRegistryEntry` struct with the application name, localized names, and logo hashes.
+
+    ```rust
+        #[pallet::call_index(1)]
+        pub fn propose_to_add_application(
+            origin: OriginFor<T>,
```
saraswatpuneet, line 159: @wilwade should we separate add and update as separate extrinsics, assuming we can handle both with an optional application index 

On similar lines do we enable updating provider data via governance in same fashion?

wilwade: Hmm....

1. I think we can just make them set functions. Simplifies the setup.
2. Re-Provider Data. This makes me wonder again about an `Application 0` default instead of having Provider Data and Application data. Is there any other way we could unify these? Perhaps a `propose_to_set_provider_application` that takes an optional application id and if no application id it sets the provider level? Not sure if that makes it more confusing over having two extrinsics or not.

wilwade: We can see how that works, but we do want to be able to update the provider information as well so not sure it gets us far unless we just "don't" have provider information and only have application information.

saraswatpuneet: That's true, it's better if we can just follow same model on both applications and provider side for issuing updates that way it's more nuanced even with some extra code.

I will capture that in this document

aramikm (approved):
LGTM

saraswatpuneet (approved):
Addressed most technical details and decisions, design may change during implementation.

PR #2312, opened 2025-04-04
Design Doc: Schemas, Intents, and Delegation Groups

### LINKED ISSUE #2265: [Discussion]: Schemas v2
### Feature Description

There are various ideas about how to improve Schemas. Let's discuss and find a way forward on it.

- Schemas have three implied and real pieces
    - Protocol namespace (`dsnp` and `frequency` on Mainnet)
    - Schema (`dsnp.update` for example)
    - Schema Version (`dsnp.update` v1 and v2 for example)
- Delegations are at the Schema Version level, the other levels are merely structure
- Protocol authors cannot upgrade independent of new schema version and asking each user to approve

### Concerns/Tradeoffs
- How can users be protected from bad schema upgrades?
- How can users not be bothered by schema upgrades that have only structural differences?
- How can the delegation model be improved/changed to handle the multi-layer schema structure (specifically different versions)?
- How can the interaction model with schemas be unified and simplified (even at the expense of complexity in schema creation)?
- How can the Ethereum integration work align with these changes?

JoeCap08055:
# Description

In the current implementation, schemas are registered with immutable numeric identifiers (`SchemaId`) and describe the layout and storage semantics (e.g., Avro/Parquet formats, on-chain/off-chain storage). These schema IDs are used as references by clients and runtime modules alike, particularly in the delegation system defined by the `msa` pallet.

Delegations currently allow a user to authorize a provider (e.g., an app or service) to act on their behalf, but this authorization is tightly bound to a specific `SchemaId`. This model has proven limiting in several ways:
-   **Coupling between specific schema versions and delegations**
-   **Schemas represent data format, not purpose**
-   **Lack of human-readable context**

These limitations have motivated a re-architecture of the schema and delegation systems to introduce the concepts of:
-   **Named intents** with schema version tracking
-   **Intent-based delegation**
-   **More expressive APIs and storage models**

Closes #2265


JoeCap08055:
Additional question/discussion point:

(Thinking mostly about on-chain storage like `Itemized` and `Paginated`, but could maybe also apply to off-chain storage locations like `IPFS`)

Currently, on-chain storage is stored in a storage child trie with the specific _schema version id_ as part of the key. This made complete sense given that, at the time, there was no way to associate on schema version with another as being representative of minor variations of the same data format.

However, with the advent of an encapulating _Schema_ for which individual _Schema Versions_ are merely minor format changes, would it make sense to reorganize on-chain storage such that the on-chain storage references the _Schema_, rather than the _Schema Version_, and simply includes the _Schema Version_ as a record header? This would have the following effects/benefits/drawbacks:
* Increase item/page overhead in storage by 2 bytes (`u16`) per item (`Itemized`) or page (`Paginated`) stored
* Make schema version upgrades simpler and avoid hefty migrations and other edge cases. For example:
    * Currently, for even a minor/trivial format change that requires publishing a new schema version, the following scenarios must be accounted for:
        * Which, of multiple Providers, bears the responsibility of migrating a user's data from the old schema version to the new one?
        * How to avoid race conditions in migrating a user's data between schema versions?
        * It is possible that a user's data could become bifurcated; that is, new records start to be written to the latest schema version before the old records in the prior version have been migrated; how then to know what the proper state is?
        Example (using a user's graph as a proxy for the generic on-chain storage):
            * User 'Alice' has delegations to Provider A. Alice's graph contains Bob and Charlie in Schema Version 1.
            * Schema version 2 is published
            * Alice onboards with Provider B, who requests a delegation for Schema Version 2. Provider B's application does not handle older schema versions.
            * Alice adds Dave to her graph via Provider B, who writes it to a new page in Schema Version 2.
            * Now, Alice has graph pages in both Schema Versions 1 & 2. Does 2 supersede 1? Or does 1 still need to be merged into 2?
            * The onus, therefore, is on the Provider app to always check to see if there is a new (empty) schema version for the graph data, and either (a) avoid writing new data until migration is complete, or (b) perform a migration of the user's entire graph before writing new data. This is difficult (at the least expensive) to enforce on-chain.
            * TL;DR: Currently, the only way to know if a user's graph has been upgraded is the presence/absence of data in the new schema version. If *any* data exists in the new version, it is impossible to know if that was the result of a full migration of the user's graph from the older version, or a one-off update/write to the new version. Requiring this scenario to be managed by a Provider app is unrealistic.
                       
            Now, if on-chain storage were indexed by _Schema_ instead of _Schema Version_, we would get something like the following:
            * Alice has Bob & Charlie in her graph on Page 1. Page 1 is written with Schema Version 1, for the schema `dsnp.private_follows`
            * Schema Version 2 is published for `dsnp.private_follows`
            * Alice adds Dave to her graph via Provider B.
            * Provider B reads the graph according to `dsnp.private_follows` and decodes according to schema version 1 (from the page header).
            * Provider B writes a new Page 1 containing Bob, Charlie & Dave, using Schema Version 2 (and indicating so in the page header).
            * Other graph pages may still be using Schema Version 1, until either opportunistically rewritten during a graph update, or a Provider may choose to proactively update all pages for a user.
            * TL;DR: In this model, migrations to new schema versions may be performed opportunistically, en masse, or not at all, at the Provider's discretion. (Granted, the prior scenario still exists if there is a _major_ (ie, material) change to the data format such that a new `Schema` (not Schema _Version_) is published, but this scenario should occur much less frequently.)
            
            
            
Thoughts?

designdocs/schemas_protocols_intents.md
```diff
@@ -0,0 +1,642 @@
...
+Each protocol is a unique identifier, typically human-readable (e.g., `"dsnp"`), mapped to an on-chain `ProtocolId` (
+e.g., `ProtocolId = 2`).
+
+- Every schema must be published under a protocol: `protocol.schema`
+- Intents may also be published under a protocol _(open question, discussed [below](#section_7))_
+
+### 👤 Ownership and Control
+
+Protocols are envisioned to be owned by an entity, but there is an open question as to whether this entity should be an
+MSA (i.e., a Provider) or a raw account (i.e., a public key). This decision will impact how access is managed,
```
enddynayn, line 102: One of the advantages of MSAs is that, in theory, we can attach permissions to them. This would allow governance to add keys with limited access, which is especially valuable for organizations that need to delegate responsibilities. Being able to grant keys with restricted functionality becomes even more important in these contexts.

aramikm:
@JoeCap08055 I can see that might be helpful if we consider the abstraction and generic usecase of stateful storage but I can not see why we might want to change the graph related schemas. 

My 2 cents are that if we don't have to do it we probably should not try. The reason is that even though it might be beneficial for generic case it's problematic for graph which is 99% of the actual use of the stateful storage for us.

We would need a deeper discussion about this.

designdocs/schemas_protocols_intents.md
```diff
@@ -0,0 +1,642 @@
...
+    pub revoked_at: Option<BlockNumber>,   // Replaces `expiration` and `revoked`
+    pub granted_at: BlockNumber,
+}
+```
+
+And:
+
+```rust
+pub enum DelegationTarget {
+    SchemaVersion(SchemaVersionId),
```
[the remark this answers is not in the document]
JoeCap08055: Actually `SchemaVersionId` is what current Frequency calls `SchemaId` and is just a `u16`

designdocs/schemas_protocols_intents.md
```diff
@@ -0,0 +1,642 @@
...
+}
+```
+
+Delegation lookup is flattened into:
+
+```rust
+pub type Delegations<T> = StorageDoubleMap<
+    _, Blake2_128Concat, MsaId,           // Delegator
+    Blake2_128Concat, ProviderId,         // Provider
+    BoundedVec<DelegationInfo, MaxDelegationsPerProvider>
```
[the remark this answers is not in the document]
JoeCap08055: There are 2 things to consider here:

1. If we ignore `DelegationInfo`, and simply store delegations the way we currently do, using a `BoundedBTreeMap`, with a `DelegationTarget` enum key instead of a `SchemaId` (u16), here are some back-of-napkin calculations:
    - current max storage for 30 schema delegations (current per-delegation max): ~180 bytes
    - max storage for 30 delegations of `DelegationTarget`: ~211 bytes
    
2. However, our current delegation storage has some deficiencies; namely, we can know that a Schema is currently delegated, or when it was revoked, but we cannot know when that delegation started (because we only store a revoked block number or zero). The new `DelegationInfo` struct allows us to store that information. To store 30 delegations (we probably would need to use `BoundedVec` instead of `BoundedBTree` due to the way we would need to do lookups), would cost ~361 bytes, so effectively double the storage we currently use for delegations.

designdocs/schemas_protocols_intents.md
```diff
@@ -0,0 +1,642 @@
...
+- **On-Chain Efficiency**
+
+## 3. Glossary of Terms <a id="section_3"></a>
+
+- **protocol** - the top level of the tripartite nomenclature _protocol.schema@version_ (ex: _dsnp_ is the protocol in
+  _dsnp.broadcast@v2_)
+- **schema** - the second level of _protocol.schema@version_ (ex: _dsnp.broadcast_ resolves to a specific schema).
+  Schemas should always be referred to by their fully-qualfied name (ie, _dsnp.broadcast_, not just _broadcast_)
+- **version** - the third level of _protocol.schema@version_, resolves to a specific _minor_ iteration of a schema
+- **intent** - an on-chain primitive representing an abstract action or operation
```
JoeCap08055, line 51: Looking for some comments on naming.

I was initially in favor of: Namespace->Protocol->Schema; however, apparently "Protocol" for the 2nd level was rejected...?

@wilwade suggested Protocol->Schema->SchemaVersion. However, already in conversations it's clear that Schema vs. SchemaVersion is very confusing.

I think perhaps we should keep the third level as it currently is, `Schema`, to avoid renaming/cognitive shift. I've got some suggestions here for the other levels:

Level 1: Namespace, Domain, Package, Module, Scope
Level 2: Model, Template, Component

Happy to consider others...

JoeCap08055: Hmm... what if we did:
1. Domain -> analagous to ENS Domain; will be an easy conversion if we ever implement/move to ENS
2. MetaSchema
3. Schema

designdocs/schemas_protocols_intents.md
```diff
@@ -0,0 +1,642 @@
...
+- `Versions`: Ordered list of `SchemaVersionId`s (e.g., v0 → v1 → v2)
+- Metadata: Optional description, links, etc.
+
+### 🧬 Versioning Rules
+
+- Versions are stored in a monotonic, gapless array
+- A new schema version ID is automatically assigned the next version
+- Older versions are preserved permanently
+- Delegations may apply to the schema, not a specific schema version ID
+- _Question: can versions be deprecated?_
```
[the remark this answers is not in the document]
JoeCap08055: I believe that's currently a config value & likely will not change, but I'll add a mention of it.

designdocs/schemas_protocols_intents.md
```diff
@@ -0,0 +1,642 @@
...
+
+### 🧭 Governance Rules
+
+- New schemas must be approved by governance
+- Minor updates (new schema version) may be published by the protocol owner
+- Major updates require a new schema (e.g., change in semantic intent)
```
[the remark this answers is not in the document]
JoeCap08055: Since this doc is being revised to defer the concept owner publishing capabilities and make everything a governance action (for now), the major/minor distinction will be validated by governance.

Whether a programmatic determination can ever be made in this regard may determine the fate of an ownership publishing (without governance) capability.

designdocs/schemas_protocols_intents.md
JoeCap08055: Interesting. Probably the subject of a separate design or issue, but the idea of adding a control key with explicit/limited rights is intriguing... though, delegations could accomplish the same thing.

designdocs/schemas_protocols_intents.md
```diff
@@ -0,0 +1,642 @@
...
+  *either** a valid intent **or** schema delegation is found.
+- The delegation resolution policy (e.g., precedence, fallback order, or required combinations) must be explicitly
+  defined by the runtime and documented for developers.
+
+This multi-path resolution allows greater flexibility, but requires care to avoid ambiguity and ensure consistent
+behavior across different APIs and use cases.
+
+### 🧪 Additional Considerations
+
+- In the future, extrinsics may support intent-based authorization **without referencing a schema** at all. These would
```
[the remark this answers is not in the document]
JoeCap08055: We're going to defer the concept of `Intents` for now; I'll separate it out to a separate issue for tracking.

designdocs/schemas_protocols_intents.md
```diff
@@ -0,0 +1,366 @@
...
+- `ProtocolId`: Which Protocol owns this Intent
+- `IntentId`: Compact numeric ID used on-chain
+- `Name`: Human-readable name (e.g., `broadcast`)
+- `Description`: Optional developer-facing documentation
+- `Link`: Optional link to spec or docs
+- `RegisteredBy`: MSA (ProviderId) that registered the intent
+- `RegisteredAt`: Block number
+- `Versions`: [optional] Ordered list of `SchemaId`s (e.g., `[7, 15, 27]`, for versions 0..2)
+
+### 🧬 Versioning Rules
```
wilwade, line 106: What if we dropped versions and instead just did schema deprecation tagging?

JoeCap08055: Not sure I understand. Elaborate?

wilwade: I have a bit more here: https://github.com/frequency-chain/frequency/pull/2312#discussion_r2219482941

Versions imply a semantic that we might not want to imply. Also some protocols might use other ways of versioning than int incrementing.

Effectively with Intents, we can just say "here are all the schemas for this intent and these ones are flagged as deprecated"

The schema naming we do could also be updated. With named Protocols and Intents, the only reason to name schemas is for ease of developer integration across mainnet and testnets.

designdocs/schemas_protocols_intents.md
```diff
@@ -0,0 +1,366 @@
...
+}
+
+pub type Intents<T> = StorageMap<
+    _, Blake2_128Concat, IntentId, IntentInfo
+>;
+
+// Q: Is this necessary as a separate map, or can be folded into the `IntentInfo` struct?
+pub type IntentVersions<T> = StorageMap<
+    _, Blake2_128Concat, IntentId, BoundedVec<SchemaId, MaxSchemaVersionsPerIntent>
+>;
```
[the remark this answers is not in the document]
saraswatpuneet: I see, so we are `BoundedVec<SchemaId, MaxSchemaVersionsPerIntent>` which are schema ids, this could use some renaming from `IntentVersions`, it could be a common state call for schema management

designdocs/schemas_protocols_intents.md
```diff
@@ -0,0 +1,379 @@
...
+
+2. **Submit by SchemaId + IntentId (Explicit):**
+
+    - Enables precision delegation checks
+    - Slightly more complex client integration
+
+3. **Submit by IntentId only (No Schema):**
+
+    - Enables non-schema-based actions (e.g., future custom actions)
+    - Runtime must validate delegation by `IntentId`
```
saraswatpuneet, line 290: from blockchain perspective, keeping things minimal, why we should support multiple ways to publish other than schemaId? that more granular for the chain and can enforce some provider side management or a library from us for schema management

JoeCap08055: For clarification:

You _publish_ to a `SchemaId`; you _delegate_ to an `IntentId`. `Schemas` implement `Intents`. After thinking about it some more, I realized that any extrinsic that publishes to a Schema already has to read the `SchemaInfo` metadata to do some basic validation. So if the `IntentId` is part of that metadata, we don't even need an extra read to look it up to validate the delegation. Thus our existing extrinsic model of just passing `SchemaId` can remain.

Publishing to a schema on-chain, then, would look like:
* Read `SchemaInfo` by `SchemaId`
* Basic validation (ie, storage location)
* Get `IntentId` from `SchemaInfo`
* Query delegations for `IntentId` just as we currently would for `SchemaId`

JoeCap08055: I'll update the doc to make ☝🏻 more clear

JoeCap08055:
Additional simplification:

Since we've already removed the concept of ownership (deferred until future implementation of ENS and DAO on Frequency), the concept of a "Protocol" or "Namespace" as a primary entity seems superfluous. It can easily be introduced later with minimal effort if we want, but for now it seems sufficient to keep a model similar to the current on-chain storage for schema names; ie  a double map `protocol/namespace` -> `intent name` -> `IntentId`

Thoughts?

designdocs/schemas_protocols_intents.md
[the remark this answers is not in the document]
shannonwells: New version makes a lot more sense to me, too.

shannonwells:
### Schema versions
I think there needs to be a way to reject messages for unsupported versions, i.e. ones that are past "deprecation".  If app developers don't have to upgrade, it's guaranteed that some never will, and it seems that for social graph in particular, that will cause problems. I would want to see another field indicating whether the schema version is still valid for the protocol, and Gateway could even reject messages for that version if it's no longer supported. Maybe that feature doesn't need to be part of this design doc though.  I thought we had discussed schema states at some point, like - current --> deprecated --> readonly.

In your example, what if Provider A just insists on updating the graph using Schema v1 while everyone else is using Schema v2?  Then I guess it's left to the user to figure out how to pressure Provider A to upgrade to the new version, or withdraw their delegation & stop using the app?

### Governance trust
This is getting even more off-piste, but I think it would be a good idea to push out announcements for these types of changes to end users. At first I thought should be a DSNP message type that is basically a system announcement (e.g. ProtocolChange, SchemaUpgrade, etc) but really, a Broadcast would be fine, and wouldn't require Providers to make any code changes.

So Frequency Foundation can have a Provider account and post messages that can be picked up by any other Provider.  Those users can go and vote on it (if they have rewards) or at least look at it, and decide what to do, if anything.

designdocs/schemas_protocols_intents.md
```diff
@@ -0,0 +1,334 @@
...
+
+For the `messages` pallet, it _may_ be possible to avoid a migration by bifurcating message storage at a specific block
+number, ie, content in blocks before the upgrade are kept in the existing storage keyed by `SchemaId`, while new content
+is stored indexed by `IntentId` and containing a `SchemaId` in the payload. Further analysis is required to determine if
+this approach is feasible or desirable.
+
+For the `stateful-storage` pallet, it would be necessary to re-write all storage pages to include the concrete
+`SchemaId` in the page header. How to accomplish this for ~1M user graphs and graph keys on-chain requires further
+analysis.
+
```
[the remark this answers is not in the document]
JoeCap08055: It's not *specifically* graph data; that's merely the main data that currently uses that pallet. In future, I imagine other sorts of data will also use `stateful-storage`; the point is on-chain data is expensive to migrate, so minimizing the impact of minor storage format changes so that migrations aren't required seems like a valuable change to make early on.

designdocs/schemas_protocols_intents.md
```diff
@@ -0,0 +1,421 @@
...
+The <a href="#delegation_group_struct">DelegationGroup structures</a> will be stored as follows:
+
+```rust
+#[pallet::storage]
+pub(super) type DelegationGroups<T: Config> =
+StorageMap<_, Twox64Concat, DelegationGroupId, DelegationGroup, OptionQuery>;
+```
+
+#### Name Registry
```
JoeCap08055, line 406: One note for discussion:

Instead of a monolithic name registry, we could define `Protocol` as a first-level storage entity with an ID and a name mapping name->ProtocolId; then define separate maps (ProtocolId, Descriptor) -> IntentId, and (ProtocolId, Descriptor) -> DelegationGroupId

Not sure what that buys us, but it's an option.

designdocs/schemas_protocols_intents.md
```diff
@@ -0,0 +1,421 @@
...
+
+### 13. **Storage**<a id="storage"></a>
+
+#### Schemas
+
+The storage structures for Schemas will not fundamentally change, except for internal changes to the `SchemaInfo`
+structure that will require migration:
+
+* `payload_location`  & `settings` will migrate to `IntentInfo`
+* addition of `intent_id`
```
saraswatpuneet, line 380: 👍🏽

saraswatpuneet (commented):
Looks good!

nit: a section for rpcs or runtime api, if any, would be useful 

Edit: noticed a separate section for runtime calls, looks good 👍🏽

saraswatpuneet (approved):

designdocs/schemas_protocols_intents.md
JoeCap08055: One thing it buys is smaller, cheaper extrinsic payloads when creating/updating Intents and DelegationGroups--instead of passing both payload and descriptor names, could pass just `ProtocolId` and descriptor name.

designdocs/schemas_protocols_intents.md
```diff
@@ -0,0 +1,445 @@
...
+as follows:
+
+1. Create a new `Intent` for every currently existing `Schema`, as follows:
+    1. The new `Intent` will have the same numeric ID value as the original Schema.
+    2. The new Intent shall inherit the `payload_location` and `settings` from the existing `SchemaInfo` object.
+2. Migrate existing schema name mappings as follows:
+    1. For each `SchemaNamespace` '\<protocol_name>'
+        1. For each `SchemaDescriptor` '\<name>' at index `n` belonging to a '\<protocol_name>'
+            * Create a new name mapping in the `NameRegistry` as `<protocol_name>.<name>_n` to `Intent(id)`
+3. Store the `messages` pallet cutover block number
```
JoeCap08055, line 445: Optionally could seed the chain with some pre-defined Delegation Groups as part of the migration.

designdocs/schemas_protocols_intents.md
```diff
@@ -0,0 +1,334 @@
...
+- Older versions are preserved permanently
+- Delegations apply to an Intent
+- _Question: can Schemas (particular versions) be deprecated?_
+
+### 🧭 Governance Rules
+
+- Intents must be approved by governance, and are immutable except for appending new SchemaIds to the version list
+- New Schemas must be approved by governance
+- Minor Schema updates (semantic-preserving format changes, etc) may be approved for the same Intent
+- Major updates (change in meaning or semantics, or significant breaking format change) require publishing under a new
```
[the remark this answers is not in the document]
JoeCap08055: Correct. A better/more standardized mechanism for this could be the subject of a future design discussion; for now, it would add too much complexity to an already large-ish undertaking.

designdocs/schemas_protocols_intents.md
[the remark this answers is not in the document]
JoeCap08055: This being tracked here: #2510

designdocs/schemas_protocols_intents.md
JoeCap08055: Tracking here: #2534

designdocs/schemas_protocols_intents.md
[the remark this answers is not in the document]
JoeCap08055: Tracking here: #2534

enddynayn (approved):
It is clear a lot of thought went into this design document. I agree that having governance as the ultimate authority for creating and modifying Intents and Delegation Groups is the right approach and key to all this.

🚀 🚀

PR #2314, opened 2025-04-05
fixes double charge, undo changes to increase balances in e2e tests

shannonwells:
# Goal
Fixes the double-charge problem and undoes some of the hacks used to make some e2e tests pass.

Followed the model of ChargeTransactionPayment and created a `can_withdraw_fee` function in the Nontransferable trait. Then this is used by `validate` and `withdraw_fee` is only used in `pre_dispatch`.

Following the guidance/direction from Parity, we do not perform the validation steps twice so  `withdraw_fee` doesn't call `can_withdraw_fee`.

This also required adding a `replenishable_balance` to the Nontransferable trait, which isn't  what I'd prefer but the trait generics required it, otherwise it was a mismatch of Balance types.

### DIFF
--- a/e2e/scaffolding/extrinsicHelpers.ts
+++ b/e2e/scaffolding/extrinsicHelpers.ts
@@ -259,7 +259,7 @@ export class Extrinsic<N = unknown, T extends ISubmittableResult = ISubmittableR
     return firstValueFrom(
       this.extrinsic()
         .paymentInfo(getUnifiedAddress(this.keys))
-        .pipe(map((info) => info.partialFee.toBigInt() * 2n))
+        .pipe(map((info) => info.partialFee.toBigInt()))
     );
   }
 
--- a/pallets/frequency-tx-payment/src/lib.rs
+++ b/pallets/frequency-tx-payment/src/lib.rs
@@ -372,7 +372,8 @@ where
 		}
 	}
 
-	fn can_withdraw_fee(
+	// simulates fee calculation and withdrawal without applying any changes
+	fn dryrun_withdraw_fee(
 		&self,
 		who: &T::AccountId,
 		call: &<T as frame_system::Config>::RuntimeCall,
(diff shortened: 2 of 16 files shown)


e2e/scaffolding/extrinsicHelpers.ts
```diff
@@ -259,7 +259,7 @@ export class Extrinsic<N = unknown, T extends ISubmittableResult = ISubmittableR
     return firstValueFrom(
       this.extrinsic()
         .paymentInfo(getUnifiedAddress(this.keys))
-        .pipe(map((info) => info.partialFee.toBigInt() * 2n))
+        .pipe(map((info) => info.partialFee.toBigInt()))
```
shannonwells, line 262: This was the main hack to get a bunch of tests passing, but it was only there because I hadn't realized withdraw_fee was happening twice, and that was due to the changes to the way <del>`SignedTransactions`</del> `TransactionExtensions` work.

pallets/frequency-tx-payment/src/payment.rs
```diff
@@ -35,11 +41,6 @@ where
 		key: &T::AccountId,
 		fee: Self::Balance,
 	) -> Result<Self::Balance, TransactionValidityError> {
-		ensure!(
```
aramikm, line 38: @enddynayn I'm not sure why this is removed. Any idea why?
 Edit: It looks like it is moved to the `can_withdraw_fee` but I don't see any problem just keeping it here as is

aramikm (approved):

pallets/frequency-tx-payment/src/payment.rs
JoeCap08055: I suppose it's just because it's redundant/inefficient (since it's in the `validate` path, and hence could be called many more times than the `pre_dispatch` patch?

JoeCap08055 (approved):
- Reviewed code
- Ran unit tests
- Ran e2e tests; pushed a fix to e2e test script & one failing test

Looks good!

pallets/frequency-tx-payment/src/lib.rs
```diff
@@ -372,6 +372,57 @@ where
...
+				self.can_withdraw_capacity_fee(who, &vec![*call.clone()], len),
+
+			Some(Call::pay_with_capacity_batch_all { calls }) =>
+				self.can_withdraw_capacity_fee(who, calls, len),
+
+			_ => self.can_withdraw_token_fee(who, call, info, len, self.tip(call)),
+		}
+	}
+
+	fn can_withdraw_capacity_fee(
```
enddynayn, line 393: Minor: I expected this to return a boolean, but it returns a value instead. A name like try_withdraw_capacity_fee might better reflect its behavior.

pallets/frequency-tx-payment/src/payment.rs
enddynayn: I believe this is likely due to the can_withdraw_x function being called before the fee is withdrawn.

enddynayn (approved):
great! 👍

PR #2316, opened 2025-04-08
fix: lint (clippy) fixes

JoeCap08055:
Lint/Clippy cleanup


pallets/msa/src/lib.rs
```diff
@@ -1244,9 +1244,8 @@ impl<T: Config> Pallet<T> {
 				let is_new = maybe_delegation_info.is_none();
 				let mut delegation = maybe_delegation_info.take().unwrap_or_default();
 
-				f(&mut delegation, is_new).map(move |result| {
+				f(&mut delegation, is_new).inspect(move |_result| {
```
aramikm, line 1247: This might work but I don't think this is the typical usecase for using the inspect. I think it was only supposed to be used for reading or inspecting/logging stuff instead of settring stuff but just want to make sure we have tests that cover this and it works as intended

JoeCap08055: Yeah, the docs say that's the most common use case, but clippy doesn't like returning an un-mutated item from a map & suggested `inspect`. You're right, though, I'll review to make sure it's covered by tests.

aramikm (approved):

PR #2331, opened 2025-04-15
Upgrade Frequency to polkadot-stable2503

JoeCap08055:
# Goal
The goal of this PR is to upgrade Frequency to the latest Polkadot `stable2503` release.

In addition to the base changes required by `stable2503`, a couple other fixes/tweaks:
* Fix `schemas` pallet benchmarks that were broken by #2321 
* Since `stable2512` introduced the requirement for frame system extensions weights to construct the runtime, add `frame_system_extensions` to our benchmarking setup (instead of hard-coding a copy from the polkadot-sdk repo)

Closes #2317 

# Checklist
- [x] Updated Pallet Readme?
- [x] Benchmarks added?
- [x] Spec version incremented?

PR #2337, opened 2025-04-18
Split the build benchmark binary and benchmark run steps

wilwade:
This updates the way benchmarks are run so that we use less time on the benchmark runner.

Benchmark runner time is now < 20 minutes. (Total run is much longer, but this is much easier to use spot instances with)

- Update the release workflow
- Update the benchmark runner
- Switch the benchmark runner to use `m6id.2xlarge` as the benchmark. This is for various reasons, but it mostly is a better equivalent for the Polkadot Validator minimum.
- Benchmark runner had been a `c7i` for the last run, so the numbers look like they increase more than they actually did.
- Ran all benchmarks


.github/workflows/benchmarks-run.yml
```diff
@@ -16,9 +16,35 @@ on:
...
+        run: |
+          echo "Running benchmark binary builder..."
+          ./scripts/run_benchmarks.sh -b
+
+      - name: Upload Binary
+        uses: actions/upload-artifact@v4
+        with:
+          name: benchmark-binary
+          path: target/release/frequency
+          retention-days: 3
```
JoeCap08055, line 42: 3 days seems excessive... I wouldn't think we would need more than 1

scripts/run_benchmarks.sh
```diff
@@ -93,7 +97,7 @@ function is_custom_pallet() {
   return 1
 }
 
-while getopts 'dh:p:st:v' flag; do
+while getopts 'd:h:p:st:vb' flag; do
```
JoeCap08055, line 100:
```suggestion
while getopts 'd:hp:st:vb' flag; do
```

.github/workflows/benchmarks-run.yml
JoeCap08055, line 42: Note, pending PR #2331 needs the runtime wasm blob in addition to the node binary; may as well upload it here

.github/workflows/benchmarks-run.yml
```diff
@@ -29,33 +55,35 @@ jobs:
...
           ref: ${{github.event.inputs.branch}}
-      - name: Install Required Packages
-        run: |
-          sudo apt-get update
-          sudo apt install -y protobuf-compiler libclang-dev clang cmake
-      - name: Install Rust Toolchain
-        # Match installation steps to CI base docker image
-        run: |
-          curl https://sh.rustup.rs -sSf | HOME=`pwd` bash -s -- -y
-          echo "PATH=`pwd`/.cargo/bin:$PATH" >> $GITHUB_ENV
```
JoeCap08055, line 44: Note, the benchmark script runs unit tests at the end, ie `cargo test ...`, so you'd either need to skip that or still have the Rust toolchain installed. Or, could extract that out to a separate post-benchmark job.

I think it's important to still run the tests post-benchmarks, to make sure the updated weights files haven't broken the build.

JoeCap08055:
FYI, PR #2331 has a [commit](https://github.com/frequency-chain/frequency/pull/2331/commits/7820196f9efb0f6c7cd22bcbffbe30f450acb2ef) that fixes the benchmarks for the `schemas` pallet

aramikm:
This appeared to work but it might be anecdotal. After checking latest failures it usually fails in 2 ways which I don't think none of those get's addressed here.

- It gets stuck when trying to get an instance to run the job

- It fails during build (which might be related to some resource exhaustion)

.github/workflows/benchmarks-run.yml
```diff
@@ -5,6 +5,8 @@ concurrency:
   group: ${{github.workflow}}-${{github.ref}}
   cancel-in-progress: true
 on:
+  push:
```
JoeCap08055, line 8:
```suggestion
```

wilwade: Up. This is a REVERT ME Testing commit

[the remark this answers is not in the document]
wilwade: Just for testing

.github/workflows/benchmarks-run.yml
```diff
@@ -29,33 +64,45 @@ jobs:
...
+        with:
+          name: wasm-artifacts
+          path: target/release/wbuild
+
+      - name: Make Binary Executable
+        run: chmod +x target/release/frequency
+
       - name: Update Weights
         run: |
           rustup show
```
wilwade, line 93:
```suggestion
```

aramikm (approved):
Awesome. Ship it

PR #2347, opened 2025-04-21
Update JS Dependencies

wilwade:
# Goal
The goal of this PR is to update JS dependencies before the new release

- e2e
- api-augment
- schemas
- tools

Also:
- Removed workerpool as no longer needed


e2e/scaffolding/globalHooks.ts
```diff
@@ -65,7 +64,6 @@ async function devSudoActions() {
 }
 
 export async function mochaGlobalSetup(context) {
-  console.log('Global Setup Start', 'Reported CPU Count: ', workerpool.cpus);
```
wilwade, line 68: This was the only use of workerpool

aramikm (approved):

Issue #2333, opened 2025-04-16
Add full relay chain + collator tests in CI pipeline


JoeCap08055:
## Details
In the course of upgrading to `polkadot-stable2412`, an incompatibility was introduced that prevents our collator node from onboarding to the relay chain. Because our CI pipeline only tests parachain functionality with no relay chain (instant-seal or interval-seal build/mode), this was only detected through manual testing. We should develop a test in CI to validate that full parachain + relay chain behaves properly.

## How to test (manually)
- Run `make run-frequency-docker`
    - This brings up 2 relay chain nodes (alice + bob), and 1 Frequency "alice" collator node as Docker containers
    - Registers parachain ID 2000 with the relay chain
    - Onboards the Frequency runtime from the collator node to the relay chain
- Expected results: parachain collator starts forming blocks

## Acceptance Criteria
- [ ] Top-level Makefile includes a target to run tests locally for above
- [ ] CI pipeline includes a workflow to test the above setup

### Searched for Related Issues

- [x] I have done a search for related issues and either found none, or noted them

PR #2350, opened 2025-04-22
Update to latest Paseo-Local spec v1.4.1

wilwade:
# Goal
The goal of this PR is to update the paseo local chain spec.

Part of #2333

# Discussion

- Updated from https://github.com/paseo-network/runtimes/tree/main/chain-specs/local
- This does change the genesis, so it breaks things like `make run-frequency-docker` until the docker images are updated

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -32,7 +32,6 @@ services:
       --rpc-external
       --rpc-cors=all
       --rpc-methods=Unsafe
-      --pool-type=fork-aware
       --alice
       -ltxpool=trace
     restart: unless-stopped
@@ -66,7 +65,6 @@ services:
       --rpc-external
       --rpc-cors=all
       --rpc-methods=Unsafe
-      --pool-type=fork-aware
       --bob
       --log="main,info"
     restart: unless-stopped
--- a/node/cli/src/command.rs
+++ b/node/cli/src/command.rs
@@ -252,7 +252,7 @@ macro_rules! construct_async_run {
 	(|$components:ident, $cli:ident, $cmd:ident, $config:ident| $( $code:tt )* ) => {{
 		let runner = $cli.create_runner($cmd)?;
 		runner.async_run(|$config| {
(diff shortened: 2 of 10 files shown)


JoeCap08055:
Does this break `make run-frequency-docker`, though? When running the containers, it gets the genesis from the running container, and the chain spec is also baked into the container image. So should be fine, no?

wilwade:
> Does this break `make run-frequency-docker`, though? When running the containers, it gets the genesis from the running container, and the chain spec is also baked into the container image. So should be fine, no?

@JoeCap08055 It breaks it because the run-frequency-docker uses the published docker image for the collator. If you build a new docker image and use that, it would work.

JoeCap08055:
> > Does this break `make run-frequency-docker`, though? When running the containers, it gets the genesis from the running container, and the chain spec is also baked into the container image. So should be fine, no?
> 
> @JoeCap08055 It breaks it because the run-frequency-docker uses the published docker image for the collator. If you build a new docker image and use that, it would work.

Hmm, to be clear (I think?), `make run-frequency-docker` was already broken by the fact that we're now running with `stable2412` relay node images (and even our `stable2412` collator release image `v1.16.0-rc1` has issues), not because this PR changes the genesis, right?

JoeCap08055 (approved):
Other comments notwithstanding, approved.

wilwade:
> Hmm, to be clear (I think?), make run-frequency-docker was already broken by the fact that we're now running with stable2412 relay node images (and even our stable2412 collator release image v1.16.0-rc1 has issues), not because this PR changes the genesis, right?

Not the relay chain node images, but something in the upgrade required the new runtime api (which is a part of the async backing). Since the old Paseo-Local didn't have that runtime api, it was failing.

Now it is broken because make run-frequency-docker uses the published docker image, but that was built with a different relay chain genesis. (Paseo-Local changes genesis on each version bump)

aramikm (approved):
This should be good if we don't have anything custom in our paseo-local genesis.

PR #2355, opened 2025-04-23
Reduce Testnet WASM Size

wilwade:
# Goal
The goal of this PR is to reduce the size of the released wasm for Testnet

## Details

Testnet wasm has always been larger than mainnet, as it includes a few more things like the sudo pallet. However in release [v1.14.0](https://github.com/frequency-chain/frequency/releases/tag/v1.14.0) it increased a large amount.

This has not been an issue, but in v1.16.0, the wasm increased to over 2.5Mb which caused issues with being able to upgrade cleanly on testnet.

A reason for the increase was that the mainnet genesis was also being included in the testnet releases. This was not needed, so this PR removes it from those builds.

## Summary

- Fixed release notes template issues
- Removed paseo local spec from the build for paseo testnet
- Limited the `frequency_genesis_config` to just `frequency` and benchmarks runs
- Bumped the spec version

## Result

Local builds of the Mainnet wasm are at ~1.7M and Testnet dropped from 2.6M to 2.1M

## Local Testing

- Mainnet: `make build-mainnet-release && ls -lah target/release/wbuild/frequency-runtime/frequency_runtime*`
- Testnet: `make build-testnet-release && ls -lah target/release/wbuild/frequency-runtime/frequency_runtime*`


JoeCap08055 (approved):
Looks good--can't hurt, even if it's not the source of the issue

node/service/src/chain_spec/frequency_paseo_local.rs
```diff
@@ -0,0 +1,59 @@
...
+		&include_bytes!("../../../../resources/frequency-paseo.raw.json")[..],
+	)
+	.unwrap()
+}
+
+// TODO: Remove once on a Polkadot-SDK with Paseo
+#[allow(clippy::unwrap_used)]
+/// Generates the Paseo Relay chain spec from the json
+pub fn load_paseo_spec() -> RelayChainSpec {
+	RelayChainSpec::from_json_bytes(&include_bytes!("../../../../resources/paseo.json")[..])
```
aramikm, line 28: nit: if we already have `Paseo-Local Relay chain spec` why do we need this? Doesn't seem to be used.

aramikm (approved):
Looks good! Our feature flags are a mess and it's becoming hard to follow.

node/service/src/chain_spec/frequency_paseo_local.rs
wilwade: Ah you are right. It is not needed on the paseo-local side. I'll remove.

wilwade: Removed

mattheworris (approved):
🚢 it!

Issue #2351, opened 2025-04-22
MSAs can Have a Balance (Blocked)


JoeCap08055:
## General Idea

It is useful to be able to send tokens to an MSA for various needs. Rewards programs from both the chain and providers would be possible.

## A/C

- [ ] Given that I know an MSA Id, I can send that MSA tokens
    - [ ] Each MSA should have a derivable Ethereum Address.
    - [ ] Sending a token to an MSA should be no different than sending to another account
- [ ] Given I known an MSA Id, I should be able to derive the Ethereum address and then ask the balances pallet for the balance of that MSA.
- [ ] Given I have an MSA, I can use a control key to set a Payment Receiver Address
    - [ ] The user must provide proof that they control the Payment Receiver Address
- [ ] After the receiver address is set, all tokens will drain to that Payment Receiver Address.
- [ ] Given I have an MSA with a Payment Receiver Address, I can remove my Payment Receiver Address with the Payment Receiver Address
- [ ] Given I have an MSA with a Payment Receiver Address, I can remove my Payment Receiver Address with any Control Key
- [ ] [UNSURE] Given I have an MSA with a Payment Receiver Address, I can change my Payment Receiver Address with the Payment Receiver Address
- [ ]  [UNSURE] Given I have an MSA with a Payment Receiver Address, there *cannot* be another MSA with the same Payment Receiver Address

### ALT: Receiver Address is ALSO a Control Key

As the Receiver Address is required to be proven as owned by the user, it could be also set as a control key

## Questions/Notes
- The Payment Receiver Address auto forwarding requirement could be dropped, but is desired

### Risks
- Token receipt to MSAs and balances are public information. Those tokens are able to be mostly tracked to any other used addresses. This represents a risk of additional financial information being discovered by the public. However, there are no absolute defenses for this without problematic services such as mixers
- The Control Keys would not just be social/MSA control, but would upgraded to enable to maintenance of the Payment Receiver Address 

### Rejected AC
- After the receiver address is set, if new tokens are sent to that address, it should automatically forward those tokens to the Payment Receiver Address

aramikm:
If each MSA should have a `derivable Ethereum Address` what is the reason for `Payment Receiver Address`? I found these confusing.

JoeCap08055:
> If each MSA should have a `derivable Ethereum Address` what is the reason for `Payment Receiver Address`? I found these confusing.

[...]

PR #2358, opened 2025-04-24
feat: custom runtime functions to generate deterministic Eth address from an MSA ID

### LINKED ISSUE #2352: Given that I know an MSA Id, I can send that MSA tokens
# Description
As a user, I want to be able to send tokens to an MSA using a known/derivable Ethereum address.

## AC
- [ ] Each MSA should have a derivable Ethereum Address.
- [ ] Sending a token to an MSA should be no different than sending to another account
- [ ] I **should not** be able to send tokens _from_ an MSA (or an Ethereum address to which an MSA maps)

### LINKED ISSUE #2353: Given I known an MSA Id, I should be able to derive the Ethereum address and then ask the balances pallet for the balance of that MSA.

JoeCap08055:
# Goal
This PR adds some internal pallet functions and custom runtime to deterministically generate an Ethereum address from an MSA ID, and vice versa.

The general approach is:
- Use a domain separator prefix to eliminate any possible conflict with addresses generated by `CREATE2`
- Use a salt that is the hash of the string "MSA Generated"
- Take the last 20 bytes of the hash of `(prefix + MSA ID + salt)`
- If returning a string for the runtime query, pass through an [ERC-55 checksum algorithm](https://github.com/ethereum/ercs/blob/master/ERCS/erc-55.md) to alter the returned hex string

Note, although addresses generated in this manner cannot be reversed to recover the related MSA ID, they can be verified (ie, given an address and an MSA ID, we can determine whether or not the address is the result of applying the above algorithm to the MSA ID). It is not envisioned that there would ever be an _on-chain_ requirement to recover the MSA ID from the address; if such a use case was desired for off-chain purposes, it would be relatively trivial to populate a static lookup table mapping all known or possible MSAs to their corresponding address, or vice-versa.

Closes #2353 
Closes #2352 

# Discussion

- Q0: Currently returning ASCII bytes; or do we want to return the binary address bytes?
    - A: Now returning a struct containing both the binary address & checksummed string

# Checklist
- [x] Updated Pallet Readme?
- [ ] Design doc(s) updated?
- [x] Unit Tests added?
- [x] e2e Tests added?
- [ ] Benchmarks added?


pallets/msa/src/lib.rs
```diff
@@ -1328,6 +1332,78 @@ impl<T: Config> Pallet<T> {
...
+		// Extract and return u64
+		let mut id_bytes = [0u8; 8];
+		id_bytes.copy_from_slice(&address[..8]);
+		Some(u64::from_be_bytes(id_bytes))
+	}
+
+	/// Converts a 20-byte synthetic Ethereum address into a checksummed string format,
+	/// using ERC-55 checksum rules.
+	/// Formats a 20-byte address into an EIP-55 checksummed `0x...` string.
+	pub fn eth_address_to_checksummed_string(addr: &[u8; 20]) -> BoundedVec<u8, ConstU32<42>> {
```
enddynayn, line 1373: Would it make sense to create an alias type for AccountId20, if one doesn’t already exist?

Also, this seems like a fairly common utility—other projects might already have solid implementations using libraries like alloy or web3. Do you think it might be worth considering one of those instead?

enddynayn:
Would you mind adding a brief explanation to the PR description about why this feature is being introduced?

JoeCap08055:
> There are a number of questions around this approach:
> 
> 1. What is the scope and goal of these ETH addresses? Are we expecting to use these in the bridges to send money to and  from Ethereum?
- In order to be able to have MSAs _receive_ tokens at a known address, that can be deterministically generated and operate with the balances pallet normall
- Since there will be no private key associated with this address, it will be impossible to transfer tokens _away_ from the MSA address except by such methods as we specifically provide

> 2. Ideas about generation
>    
>    1. I don't like any addresses prefixed or postfixed with zeros since those are likely the precompiles in some chain or another
- We can prefix/postfix with whatever value we want
>    2. Another alternative might be to have the first or last 8 bytes for MsaID and either use it with some kind of tag to hash or use as a seed to generate the remaining 12 bytes.
- Using a hash instead of a known pre/post-fix makes it difficult to determine that the key was generated with this algorithm & therefore reverse the process. However, if we determine reversal (ie, get MSA ID from address) is not important, or if we want to simply maintain a mapping (perhaps off-chain), then we could fill with a hash.

>    3. Another alternatives would be that If they behave like a smart contract then maybe it should get generated like a smart contract or at least in a similar way.
>    4. In general, l  think there might be better ways to generate these addresses that are more natural with more entropy. I think it would worth doing some research since we are going to be stuck with them for a long time.
- Is entropy important just for an address, as long as it's not possible to have a conflict in the address space? Remember, it's not an actual public key, just an address to send tokens to.

> 3. We should probably use something similar `trait AccountAddressMapper`

JoeCap08055:
> Would you mind adding a brief explanation to the PR description about why this feature is being introduced?

It's in the linked issue #2353 and its parent #2351

pallets/msa/src/lib.rs
JoeCap08055: `web3` is deprecated, but `alloy` has some good stuff--however, given the size/simplicity of the implementation here, I don't know if we really need to take a dependency unless we see ultimate value in the rest of the `alloy` crate(s) in the future (which we may...)

I could easily swap in alloy for pieces of this, if that's the way we want to go...

wesbiggs:
Can any user can send tokens to these addresses at any time (as long as they know the MSA and the algorithm to determine the account)? Or will the accounts be marked in some way that prevents this from happening other than via chain code?

If not, two cases to consider, neither of which may prevent a threat, but just thinking it through...

1. Send tokens to a known MSA (prior to that MSA receiving a reward)
2. Send tokens to a not-yet-allocated MSA which is later allocated

wilwade:
@wesbiggs 

> Can any user can send tokens to these addresses at any time (as long as they know the MSA and the algorithm to determine the account)?

This would be the idea. So yes, it is possible for anyone to send tokens to any MSA, even future MSAs, (although there is no way to specifically select an MSA Id or batch transactions in such a way that ensures a specific MSA Id at this time)

harry-evans:
Most of this sounds fine to me. I don’t love the idea of being able to send tokens to a non existent MSA, given that they are assigned vs claimed (token accounts are “claimed” by proving you have the private key, while MSAs are assigned based on the next available number/id). 

I don’t think that really matters, but in ideal world that would be an obvious error that we would just reject when the txn was submitted. Not sure it is worth the work to implement or the runtime efficiency overhead, but that is the only thing that I would really change.

pallets/msa/Cargo.toml
```diff
@@ -35,6 +35,7 @@ common-runtime = { path = "../../runtime/common", default-features = false }
 pallet-schemas = { path = "../schemas", default-features = false }
 pallet-handles = { path = "../handles", default-features = false }
 pallet-collective = { workspace = true }
+pallet-revive = { workspace = true }
```
mattheworris, line 38: Move this to `[dependencies]

pallets/msa/src/lib.rs
```diff
@@ -1328,6 +1332,73 @@ impl<T: Config> Pallet<T> {
...
+		bytes[53..85].copy_from_slice(&init_code_hash);
+		let hash = keccak_256(&bytes);
+		H160::from_slice(&hash[12..])
+	}
+
+	/// Converts an MSA ID into a synthetic Ethereum address (raw 20-byte format) as follows:
+	/// [0..8]:   u64 (big endian)
+	/// [8..17]:  b"Generated"
+	/// [17..20]: all zeroes
+	pub fn msa_id_to_eth_address(id: MessageSourceId) -> H160 {
```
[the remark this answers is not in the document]
wilwade: For salt, I wonder if we just want to generate a value, but have it different per local, testnet, and mainnet.

salt is usually used on Ethereum to allow you to semi-pick an address or make sure that there isn't a collision for the same contract code deployed twice.

Deployer Address perhaps `0x00...00`? Although all zeros can cause issues sometimes, but that's a deployer that just "never" exists and is easy to remember. Another option would be taking 20 bytes from the palletId? `PalletId(*b"py/trsry")` style?

Code? If we don't have to have anything here, that would be nice. Perhaps instead of a network salt, then we use genesis here?

I think my tldr: The easier these are, the better, but feel like something should be different between mainnet/testnet/local.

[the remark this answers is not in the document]
JoeCap08055: Hmm, okay... I think this will work, then:
* Deployer address: `0x00..00` should be fine
* Code: genesis
* Salt: we'll just generate some value and add it to the code as a constant

For the genesis: should we hard-code based on features, or retrieve at runtime? (this is not on-chain, so not concerned much about cost)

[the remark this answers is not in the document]
wilwade: Re genesis: Runtime? Local would have a harder time with using it otherwise as it doesn't have a hard coded value. However, perhaps (for tests and such) we just want to hard code local to something.

pallets/msa/src/lib.rs
JoeCap08055: Ended up going with:
- The `H160` type from the [`primitive-types` create](https://docs.rs/primitive-types/0.13.1/primitive_types/). This is a standalone crate by Parity with types shared between Substrate and Ethereum. It's also the address type that `pallet-revive` uses.
- Cloned the `create2()` function from `pallet-revive`, rather than import the entire crate. It's a simple, standard algorithm for Ethereum. We can replace it in the future if we take a dependency that provides an implementation.

pallets/msa/src/lib.rs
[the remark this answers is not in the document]
JoeCap08055: Right--I think for local/dev chain, we want a hard-coded value so that test data is deterministic. Real live chains can use the actual genesis.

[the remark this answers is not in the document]
JoeCap08055: Went with:
* Deploy address = 0x00..00
* Code = genesis for mainnet/testnet; canned value for dev chain
* Input data = MSA ID big-endian bytes
* Salt = same as Code

pallets/msa/src/lib.rs
```diff
@@ -1357,6 +1359,89 @@ impl<T: Config> Pallet<T> {
...
+		// Use a canned value for all but mainnet/testnet builds
+		let genesis_hash: [u8; 32] = if cfg!(any(
+			feature = "frequency-local",
+			feature = "frequency-no-relay",
+			feature = "frequency-lint-check",
+			feature = "runtime-benchmarks"
```
JoeCap08055, line 1389: Could use a double-check here that this is correct, and that I added configs that were previously unused by this pallet correctly to the appropriate Cargo.toml files.

pallets/msa/src/lib.rs
```diff
@@ -1357,6 +1359,89 @@ impl<T: Config> Pallet<T> {
...
+		)) {
+			hex::decode("391116a649bc9bb87cde5ccda6a70e352b665e96d7308f04ff932677de96d39c")
```
wilwade, line 1391: Wonder if we want to make this all `11` or something easy to know?

e2e/msa/msaTokens.test.ts
```diff
@@ -0,0 +1,59 @@
...
+let checksummedEthAddress: HexString;
+
+if (isTestnet()) {
+  checksummedEthAddress = '0x05500A07f5fD359e9E785c74E21d5b180e63B63b'; // Example checksummed Ethereum address for MSA ID 1234 on Paseo testnet
+} else {
+  checksummedEthAddress = '0xF5b82ff246a2F4226749bd78B1bDaE28Cfffb9f7'; // Example checksummed Ethereum address for MSA ID 1234 on a development chain
+}
+
+describe('MSAs Holding Tokens', function () {
+  describe('getEthereumAddressForMsaId', function () {
```
wilwade, line 18: While not quite an e2e test (and perhaps it is coming), but it would be a good idea to have a test that shows an external js library generating the same value so that we make sure that a external implementation of the library generates the same result as the chain. Also as an example of how to use it to generate the value.

pallets/msa/src/runtime-api/src/lib.rs
```diff
@@ -29,7 +29,7 @@ sp_api::decl_runtime_apis! {
 	/// Runtime Version for MSAs
 	/// - MUST be incremented if anything changes
 	/// - See: https://paritytech.github.io/polkadot/doc/polkadot_primitives/runtime_api/index.html
-	#[api_version(2)]
+	#[api_version(3)]
```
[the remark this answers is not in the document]
JoeCap08055: I read @aramikm 's PR that bumped the version to "2", and unless I misread, [you made the same comment](https://github.com/frequency-chain/frequency/pull/2022#pullrequestreview-2112836141), and lthe issue was that, despite documentation to the contrary, it doesn't actually work that way; there were unresolvable build errors.

enddynayn:
I actually preferred your earlier approach using a simple prefix string + MSA ID + padding hash. It was straightforward, easy to follow, and it worked well for what we needed.

Using CREATE2 feels like overkill in this case since we’re not deploying contracts at these addresses. It also adds some extra code we’ll now have to maintain, without a clear benefit for our use case.

The simpler version gave us everything we needed — deterministic, predictable addresses that clearly link back to the MSA without the added complexity.

wilwade:
@enddynayn  Re: CREATE2

Here's the reasons for me at least for why I liked CREATE2 better:

1. More Ethereum-y way to do it.
2. All libraries for Ethereum already have the logic and that logic is very well documented. (This is likely the biggest one for me)
3. Results in a random spread of addresses instead of a block. While in theory there is no differences between the two in security, a block of addresses is always a smell to me.
4. Means the MSA addresses are closer to contract addresses.

enddynayn:
@wilwade thanks for sharing helping me understand your thinking. I just want to see if I understand a little bit more

1. More Ethereum-y way to do it.
I get what your saying but here the case is not actually deploying a contract and we are simply deriving a keyless address. I am unsure if this actually fits the mechanics of what we are doing.

2. Libraries and docs are well supported
True but it seems we are forcing a pattern and adding duplicate code that we need to maintain just to take a advantage. The simpler approach is even easier to implement and do not require users to import other tools to use.

3. Random spread...
True but we can still introduce a salt without adding CREATE2

4. Closer to contract address
Interesting but I do not see the benefits from the similarity we are not treating these like contracts. We are managing these through custom extrinsics.

Overall, I see where you are coming from. The simpler approach seems more pupose fit and avoid extra maintance burden and does not really fit our use case. What happens when CREATE3 comes out?


> @enddynayn Re: CREATE2
> 
> Here's the reasons for me at least for why I liked CREATE2 better:
> 
> 1. More Ethereum-y way to do it.
> 2. All libraries for Ethereum already have the logic and that logic is very well documented. (This is likely the biggest one for me)
> 3. Results in a random spread of addresses instead of a block. While in theory there is no differences between the two in security, a block of addresses is always a smell to me.
> 4. Means the MSA addresses are closer to contract addresses.

I see your point

pallets/msa/src/lib.rs
aramikm: Another maybe simpler way of doing is to use the chainID to differentiate between chains. We would need to use it for domain separators in e-712 signatures anyway. so Why not use it here?

aramikm (approved):
Looked good just some comments.

pallets/msa/src/lib.rs
```diff
@@ -1357,6 +1359,89 @@ impl<T: Config> Pallet<T> {
...
+		H160::from_slice(&hash[12..])
+	}
+
+	/// Converts an MSA ID into a synthetic Ethereum address (raw 20-byte format) as follows:
+	/// [0..8]:   u64 (big endian)
+	/// [8..17]:  b"Generated"
+	/// [17..20]: all zeroes
+	pub fn msa_id_to_eth_address(id: MessageSourceId) -> H160 {
+		// Use a canned value for all but mainnet/testnet builds
+		let genesis_hash: [u8; 32] = if cfg!(any(
```
aramikm, line 1385: To simplify this I think you could do it the other way around and checking to see if main-net or testnet and everything else uses the default genesis

pallets/msa/src/lib.rs
enddynayn: I like something easy to know.

JoeCap08055: @aramikm if chain-id is easily fetchable from inside the runtime, that might be prefereable to using the genesis hash, as we could use it even for a development chain, I think--yes? We would just use the `keccak256` of the chain-id; that's 32 bytes.

@enddynayn & @wilwade that would eliminate the need for any hard-coded value at all

enddynayn: That works too. I think ParachainInfo should ideally contain the chain ID, but using it would require adding another dependency to the pallet. It feels like a bit of extra work without much bang. That said, I am not opposed — Ill pass it along to Joe for his take.

JoeCap08055: I agree with @enddynayn -- adding the `parachain-info` dependency to this pallet is a bit more involved than just adding that one crate dependency. I think the way to go is:
* Use genesis block for mainnet/testnet
* Use `keccack_256(some-well-known-value)` for dev/local

pallets/msa/src/lib.rs
JoeCap08055: Flipped.

pallets/msa/src/lib.rs
JoeCap08055: Now using `keccak_256(b"frequency-development")` as the fixed value

e2e/msa/msaTokens.test.ts
JoeCap08055: I'll add JS functions to the scaffolding that implements both `CREATE2` and our MSA address derivation logic.

e2e/scaffolding/foo.ts
```diff
@@ -0,0 +1,12 @@
...
+import { Keyring } from '@polkadot/api';
+import { mnemonicGenerate } from '@polkadot/util-crypto';
+
+const ethKeys = new Keyring({ type: 'ethereum' }).createFromUri(mnemonicGenerate());
+const ethAddress = getUnifiedAddress(ethKeys);
+const ethPublicKey = getUnifiedPublicKey(ethKeys);
+
+const ethAddrss2 = getUnifiedAddress({ type: 'ethereum', publicKey: ethPublicKey } as unknown as KeyringPair);
+
+console.log('Results: ', { ethPublicKey, ethAddress, ethAddrss2 });
```
enddynayn, line 12: leftover?

pallets/msa/src/lib.rs
```diff
@@ -1357,6 +1359,68 @@ impl<T: Config> Pallet<T> {
...
+	/// [0..1]: 0xD9 (first byte of the keccak256 hash of the domain prefix "Frequency")
+	/// [1..9]:   u64 (big endian)
+	/// [9..41]:  keccack256("MSA Generated")
+	pub fn msa_id_to_eth_address(id: MessageSourceId) -> H160 {
+		/// First byte of the keccak256 hash of the domain prefix "Frequency"
+		/// This "domain separator" ensures that the generated address will not collide with Ethereum addresses
+		/// generated by the standard 'CREATE2' opcode.
+		const DOMAIN_PREFIX: u8 = 0xD9;
+		/// keccak256 of "MSA Generated"
+		const SALT: [u8; 32] = [
```
enddynayn, line 1373: We can convert it to bytes: b"MSA generated"

enddynayn (approved):
👍 great!

aramikm (approved):
Looks good, added some nits

e2e/scaffolding/foo.ts
JoeCap08055: lol, whole file is an accidental commit

shannonwells (approved):
Nothing blocking, leave it up to you to incorporate any suggestions, otherwise GTG!

pallets/msa/src/lib.rs
```diff
@@ -1357,6 +1360,67 @@ impl<T: Config> Pallet<T> {
...
+	/// [0..1]: 0xD9 (first byte of the keccak256 hash of the domain prefix "Frequency")
+	/// [1..9]:   u64 (big endian)
+	/// [9..41]:  keccack256("MSA Generated")
+	pub fn msa_id_to_eth_address(id: MessageSourceId) -> H160 {
+		/// First byte of the keccak256 hash of the domain prefix "Frequency"
+		/// This "domain separator" ensures that the generated address will not collide with Ethereum addresses
+		/// generated by the standard 'CREATE2' opcode.
+		const DOMAIN_PREFIX: u8 = 0xD9;
+
+		lazy_static! {
```
aramikm, line 1374: I'm unsure about the behavior of `lazy_static` inside a scope such as a function.  I've only seen it in global scope.

JoeCap08055: Pretty sure it just limits the scope of the the resulting `static ref` variable to the enclosing function. I prefer to keep things scoped as minimally as possible.

e2e/msa/msaTokens.test.ts
```diff
@@ -0,0 +1,113 @@
...
+        `Expected checksummed address ${checksummedEthAddress}, but got ${accountIdChecksummed.toString()}`
+      );
+    });
+
+    it('should validate the Ethereum address for an MSA ID', async function () {
+      const isValid = await ExtrinsicHelper.apiPromise.call.msaRuntimeApi.validateEthAddressForMsa(
+        generateMsaAddress(msaId),
+        msaId
+      );
+      assert.equal(isValid, true, 'Expected the Ethereum address to be valid for the given MSA ID');
```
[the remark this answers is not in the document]
JoeCap08055: Updated.

PR #2363, opened 2025-04-28
Fix: check_handle should apply the whitespace trimming and collapsing

### LINKED ISSUE #2362: BUG: check_handle runtime RPC was not applying the trim_and_collapse_whitespace function
`check_handle` Runtime RPC is supposed to return the display handle as it would be recorded, but it was not using the `trim_and_collapse_whitespace` function, thus the whitespace trim application was not happening, thus the values returned are incorrect.

- Expected: `handles.check_handle("   hell  o   ")` -> `"hell o"`
- Actual: `handles.check_handle("   hell  o   ")` -> `"   hell  o   "`

wilwade:
# Goal
The goal of this PR is to update the check_handle Runtime RPC to correctly use the whitespace trim and collapse

Closes #2362

# Discussion

- I thought about updating how this gets the value to be more aligned, but in the end the simple fix won out.

# Checklist
- [x] Unit Tests added?


mattheworris (approved):
🚢 it!

aramikm (approved):
Looks good! I'm just surprised that we are not using the same functions to do the same checks or transformations in the first place.

enddynayn (approved):
👍

PR #2373, opened 2025-04-29
feat(bridging): add config for local westend

### LINKED ISSUE #2372: Bridging: All required configurations (spec file) for bridging to/from Westend local added.
**Description**
Add the ability to test bridging on local westend.

**Acceptance Criteria**
- [x] Configure Frequency to generate chain spec file for westend-local.

enddynayn:
Add chain spec configuration to enable running Frequency
with a local Westend network. This setup supports testing
bridging functionality between Frequency and Snowbridge.

Closes #2372


node/cli/src/command.rs
```diff
@@ -75,6 +80,9 @@ fn load_spec(id: &str) -> std::result::Result<Box<dyn ChainSpec>, String> {
 		#[cfg(feature = "frequency-local")]
 		"frequency-paseo-local" =>
 			Ok(Box::new(chain_spec::frequency_paseo_local::local_paseo_testnet_config())),
+		#[cfg(feature = "frequency-bridging")]
```
wilwade, line 83: Perhaps this was for later, but I thought we were going to add a different feature flag for the spec vs the actual bridging code

wilwade: Or I guess here it should be this right?

```suggestion
		#[cfg(feature = "frequency-local")]
```

wilwade (commented):
Looks good to me other than a lot of the `frequency-bridging` flags I think should be `frequency-local`?

node/cli/src/command.rs
enddynayn: Yes, thanks for catching this, we plan on revisiting this as we add `frequency-westend-testnet`.

mattheworris (approved):

PR #2378, opened 2025-04-29
allow two freezes so capacity stakers can vote

### LINKED ISSUE #2357: BUG: Capacity `get_stakable_amount_for` needs to be modified to allow overlapping freezes
`get_stakable_amount_for` uses `reducible_balance`, but `reducible_balance` excludes all freezes.

The result is that the Capacity staking uses freezing, but you cannot overlap with existing freezes (although freezes from elsewhere CAN overlap with Stake Freezes).

https://github.com/frequency-chain/frequency/blob/758d9516daae4504b4fb7b263a514cf573dd6bdd/pallets/capacity/src/lib.rs#L827-L839

## A/C
- [ ] I should be able to have 100 tokens, with 90 frozen due to other reasons, and still stake > 10 tokens for capacity

## Notes

- I think we want to use [`balance_freezable`](https://paritytech.github.io/polkadot-sdk/master/frame_support/traits/tokens/fungible/freeze/trait.Inspect.html#method.balance_freezable) AND subtract the existing stake to get the correct value.

shannonwells:
# Goal
The goal of this PR is 

Closes #2357 

# Discussion
- implement the change to allow staking on already frozen funds
- Update mocks to allow two freezes on accounts.
- Add and update some tests to validate that a time release lets you stake and vice versa.

## To verify
To do the manual verification for voting + staking:
- Run a local chain and manually verify that an account that has all its token staked can vote on a referendum.  The easiest way to create a referendum is: 

1. create and approve a council external proposal
2. Fast Track it (don't forget to make the voting period and delay small!).  DM me if you have forgotten how to do this, but it's in our Wiki, too.
3. and then mine blocks until it becomes a referendum. 
4. Vote with all of the staker's token - this is the default anyway.
5. Verify the extrinsic succeeds.  Bonus: View accounts and expand the account tab and you should see something like this:

Screenshot 2025-04-29 at 1 51 31 PM

The proposal was to transfer 200 UNIT from Alice to Eve, which is why there is a reserved balance.

Voting on a TC or Council proposal costs only milliUnits; you can't do conviction voting so it won't help you validate the change.

For convenience, if you want to use one of the default accounts, stake nearly all their token with the value `1152921500000000000`

**NOTES:** 
deferred a full e2e test that checks voting can occur with capacity-staked token, but this was verified manually on a local machine.

Capacity stakers may vote, but not _propose_ with their capacity-staked token. This is a feature of the propose extrinsic and can't be controlled by us.

# Checklist
- [x] Unit Tests updated?
- [x] e2e Tests added/updated?


e2e/capacity/staking.test.ts
```diff
@@ -293,4 +294,45 @@ describe('Capacity Staking Tests', function () {
...
+
+      const vestedTransferTx = ExtrinsicHelper.timeReleaseTransfer(fundingSource, vesterKeys, schedule);
+
+      await assert.doesNotReject(vestedTransferTx.signAndSend());
+      await assertFrozen(vesterKeys, 100n * DOLLARS);
+
+      await assert.doesNotReject(ExtrinsicHelper.stake(vesterKeys, providerId, 80n * DOLLARS).signAndSend());
+
+      const spendable = await getSpendableBalance(vesterKeys);
+      // after txn fees
```
shannonwells, line 333: Check is done this way to account for variable txn fees due to changing weights.

e2e/scaffolding/helpers.ts
```diff
@@ -664,7 +665,41 @@ export async function getFreeBalance(source: KeyringPair): Promise<bigint> {
...
+}
+
 export async function assertExtrinsicSucceededAndFeesPaid(chainEvents: any) {
   assert.notEqual(chainEvents['system.ExtrinsicSuccess'], undefined, 'should have returned an ExtrinsicSuccess event');
   assert.notEqual(chainEvents['balances.Withdraw'], undefined, 'should have returned a balances.Withdraw event');
 }
+
+
+export function getBlocksInMonthPeriod(blockTime: number, periodInMonths: number) {
+  const secondsPerMonth = 2592000; // Assuming 30 days in a month
```
shannonwells, line 684: moved from timeRelease.test.ts

e2e/sudo/sudo.test.ts
```diff
@@ -187,8 +187,8 @@ describe('Sudo required', function () {
         });
       });
 
-      describe('Capacity should not be affected by a hold being slashed', function () {
-        it('stake should fail when overlapping tokens are on hold', async function () {
+      describe('Capacity staking is affected by a hold being slashed', function () {
```
shannonwells, line 190: description wasn't quite correct

pallets/capacity/src/lib.rs
```diff
@@ -828,9 +825,12 @@ impl<T: Config> Pallet<T> {
 		staker: &T::AccountId,
 		proposed_amount: BalanceOf<T>,
 	) -> BalanceOf<T> {
-		let account_balance =
-			T::Currency::reducible_balance(staker, Preservation::Preserve, Fortitude::Polite);
-		let stakable_amount = account_balance.saturating_sub(T::MinimumTokenBalance::get());
+		let freezable_balance = T::Currency::balance_freezable(staker);
```
shannonwells, line 828: The meaty bit

deny.toml
```diff
@@ -107,6 +107,7 @@ allow = [
   "OpenSSL",
   "Zlib",
   "Unicode-3.0",
+  "CDLA-Permissive-2.0"
```
shannonwells, line 110: CI rejected a license, "CDLA-Permissive-2.0" for webpki-root, okayed for this list by @wilwade

pallets/capacity/src/tests/mock.rs
```diff
@@ -81,7 +81,7 @@ impl pallet_balances::Config for Test {
 	type AccountStore = System;
 	type WeightInfo = ();
 	type FreezeIdentifier = RuntimeFreezeReason;
-	type MaxFreezes = ConstU32<1>;
+	type MaxFreezes = ConstU32<2>;
```
wilwade, line 84: This is still set to 1 in a lot of the mocks, although it doesn't matter for those I guess.

pallets/capacity/src/lib.rs
```diff
@@ -828,9 +824,12 @@ impl<T: Config> Pallet<T> {
...
+		let current_staking_balance =
+			StakingAccountLedger::<T>::get(staker).unwrap_or_default().active;
```
wilwade, line 829: Does this increase the db reads for the call? I would think not as surely it is already in there somewhere right?

wilwade (approved):
This all works for me! Two questions, but they are not blocking and likely just informational.

Ran a bunch of manual tests (even some just adjacent) locally to double check things worked as expected:

* [x] Vote -> Stake
* [x] Stake -> Vote
* [x] Stake -> Not able to Stake > total (aka no overlapping capacity freezes)
* [x] Lock -> Stake -> Vote
* [x] Lock transfer -> stake -> Unlock transfer -> freeze lowered to stake amount
* [x] stake -> unstake partial -> freeze lowered

pallets/capacity/src/tests/mock.rs
shannonwells: I had originally set them all to 2, but I put it back for that reason.

pallets/capacity/src/lib.rs
shannonwells: I ran the benchmarks locally and it does not increase reads.

JoeCap08055 (approved):
- Reviewed code & tests
- Built locally
- Manually ran through propose/stake/vote scenario locally
👍🏻

Issue #2281, opened 2025-03-20
Create New JS/TS Library for Ethereum-Compatible Keys and Signatures


p5150j:
### Description
Develop a new JavaScript/TypeScript library that will handle Ethereum-compatible keys and signatures for our ecosystem. This library will be a foundation for multiple services.

Tasks

- Design library architecture and APIs
- Implement key generation compatible with Ethereum
- Implement signature creation and verification compatible with EIP-712
- Create functions for address derivation and validation
- Build comprehensive test suite
- Should work in the browser and NodeJS apps

Acceptance Criteria

- [ ]  Library successfully generates Ethereum-compatible keys
- [ ]  Library correctly signs and verifies messages using Ethereum standards
- [ ]  Address derivation matches Ethereum specifications
- [ ]  All functions have >95% test coverage
- [ ]  Documentation with examples for all main functions
- [ ]  Published as an NPM package with proper versioning

Engineering Notes
[To be filled by lead developer]

Blocked by #2278

Issue #2278, opened 2025-03-20
Research RLP/SCALE Encoding for Ethereum Compatibility


p5150j:
## Description
Research and document how RLP (Recursive Length Prefix) encoding used in Ethereum can work with SCALE (Simple Concatenated Aggregate Little-Endian) encoding in our system. This is foundational for creating Ethereum-compatible keys and signatures.

Tasks

- Research Ethereum's RLP encoding structure and specifications
- Document how SCALE encoding functions in our current system
- Identify compatibility issues between the two encoding methods
- Propose existing solutions for converting between encoding formats and propose a new one if there is no existing solution
- Create proof-of-concept code for encoding/decoding between formats

Acceptance Criteria

- [ ]  Comprehensive technical document explaining both encoding methods
- [ ]  Detailed analysis of compatibility challenges
- [ ]  Proposed solution with examples for encoding/decoding between formats
- [x]  Working proof-of-concept code demonstrating the proposed solution
- [ ]  Performance impact assessment of encoding/decoding operations

Engineering Notes

- We mostly like to know how RLP and SCALE conversion happens in new written  ETH proxy and Pallet Revive. This is important for signature verification since we have to encode the types to binary to be able to generate or verify a signature.

aramikm:
##  Eth RPC and pallet-revive
Eth RPC support 4 kinds of transactions with RLP encoding and decoding. 

1. Legacy
2. EIP-2930
3. EIP-1559
4. EIP-4844

All of the mentioned transaction types are unified into a `GenericTransaction` struct which is also supports RLP encoding/decoding.

To see the transaction data if we convert the raw transaction bytes into hex we might get something like the following example:

```hex
0x02f89384190f1b44088203e88203e8870570b6627d281c94c01ee7f10ea4af4673cfff62710e1d7792aba8f380a46057361d000000000000000000000000000000000000000000000000000000000000dce2c080a024151020a7c742820c6eb403b320d605ec36406f653a22bbb9b968e8de96d525a0507f0100cf327eec69815da9d1214def7bc7d2dcc5c5e4e2802fdb60fe11011b
```

using a tool such as https://tools.deth.net/tx-decoder we can get the following data

```json
{
  "chainId": "0x190f1b44",
  "nonce": "0x8",
  "maxPriorityFeePerGas": "0x3e8",
  "maxFeePerGas": "0x3e8",
  "gasLimit": "0x570b6627d281c",
[...]

PR #2385, opened 2025-05-02
init: eip-712 support

aramikm:
# Goal
The goal of this PR is to allow encoding and verification of EIP-712 signatures for our custom signed payloads.
Related to #2278 

# Discussion
- The frontend side of these changes will be implemented in #2281 
- I didn't know which chain ID to use so used the test chainID for pallet revive. We need to set these once we figure out those values.

# Checklist
- [x] Unit Tests added?
- [x] e2e Tests added?


common/primitives/src/handles.rs
```diff
@@ -58,6 +63,31 @@ impl<BlockNumber> ClaimHandlePayload<BlockNumber> {
...
 
+impl<BlockNumber> EIP712Encode for ClaimHandlePayload<BlockNumber>
+where
+	BlockNumber: Into<U256> + TryFrom<U256> + Copy,
+{
+	fn encode_eip_712(&self) -> Box<[u8]> {
+		lazy_static! {
+			// get prefix and domain separator
+			static ref PREFIX_DOMAIN_SEPARATOR: Box<[u8]> =
+				get_eip712_encoding_prefix("0xcccccccccccccccccccccccccccccccccccccccc");
```
aramikm, line 74: These are hardcoded for each payload and most probably would need to be replaced with precompile addresses when implemented.

common/primitives/src/signatures.rs
```diff
@@ -345,14 +360,57 @@ fn check_ethereum_signature<L: Lazy<[u8]>>(
...
+	lazy_static! {
+		// domain separator
+		static ref DOMAIN_TYPE_HASH: [u8; 32] = sp_io::hashing::keccak_256(
+			b"EIP712Domain(string name,string version,uint256 chainId,address verifyingContract)",
+		);
+
+		static ref DOMAIN_NAME: [u8; 32] = sp_io::hashing::keccak_256(b"Frequency");
+		static ref DOMAIN_VERSION: [u8; 32] = sp_io::hashing::keccak_256(b"1");
+		// TODO: USE correct chain ids for different networks
+		static ref CHAIN_ID: [u8; 32] = to_abi_compatible_number(420420420u32);
```
aramikm, line 380: using pallet_revive test chain ID for now.

e2e/scaffolding/helpers.ts
```diff
@@ -93,6 +97,50 @@ export function signPayload(keys: KeyringPair, data: Codec): MultiSignatureType
   }
 }
 
+/**
+ * Signing EIP-712 compatible signature for AddKeyData
+ * @param keys
+ * @param payload
+ */
+export async function signEip712AddKeyData(keys: Keypair, payload: AddKeyData): Promise<MultiSignatureType> {
```
aramikm, line 105: a temporary function until we implement the common library for Ethereum utilities

e2e/scaffolding/ethereum.ts
```diff
@@ -55,6 +55,14 @@ export function getUnifiedPublicKey(pair: KeyringPair): Uint8Array {
   return pair.publicKey;
 }
 
+export function reverseUnifiedAddressToEthereumAddress(unifiedAddress: Uint8Array) : string {
+  const hex = Buffer.from(unifiedAddress).toString("hex");
+  if (!hex.toLowerCase().endsWith("eeeeeeeeeeeeeeeeeeeeeeee")) {
```
JoeCap08055, line 60: nit: might be more readable as `'ee'.repeat(12)` to make it clear that we're looking for 12 bytes of 0xEE.
```suggestion
  if (!hex.toLowerCase().endsWith("ee".repeat(12))) {
```

pallets/msa/src/types.rs
```diff
@@ -30,6 +35,39 @@ pub struct AddKeyData<T: Config> {
 	pub new_public_key: T::AccountId,
 }
 
+impl<T: Config> EIP712Encode for AddKeyData<T> {
+	fn encode_eip_712(&self) -> Box<[u8]> {
+		lazy_static! {
+			// get prefix and domain separator
+			static ref PREFIX_DOMAIN_SEPARATOR: Box<[u8]> =
+				get_eip712_encoding_prefix("0xcccccccccccccccccccccccccccccccccccccccc");
```
JoeCap08055, line 43: nit: seems like this could be refactored, as it's in more than one place

aramikm: So it depends. I kept them separated because in EIP-712 the idea is that each one of them (at least for each pallet) we will have different contract addresses. We can make a decision that maybe we would like to simplify everything and just use the same contract address or not but that is a bigger decision.

tools/eth-migration/metamask.html
aramikm, line 1: To use this html you can run `npm install -g serve` and go to `tool/migrations` folder and run `serve`
and open `http://localhost:3000/metamask.html`

common/primitives/src/handles.rs
```diff
@@ -58,6 +63,31 @@ impl<BlockNumber> ClaimHandlePayload<BlockNumber> {
 	}
 }
 
+impl<BlockNumber> EIP712Encode for ClaimHandlePayload<BlockNumber>
+where
+	BlockNumber: Into<U256> + TryFrom<U256> + Copy,
+{
+	fn encode_eip_712(&self) -> Box<[u8]> {
+		lazy_static! {
```
mattheworris, line 71: Nice, I didn't know about this macro.

mattheworris (approved):
Nice work, 🚢 it!

pallets/passkey/src/tests.rs
```diff
@@ -543,3 +547,40 @@ fn pre_dispatch_with_exceeding_weight_should_fail() {
...
+		let payload = PasskeyPublicKey (from_hex("0x40a6836ea489047852d3f0297f8fe8ad6779793af4e9c6274c230c207b9b825026").unwrap().try_into().unwrap());
+		let encoded_payload = payload.encode_eip_712();
+
+		// following signature is generated via Metamask using the same input to check compatibility
+		let signature_raw = from_hex("0xbafaf5e21695a502b2d356b4558da35245aa1be7161f01a5f0224fbfdf85b5c52898fc495ab1ca9b68c3b07e23d31a5fe1686165344b22bc14201f293d54f36b1b").expect("Should convert");
+		let unified_signature = UnifiedSignature::from(ecdsa::Signature::from_raw(
+			signature_raw.try_into().expect("should convert"),
+		));
+
+		// 0x509540919faacf9ab52146c9aa40db68172d83777250b28e4679176e49ccdd9fa213197dc0666e85529d6c9dda579c1295d61c417f01505765481e89a4016f02
```
shannonwells, line 576: What is the purpose of this comment?

aramikm: That is the non compressed public key of keypair used in metamask for signature. It is not required but it is helpful to have. I'll this in comments

shannonwells (approved):
Ship it...I read through the code, and I ran the test with the Metamask example, looking at all the EIP-712 payload info & signatures

JoeCap08055 (approved):
Nice work. Approved.

PR #2402, opened 2025-05-12
feat: implement pallet_msa::withdraw_tokens extrinsic

### LINKED ISSUE #2388: Tokens held by MSA-owned accounts can be withdrawn
## Details
Related to #2351, a holder of a control key for an MSA should be able to "withdraw" tokens held by the associated address to another address

## A/C
- Only the complete balance held by the MSA account may be withdrawn
- The withdrawal MUST be approved by a Control Key of the MSA that is the source of the withdrawal
- The withdrawal MUST be from the holder of the private key of the address the tokens are withdrawing to
- The MSA Control Key MUST approve the specific address the withdrawal is going to
- MSAs holding tokens may not be retired until tokens are withdrawn
- Allow Withdrawal as free extrinsic

### Options

Winner: **Withdrawal Tx is submitted by the token withdrawal target (with signature from MSA Control Key)**

1. ~Withdrawal Tx is submitted by the MSA Control Key (with signature from withdrawal target)~ Don't want some control keys to have to do the act of moving tokens
2. Withdrawal Tx is submitted by the token withdrawal target (with signature from MSA Control Key)
3. ~Withdrawal Tx takes a long-term, persistent approval step submitted by the MSA Control Key and then a withdrawal step submitted by the token withdrawal target~ Control Key permissions is effectively the same
4. ~Withdrawal Tx takes a one-time per withdrawal approval step submitted by the MSA Control Key and then a withdrawal step submitted by the token withdrawal target~ Uselessly complex
5. ~Withdrawal Tx via Control Key permissions (see below)~ Future

## Withdrawal Tx via Control Key permissions
- Add permissions to control keys (specifically payout permission, but future support for capacity only permission setting)
- Add way to add generic control key to MSA with permissions (non-permission version exists on chain, not in FA yet)
- Withdrawal as free extrinsic with fees paid by the MSA balance

## Discussion
(to be moved to A/C once requirements are fully determined):
- Q: _Any restrictions on the destination address?_ **Nope**
- Q: _Do we want to require a signature to prove control of the destination address?_ **Is transactor, so has tx signature**
- It's currently possible to send tokens to an MSA that does not exist (either not-yet-created, or retired)
    - Q: _Do we want to prohibit, in any way, sending tokens to a retired MSA?_ **No**
    - Q: _Do we want to prohibit retiring an MSA that has a token balance (ie, require to withdraw tokens before retiring)?_ **Yes**

JoeCap08055:
# Goal
The goal of this PR is to implement an initia `withdraw_tokens` extrinsic that allows MSAs holding tokens to authorize their tokens to be withdrawn to a specified account.

Closes #2388 

# Discussion

# Checklist
- [ ] Updated Pallet Readme?
- [ ] Design doc(s) updated?
- [ ] Unit Tests added?
- [x] e2e Tests added?
- [ ] Benchmarks added?

### DIFF
--- a/e2e/msa/msaTokens.test.ts
+++ b/e2e/msa/msaTokens.test.ts
@@ -1,14 +1,24 @@
-/* eslint-disable mocha/no-skipped-tests */
 import '@frequency-chain/api-augment';
 import assert from 'assert';
-import { ExtrinsicHelper } from '../scaffolding/extrinsicHelpers';
-import { ethereumAddressToKeyringPair } from '../scaffolding/ethereum';
+import { AddKeyData, ExtrinsicHelper } from '../scaffolding/extrinsicHelpers';
+import { ethereumAddressToKeyringPair, getUnifiedAddress, getUnifiedPublicKey } from '../scaffolding/ethereum';
 import { getFundingSource } from '../scaffolding/funding';
 import { H160 } from '@polkadot/types/interfaces';
 import { bnToU8a, hexToU8a, stringToU8a } from '@polkadot/util';
 import { KeyringPair } from '@polkadot/keyring/types';
 import { keccak256AsU8a } from '@polkadot/util-crypto';
-import { getExistentialDeposit } from '../scaffolding/helpers';
+import {
+  CENTS,
+  createAndFundKeypair,
+  createKeys,
+  DOLLARS,
+  generateAddKeyPayload,
+  getExistentialDeposit,
+  signPayloadSr25519,
+  Sr25519Signature,
+} from '../scaffolding/helpers';
+import { u64 } from '@polkadot/types';
+import { Codec } from '@polkadot/types/types';
 
 const fundingSource = getFundingSource(import.meta.url);
 
@@ -81,7 +91,7 @@ describe('MSAs Holding Tokens', function () {
 
   describe('Send tokens to MSA', function () {
     it('should send tokens to the MSA', async function () {
-      const ed = await getExistentialDeposit();
+      const ed = getExistentialDeposit();
       const transferAmount = 1n + ed;
       let accountData = await ExtrinsicHelper.getAccountInfo(ethKeys);
       const initialBalance = accountData.data.free.toBigInt();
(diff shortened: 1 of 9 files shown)


pallets/msa/src/lib.rs
```diff
@@ -905,6 +924,99 @@ pub mod pallet {
...
+			// Make sure either:
+			// - Balance to be transferred > existential deposit, OR
+			// - Destination account balance  + ammount to be transferred > existential deposit
+			let amount_alone_is_insufficient = msa_balance < ed;
+			let dest_will_be_reaped = if amount_alone_is_insufficient {
+				let dest_balance = T::Currency::total_balance(&public_key);
+				(dest_balance + msa_balance) < ed
+			} else {
+				amount_alone_is_insufficient
+			};
```
aramikm, line 1007: these conditions are written in a confusing way

```
let y = if x {
    z
} else {
   x
}
```

pallets/msa/src/lib.rs
```diff
@@ -905,6 +924,99 @@ pub mod pallet {
...
+		///
+		#[pallet::call_index(14)]
+		#[pallet::weight(T::WeightInfo::withdraw_tokens())]
+		pub fn withdraw_tokens(
+			origin: OriginFor<T>,
+			msa_owner_public_key: T::AccountId,
+			msa_owner_proof: MultiSignature,
+			authorization_payload: AddKeyData<T>,
+		) -> DispatchResult {
+			let public_key = ensure_signed(origin)?;
```
aramikm, line 961: this must not be a control key since we don not allow control keys to have any tokens.

pallets/msa/src/lib.rs
```diff
@@ -905,6 +924,99 @@ pub mod pallet {
...
+		/// * [`Error::InsufficientBalanceToFundDestination`] - the transfer would result in an account that would be reaped due to existential balance requirements
+		/// * [`Error::UnexpectedTokenTransferError`] - the balances pallet threw an error during the attempted transfer
+		///
+		#[pallet::call_index(14)]
+		#[pallet::weight(T::WeightInfo::withdraw_tokens())]
+		pub fn withdraw_tokens(
+			origin: OriginFor<T>,
+			msa_owner_public_key: T::AccountId,
+			msa_owner_proof: MultiSignature,
+			authorization_payload: AddKeyData<T>,
```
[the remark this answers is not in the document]
JoeCap08055: Well, I don't think it simplifies anything, but you're right; a type alias would be more readable & more clear about the intent. 👍🏻

pallets/msa/src/lib.rs
JoeCap08055: I'm wondering if these checks are overkill... I can't even engineer a test for them, because it's impossible to fund an account with < ED. Is there any case where an address could be left with dust where it *wouldn't* be reaped? Is there an edge case where an account is scheduled to be reaped, but hasn't yet? I know there is some special handling if free balance < ED but there is locked balance as well, but since these accounts have no control keys, it would be impossible to lock a balance in them.

pallets/msa/src/lib.rs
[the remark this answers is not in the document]
aramikm: Yeah it doesn't appear the effort justifies the benefits. I would just keep it as is if not already invested in it.

pallets/msa/src/benchmarking.rs
```diff
@@ -369,6 +369,40 @@ mod benchmarks {
...
+		// let balance = <<T as Config>::Currency as Inspect<<T as frame_system::Config>::AccountId>>::Balance.from(10_000_000u128);
+		let balance = <T as Config>::Currency::minimum_balance();
+		T::Currency::set_balance(&msa_account_id, balance);
+		assert_eq!(T::Currency::balance(&msa_account_id), balance);
+
+		let (add_key_payload, _, new_account_id) = add_key_payload_and_signature::<T>(msa_id);
+
+		let encoded_add_key_payload = wrap_binary_data(add_key_payload.encode());
+		let owner_signature =
+			MultiSignature::Sr25519(msa_key_pair.sign(&encoded_add_key_payload).unwrap().into());
```
aramikm, line 392: nit: Since we are trying to move to Ethereum based keys. If it's not a lot of effort it might make sense to use ECDSA keys and eip712 signatures going forward but not a big deal if it's a bigger lift. Please let me know if you have any questions.

pallets/msa/src/lib.rs
aramikm: I think I've seen some code that indicates that there might be some edge cases that the account might have less than ED amount but that might be legacy stuff.
@enddynayn  do you have anything to add in here?

pallets/msa/src/lib.rs
```diff
@@ -905,6 +929,66 @@ pub mod pallet {
...
+		///
+		/// The `origin` parameter represents the account from which the function is called and must be the account receiving the tokens.
+		///
+		/// The function requires one signature: `msa_owner_proof`, which serve as proof that the owner of the MSA authorizes the transaction.
+		///
+		/// The necessary information for the withdrawal authorization, the destination account and the MSA ID, are contained in the `authorization_payload` parameter of type [AddKeyData].
+		/// It also contains an expiration block number for the proof, ensuring it is valid and must be greater than the current block.
+		///
+		/// # Events
+		/// * `pallet_balances::Event::<T,I>::Transfer` - Transfer token event
```
JoeCap08055, line 943: Anyone know how to make this a doc link correctly (ie, to another pallet)? If I add the brackets to make this a link, it fails linting as broken doc link.

pallets/msa/src/lib.rs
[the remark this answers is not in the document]
JoeCap08055: I created `AuthorizedKeyData`, with the same structure but appropriately-named members. Should be good for any future extrinsic for which we want to sign over an authorized key.

pallets/msa/src/lib.rs
```diff
@@ -1501,7 +1595,7 @@ impl<T: Config> Pallet<T> {
 		}
 
 		// Add the newest signature if we are not the first
-		if pointer.count != 0 {
+		if pointer.count != 0 && current_block.le(&pointer.newest_expires_at) {
```
JoeCap08055, line 1598: @wilwade while examining the signature registry methods in order to split checking (read) from recording (read/write) signatures, I came up with this tiny optimization--we don't need to add the previously-most-recent signature to the larger list if it's already expired.

Double-check me on that?

pallets/msa/src/lib.rs
```diff
@@ -1501,7 +1595,7 @@ impl<T: Config> Pallet<T> {
 		}
 
 		// Add the newest signature if we are not the first
```
wilwade, line 1597:
```suggestion
		// Add the newest signature if we are not the first AND that signature is still not expired
```

pallets/msa/src/lib.rs
wilwade: Yes. It took me a bit, but it works. Added a change to the comment above it to make it clear what it is doing, but that is optional.

wilwade: I think I take this back.

It doesn't work and apparently we don't have a test for it. I think...

wilwade: For others reading this, this might help:

The comment on the top of `register_signature` says this:
```
1,2 (oldest)
2,3
3,4
4 (pointer storage)
```

I thought Joe's idea is that when 5 comes in, if 4 is ALREADY expired, instead of adding `4,5` and `5 (pointer storage)` we ONLY need to keep `5 (pointer storage)`

However, this "breaks the loop" So does it work or not?

The answer is yes. Because: The comment is wrong/misleading. It should be

```
/// Example list:
/// - `1,2 (oldest)`
/// - `2,3`
/// - `3,4` (newest, is only in pointer storage)`
```
So Joe's actual idea is that when 5 comes in, if 4 is ALREADY expired, instead of adding `3,4` and `4,5 (pointer storage)` we ONLY update the pointer storage to `3,5 (pointer storage)`

That keeps the loop intact.

So @JoeCap08055 you are good to make the change, but please update the comments a bit to make it more clear (and fix my bad original commenting)

pallets/msa/src/lib.rs
JoeCap08055: After some discussion, I'm going to remove the ED check for now. It saves us 1 read (destination balance), and we can't envision any edge cases where an MSA account could possibly have < ED in free balance.

pallets/msa/src/lib.rs
JoeCap08055: @wilwade (and maybe this is a bit out of scope here, but since we're messing with this algorithm anyway...) I notice that the signature storage, once it fills up, will from that point on always contain the max # of entries, even once they've all expired. Have we considered the comparative costs of maintaining that steady-state storage for all time, versus opportunistically pruning any time a new signature comes in?

(I understand that the pruning cost *could* end up being expensive for a single "free" transaction if we pruned all expired entries. Possibly we could strike a balance, say... any time a new signature comes in, prune the `N` oldest instead of just the oldest, which would lead to a, if not empty, at least lower steady-state storage cost over time...)

If this seems of interest I could write up an issue if we don't want to include here...

JoeCap08055: (Also, I'm going to revert my change: since, currently, storage will always => max entries, and in order to make my change work, it still needs to perform a write (update instead of insert, but same cost), there seems no need to mess with this method unless we're going to address my comment above.)

e2e/msa/msaTokens.test.ts
```diff
@@ -97,9 +129,207 @@ describe('MSAs Holding Tokens', function () {
...
+        startingBalance.toBigInt() + TRANSFER_AMOUNT === endingBalance.toBigInt(),
+        'balance of recieve should have increased by the transfer amount minus fee'
+      );
+    });
+
+    it('should fail for duplicate signature submission (MsaOwnershipInvalidSignature)', async function () {
+      // In order to test this, we need to create a new keypair and fund it, because otherwise the nonce will
+      // be the same for both transactions (and, because we're using Edcs signatures, the signature will be the same).
+      // Not sure exactly what happens in this case, but it seems to be that the second transaction is siliently dropped
+      // by the node, but the status call back in polkadot.js still resolves (ie, gets 'isInBlock' or 'isFinalized')
```
JoeCap08055, line 310:
```suggestion
      // In order to test this, we need to create a new keypair and fund it, because otherwise:
      // - the nonce will be the same for both transactions (ie, zero)
      // - in our E2E harness we submit transactions as immortal by default `{ era: 0 }`, the era will be the same
      // - ...resulting in an identical transaction hash
      // - therefore, the node processes this as a duplicate transaction submission and simply returns
      //    the prior (successful) event stream, rather than performing new transaction validations
```

pallets/msa/src/lib.rs
```diff
@@ -899,6 +916,67 @@ pub mod pallet {
...
+			Self::register_signature(&msa_owner_proof, authorization_payload.expiration)?;
+
+			let msa_id = authorization_payload.msa_id;
+
+			// - Get account address for MSA
+			let msa_address = Self::msa_id_to_eth_address(msa_id);
+
+			// - Convert to AccountId
+			let mut bytes = &EthereumAddressMapper::to_bytes32(&msa_address.0)[..];
+			let msa_account_id = T::AccountId::decode(&mut bytes).unwrap();
```
aramikm, line 959: Must not do unwrap here

pallets/msa/src/weights.rs
```diff
@@ -238,6 +239,13 @@ impl<T: frame_system::Config> WeightInfo for SubstrateWeight<T> {
 		// Minimum execution time: 4_000_000 picoseconds.
 		Weight::from_parts(8_000_000, 0)
 	}
+	fn withdraw_tokens() -> Weight {
+		// Proof Size summary in bytes:
+		//  Measured:  `0`
+		//  Estimated: `0`
+		// Minimum execution time: 4_000_000 picoseconds.
+		Weight::from_parts(8_000_000, 0)
```
aramikm, line 247: DB read write don't match. Need to run benchmarks for accurate weights.

aramikm (commented):
Great effort but there are some blockers.

pallets/msa/src/lib.rs
JoeCap08055: Fixed.

.github/workflows/verify-pr-commit.yml
```diff
@@ -243,7 +243,8 @@ jobs:
           cargo install --force --locked cargo-deny@0.17.0
           cargo generate-lockfile
       - name: Run Cargo Deny
-        run: cargo deny check --hide-inclusion-graph -c deny.toml
+        # TODO: Remove NIGHTLY once we upgrade to Rust 1.85+
+        run: cargo +${{env.NIGHTLY_TOOLCHAIN}} deny check --hide-inclusion-graph -c deny.toml
```
mattheworris, line 247: nit: Since nightly technically changes the check of the lock file, recommend using the following patch. 

However, it may be preferred to move to 1.85 soon after this PR to use `base64ct` at v1.8.0, instead of pinning at v1.7.2. [Currently used version v1.7.3 doesn't have a branch/tag 🤷, doesn't seem like any fundamental differences between .2. and .3 ] [Note: `base64ct:v1.8.0` is the crate that requires `edition2024` which is supported as of Rust 1.85]

```toml
[patch.crates-io]
base64ct = { git = "https://github.com/RustCrypto/formats.git", tag = "base64ct/v1.7.2", package = "base64ct" }
```

pallets/msa/src/lib.rs
```diff
@@ -899,6 +916,70 @@ pub mod pallet {
...
+
+			let msa_id = authorization_payload.msa_id;
+
+			// - Get account address for MSA
+			let msa_address = Self::msa_id_to_eth_address(msa_id);
+
+			// - Convert to AccountId
+			let mut bytes = &EthereumAddressMapper::to_bytes32(&msa_address.0)[..];
+			let msa_account_id = T::AccountId::decode(&mut bytes).map_err(|_| {
+				log::error!("Failed to decode MSA account ID from Ethereum address");
```
enddynayn, line 960: This is not possible?

pallets/msa/src/tests/msa_token_tests.rs
```diff
@@ -0,0 +1,105 @@
...
+
+use common_primitives::{
+	msa::H160,
+	signatures::{AccountAddressMapper, EthereumAddressMapper},
+};
+
+use pallet_balances::Event as BalancesEvent;
+
+#[test]
+fn it_succeeds_when_balance_is_sufficient() {
```
enddynayn, line 21: I suggest adding another another test when the balance is 0.

.github/workflows/verify-pr-commit.yml
JoeCap08055: Applied; reverted the change to use nightly

pallets/msa/src/tests/msa_token_tests.rs
JoeCap08055: I tried; you can't set the balance to zero unless you do it in the genesis, and even then I'm not sure the account wouldn't get reaped before you could execute test method. Didn't seem worth it to go to all that effort.

enddynayn (approved):
Nice work here 👍 ! Definitely not a simple PR.

aramikm (approved):

PR #2403, opened 2025-05-12
Subsidize adding an Ethereum compatible control key

### LINKED ISSUE #2286: Make add_key Free On-Chain for Eligible Keys
### Description
Modify on-chain logic to make the `add_key` operation free for eligible keys, reducing friction in the migration process.
In the first phase of ETH key migrations we need to allow users to add one ETH compatible key as a free transaction.

Tasks

- Implement fee waiver logic in on-chain code
- Create validation mechanisms to prevent abuse
- Test edge cases for eligibility determination

Acceptance Criteria

- [x]  `add_key` operation is free for all eligible keys (a tiny capacity amount)
- [x]  Non-eligible keys still require appropriate fees
- [x]  Abuse prevention mechanisms are in place
- [x]  Transaction processing correctly handles fee-free operations (a tiny capacity amount)
- [x]  All edge cases tested and working correctly
- [x]  Documentation updated to explain eligibility criteria

Engineering Notes
- eligibility: Any MsaId that has only one key before a certain block_number or msa_id (this block number or msa id would be the cutover window which all new Accounts are created with a ETH compatible key after that). This cutover value or criteria should be configurable

shannonwells:
# Goal
The goal of this PR is to allow adding a second key to be a <del>free</del> subsidized capacity transaction under specific circumstances.  This change reduces the capacity fee by about 70% when there is one key, and the new key is Ethereum-compatible.

Most of #2286 

# Discussion
Note on something easy to miss:  the expiration block will default to 0, which means this feature is off by default, until the value is set to a block in the future.  There is no on-chain monitoring of the usage of this feature.

# Checklist
- [x] As Sudo or Council, can set the expiration block for the <del>free</del> subsidized key addition period.
- [x] Updated frequency-tx-payment SignedExtension to use a filter for the specific call
- [x] frequency-tx-payment uses an Msa call to determine eligibility
- [x] Updated Pallet Readme?
- [x] Unit Tests added?
- [x] e2e Tests added?
- [x] Benchmarks added?

### not applicable
<del>
- [ ] Design doc(s) updated?
</del>


e2e/package.json
```diff
@@ -46,5 +46,8 @@
     "tsx": "^4.19.3",
     "typescript": "^5.8.3",
     "typescript-eslint": "^8.30.1"
+  },
+  "optionalDependencies": {
```
shannonwells, line 50: Needed for running e2e tests locally on Ubuntu

e2e/msa/msaKeyManagement.test.ts
```diff
@@ -132,9 +134,10 @@ describe('MSA Key management', function () {
       newSig = signPayloadSr25519(secondaryKey, addKeyData);
       const addPublicKeyOp = ExtrinsicHelper.addPublicKeyToMsa(keys, ownerSig, newSig, payload);
 
-      const { target: publicKeyEvents } = await addPublicKeyOp.fundAndSend(fundingSource);
+      const { target: publicKeyEvents, eventMap: chainEvents } = await addPublicKeyOp.fundAndSend(fundingSource);
 
       assert.notEqual(publicKeyEvents, undefined, 'should have added public key');
+      assertExtrinsicSucceededAndFeesPaid(chainEvents);
```
shannonwells, line 140: Checks the opposite case of the new test below.

pallets/frequency-tx-payment/src/lib.rs
```diff
@@ -423,12 +443,27 @@ where
...
 				self.withdraw_capacity_fee(who, calls, len),
-			_ => self.withdraw_token_fee(who, call, info, len, self.tip(call)),
+			_ => {
+				if let Some((owner_account_id, msa_id)) = T::MsaCallFilter::get_add_key_data(call) {
+					if T::MsaKeyProvider::key_eligible_for_free_addition(
+						owner_account_id.clone(),
+						msa_id,
+					) {
+						let tip = self.tip(call);
+						let fee = pallet_transaction_payment::Pallet::<T>::compute_fee(
```
shannonwells, line 459: Unsure if this should be forced to zero.

e2e/msa/msaKeyManagement.test.ts
```diff
@@ -170,6 +174,41 @@ describe('MSA Key management', function () {
...
+      const { target: event, eventMap: chainEvents } = await op.fundAndSend(fundingSource);
+
+      // assert
+      assert.notEqual(event, undefined, 'should have added public key');
+      assert.notEqual(
+        chainEvents['system.ExtrinsicSuccess'],
+        undefined,
+        'should have returned an ExtrinsicSuccess event'
+      );
+      assert.equal(chainEvents['balances.Withdraw'], undefined, 'should not have withdrawn a balance');
```
JoeCap08055, line 208: would it be more correct to look for the event `transactionPayment.TransactionFeePaid`?

e2e/msa/msaTokens.test.ts
```diff
@@ -52,8 +52,10 @@ describe('MSAs Holding Tokens', function () {
   describe('getEthereumAddressForMsaId', function () {
     it('should return the correct address for a given MSA ID', async function () {
       const expectedAddress = CHECKSUMMED_ETH_ADDR_1234.toLowerCase();
-      const { accountId, accountIdChecksummed } =
-        await ExtrinsicHelper.apiPromise.call.msaRuntimeApi.getEthereumAddressForMsaId(MSA_ID_1234);
+``      const result: any = await ExtrinsicHelper.apiPromise.call.msaRuntimeApi.getEthereumAddressForMsaId(MSA_ID_1234);
+      const accountId = result?.accountId;
+      const accountIdChecksummed = result?.accountIdChecksummed;
+
```
JoeCap08055, line 58: Looks like some errant backticks here...?

e2e/scaffolding/helpers.ts
```diff
@@ -478,6 +478,11 @@ export async function setEpochLength(keys: KeyringPair, epochLength: number): Pr
   }
 }
 
+export async function setFreeKeyAddExpirationBLock(keys: KeyringPair, expires_at: number): Promise<void> {
```
JoeCap08055, line 481:
```suggestion
export async function setFreeKeyAddExpirationBlock(keys: KeyringPair, expires_at: number): Promise<void> {
```

e2e/msa/msaKeyManagement.test.ts
shannonwells: I thought so too, but that event isn't in the event list.

shannonwells: For some reason I was not seeing that event before, but added, checks for both now.

e2e/msa/msaTokens.test.ts
shannonwells: fixed

e2e/scaffolding/helpers.ts
shannonwells: fixed

pallets/frequency-tx-payment/src/lib.rs
aramikm: I think we can test it maybe manually to see what happens with the fee or with just zero. The idea is that the caller should not have any token and it should work.
The other question that comes up is how about the capacity? Should it be free also if called with capacity? My impression was that since these are going to be called by custodial wallet under the hood we probably want to make it free only for capacity calls.

shannonwells:
> I think we can test it maybe manually to see what happens with the fee or with just zero. The idea is that the caller should not have any token and it should work.

It definitely does not withdraw a fee, but I have not tried using a non-token account so I don't know if it would be rejected for that reason.

pallets/frequency-tx-payment/src/lib.rs
JoeCap08055: Is the custodial wallet a Provider? I think maybe not...? They shouldn't have to be a Provider, they currently don't actually execute any calls on-chain directly, so for the cutover it might be better to have this be free for token-based calls, not Capacity.

pallets/msa/src/lib.rs
```diff
@@ -1674,6 +1704,25 @@ impl<T: Config> SchemaGrantValidator<BlockNumberFor<T>> for Pallet<T> {
...
+	// - It has only one key associated with it
+	fn key_eligible_for_free_addition(old_key: Self::AccountId, msa_id: MessageSourceId) -> bool {
+		if let Some(stored_msa_id) = Self::get_msa_id(&old_key) {
+			let block = frame_system::Pallet::<T>::block_number();
+
+			return FreeKeyAddExpirationBlock::<T>::get().gt(&block) &&
+				stored_msa_id == msa_id &&
+				PublicKeyCountForMsaId::<T>::get(msa_id).eq(&1u8);
+		}
+		false
```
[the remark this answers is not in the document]
JoeCap08055: I'm fuzzy on this; remind me--does this get called as part of a Signed Extension? Regardless, do we want to optimize this so that tests that require storage reads all come *after* other check, to minimize cost?
1. In the free case, ALL checks must pass, so it doesn't matter what order they are done; all reads will be incurred
2. In the non-free case, we might bail early, so saving some reads will ultimately save the transactor on cost
    a. Do we really care that much? 99% of non-free use of this call is via Capacity, which doesn't take this path anyway (though there is some other discussion around this...)
    b. Presumably, this path will be removed once the cutover is complete, so maybe not that important to optimize to that degree (but if it's simple, why not optimize anyway?)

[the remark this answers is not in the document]
shannonwells: Yes this is called in SignedExtension, hopefully soon to be TransactionExtension.

shannonwells:
putting back to draft while I get all tests passing, conflicts resolved, etc

e2e/msa/msaTokens.test.ts
```diff
@@ -51,8 +51,10 @@ describe('MSAs Holding Tokens', function () {
   describe('getEthereumAddressForMsaId', function () {
     it('should return the correct address for a given MSA ID', async function () {
       const expectedAddress = CHECKSUMMED_ETH_ADDR_1234.toLowerCase();
-      const { accountId, accountIdChecksummed } =
-        await ExtrinsicHelper.apiPromise.call.msaRuntimeApi.getEthereumAddressForMsaId(MSA_ID_1234);
+      const result: any = await ExtrinsicHelper.apiPromise.call.msaRuntimeApi.getEthereumAddressForMsaId(MSA_ID_1234);
```
shannonwells, line 54: lint was failing on this :roll_eyes:

common/primitives/src/msa.rs
```diff
@@ -295,7 +295,19 @@ pub trait SchemaGrantValidator<BlockNumber> {
 	) -> DispatchResult;
 }
 
-/// RPC Response for getting getting MSA keys
+/// A trait that allows checking whether adding a key would be free
+pub trait MsaKeyProvider {
+	/// the type to use for looking up keys in storage.
+	type AccountId;
+	/// Returns whether adding `new_key` to `msa_id` would be free
+	fn key_eligible_for_free_addition(
```
shannonwells, line 303: I considered having this accepting `AddKeyData`, but `AddKeyData` takes a `Config` generic, and you wind up with dueling Configs (between frequency-tx-payment and msa).

e2e/capacity/transactions.test.ts
```diff
@@ -53,16 +53,6 @@ describe('Capacity Transactions', function () {
         assert.notEqual(schemaId, undefined, 'setup should populate schemaId');
       });
 
-      function getCapacityFee(chainEvents: EventMap): bigint {
```
shannonwells, line 56: moved to extrinsicHelpers.

pallets/frequency-tx-payment/src/tests/mock.rs
```diff
@@ -252,6 +260,8 @@ impl GetStableWeight<RuntimeCall, Weight> for TestCapacityCalls {
 			RuntimeCall::Balances(BalancesCall::transfer_allow_death { .. }) =>
 				Some(Weight::from_parts(11, 0)),
 			RuntimeCall::Msa(pallet_msa::Call::create { .. }) => Some(Weight::from_parts(12, 0)),
+			RuntimeCall::Msa(pallet_msa::Call::add_public_key_to_msa { .. }) =>
```
shannonwells, line 263: This is for a more realistic test of the call; otherwise it just errors saying the call isn't a Capacity eligible call.

pallets/frequency-tx-payment/src/tests/mock.rs
```diff
@@ -116,7 +116,13 @@ pub type MaximumCapacityBatchLength = ConstU8<10>;
 
 pub struct TestAccountId;
 impl Convert<u64, AccountId> for TestAccountId {
-	fn convert(_x: u64) -> AccountId32 {
+	fn convert(x: u64) -> AccountId32 {
+		// force this key to be an ethereum-compatible key
+		if x == 999 {
```
shannonwells, line 121: Set the converter to look for a magic number rather than doing numerical gymnastics.

pallets/frequency-tx-payment/src/lib.rs
```diff
@@ -441,17 +461,36 @@ where
 	) -> Result<(BalanceOf<T>, InitialPayment<T>), TransactionValidityError> {
 		let mut calls_weight_sum = Weight::zero();
 		for call in calls {
-			let call_weight = T::CapacityCalls::get_stable_weight(call)
-				.ok_or(ChargeFrqTransactionPaymentError::CallIsNotCapacityEligible.into())?;
+			let call_weight = if self.call_is_adding_eligible_key_to_msa(call) {
+				// "free-ish"
+				Weight::from_parts(1_000, 0)
```
shannonwells, line 466: This doesn't seem to work out to much of a discount. See e2e test.

e2e/msa/addKeyLowCost.ethereum.test.ts
```diff
@@ -0,0 +1,123 @@
...
+      providerKeys,
+      ExtrinsicHelper.api.events.msa.PublicKeyAdded
+    );
+
+    // ACT pay with capacity to add a third key
+    const { eventMap: eventMap2 } = await addThirdKeyOp.payWithCapacity();
+    assertEvent(eventMap2, 'msa.PublicKeyAdded');
+    const thirdKeyCapacityFee = ExtrinsicHelper.getCapacityFee(eventMap2);
+    // 4260363n vs
+    // 3258176n
```
shannonwells, line 119: This doesn't seem like much of a discount... but look at the Weight returned [here](https://github.com/frequency-chain/frequency/pull/2403/files#diff-4a481e841703b9d88a0508a5d2e834def9a2fa7bcea617cdfc4243796067e82cR466)

pallets/frequency-tx-payment/src/types.rs
```diff
@@ -10,3 +10,10 @@ pub trait GetStableWeight<RuntimeCall, Weight> {
 	/// e.g. in case of `pay_with_capacity` and `pay_with_capacity_batch_all`
 	fn get_inner_calls(outer_call: &RuntimeCall) -> Option<Vec<&RuntimeCall>>;
 }
+
+/// Filters calls that match Msa::add_public_key_to_msa
+pub trait GetAddKeyData<RuntimeCall, AccountId, MessageSourceId> {
+	/// If the fall matches Msa::add_public_key_to_msa, return the owner account id and msa id
```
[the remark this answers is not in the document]
shannonwells: fixed

pallets/frequency-tx-payment/src/lib.rs
shannonwells: NB for future us:  they are a Provider, they don't normally make capacity calls, but we decided to make this a subsidized capacity call for better control over who's calling & spam prevention.

pallets/frequency-tx-payment/src/types.rs
```diff
@@ -10,3 +10,10 @@ pub trait GetStableWeight<RuntimeCall, Weight> {
 	/// e.g. in case of `pay_with_capacity` and `pay_with_capacity_batch_all`
 	fn get_inner_calls(outer_call: &RuntimeCall) -> Option<Vec<&RuntimeCall>>;
 }
+
+/// Filters calls that match Msa::add_public_key_to_msa
+pub trait GetAddKeyData<RuntimeCall, AccountId, MessageSourceId> {
+	/// If the xall matches Msa::add_public_key_to_msa, return the owner account id and msa id
```
JoeCap08055, line 16:
```suggestion
	/// If the call matches Msa::add_public_key_to_msa, return the owner account id and msa id
```

runtime/frequency/src/lib.rs
```diff
@@ -337,6 +337,28 @@ impl Contains<RuntimeCall> for PasskeyCallFilter {
...
 }
 
+pub struct MsaCallFilter;
+use pallet_frequency_tx_payment::types::GetAddKeyData;
+impl GetAddKeyData<RuntimeCall, AccountId, MessageSourceId> for MsaCallFilter {
+	fn get_add_key_data(call: &RuntimeCall) -> Option<(AccountId, AccountId, MessageSourceId)> {
+		match call {
+			#[cfg(feature = "runtime-benchmarks")]
+			RuntimeCall::System(frame_system::Call::remark { .. }) => None,
+
```
JoeCap08055, line 347: curious, why do we need this if the default arm of the match returns `None` anyway?

shannonwells: Good catch.  This originally returned something else, and during a debugging session I set it to None then forgot about it.  I think it's fine just to delete it but let me double check the benchmarks.

pallets/frequency-tx-payment/src/lib.rs
shannonwells: Per conversation in Wednesday's meeting, rather than worrying with the weight calculation at all, we will just halve the calculated capacity fee.

pallets/frequency-tx-payment/src/lib.rs
```diff
@@ -440,18 +452,48 @@ where
...
 
 		Ok((fee.into(), InitialPayment::Capacity))
 	}
 
+	// Reduction for subsidized calls.
+	fn subsidized_calls_reduction(len: usize, eligible_call_weight: Weight) -> BalanceOf<T> {
+		if eligible_call_weight.is_zero() {
+			0u32.into()
+		} else {
+			let reduction: Permill = Permill::from_percent(70u32);
```
shannonwells, line 479: See e2e test for how much this reduces cost.

pallets/frequency-tx-payment/src/tests/pallet_tests.rs
```diff
@@ -1010,3 +1011,76 @@ fn can_withdraw_fee_errors_on_token_txn_witout_enough_funds() {
...
+		.execute_with(|| {
+			let msa_id = 2u64;
+			let owner_id = 2u64;
+			let pay_with_capacity_add_public_key_call =
+				generate_add_public_key_call(msa_id, owner_id);
+			// ask if the returned weight is much less
+			let withdraw_fee = ChargeFrqTransactionPayment::<Test>::from(0u64)
+				.withdraw_fee(&owner_id, &pay_with_capacity_add_public_key_call, &dispatch_info, 10)
+				.unwrap();
+			assert!(withdraw_fee.0.lt(&106_000_000u64));
```
aramikm, line 1060: nit: Might worth comparing the non-ethereum vs ethereum weights for the same call instead of using a magic number which might change in feature.

pallets/msa/src/weights.rs
aramikm, line 1: if no additional functions are added or removed then maybe this should not get committed!?

pallets/frequency-tx-payment/src/lib.rs
aramikm: nit: I think it might be helpful to have a comment about the weight reductions or percentages and such in here. 

Overall is it fair to say that this reduces the fee to around 0.3 or a bit more for subsidized?

mattheworris (approved):
- Read through changes.
🚢 it!

pallets/frequency-tx-payment/src/lib.rs
shannonwells: Yes, that's about right, somewhere between 25-30%

pallets/msa/src/weights.rs
shannonwells: oversight, thanks

JoeCap08055 (approved):
Nice work, approved.

PR #2404, opened 2025-05-13
ethereum utility library

### LINKED ISSUE #2281: Create New JS/TS Library for Ethereum-Compatible Keys and Signatures

aramikm:
# Goal
The goal of this PR is to create a ethereum utility library which can help us dealing with ethereum addresses and signatures and etc.

Closes #2281

## What is added?
- Generates CJS and ESM and browser compatible JS.
- Be able to generate EIP-712 signatures and verify them
- Generate Ethereum keys with different address encodings

# Checklist
- [x] Unit Tests added?
- [x] e2e Tests modified


js/ethereum-utils/src/index.ts
```diff
@@ -0,0 +1,11 @@
...
+import * as payloads from './payloads';
+import * as signature from './signature';
+import * as signatureDefinitions from './signature.definitions';
+
+export * from './payloads';
+export * from './signature';
+export * from './signature.definitions';
+export * from './address';
+
+export default { ...payloads, ...address, ...signatureDefinitions, ...signature };
```
aramikm, line 11: this flattens all the definitions and functions which if duplicated might cause some collisions in future. If you have some better organization ideas please let me know.

scripts/generate_js_definitions.sh
```diff
@@ -46,3 +46,12 @@ fi
 # Generate the new packed tgz
 cd dist
 npm pack
+
```
wilwade, line 49: Next door to this file should also be updated: `scripts/update-js-dependencies.sh`

js/ethereum-utils/src/index.ts
wilwade: I think this works.

I was able to do both of these, which is good:

- `import { getEip712BrowserRequestAddKeyData } from "@frequency-chain/ethereum-utils"`
- `import frqEth from "@frequency-chain/ethereum-utils"`

js/ethereum-utils/src/payloads.ts
```diff
@@ -0,0 +1,143 @@
+export type HexString = `0x${string}`;
+
+export interface EcdsaSignature {
+  Ecdsa: HexString;
+}
+
+export type ChainType = 'Mainnet-Frequency' | 'Paseo-Testnet-Frequency';
```
wilwade, line 7: Do we need local or "dev" here?

js/ethereum-utils/src/signature.definitions.ts
```diff
@@ -0,0 +1,158 @@
...
+
+// using pallet_revive test chain ID for now.
+export const EIP712_DOMAIN_DEFAULT: EipDomainPayload = {
+  name: 'Frequency',
+  version: '1',
+  chainId: '0x190f1b44',
+  verifyingContract: '0xCcCCccccCCCCcCCCCCCcCcCccCcCCCcCcccccccC',
+};
+
+export const ADD_PROVIDER_DEFINITION = {
```
wilwade, line 32: Again, I don't think so, but I wonder if we need to version these in any way? I guess we can later add `_V2` on the end without issue.

wilwade (approved):
A few questions, but nothing blocking. Good work.

js/ethereum-utils/src/signature.definitions.ts
JoeCap08055: I would say, any structs that already are at "V2" should get `_V2` appended to the const name now.

js/ethereum-utils/src/signature.definitions.ts
```diff
@@ -0,0 +1,158 @@
...
+export const PASSKEY_PUBLIC_KEY_DEFINITION = {
+  PasskeyPublicKey: [
+    {
+      name: 'publicKey',
+      type: 'bytes',
+    },
+  ],
+};
+
+export const PAGINATED_DELETE_SIGNATURE_PAYLOAD_DEFINITION = {
```
JoeCap08055, line 88:
```suggestion
export const PAGINATED_DELETE_SIGNATURE_PAYLOAD_DEFINITION_V2 = {
```

JoeCap08055: etc...

js/ethereum-utils/src/signature.ts
```diff
@@ -0,0 +1,497 @@
...
+  return {
+    types: {
+      ...EIP712_DOMAIN_DEFINITION,
+      ...ADD_KEY_DATA_DEFINITION,
+    },
+    primaryType: message.type,
+    domain: domain,
+    message: normalized,
+  };
+}
```
JoeCap08055, line 333: nit: this seems like a good refactoring target, since it's duplicated for every payload type. Perhaps:
```js
function createEip712Payload(domain: EipDomainPayload, primaryType: any, message: any, data: any) {
   return {
      types: {
         ...EIP712_DOMAIN_DEFINITION,
         ...data,
      },
      primaryType,
      domain,
      message,
   };
}
```

js/ethereum-utils/src/utils.ts
```diff
@@ -0,0 +1,42 @@
...
+
+/**
+ * Validate that a string is a valid hex string
+ */
+export function isValidHexString(value: string): boolean {
+  // Check if string starts with '0x' and contains only hex characters
+  const hexRegex = /^0[xX][0-9a-fA-F]*$/;
+  const isHex = hexRegex.test(value);
+  return isHex && value.length % 2 === 0;
+}
```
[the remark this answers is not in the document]
JoeCap08055: Though not much added benefit here; the main benefit of type guards is you can use it as a conditional test, and then after that test you can access type-specific methods and properties on the object inside the block (as well as pass to other functions that require that type, without casting). If there are no unique methods or properties, merely an assertion that the value conforms to some acceptable range, then there's really no benefit to a type guard over a boolean test function (I guess, unless you want to be able to pass into a function that takes a `HexString` parameter).

Then again, there's no harm or cost to a type guard over a boolean function, either... :man_shrugging:

scripts/generate_js_definitions.sh
aramikm: Good catch. Thanks

js/ethereum-utils/src/payloads.ts
aramikm: I wasn't sure we are going to need it but I'll add it.
These are going to be used for ChainId and maybe other options. Some chains have different chainIds for their mainnet and testnet and some only have one for the mainnet. I think we can use the existing pallet-revive default config for `dev` and if we don't get a new chainId for testnet then maybe we use the same for testnet also but that is a decision we can make later.

js/ethereum-utils/src/utils.ts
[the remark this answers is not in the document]
aramikm: Good suggestion. I'll update.

js/ethereum-utils/src/signature.definitions.ts
aramikm: If it the only version based on our existing convention I didn't add any version but all versions `> 1` should have the version in the name unless there's been a mistake (just like what Joe suggested)

js/ethereum-utils/src/signature.ts
aramikm: applied

js/ethereum-utils/src/signature.ts
```diff
@@ -485,13 +439,17 @@ export function getEip712BrowserRequestPasskeyPublicKey(
...
+
+function createEip712Payload(typeDefinition: any, primaryType: any, domain: EipDomainPayload, message: any): unknown {
   return {
     types: {
       ...EIP712_DOMAIN_DEFINITION,
-      ...PASSKEY_PUBLIC_KEY_DEFINITION,
+      ...typeDefinition,
     },
-    primaryType: message.type,
+    primaryType: primaryType,
```
JoeCap08055, line 451: nit: redundant
```suggestion
    primaryType,
```

js/ethereum-utils/src/signature.ts
```diff
@@ -485,13 +439,17 @@ export function getEip712BrowserRequestPasskeyPublicKey(
...
     domain: domain,
-    message: normalized,
+    message: message,
```
JoeCap08055, line 453: same
```suggestion
    message,
```

js/ethereum-utils/.tool-versions
```diff
@@ -0,0 +1 @@
+nodejs 20.12.2
```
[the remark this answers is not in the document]
aramikm: correct, but I think since we have other clients that are using node 20 it made sense to use the older version since this is going to be a dependency for those.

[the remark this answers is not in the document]
shannonwells: +1

mattheworris (approved):
- Read through changes.
- Nice work!
🚢 it!

PR #2421, opened 2025-05-29
Chore Update JS dependencies

wilwade:
# Goal
The goal of this PR is regular updates of JavaScript dependencies.


mattheworris (approved):
- Read over changes.
- Suggestion: Add sub-issues to flow the updates into gateway, sat, siwf?
🚢 it!

wilwade:
@mattheworris 
> Suggestion: Add sub-issues to flow the updates into gateway, sat, siwf?

I don't think we want to push updates, but I think we could setup dependabot better to handle this

mattheworris:
> @mattheworris
> 
> > Suggestion: Add sub-issues to flow the updates into gateway, sat, siwf?
> 
> I don't think we want to push updates, but I think we could setup dependabot better to handle this

Sounds good. Polkadot API fails in opaque ways and it has bitten me too many times, some automation can only improve things.

shannonwells (approved):
Just noting here the main update is a major version upgrade from polkadotjs-api v15 to v16

PR #2425, opened 2025-06-02
merged unlock chunks for same thaw at

### LINKED ISSUE #2416: [Capacity/Boosting Governance] Unlock Chunks for the same `thaw_at` should merge to one unlock chunk
As a user unstaking, I might need to unstake more than I thought. Since the number of unlock chunks is limited, I want to make sure that if I unstake twice in the same capacity epoch, that it only uses one of the possible unlock chunks.

## Technical Notes

tldr: adjust `add_unlock_chunk` to be an update instead of a pure add by adding to an existing chunk that has the same `thaw_at`

aramikm:
# Goal
The goal of this PR is to close #2416 

# Checklist
- [x] Unit Tests added?


enddynayn (approved):
👍 great!

PR #2436, opened 2025-06-06
Committed boost extrinsic

aramikm:
# Goal
The goal of this PR is to create an extrinsic for committed_boosting

# Discussion
- All the inner logic for provider boosting is essentially the same for committed boosting


pallets/capacity/src/lib.rs
```diff
@@ -679,6 +694,43 @@ pub mod pallet {
...
+		pub fn committed_boost(
+			origin: OriginFor<T>,
+			target: MessageSourceId,
+			amount: BalanceOf<T>,
+		) -> DispatchResult {
+			let staker = ensure_signed(origin)?;
+
+			let (mut staking_details, stakable_amount) =
+				Self::ensure_can_stake(&staker, target, amount, StakingType::CommittedBoost)?;
+			staking_details.staking_type = StakingType::CommittedBoost;
```
JoeCap08055, line 715: nit: I think this should be done inside `ensure_can_stake` (but I don't feel super strongly about it...)

PR #2439, opened 2025-06-06
Feat/consistent committed boosting release

JoeCap08055:
# Goal
The goal of this PR is to re-work the staged release algorithm for Committed Boosting such that the same amount can be released at each stage, irrespective of whether the user has previously unstaked released amounts or not.


pallets/capacity/src/lib.rs
```diff
@@ -904,79 +912,100 @@ impl<T: Config> Pallet<T> {
...
-				}
-				// PTE block is not set, and we're past the failsafe block; the Committed Boosting program is ended
-				else if curr_block >= T::CommittedBoostFailsafeUnlockBlockNumber::get().into() {
-					(StakingType::FlexibleBoost, Perbill::from_percent(100))
-				}
-				// PTE block is not set, but we haven't passed the failsafe block yet--we're in the Pre-Commitment phase
-				else {
-					(current_staking_type, Perbill::zero())
+	) -> CommittmentPhase {
+		let commitment_phase = if let Some(pte_block) = PrecipitatingEventBlockNumber::<T>::get() {
```
wilwade, line 920: I had a hard time following the if statements. I looked into if a match could make it cleaner:

```rust
match PrecipitatingEventBlockNumber::<T>::get() {
    // We have a block that is the event
    Some(pte_block) => {
        let initial_commitment_block =
            pte_block.saturating_add(staking_config.initial_commitment_blocks.into());
        let max_staged_release_blocks = initial_commitment_block.saturating_add(
            staking_config
                .commitment_release_stage_blocks
                .mul(staking_config.commitment_release_stages)
                .into(),
        );

        match block_number {
            // End of the program
            n if n >= max_staged_release_blocks => CommittmentPhase::RewardProgramEnded,
            // Before the end, but after the commitment
            n if n >= initial_commitment_block => CommittmentPhase::StagedRelease,
            // Still before the commitent
            _ => CommittmentPhase::InitialCommitment,
        }
    },
    None if block_number >= T::CommittedBoostFailsafeUnlockBlockNumber::get().into() => {
        CommittmentPhase::Failsafe
    },
    // No event, and before the failsafe, we are pre-commitment
    _ => CommittmentPhase::PreCommitment,
}
```

Perhaps helped some? Might be worth making some helper functions?

Anyway, this is partially just golf. Mostly looking for a way to make the phases read better since the optimal calculation is not in temporal order.

pallets/capacity/src/lib.rs
```diff
@@ -904,79 +912,100 @@ impl<T: Config> Pallet<T> {
...
+			} else {
+				CommittmentPhase::InitialCommitment
+			}
+		} else if block_number >= T::CommittedBoostFailsafeUnlockBlockNumber::get().into() {
+			CommittmentPhase::Failsafe
+		} else {
+			CommittmentPhase::PreCommitment
+		};
+
+		commitment_phase
```
wilwade, line 947: Suggest to just return the if statement above.

pallets/capacity/src/lib.rs
```diff
@@ -904,79 +912,100 @@ impl<T: Config> Pallet<T> {
 		CapacityLedger::<T>::insert(target, capacity_details);
 	}
 
-	fn get_staking_type_and_releasable_percent_in_force(
-		current_staking_type: StakingType,
+	/// Get the phase of Committed Boosting at the indicated block
+	fn get_committed_boosting_phase(
```
wilwade, line 916: I was surprised there wasn't a nice unit test for this. Suggestion that it might be worth having an AI generate a set to cover the cases as a separate set of tests instead of just testing a level up.

wilwade (approved):
Few suggestions, but just that. Take them as you wish. Logic is all sound.

pallets/capacity/src/lib.rs
JoeCap08055: Tests are coming 😄

JoeCap08055: Once I've got the final functional updates merged into the feature branch, I'm going to do a unit & e2e test PR.

pallets/capacity/src/lib.rs
JoeCap08055: Refactor for readability coming, along with better tests.

PR #2457, opened 2025-06-13
fix: correct logic for determining whether there are unclaimed rewards for an account

### LINKED ISSUE #2295: [Bug]: Unstaking a second time in an era after you claimed rewards fails with `MustFirstClaimRewards`
### What happened?

Thanks to @julian-amplica for finding this one.

Steps to reproduce:
1. Do a ProviderBoost for a target Provider for at least 3X the minimum amount.
2. Wait 2 RewardEras so you can claim rewards.
3. Claim rewards - this succeeds
4. Unstake a small amount, less than what you boosted.  This succeeds.
5. Unstake another small amount.  This fails.

It fails b/c I didn't take into account the fact that a new boost history would be added if there are additional boosting events (anything that will add a history item, a boost or an unstake) that occur in the same era as claiming rewards.

It works for 1 unstake event because the new ProviderBoostHistory hasn't been added yet.  It fails on the second unstake event, because the code acts as as if you are coming back in a later era and you have rewards.

### Relevant log output

```shell

```

JoeCap08055:
# Goal
The goal of this PR is to correct the logic used to determine whether an account has unclaimed rewards.

Closes #2295 

# Checklist
- [ ] Updated Pallet Readme?
- [ ] Design doc(s) updated?
- [x] Unit Tests added?
- [ ] e2e Tests added?
- [ ] Benchmarks added?

### DIFF
--- a/pallets/capacity/src/lib.rs
+++ b/pallets/capacity/src/lib.rs
@@ -1042,14 +1042,14 @@ impl<T: Config> Pallet<T> {
 	pub(crate) fn has_unclaimed_rewards(account: &T::AccountId) -> bool {
 		let current_era = CurrentEraInfo::<T>::get().era_index;
 		match ProviderBoostHistories::<T>::get(account) {
+			// We can ignore any entries for the current or prior era, since:
+			//   - if it's for the previous era, it means we've already paid out rewards for that era,
+			//     or they just staked in that era & hence aren't eligible for rewards yet.
+			//   - if it's for the current era, then they've only just started staking
+			// If there are any entries for eras earlier than one prior to the current era, then there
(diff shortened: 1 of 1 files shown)


wilwade (approved):
The fix is correct, but I was surprised this doesn't break or change any tests.

shannonwells (changes requested):
Needs a regression test for the bug... otherwise gtg, thanks!

mattheworris (approved):
- Read through changes.
🚢 it!

shannonwells:
Added regression test, passed the red-green check before and after fix.

shannonwells (approved):
Added regression test

PR #2460, opened 2025-06-16
2453 signed payload structures need a type discriminator

### LINKED ISSUE #2453: Signed payload structures need a type discriminator
### What happened?

# Description
There has been past discussion about whether structures that represent payloads to be signed need a type discriminator field. Previously, it had been decided that since no two payload structures that we would sign had the same/similar structures, or could be used for more than one purpose, it was not necessary.

Recently, we added `AuthorizedKeyData`, a payload with the same internal structure as `AddKeyData` (different field names, etc--but, critically, the same SCALE-encoded representation).

Because we now have an identical structure that could be used for two different purposes, it is now necessary to add a type discriminator to at least one of those structures. Otherwise it would be possible to tell a user that they were signing a payload for one purpose, but then use the signed payload for a different purpose.

###Note
(I think) this discriminator would not be necessary if all of our signed payloads used EIP-712 signing; however, since we still support both EIP-712 _and_ sr25519 signatures for signed payloads, it is necessary.

JoeCap08055:
# Goal
The goal of this PR is to add a type discriminator to the `AuthorizedKeyData` signed payload structure, so that it can be differentiated from the `AddKeyData` structure, which has different field names, but the same internal structure, and therefore the same SCALE encoding.

Closes #2453 

# Discussion

- 

# Checklist
- [ ] Updated Pallet Readme?
- [ ] Design doc(s) updated?
- [x] Unit Tests added?
- [x] e2e Tests added?
- [ ] Benchmarks added?


pallets/msa/src/tests/signed_extension_tests.rs
```diff
@@ -507,18 +508,45 @@ fn signed_ext_validate_fails_when_withdraw_tokens_caller_key_does_not_match_payl
 	});
 }
 
+#[test]
+fn signed_ext_validate_fails_when_withdraw_tokens_payload_payload_is_has_wrong_type_discriminant() {
```
mattheworris, line 512: nit: this sounds funny to me
```suggestion
fn signed_ext_validate_fails_when_withdraw_tokens_payload_has_wrong_type_discriminant() {
```

mattheworris (approved):
- Read through changes.
- Non-blocking comment.
🚢 it!

pallets/msa/src/types.rs
```diff
@@ -69,13 +69,18 @@ impl<T: Config> EIP712Encode for AddKeyData<T> {
...
+
 /// A type definition for the payload for authorizing a public key for the following operations:
 /// -  Authorizing a token withdrawal to an address associated with a public key - `pallet_msa::withdraw_tokens`
 #[derive(
 	TypeInfo, RuntimeDebugNoBound, Clone, Decode, DecodeWithMemTracking, Encode, PartialEq, Eq,
 )]
 #[scale_info(skip_type_params(T))]
 pub struct AuthorizedKeyData<T: Config> {
+	/// type discriminator, because otherwise this struct has the same SCALE encoding as `AddKeyData`
+	pub discriminant: [u8; 17],
```
aramikm, line 83: Why not use a global enum for discriminant?

aramikm (approved):

shannonwells (approved):
I reviewed the changes, looks good to go!

PR #2461, opened 2025-06-17
2456 bug can unstake and then immediately restake entire amount

### LINKED ISSUE #2456: [Bug]: Can unstake and then immediately restake entire amount
### What happened?
This was a test to see if there was a staking-unstaking loop bug that could be exploited to game the extended locking period. Fortunately there isn't, however, you still should be unable to re-stake until your staked token has thawed, otherwise IMO this could cause some other cascading failures or at minimum, inexplicable and surprising behavior for boosters.
This test as is passes. It should fail where indicated - of course once it does, then the rest of it can go away.

```rust
#[test]
fn cannot_stake_loop_to_avoid_thaw_period() {
		// provider_boost and then unstake, before the thaw period check if you can boost again.
		// if you can boost again, see if you can call withdraw_thawed_token
		let staker = 600;
		let target: MessageSourceId = 1;
		let amount = 500; // ensure >> half
		let epoch_length = 3;

		register_provider(target, String::from("Foo"));
		EpochLength::<Test>::set(epoch_length);
		assert_ok!(Capacity::provider_boost(RuntimeOrigin::signed(staker), target, amount));

		// run to next RewardEra and unstake entire amount
		system_run_to_block(11);
		assert_ok!(Capacity::unstake(RuntimeOrigin::signed(staker), target, amount));

		// restake in same block
		// a stake of any type should fail, but doesn't
		assert_ok!(Capacity::provider_boost(RuntimeOrigin::signed(staker), target, amount));
		// assert_noop!(
		// 	Capacity::committed_boost(RuntimeOrigin::signed(staker), target, amount),
		// 	Error::<Test>::BalanceTooLowtoStake,
		// );
		
		assert_eq!(UnstakeUnlocks::<Test>::get(staker).unwrap_or_default().len(), 1);

		// wait for thaw (2 epochs in test)
		let blocks_to_thaw: u32 = epoch_length *2u32;
		run_to_block(12u32 + blocks_to_thaw*3);

		// withdraw unlocked token succeeds...
		assert_ok!(Capacity::withdraw_unstaked(RuntimeOrigin::signed(staker)));

		// check transferable balance
		// If uncommented, FAILS here b/c WithdrawConsequence::Frozen, 
                // no gaming attack, because there was an overlapping freeze. See below.
		// assert_transferable::<Test>(&staker, amount);
		
		
		let tboh = <Test as Config>::Currency::balance_frozen(&FreezeReason::CapacityStaking.into(), &staker);
		assert_eq!(tboh, amount);
	});
}
```

### Relevant log output

```shell

```

JoeCap08055:
# Goal
The goal of this PR is to:
- Account for unlocking token balance when determining available amount to stake (ie, cannot re-stake an prevoiusly-staked token until it has thawed & been withdrawn from staking)

Closes #2456 

# Discussion

- 

# Checklist
- [ ] Updated Pallet Readme?
- [ ] Design doc(s) updated?
- [x] Unit Tests added?
- [x] e2e Tests added?
- [ ] Benchmarks added?
- [x] Weights updated?


pallets/capacity/src/lib.rs
```diff
@@ -831,11 +830,16 @@ impl<T: Config> Pallet<T> {
...
+		let unlocks = UnstakeUnlocks::<T>::get(staker).unwrap_or_default();
+
+		let unthawed_unstaked_balance =
+			unthawed_unlock_chunks_total::<T>(&unlocks, CurrentEpoch::<T>::get());
 		let freezable_balance = T::Currency::balance_freezable(staker);
 		let current_staking_balance =
 			StakingAccountLedger::<T>::get(staker).unwrap_or_default().active;
 		let stakable_amount = freezable_balance
 			.saturating_sub(current_staking_balance)
+			.saturating_sub(unthawed_unstaked_balance)
```
JoeCap08055, line 842: @shannonwells I updated this so that it now includes thawed unlock chunks in the stakeable balance. Also updated unit tests accordingly.

However, there is one side-effect of this that we may want to consider more thoroughly: this setup allows overlapping `CapacityFreeze` locks on the same tokens. I don't think it currently causes any weird effects, but it could be something we trip over in the future.

scripts/generate_js_definitions.sh
```diff
@@ -55,3 +55,12 @@ npm run build
...
 npm pack
+
+# build recovery-sdk
+cd "../../recovery-sdk"
+npm install
+npm run build
+
+# Generate recovery-sdk packed tgz
+cd dist
+npm pack
```
JoeCap08055, line 66: Looks like this got missed when the recovery-sdk was added; not needed in CI because it uses previously built artifacts, but necessary if you want to run E2E tests locally.

pallets/capacity/src/lib.rs
```diff
@@ -862,6 +865,9 @@ impl<T: Config> Pallet<T> {
 			UnstakeUnlocks::<T>::set(staker, Some(unlocks));
 		}
 
+		// Set the total locked amount. We do not simply apply the delta of the withdrawn chunks, as this
+		// is error-prone due to the double-locking issue; instead, we simply calculate the total locked
+		// amount as the `total staked + total remaining unthawed unlocking chunks`.
```
aramikm, line 870: I'm a bit confused about the location of this comment. Is this comment related to the changes for this PR or just a clarification of the logic which was implemented previously?

aramikm (approved):
Read the code. Looks good to me

pallets/capacity/src/lib.rs
JoeCap08055: It was a clarification comment that I added, more for my own benefit. Can remove if confusing.

saraswatpuneet (approved):
lgtm

shannonwells (approved):
I looked at the new code, I ran the local tests and my example test and it passed.  Let 'er rip!

Issue #2284, opened 2025-03-20
Add ETH Key Support to SIWF (Sign-In With Framework)


p5150j:
### Description
Extend the Sign-In With Framework functionality to support Ethereum keys and signatures, enabling seamless authentication.

Tasks

- Update SIWF message format to be compatible with ETH signing
- Implement ETH signature verification in SIWF
- Modify session management for ETH keys
- Update frontend components to support ETH wallet connections
- Create documentation for developers integrating with SIWF using ETH

Acceptance Criteria

- [ ]  SIWF accepts and verifies ETH signatures
- [ ]  Authentication flow works end-to-end with ETH keys
- [ ]  Frontend components display correctly for ETH connections
- [ ]  Documentation updated with ETH-specific integration guidelines
- [ ]  Test suite covering ETH authentication scenarios
- [ ]  Security review of ETH implementation completed

Engineering Notes
[To be filled by lead developer]

PR #2462, opened 2025-06-18
ethereum-utils: siwf support

aramikm:
# Goal
The goal of this PR is update `ethereum-utils` library to support payloads used in SIWF

Related to #2284 

# Discussions
- you can ignore e2e test reviews. We had to change the way funding was done due to some changes on the mocha side.

# Checklist
- [x] Unit Tests added?


js/ethereum-utils/src/signature.ts
```diff
@@ -128,6 +138,10 @@ function getTypesFor(payloadType: string): SupportedPayloadDefinitions {
     AddKeyData: ADD_KEY_DATA_DEFINITION,
     AuthorizedKeyData: AUTHORIZED_KEY_DATA_DEFINITION,
     AddProvider: ADD_PROVIDER_DEFINITION,
+
+    // offchain signatures
```
wilwade, line 142: We'll need to add the one for stateful storage and handles, although A. this is just the initial draft, and B. those are close to the `SIWF_SIGNED_REQUEST_PAYLOAD_DEFINITION`

wilwade:
Ah, I had forgotten that those were already added. That is probably good
then.

js/ethereum-utils/src/signature.ts
```diff
@@ -33,49 +36,76 @@ import {
...
   const normalizedPayload = normalizePayload(payload);
   const wallet = new ethers.Wallet(secretKey);
-  // TODO: use correct chainID for different networks
-  // TODO: use correct contract address for different payloads
-  const signature = await wallet.signTypedData(EIP712_DOMAIN_DEFAULT, types, normalizedPayload);
+  let signature;
+  switch (signatureType) {
+    case 'EIP-712':
+      // TODO: use correct chainID for different networks
+      // TODO: use correct contract address for different payloads
```
wilwade, line 63: We really should figure out these todos. I know the chainID we can get setup, but less sure about what contract address we should use as we will not have those until we have the full pallet->contract precompiles setup.

Any ideas on what it would take to alter these values? I guess we'd have to support both for a while.

wilwade (approved):
The draft is looking good to me.

js/ethereum-utils/src/signature.ts
aramikm: We decided to keep it as is for now.

js/ethereum-utils/src/signature.ts
```diff
@@ -139,6 +178,14 @@ function getTypesFor(payloadType: string): SupportedPayloadDefinitions {
   return definition;
 }
 
+function getSignatureType(payloadType: string): SignatureType {
+  if (payloadType === 'SiwfLoginRequestPayload') {
+    return 'EIP-191';
+  } else {
```
enddynayn, line 184: Minor: we can drop the else here.

e2e/scaffolding/globalHooks.ts
```diff
@@ -66,9 +66,10 @@ async function devSudoActions() {
 }
 
 export async function mochaGlobalSetup(context) {
+  console.log('globalHooks.ts mochaGlobalSetup');
```
shannonwells, line 69: remove console log?

js/ethereum-utils/src/payloads.ts
```diff
@@ -141,6 +141,24 @@ export interface AddProvider {
   expiration: number;
 }
 
+export interface SiwfSignedRequestPayload {
+  // type discriminator
+  type: 'SiwfSignedRequest';
```
shannonwells, line 146: nit: will this cause confusion b/c it's not SiwfSignedRequestPayload?

shannonwells (approved):
I reviewed the code and ran the Metamask tool.  Just had non-blocking comments, LGTM

PR #2463, opened 2025-06-19
Design Docs for Account Recovery System

mattheworris:
The Goal of this PR is to outline the design of a proposed Account Recovery System.

## Discussion

- Mostly finalized design to satisfy fluid design requirements.
- Pseudo-code included


designdocs/recovery_system.md
```diff
@@ -0,0 +1,505 @@
...
+
+Option A (hash-based) is favored for simplicity and on-chain verification capability. Hashing the recovery data means the chain can later verify a recovery attempt by recomputing a hash, without needing to store any encryption keys. Hash storage is also aligned with security best practices. The only data needed for recovery is a successful hash match, rather than extracting plaintext.
+
+### Hash Algorithm and Salting
+
+This design uses a modern hash or Key Derivation Function (KDF) to protect the recovery code. A good choice is Argon2id or PBKDF2-HMAC-SHA256 with a high iteration count, as recommended by OWASP. For instance, Argon2id with ~ (2^16) memory and 3 iterations, or PBKDF2 with 200,000+ iterations, will slow down any brute-force attempt substantially. In our design, the “salt” for the hash can be the user’s email or MSA ID (to ensure uniqueness per user) combined with a global constant. However, since the seed itself is random and unique, an explicit additional salt is optional – Argon2/PBKDF2 internally use a salt. The main point is to avoid storing a raw unsalted hash of a low-entropy value. Here the entropy comes mostly from the seed phrase, which is very high.
```
wesbiggs, line 52: I don't think a "global salt" adds any security since it would be known by any attacker; it still comes down to the guessability of the identifier and seed phrase.

designdocs/recovery_system.md
```diff
@@ -0,0 +1,505 @@
...
+- Their email address (the one they used at account creation).
+- Their recovery code (the secret phrase/UUID they saved).
+  
+The provider likely presents a form for the user to input these. At this point, the provider does not yet know the user’s MSA ID or whether the data is valid – that will be determined next.
+
+#### 2. On-Chain Validation (Recovery Key)
+
+The recovery provider uses the submitted email + recovery code to validate against the blockchain:
+
+- It submits the recovery data to the blockchain and determine if any MSA record has a matching recovery hash. This can be done via an extrinsic `verify_recover_code(email, recovery_code)` as the hash is computed on-chain.
```
wesbiggs, line 184: This implies a reverse index from recovery hash to MSA, which sounds expensive. An alternative approach would be to submit the MSA along with the recovery code; it would need to be part of the recovery data the user recorded. (Or, if they know their handle, they can look it up.)

wesbiggs, line 184: I am uncomfortable with the implications of submitting these raw values to a RPC node over the network. I think the recovery service should calculate and submit the _hash_ of these values via a known formula, and the on-chain value should be the _hash of the hash_.

designdocs/recovery_system.md
```diff
@@ -0,0 +1,505 @@
...
+
+- User-specified key: If the user is somewhat technical, they might generate a new key pair on their own device (e.g. via a wallet app) and only give the public key to the provider. This is ideal for security, as the private key never leaves the user. The recovery provider would ask “Please paste a new public key to add to your account.” However, this requires the user to know how to create a key pair, which may not be user-friendly in an emergency.
+- Provider-generated key: The recovery provider can create a new key pair on behalf of the user. For instance, it generates a new Ed25519, Sr25519, or Ethereum key (compatible with Frequency). The provider will then securely transmit the private key to the user after the on-chain update. This approach is simpler for the user but requires trust that the provider deletes its copy. In practice the provider could generate a mnemonic and show it to the user over a secure TLS connection for them to copy, minimizing its own storage of the key.
+
+This design recommends giving the user the option, but in either case, a public key for the new control key is decided at this step.
```
wesbiggs, line 207: I would say the choice of recovery provider dictates this interaction. We want to make it possible for a custodial wallet to be a recovery provider, in which case the user would not receive the private key at all.

designdocs/recovery_system.md
```diff
@@ -0,0 +1,505 @@
...
...w public key to the list of control keys for `target_msa`. Frequency allows multiple control keys per MSA, so the chain will allow a recovery provider to add a new public key. The user must delegate this authority to the recovery provider. In this hash design, an elegant solution is: the extrinsic call could include the plaintext recovery code (or its hash) as a parameter; the runtime hashes it and verifies it matches the stored hash for that MSA. This way, the provider proves on-chain that it indeed has the secret. To avoid exposing the secret in plaintext on-chain, the extrinsic could take H’ (the hash) as input instead. But since H’ is stored and public, an attacker provider could try to reuse it; thus requiring the actual secret or a signature is safer. Another approach is having the provider sign the transaction with its own key (proving it’s authorized) and relying on off-chain enforcement of email verification (since governance oversight exists). For maximal security, we can do: recoverAccount(msa_id, recovery_code, new_pubkey) → runtime computes hash of (email||secret) and checks match & provider authorization. This ensures only a correct recovery_code unlocks the addition.
```
wesbiggs, line 214: Per note above, I prefer submitting `H'` with a provider signature (we need proof that the provider is an authorized recovery provider); the chain would store `H(H')`.

designdocs/recovery_system.md
```diff
@@ -0,0 +1,505 @@
...
+
+### 7. Post-Recovery: Re-Issuing Recovery Key
+
+After a successful recovery, if the user continues with a new wallet/provider, they may want to set up a new recovery key for future safety (since the old one was consumed). The flow would essentially be the same as during account creation:
+
+- The user’s new provider (or the recovery provider, if it doubles as a wallet provider) can generate a new secret, hash it with the user’s (same) email, and update the on-chain record. The chain will accept it because the previous one was invalidated. The new recovery_hash replaces the old.
+- This ensures the user isn’t left without a recovery option going forward. It’s an optional step but recommended. The user should be informed that they have a new recovery code and the old one will no longer work (consistent with the one-at-a-time rule).
+
+If the user chooses not to set a new recovery key, that’s also acceptable (maybe they prefer not to trust any provider and will manage their keys on their own). The system does not force a new recovery key issuance; it simply allows it.
```
wesbiggs, line 236: While this reasoning is sound from a web3 perspective, a custodial wallet acting as a recovery provider should _always_ create a new recovery key. But I suppose demonstrating this can be part of the governance approval process rather than chain-enforced. On the other hand, it would be more efficient on the chain to submit the new recovery hash to be stored as part of the `recoverAccount(...)` extrinsic.

designdocs/recovery_system.md
```diff
@@ -0,0 +1,505 @@
...
+- Misconduct by a provider (e.g. recovering accounts without permission, or failing to verify identity properly) would be detectable through user reports. The governance committee could then remove that provider from the approved list, and possibly even revert malicious changes if community consensus deems it necessary (though reverting blockchain state is extreme, ideally the design prevents unauthorized changes in the first place).
+- Each recovery action could incur an on-chain event log for transparency. Users could even opt to be notified (to a different contact) if their account was recovered, adding an extra layer of security (out of band notification in case of malicious recovery attempt).
+
+### One-time Key Use Enforcement
+
+The contract that the recovery key is one-time is enforced by design:
+
+- The chain will not allow a second use of the same recovery hash. After it’s cleared, if someone tried to reuse the same secret, the lookup would fail (no match).
+- If an attacker somehow eavesdropped on the secret during a legitimate recovery (which should be very hard if TLS is used), trying to use it again would not work because it’s already consumed.
```
wesbiggs, line 267: Double-hashing further mitigates against this.

Are there any scenarios where a compromised recovery provider could front-run the data from another recovery provider's transaction in the pool?

designdocs/recovery_system.md
```diff
@@ -0,0 +1,505 @@
...
+
+2. Secure Email Verification: The email-based authentication step must be implemented securely to avoid vulnerabilities. The system should ensure that possession of the email account is verified via robust methods (e.g. time-bound one-time codes or magic links) without exposing the recovery key or other secrets. No sensitive data is sent via email; only random verification tokens are.
+
+3. Brute Force Resistance: Recovery keys (as stored on-chain) must be resistant to brute force and offline attacks. An attacker who obtains the on-chain hash should not be able to feasibly crack it to derive the email or recovery code. The design should incorporate strong cryptographic hashing or encryption to protect against precomputation and exhaustive search.
+
+The following sections detail the cryptographic design meeting these requirements, the recovery workflows (with sequence diagrams), security analysis, and an example implementation snippet.
+
```
wilwade, line 32: Note that we should add a Terms section so we can make sure we are aligned and consistent with them.

designdocs/recovery_system.md
```diff
@@ -0,0 +1,505 @@
...
+## Key Requirements
+
+### Functional Requirements
+
+1. Recovery Key Generation: On MSA creation, the wallet provider collects the user’s email and generates a recovery key using a reversible combination of the email with a random salt/seed (e.g. a BIP-39 seed phrase). The resulting key (or a cryptographic derivative) is securely stored on-chain (encrypted or hashed) to allow future recovery.
+
+2. Single Active Key: Only one recovery key exists per user at any time. If a recovery key is ever used (consumed) during account recovery, it is invalidated on-chain. The user or new provider may then issue a new recovery key for future use, but duplicate or old keys cannot remain active simultaneously.
+
+3. Governance-Approved Providers: Recovery operations are performed by special recovery providers that are approved via on-chain governance (listed in a registry). When a user needs account recovery, they contact one of these trusted providers and submit their recovery key along with their email address.
```
wilwade, line 15:
```suggestion
3. Governance-Approved Providers: Recovery operations are performed by special recovery providers that are approved via on-chain governance (listed in a registry). When a user needs account recovery, they contact one of these providers trusted to do the email verification and submit their recovery key along with their email address.
```

designdocs/recovery_system.md
```diff
@@ -0,0 +1,505 @@
...
+
+4. Validation by Provider: The recovery provider must validate two things off-chain: (a) the submitted recovery key (in combination with the email) matches the on-chain record for some MSA ID (proving the key is valid for that account), and (b) the provided email is truly controlled by the user (via a secure email verification challenge). Only if both checks pass is recovery authorized.
+
+5. On-Chain Key Addition: After successful validation, the recovery provider generates a new control key for the user (typically a new key pair) and uses a blockchain transaction to associate this new key with the user’s MSA ID on-chain. This effectively restores user control. The old compromised or inaccessible key can be removed or left unused; the new key will have full control of the MSA.
```
wilwade, line 19: > recovery provider generates a new control key for the user

I wonder if we instead want this to be that the user must provide an address and signed payload by that address to the recovery provider (the provider may help).

Aka make it the same as adding a control key, just doesn't require the signature from the current control key.

wilwade: This would also remove step 6

designdocs/recovery_system.md
wilwade, line 52: I think we have a few things here we'll want to alter.

- Likely should use one of the secure hash functions we already use such as blake2/3 or one of the keccak
- In reading this, to make sure:
    - A. The recovery submission is verifiable
    - B. The recovery submission doesn't expose the email or a way to run hashes to the email 
- So pieces
    - A. The recovery key: known only to the user and (if used) the recovery provider
    - B. The email: known only to the user  and (if used) the recovery provider
    - C. The `hash(recovery key)`: Known only the user and the recovery provider until recovery, then known by the chain
    - D. The `hash(email+recovery key)`: Known only to the user and recovery provider until recovery, then known by the chain
    - E. The `hash(<hash(email+recovery key)> + <hash(recovery key)>)` value. This is the value that is published to the chain for verification
    - F: Verification is by the recovery provider submitting the `hash(email+recovery key)` and the `hash(recovery key)` Which is then hashed on chain to match the `hash(<hash(email+recovery key)> + <hash(recovery key)>)`
    
I think that would ensure that the email is always private as the value that goes with the email into that hash is also always private

wilwade (commented):
WIP

designdocs/recovery_system.md
```diff
@@ -0,0 +1,505 @@
...
+    B->>App: Return handle
+    App->>U: Show confirmed handle
+    G->>U: (Async) Send Frequency intro/recovery email including recovery code
+```
+
+#### Sequence diagram – Account creation and recovery key setup
+
+1. User Account Creation: The user signs up with a wallet/app (the provider) and provides their email address as part of registration. The provider creates a new MSA on the Frequency blockchain for the user. During this on-chain account creation, the user’s initial control key (public key) is linked to the MSA (this may be a key managed by the provider).
+
+2. Recovery Key Generation: The blockchain generates a random secret (e.g. a 128-bit entropy value). For example, it may call a secure random generator to get 16 bytes and then encode as a 12-word BIP-39 phrase for usability, or a UUIDv4 code. This secret is combined with the user’s email (e.g. by hashing as described in the Cryptographic Design) to produce the recovery hash record. The blockchain implements the hash option, e.g. it computes H = SHA-256(email || recovery_code || global salt).
```
enddynayn, line 127: Do you mean a user generates a secret key?

designdocs/recovery_system.md
```diff
@@ -0,0 +1,505 @@
...
+
+3. On-Chain Storage: The blockchain stores the recovery hash record (H) as part of the MSA creation extrinsic or a separate call right after. The blockchain will store H in association with that MSA ID. If this is part of the creation, the “Create MSA” call would include the hashed recovery key as an attribute. The chain may also stake the provider’s identity to ensure only authorized providers can do this (since only a registered provider should write recovery keys, protecting users from malicious data injection).
```
enddynayn, line 129: What does it mean when it says 'The chain may also stake the provider’s identity'?

designdocs/recovery_system.md
```diff
@@ -0,0 +1,505 @@
...
+
+4. User Receives Recovery Code: Once the MSA is successfully created and the recovery hash stored, the provider delivers the actual recovery code to the user. Typically, the wallet UI would display the 12-word phrase or UUID code and prompt the user to save it. For security, the provider does not send this over email (to avoid having the secret in the inbox). The user is instructed: “Save this recovery phrase offline. It will allow you to recover your account if needed.” This two-factor approach is still secure if the recovery code is sent via email, as long as the email verification step is robust. It is however, more secure to avoid sending the recovery code via email entirely.
```
enddynayn, line 131: >  hash stored, the provider delivers the
Slight modification, I think you mean recovery providers?
 hash stored, the _recovery_ provider delivers the

wilwade: > Actually, are Recovery Providers different than Regular providers? Can providers be both a Regular Provider and Recovery Provider?

I would say currently there is nothing that would prevent a provider from being just one or both. Don't think there needs to be an exclusivity here?

enddynayn: Thanks! Like any design decision, there are tradeoffs. Having a provider take on both roles can increase centralization and the level of trust placed in that provider. That said, since Recovery Providers are approved through governance, theres already an additional layer of vetting in place which helps mitigate risk.

designdocs/recovery_system.md
```diff
@@ -0,0 +1,505 @@
...
+
+In summary, the system meets the security requirements by layering cryptographic protections and real-world identity verification. It is designed such that no single point of failure (neither just an email, nor just a recovery code, nor a rogue provider) can compromise an account alone:
+
+- The cryptography ensures that only a user with the correct secret can ever pass the hash check.
+- The email check ensures the user’s physical identity is tied to the request.
+- Governance oversight ensures providers operate within strict guidelines.
+
+## On Chain Structure
+
+### Primitives
```
enddynayn, line 298: For Recovery Providers, are you thinking of adding a typeto the existing Provider struct, or should Recovery Providers be stored separately?

designdocs/recovery_system.md
```diff
@@ -0,0 +1,505 @@
...
+
+- Option B: Encryption (Two-Way): Derive an encryption key from the email and seed, and use it to encrypt a known plaintext (such as the user’s MSA ID or a fixed marker). For example, derive a symmetric key K = PBKDF2(email || seed, salt=networkSalt, iter=100k) and use AES-256 to encrypt the user’s MSA ID. Store the ciphertext on-chain. In the future, a recovery provider who is given the correct email and seed can derive K and decrypt the blob to retrieve the MSA ID (achieving “reversibility” by decryption). This approach hides the MSA ID and requires possession of the seed to decrypt. It effectively binds the MSA ID to the secret.
+
+Option A (hash-based) is favored for simplicity and on-chain verification capability. Hashing the recovery data means the chain can later verify a recovery attempt by recomputing a hash, without needing to store any encryption keys. Hash storage is also aligned with security best practices. The only data needed for recovery is a successful hash match, rather than extracting plaintext.
```
harry-evans, line 48: we will need to specify the importance of (and/or specific steps for) email normalization. it is not unusual for emails the get written "John.Doe@Example.com" vs "john.doe@example.com". Those would obviously hash differently. Given that the user will be entering their email address at Provider 1 to have the recovery key generated, and at Provider 2 to recover it, it feels like we can't just count on them both doing things the same way.

harry-evans: Would the recovery key the user receives just be something like "word1word2word3...word12" all smashed together? Or would it resemble how you currently manage a passphrase? or is it already hashed some number of times before being sent to the user's email?

designdocs/recovery_system.md
```diff
@@ -0,0 +1,505 @@
...
+    F->>G: Return handle
+    G->>B: Return handle
+    B->>App: Return handle
+    App->>U: Show confirmed handle
+    G->>U: (Async) Send Frequency intro/recovery email including recovery code
+```
+
+#### Sequence diagram – Account creation and recovery key setup
+
+1. User Account Creation: The user signs up with a wallet/app (the provider) and provides their email address as part of registration. The provider creates a new MSA on the Frequency blockchain for the user. During this on-chain account creation, the user’s initial control key (public key) is linked to the MSA (this may be a key managed by the provider).
```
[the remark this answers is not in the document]
harry-evans: I think the assumption is that it is technically possible to generate an account without generating a recovery key, from a chain perspective, but the best practice and our tooling (gateway, SDK, etc) would ensure that it was done when an MSA is created. I think of this like handle, where you can make an MSA without one, but that isn't the default path, tooling / experience wise.

designdocs/recovery_system.md
harry-evans: Yes. Theoretically, a "compromised" recovery provider could observe the mempool, submit their own version of the same transaction, and be ordered in the block by the collator ahead of the "real" recovery provider. Easier to pull off with a colluding collator, but statistically you could count on it working close to 50% of the time even without that.
This isn't how the mempool works, but you kind want a "first one into the pool wins". The only way to avoid for now, is if the hashed data can somehow be tied to the "real" recovery provider's id, so that the "compromised" recovery provider can't formulate a valid transaction. I can't figure out how to do that with hash construction we have now, given that the recovery provider id isn't known in advance. You can hash provider id with email or email + recovery key, but there isn't anything for the chain to compare to (since it didn't know provider id before). Maybe you could do it with a zkproof (haven't thought it through enough), but short of that, I am not sure how to avoid front running concerns.

designdocs/recovery_system.md
```diff
@@ -0,0 +1,505 @@
...
+}
+```
+
+### Add Recovery Key
+
+This extrinsic allows a provider to add a new recovery key to an existing MSA. It can be used during account creation or later to update the recovery key.
+
+```rust
+// Pseudo-code for adding a recovery key to an existing MSA
+pub fn add_recovery_key(
```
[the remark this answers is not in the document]
harry-evans: I think yes, since MSA provisioning and related operations can be done with Capacity now.

designdocs/recovery_system.md
```diff
@@ -0,0 +1,505 @@
...
+
+## Context and Scope
+
+When a Frequency blockchain user’s wallet provider becomes unavailable, recovering control of their on-chain identity (Message Source Account, MSA) is critical. This document designs a secure recovery key system that lets users regain access to their MSA ID through governance-approved recovery providers. It outlines the cryptographic scheme, end-to-end workflows, security rationale, and implementation considerations. The goal is a robust recovery process that preserves decentralization and user security, even if the original wallet or provider is lost.
+
+## Key Requirements
+
+### Functional Requirements
+
+1. Recovery Key Generation: On MSA creation, the wallet provider collects the user’s email and generates a recovery key using a reversible combination of the email with a random salt/seed (e.g. a BIP-39 seed phrase). The resulting key (or a cryptographic derivative) is securely stored on-chain (encrypted or hashed) to allow future recovery.
```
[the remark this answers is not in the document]
mattheworris: Will include in coming updates.

designdocs/recovery_system.md
mattheworris: @wilwade suggestion added.
@harry-evans Thanks for the input. I think modifications changing `email` to become `the preferred contact method`, will allow the addition of attestations in the future. I will also keep in mind as design and implementation continue.

designdocs/recovery_system.md
mattheworris: Yes, I like that idea. I don't like having to provide the user with a private key or seed phrase for a wallet more than is absolutely necessary.

designdocs/recovery_system.md
mattheworris: Will include.

designdocs/recovery_system.md
mattheworris: Hashing updates coming.

mattheworris: Will be addressed in hashing updates.

designdocs/recovery_system.md
mattheworris: - Added note about normalization for Authentication Contact
- The Secret will be a random 32 byte hexadecimal value, open to formatting, but will have smaller blocks for readability.

designdocs/recovery_system.md
mattheworris: - Recovery Providers were added in new storage.

designdocs/recovery_system.md
```diff
@@ -0,0 +1,507 @@
...
+        who: T::AccountId,
+        msa_id: MsaId,
+    },
+    /// An account was recovered with a new control key
+    AccountRecovered {
+        who: T::AccountId,
+        msa_id: MsaId,
+        new_control_key: T::PublicKey, // New control key added to the MSA
+    },
+    /// A Recovery Secret was invalidated after use
```
wilwade, line 348:
```suggestion
    /// A Recovery Secret was invalidated after use or removal
```

designdocs/recovery_system.md
```diff
@@ -0,0 +1,507 @@
...
+
+## Key Requirements
+
+### Functional Requirements
+
+1. Recovery Secret Generation: On MSA creation, the wallet provider collects the user’s Authentication Contact method and generates a Recovery Secret using a reversible combination of the Authentication Contact with a random salt/seed. The resulting hash is securely stored on-chain to allow future recovery.
+
+2. Single Active Recovery Secret: A maximum of one Recovery Secret exists per user at any time. If a Recovery Secret is ever used (consumed) during account recovery, it is invalidated on-chain. The user or new provider may then issue a new Recovery Secret for future use, but duplicate or old keys cannot remain active simultaneously.
```
wilwade, line 13: We should likely be careful where we use Secret vs Hash
```suggestion
2. Single Active Recovery Hash: A maximum of one Recovery Hash exists per user at any time. If a Recovery Hash is ever used (consumed) during account recovery, it is invalidated on-chain. The user or new provider may then issue a new Recovery Secret for future use, but duplicate or old hashes cannot remain active simultaneously.
```

designdocs/recovery_system.md
```diff
@@ -0,0 +1,507 @@
...
+- **Recovery Provider**: A third-party service approved by Frequency governance to perform account recovery operations.
+- **Recovery Hash**: A cryptographic hash derived from the Recovery Secret and Authentication Contact, stored on-chain to allow recovery.
+- **Control Key**: A cryptographic key that allows the user to control their MSA. It can be a public key used for signing transactions or messages on the Frequency blockchain.
+- **Governance**: The on-chain mechanism that approves and manages recovery providers, ensuring they follow the protocol rules and security standards.
+- **Authentication Contact**: The user’s chosen method for receiving recovery-related communications, typically an email address or phone number.
+- **Contact Verification**: The process of confirming the user’s Authentication Contact method for recovery communications.
+- **Provider Approved Login Credentials**: Alternative authentication method (such as username/password or OAuth tokens) provided by the Recovery Provider when the user does not directly manage their own Control Key.
+
```
wilwade, line 45: Not sure how I feel about this, but perhaps a good term for the pair of hashes?
```suggestion
- **Recovery Intermediary Hashes**: A pair of cryptographic hashes that are used to protect the Secret and the Authentication Contact from needed to be published on chain during recovery.

```

designdocs/recovery_system.md
```diff
@@ -0,0 +1,507 @@
...
+This extrinsic allows a provider to add a new Recovery Hash to an existing MSA. It can be used during account creation or later to update the Recovery Hash.
+
+```rust
+// Payload for adding a Recovery Hash to an existing MSA
+#[derive(Encode, Decode, Clone, PartialEq, Eq, RuntimeDebug, TypeInfo)]
+pub struct AddRecoverySecretPayload<T: Config> {
+    pub recovery_hash: RecoveryHash, // The new Recovery Hash to be added
+    pub email: Vec<u8>, // The user’s email address used for recovery
+    pub recovery_code: Vec<u8>, // The recovery code provided by the user
```
wilwade, line 384: Also needs an expiration like other signature payloads

designdocs/recovery_system.md
```diff
@@ -0,0 +1,507 @@
...
+### Add Recovery Secret
+
+This extrinsic allows a provider to add a new Recovery Hash to an existing MSA. It can be used during account creation or later to update the Recovery Hash.
+
+```rust
+// Payload for adding a Recovery Hash to an existing MSA
+#[derive(Encode, Decode, Clone, PartialEq, Eq, RuntimeDebug, TypeInfo)]
+pub struct AddRecoverySecretPayload<T: Config> {
+    pub recovery_hash: RecoveryHash, // The new Recovery Hash to be added
+    pub email: Vec<u8>, // The user’s email address used for recovery
```
wilwade, line 383: Update to the hashes

designdocs/recovery_system.md
mattheworris: - updated

designdocs/recovery_system.md
mattheworris: - updated

designdocs/recovery_system.md
```diff
@@ -0,0 +1,507 @@
...
+Note: The Authentication Contact is typically the user’s email address, but could also be a phone number or other identifier. The hash ensures that even if the Authentication Contact is known, the Recovery Secret remains secure. The Authentication Contact must be normalized (e.g. lowercased for emails, whitespace removed, etc., or convert to [E.164](https://en.wikipedia.org/wiki/E.164) format for phone numbers) before hashing to ensure consistent results.
+
+Hashing the recovery data means the chain can later verify a recovery attempt by recomputing a hash, without needing to store any encryption keys. Hash storage is also aligned with security best practices. The only data needed for recovery is a successful hash match, rather than extracting plaintext.
+
+### Hash Algorithm
+
+This design uses a keccak256 hash function for the Recovery Hash (RH) due to its widespread use in blockchain applications and its security properties. The hash is computed as follows:
+
+```rust
+// Pseudocode for computing the Recovery Hash
```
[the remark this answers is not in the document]
mattheworris: - Applied

designdocs/recovery_system.md
mattheworris: For now, I left the addition of a new Recovery Hash as a separate extrinsic, `add_recovery_hash`. If we need to optimize blockchain transactions, we could revisit this in the future. I think keeping it separate is the most flexible at this time.

designdocs/recovery_system.md
mattheworris: All Recovery Providers (RP) have to be added by governance, the pool will probably be small. All Recovery Operations will log events that can be audited to keep RP's honest. I'd be interested in looking at ZK operations in the future.

designdocs/recovery_system.md
mattheworris: - language removed, but it's meant to suggest that Governance could require a stake that could be slashed for bad behavior. This is not included in this design.

designdocs/recovery_system.md
mattheworris: - `expiration` included in the payload refactor

designdocs/recovery_system.md
mattheworris: - applied in the refactor

designdocs/recovery_system.md
```diff
@@ -0,0 +1,587 @@
...
+
+### Recovery Hash Generation
+
+The blockchain combines the user’s Authentication Contact with the Recovery Secret, in the form of the Recovery Intermediary Hashes to preserve secrecy, in a secure way.
+
+Implementation: Deterministic Hash (One-Way): Compute a cryptographic hash
+
+#### Recovery Intermediary Hashes
+
+- H(s) = keccak256(Recovery Secret)
```
shannonwells, line 61: Inconsistency: Here the intermediary hash is defined as just the hash of the recovery secret, but below, the pseudocode says:
```
fn compute_recovery_hash(hash_recovery_secret: [u8; 32], hash_contact_secret: [u8; 32]) -> [u8; 32] {
```

and later:
```
pub struct RecoveryHashesPayload<T: Config> {
    /// The hash of the user's Authentication Contact
    pub contact_hash: Vec<u8>,
    /// The hash of (Recovery Secret || Authentication Contact)
    pub secret_hash: Vec<u8>,
    /// The block number at which a signed proof of this payload expires
    pub expiration:  BlockNumberFor<T>,
}
```

nit: Also I would prefer 'hash' or 'secret' but not both

designdocs/recovery_system.md
```diff
@@ -0,0 +1,587 @@
...
+- **Authentication Contact**: The user’s chosen method for receiving recovery-related communications, typically an email address or phone number.
+- **Contact Verification**: The process of confirming the user’s Authentication Contact method for recovery communications.
+- **Provider Approved Login Credentials**: Alternative authentication method (such as username/password or OAuth tokens) provided by the Recovery Provider when the user does not directly manage their own Control Key.
+- **Recovery Intermediary Hashes**: A pair of cryptographic hashes that are used to protect the Secret and the Authentication Contact from needing to be published on chain during recovery.
+
+## Cryptographic Design & Rationale
+
+### Recovery Secret Composition
+
+Each user’s Recovery Secret is derived from a combination of their Authentication Contact and a random secret. This design proposes using a 32 byte random hexadecimal value as the random secret component due to creating high entropy and user-friendliness.
```
shannonwells, line 51: I'm not sure why we are requiring both hashes. Seems like they could be part of the recovery process where you input your recovery email and secret at the same time, and the ORed hash only is submitted.

designdocs/recovery_system.md
```diff
@@ -0,0 +1,587 @@
...
+            Ok(())
+        },
+        None => Err(Error::<T>::RecoveryHashNotFound.into()),
+    }
+}
+```
+
+### Recover Account
+
+This extrinsic is called by the Recovery Provider after verifying the user’s Recovery Intermediary Hashes. It adds a new Control Key to the MSA and invalidates the old Recovery Hash.
```
shannonwells, line 477: We might also want to remove all existing Control Keys for that MSA.

designdocs/recovery_system.md
```diff
@@ -0,0 +1,587 @@
...
+
+    - Their MSA ID, if the user didn’t know it (users may not normally need to know their numeric MSA ID, but for transparency the Recovery Provider should tell them which account was recovered).
+    - The new Control Key credentials or Provider Approved Login Credentials.
+
+    At this point, the user can use a wallet, possibly the recovery provider’s interface or any Frequency-compatible wallet, to load the new Control Key and access their account (MSA). They should see that their data, messages, and any delegations under that MSA are intact. They have essentially regained control without needing the original provider.
+
+    Additionally, the governance or Frequency network can record that “Recovery Provider X performed recovery for MSA Y at time Z.” Recovery Providers are trusted, but such events may be monitored to ensure no Recovery Provider is abusing the system. If a Recovery Provider is found adding keys without proper user requests, governance can revoke their recovery role. This provides an accountability layer.
```
shannonwells, line 248: If we end up asking for a bonded stake to be a recovery provider then the penalty needs to include slashing.  If they're a Provider, governance might consider revoking their Provider status, slashing their stake for Capacity, or otherwise disallowing new delegations.

shannonwells (changes requested):
Question about the hashes, and an inconsistency that needs to be addressed.  Also wondering what people think about removing old Control Keys.  

The other comments about governance and slashing are non-blocking (also just saw in the resolved comments that that part is not part of this design doc)

designdocs/recovery_system.md
shannonwells: Re: intermediary hashes, I meant that in your docs here, it says the first hash is of the recovery secret, but in the pseudocode examples, it's of the authorized contact.  So that's why I was confused.

designdocs/recovery_system.md
```diff
@@ -0,0 +1,587 @@
...
+    Email = 0x00,
+    /// Phone number contact type (E.164 format)
+    PhoneNumber = 0x01,
+    /// Future extension for other contact types
+    /// Could include social media handles, attestation-based contacts, etc.
+}
+
+/// Authentication Contact structure with type-safe encoding
+/// Format: [type_byte] + [contact_data_bytes]
+/// Example: 0x01 + utf8_to_bytes("user@example.com")
```
wilwade, line 346:
```suggestion
/// Example: 0x00 + utf8_to_bytes("user@example.com")
```

designdocs/recovery_system.md
```diff
@@ -0,0 +1,587 @@
...
+
+This extrinsic allows a provider to add a new Recovery Hash to an existing MSA. It can be used during account creation or later to update the Recovery Hash.
+
+```rust
+// Pseudo-code for adding a Recovery Hash to an existing MSA
+pub fn add_recovery_hash(
+    origin: OriginFor<T>,
+    msa_owner_key: T::AccountId,
+    proof: MultiSignature,
+    payload: RecoveryHashesPayload,
```
wilwade, line 410: Only has the final recovery hash

designdocs/recovery_system.md
```diff
@@ -0,0 +1,587 @@
...
+
+```rust
+// Pseudo-code for verifying the Recovery Intermediary Hashes
+pub fn verify_recovery_hashes(
+    origin: OriginFor<T>,
+    msa_owner_key: T::AccountId,
+    proof: MultiSignature,
+    payload: RecoveryHashesPayload,
+) -> DispatchResult {
+    ensure_signed(origin)?;
```
wilwade, line 458:
```suggestion
    let who = ensure_signed(origin)?;
    
       // Verify that the Recovery Provider is a registered Recovery Provider
    ensure!(
        T::RecoveryProviderRegistry::is_recovery_provider(&who),
        Error::<T>::NotAuthorizedRecoveryProvider
    );
```

designdocs/recovery_system.md
mattheworris: Good catch. The pseudocode is incorrect, updated with new names. Also clarifying the concatenation.

designdocs/recovery_system.md
mattheworris: `||` means concatenated. I will update to use `+` and add a clarification that it means concatenation.

Using only `H(sc)` has several vulnerabilities:
1. Concatenation boundary attacks: Without proper separation, different secret+contact combinations could potentially collide
2. No proof of dual knowledge: The system can't verify the user knows both the secret AND the correct contact independently  
3. Weaker cryptographic commitment: A single hash doesn't create a robust commitment scheme that binds both factors securely

The dual-hash design `RC = keccak256(H(s) + H(sc))` creates a cryptographic commitment that:
- Requires proof of the Recovery Secret alone (via H(s))
- Requires proof of the Secret+Contact combination (via H(sc))  
- Binds both proofs together with fixed-length inputs (eliminating boundary issues)
- Ensures recovery requires knowledge of BOTH factors, not just a successful hash match

For account Recovery, `H(s)` and `H(sc)` will be computed and submitted by the Recovery Provider (RP). The user will provide the Recovery Secret and the Authentication Contact, according to the RP user interface. 

The blockchain should confirm that the Recovery Commitment (RC) is valid by computing it from the separate Recovery Intermediary Hashes before it recovers an account.

designdocs/recovery_system.md
```diff
@@ -0,0 +1,587 @@
...
+/// RecoveryHash is a 32-byte hash computed from the Recovery Intermediary Hashes.
+pub type RecoveryHash = [u8; 32];
+
+// Payload for adding a Recovery Hash to an existing MSA
+#[derive(Encode, Decode, Clone, PartialEq, Eq, RuntimeDebug, TypeInfo)]
+pub struct RecoveryHashesPayload<T: Config> {
+    /// The hash of the user's Authentication Contact
+    pub contact_hash: Vec<u8>,
+    /// The hash of (Recovery Secret || Authentication Contact)
+    pub secret_hash: Vec<u8>,
```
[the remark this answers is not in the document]
mattheworris: - not relevant, this struct was updated and used differently

shannonwells (approved):
A lot clearer, thanks

designdocs/recovery_system.md
```diff
@@ -0,0 +1,609 @@
...
+
+// Payload for verifying a Recovery Commitment for an existing MSA
+#[derive(Encode, Decode, Clone, PartialEq, Eq, RuntimeDebug, TypeInfo)]
+pub struct IntermediaryHashesPayload<T: Config> {
+    /// The hash of (Recovery Secret)
+    pub intermediary_hash_a: Vec<u8>,
+    /// The hash of (Recovery Secret + Authentication Contact)
+    pub intermediary_hash_b: Vec<u8>,
+    /// The block number at which a signed proof of this payload expires
+    pub expiration:  BlockNumberFor<T>,
```
wilwade, line 342: Since this doesn't get signed externally, we don't need an expiration.
```suggestion
```

designdocs/recovery_system.md
```diff
@@ -0,0 +1,609 @@
...
+#[pallet::storage]
+pub type RecoveryCommitmentToMsaId<T: Config> = StorageMap<_, Twox64Concat, RecoveryCommitment, MsaId>;
+```
+
+## Required Extrinsics
+
+|Name/Description|Caller|Payment|Key Events|Runtime Added|
+|---|---|---|---|---|
+| `add_recovery_commitment`<br />Add a new Recovery Commitment to an existing MSA| Provider| Capacity or Tokens | [`RecoveryCommitmentAdded`](https://frequency-chain.github.io/frequency/pallet_msa/pallet/enum.Event.html#variant.RecoveryCommitmentAdded)| 166|
+| `verify_intermediary_hashes`<br />Verify Recovery Intermediary Hashes against stored Recovery Commitment| Recovery Provider| Capacity or Tokens | [`RecoveryCommitmentVerified`](https://frequency-chain.github.io/frequency/pallet_msa/pallet/enum.Event.html#variant.RecoveryCommitmentVerified)| 166|
```
wilwade, line 416: This wouldn't be an Extrinsic, just an internal function or perhaps a Runtime RPC right?

designdocs/recovery_system.md
```diff
@@ -0,0 +1,609 @@
...
+```
+
+### Verify Intermediary Hashes
+
+This extrinsic is used by the Recovery Provider to verify if a user’s Recovery Intermediary Hashes compute a match to any existing MSA.
+
+```rust
+// Pseudo-code for verifying the Recovery Intermediary Hashes
+pub fn verify_intermediary_hashes(
+    payload: IntermediaryHashesPayload,
```
wilwade, line 463: I'm tempted to say we should just do two parameters instead of a payload system as it doesn't have an external signer.

designdocs/recovery_system.md
```diff
@@ -0,0 +1,609 @@
...
+### Recover Account
+
+This extrinsic is called by the Recovery Provider after verifying the user’s Recovery Intermediary Hashes. It adds a new Control Key to the MSA and invalidates the old Recovery Commitment.
+
+```rust
+// Pseudo-code for recovering an account with a new control key
+pub fn recover_account(
+    origin: OriginFor<T>,
+    new_control_key: Vec<u8>,
+    payload: IntermediaryHashesPayload,
```
wilwade, line 500: I'm tempted to say we should just do two parameters instead of a payload system as it doesn't have an external signer.

designdocs/recovery_system.md
```diff
@@ -0,0 +1,609 @@
...
+    /// A Recovery Commitment was added to an existing MSA
+    RecoveryCommitmentAdded {
+        who: T::AccountId,
+        msa_id: MsaId,
+    },
+    /// Recovery Intermediary Hashes were verified successfully
+    RecoveryCommitmentVerified {
+        who: T::AccountId,
+        msa_id: MsaId,
+    },
```
wilwade, line 379:
```suggestion
```

wilwade (changes requested):
I think we're down to the last changes. The language changes have been really helpful.

Only required changes:
- the removal /switch to rpc for the verification extrinsic.
- Removing the expiration value from the `IntermediaryHashesPayload`

designdocs/recovery_system.md
```diff
@@ -0,0 +1,609 @@
...
+      msa_id,
+    });
+
+    Ok(())
+}
+```
+
+### Verify Intermediary Hashes
+
+This extrinsic is used by the Recovery Provider to verify if a user’s Recovery Intermediary Hashes compute a match to any existing MSA.
```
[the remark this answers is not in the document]
mattheworris: Agreed, if H(s) and H(sc) are exposed, then the Recovery Commitment must be invalidated. 

Is an RPC secure enough for Recovery Providers to use without exposing H(s) and H(sc), or even RC?

Maybe we should alter the flow such that RP must validate the Authentication Contact before attempting to call `recover_account` and the extrinsic will return an error if it fails to find a Recovery Commitment? 

Why should the RP have a separate check for RC validation?

mattheworris:
> I think we're down to the last changes. The language changes have been really helpful.
> 
> Only required changes:
> 
> * the removal /switch to rpc for the verification extrinsic.
> * Removing the expiration value from the `IntermediaryHashesPayload`

Latest refactor implements these changes.

designdocs/recovery_system.md
```diff
@@ -0,0 +1,477 @@
...
+
+3. Governance-Approved Providers: Recovery operations are performed by special Recovery Providers that are approved via on-chain governance (listed in a registry). When a user needs account recovery, they contact one of these providers trusted to do the Contact Verification and submit their Recovery Secret along with their Authentication Contact information.
+
+4. Validation by Recovery Provider: The Recovery Provider must validate two things off-chain: (a) the submitted Recovery Secret (in combination with the Authentication Contact information) matches the on-chain Recovery Commitment for some MSA ID (proving the key is valid for that account), and (b) the provided Authentication Contact is truly controlled by the user (e.g. via a secure email verification challenge). Only if both checks pass is recovery authorized.
+
+    Note: The Recovery Provider must maintain an off-chain mapping of Recovery Commitments to MSA IDs. The blockchain will not store this mapping directly. The MSA ID is required to execute the `recover_account` extrinsic.
```
wilwade, line 19:
```suggestion
    Note: The Recovery Provider must maintain an off-chain mapping of MSA IDs to Recovery Commitments. The blockchain will not store this mapping directly. The MSA ID is required to execute the `recover_account` extrinsic.
```

designdocs/recovery_system.md
```diff
@@ -0,0 +1,476 @@
...
+Note: Here `+` denotes concatenation of byte arrays.
+
+#### Recovery Intermediary Hashes
+
+- H(s) = keccak256(Recovery Secret)
+- H(sc) = keccak256(Recovery Secret + Authentication Contact)  
+
+#### Final Recovery Commitment (RC)
+
+- RC = keccak256(H(s) + H(sc))
```
aramikm, line 69:
```suggestion
- RC = keccak256(H(s) || H(sc))
```

designdocs/recovery_system.md
```diff
@@ -0,0 +1,476 @@
...
+    B->>App: Return handle
+    App->>U: Show confirmed handle
+    B->>U: (Async) Send Frequency intro/recovery email including recovery code
+```
+
+#### Sequence diagram – Account creation and Recovery Secret setup
+
+1. User Account Creation: The user signs up with a wallet/app (the provider) and provides their Authentication Contact as part of registration. The provider creates a new MSA on the Frequency blockchain for the user. During this on-chain account creation, the user’s initial Control Key (public key) is linked to the MSA (this may be a key managed by the provider).
+
+2. Recovery Secret Generation: The blockchain generates a random secret (e.g. a 256-bit entropy value). It will call a secure random generator to get 32 bytes and then encode as a hexadecimal phrase.
```
aramikm, line 163: blockchain can not generate a secure random secret as far as I know. This has to be generated offchain

aramikm (approved):
Added some questions and suggestions but approving this to unblock

designdocs/recovery_system.md
```diff
@@ -0,0 +1,476 @@
...
+    )
+
+    // Register the signature to prevent replay attacks
+    Self::register_signature(&proof, payload.expiration.into());
+    
+    // Recover the MSA ID from the msa_owner_key
+    let msa_id = ensure_valid_msa_key(&msa_owner_key);
+
+    // Store the new RecoveryCommitment
+    MsaIdToRecoveryCommitment::<T>::insert(msa_id, payload.recovery_commitment);
```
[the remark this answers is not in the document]
wilwade: Override. It currently is a `set` operation, although may need to be different as the recovery system evolves.

designdocs/recovery_system.md
```diff
@@ -0,0 +1,476 @@
...
+
+    Additionally, the governance or Frequency network can record that “Recovery Provider X performed recovery for MSA Y at time Z.” Recovery Providers are trusted, but such events may be monitored to ensure no Recovery Provider is abusing the system. If a Recovery Provider is found adding keys without proper user requests, governance can revoke their recovery role. This provides an accountability layer.
+
+7. Post-Recovery: Re-Issuing Recovery Secret
+
+    After a successful recovery, if the user continues with a new wallet/provider, they may want to set up a new Recovery Secret for future safety (since the old one was consumed). The flow would essentially be the same as during account creation:
+
+    - The user’s new provider (or the recovery provider, if it doubles as a wallet provider) can generate a new secret, and update the on-chain record. The chain will accept it because the previous one was invalidated. The new Recovery Commitment replaces the old.
```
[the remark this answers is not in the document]
wilwade: The setting of a RC requires the control key signoff. So unless the provider can get the user to sign-off with the control key, it cannot do this.

Even if the provider were able to set it or maintained access to the recovery secret, it would still need to go through one of the approved recovery providers that would presumably request the contact method (under the control of the user) be validated.

Issue #2464, opened 2025-06-20
Recovery Provider Management


enddynayn:
**Description**
As a Frequency governance member, I want to approve and manage recovery providers so that only trusted entities can perform account recovery operations.

**Acceptance Criteria**
- [x] Governance can approve recovery providers via extrinsic
- [x] Governance can remove recovery providers via extrinsic
- [x] Only governance can approve/remove providers
- [x] Events are emitted for provider approval/removal
- [x] Provider status can be queried

**Design Notes:**
_Extrinsics_
- Implement approve_recovery_provider()
- Implement remove_recovery_provider() extrisic

_Origin_
- Add CreateProviderViaGovernanceOrigin

Storage
- Add RecoveryProviders

Events
- RecoveryProviderApproved 
- RecoveryProviderRemoved

PR #2469, opened 2025-06-23
feat(account): add Recovery Provider support

enddynayn:
Introduce support for designating and managing Recovery Providers,
allowing only trusted entities to perform account recovery actions.

Recovery Providers must be approved through governance and must already
be registered as a regular provider before they can take on the recovery role.

#2464


pallets/msa/src/weights.rs
```diff
@@ -454,13 +472,29 @@ impl WeightInfo for () {
...
+		//  Measured:  `1234`
 		//  Estimated: `6691`
-		// Minimum execution time: 83_233_000 picoseconds.
-		Weight::from_parts(88_177_000, 6691)
+		// Minimum execution time: 83_449_000 picoseconds.
+		Weight::from_parts(89_964_000, 6691)
 			.saturating_add(RocksDbWeight::get().reads(5_u64))
 			.saturating_add(RocksDbWeight::get().writes(5_u64))
 	}
+
```
enddynayn, line 482: Update Weights

pallets/msa/src/lib.rs
```diff
@@ -302,6 +309,16 @@ pub mod pallet {
 			/// The Delegator MSA Id
 			delegator_id: DelegatorId,
 		},
+		/// A recovery provider has been approved.
+		RecoveryProviderApproved {
+			/// The provider account ID
+			provider_id: ProviderId,
```
mattheworris, line 315: @wilwade Should we include some metadata with a URL for a recovery "app"? 

My concern here is once a user looks up a given Recovery Provider, they have access to 1) `ProviderId/MsaId` and 2) `Provider Name`. However, a user would still need to find the "app" where they can begin the recovery process. 

It may be enough that governance publishes an off-chain list of URL's for approved Recovery Providers, or we might want it all to be on-chain.

pallets/msa/src/lib.rs
```diff
@@ -981,10 +998,76 @@ pub mod pallet {
...
+		///
+		/// * [`DispatchError::BadOrigin`] - Caller is not authorized to remove recovery providers.
+		///
+		#[pallet::call_index(16)]
+		#[pallet::weight(T::WeightInfo::remove_recovery_provider())]
+		pub fn remove_recovery_provider(
+			origin: OriginFor<T>,
+			provider: ProviderId,
+		) -> DispatchResult {
+			T::CreateProviderViaGovernanceOrigin::ensure_origin(origin)?;
```
enddynayn, line 1050: RecoveryProviderApprovalOrigin

enddynayn: EitherOfDiverse<
			EnsureRoot<AccountId>,
			pallet_collective::EnsureProportionAtLeast<AccountId, CouncilCollective, 1, 2>,
		>,

pallets/msa/src/tests/mock.rs
```diff
@@ -234,6 +234,11 @@ impl pallet_msa::Config for Test {
 		EnsureRoot<AccountId>,
 		pallet_collective::EnsureMembers<AccountId, CouncilCollective, 1>,
 	>;
+
+	type RecoveryProviderApprovalOrigin = EitherOfDiverse<
+		EnsureRoot<AccountId>,
+		pallet_collective::EnsureProportionAtLeast<AccountId, CouncilCollective, 2, 3>,
```
mattheworris, line 240: Why the different proportion from mainnet (`2, 3` vs `1, 2`)?

mattheworris (approved):
- Read through changes, looks good.
- One non-blocking comment.
🚢 it!

PR #2472, opened 2025-06-26
Recovery System JS SDK

### LINKED ISSUE #2466: Recovery Code Verification (SDK?)
**Description**
As a recovery provider, I want to verify recovery codes without exposing private data so that I can validate recovery requests securely.

**Notes**
While this is done on-chain as part of recovery method perhaps this is an SDK?

wilwade:
# Goal
The goal of this PR is to generate the JS SDK that makes it easy to generate and use the recovery hashes and commitments

Closes #2466 

# Discussion

- Added CI
- Added Recovery SDK


js/recovery-sdk/src/standardize.ts
```diff
@@ -0,0 +1,42 @@
...
+/**
+ * Cleans up the email by lowercasing and removing periods in the username
+ * Does NOT validate the email
+ *
+ * @param email Raw email
+ * @returns standardized email string
+ */
+function standardizeEmail(email: string): string {
+  const [user, domain] = email.toLowerCase().split('@');
+  return user.replaceAll('.', '') + '@' + domain;
```
mattheworris, line 13: I think "dots optional" is a gmail-only feature.
I suggest that we leave "dots" for now, as it does not appear to be a universal.
Also suggest removing leading/trailing whitespace.

mattheworris (approved):

PR #2490, opened 2025-07-07
bug: remove unwrap; update lints

### LINKED ISSUE #2489: [Bug]: `unwrap` used in `ensure_msa_can_retire`
### What happened?

The lint for `unwrap_used` was inadvertently removed, which allowed an `unwrap` to be merged. When the lint was re-enabled, the code fails to pass CI.

## Acceptance Criteria

- [x] Add `unwrap_used` lint back to `Cargo.toml` as in this [example](https://github.com/frequency-chain/frequency/pull/2364/files#diff-2e9d962a08321605940b5a657135052fbcef87b5e360662bb527c96d9a615542R195).
- [x] Remove `unwrap` from this [line](https://github.com/frequency-chain/frequency/blob/main/pallets/msa/src/lib.rs#L2043).

### Relevant log output

```shell
error: used `unwrap()` on a `Result` value
    --> pallets/msa/src/lib.rs:1970:24
     |
1970 |         let msa_account_id = T::AccountId::decode(&mut bytes).unwrap();
     |                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
     |
     = note: if this value is an `Err`, it will panic
     = help: consider using `expect()` to provide a better panic message
     = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#unwrap_used
     = note: requested on the command line with `-D clippy::unwrap-used`

error: could not compile `pallet-msa` (lib) due to 1 previous error
warning: build failed, waiting for other jobs to finish...
Error: Process completed with exit code 101.
```

mattheworris:
# Goal
The goal of this PR is to update the `unwrap_used` lint and fix `unwrap` errors.

Closes #2489

# Discussion

- 

# Checklist
- [ ] Updated Pallet Readme?
- [ ] Design doc(s) updated?
- [ ] Unit Tests added?
- [ ] e2e Tests added?
- [ ] Benchmarks added?

### DIFF
--- a/node/service/src/chain_spec/frequency_paseo_local.rs
+++ b/node/service/src/chain_spec/frequency_paseo_local.rs
@@ -15,7 +15,7 @@ pub type RelayChainSpec = sc_service::GenericChainSpec<RelayChainExtensions>;
 /// Generates the Paseo-Local Relay chain spec from the json
 pub fn load_paseo_local_spec() -> RelayChainSpec {
 	RelayChainSpec::from_json_bytes(&include_bytes!("../../../../resources/paseo-local.json")[..])
-		.unwrap()
+		.expect("Failed to parse paseo-local.json chain spec")
 }
 
 /// Generates the chain spec for a local testnet
--- a/pallets/msa/src/lib.rs
+++ b/pallets/msa/src/lib.rs
@@ -2040,7 +2040,8 @@ impl<T: Config + Send + Sync> CheckFreeExtrinsicUse<T> {
 
 		// - Convert to AccountId
 		let mut bytes = &EthereumAddressMapper::to_bytes32(&msa_address.0)[..];
-		let msa_account_id = T::AccountId::decode(&mut bytes).unwrap();
+		let msa_account_id = T::AccountId::decode(&mut bytes)
+			.map_err(|_| InvalidTransaction::Custom(ValidityError::InvalidMsaKey as u8))?;
 
 		// - Check that the MSA does not have a token balance
 		let msa_balance = T::Currency::reducible_balance(
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -176,6 +176,7 @@ complexity = { level = "deny", priority = -1 }
 perf = { level = "deny", priority = -1 }
 suspicious = { level = "deny", priority = -1 }
 # These should stay
+unwrap_used = "deny"
 wildcard_dependencies = "deny"
 from_over_into = "allow"
 type_complexity = "allow"


wilwade (approved):
👍

saraswatpuneet (approved):

node/service/src/chain_spec/frequency_paseo_local.rs
```diff
@@ -15,7 +15,7 @@ pub type RelayChainSpec = sc_service::GenericChainSpec<RelayChainExtensions>;
 /// Generates the Paseo-Local Relay chain spec from the json
 pub fn load_paseo_local_spec() -> RelayChainSpec {
 	RelayChainSpec::from_json_bytes(&include_bytes!("../../../../resources/paseo-local.json")[..])
-		.unwrap()
+		.expect("Failed to parse paseo-local.json chain spec")
```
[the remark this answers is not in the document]
JoeCap08055: This is _maybe_ OK, since it's in the node, not the runtime? Are we okay panicking here?

pallets/msa/src/lib.rs
```diff
@@ -2040,7 +2040,8 @@ impl<T: Config + Send + Sync> CheckFreeExtrinsicUse<T> {
 
 		// - Convert to AccountId
 		let mut bytes = &EthereumAddressMapper::to_bytes32(&msa_address.0)[..];
-		let msa_account_id = T::AccountId::decode(&mut bytes).unwrap();
+		let msa_account_id = T::AccountId::decode(&mut bytes)
+			.map_err(|_| InvalidTransaction::Custom(ValidityError::InvalidMsaKey as u8))?;
```
JoeCap08055, line 2044: Not sure what the error should be here. Theoretically, this `.unwrap()` can _never_ fail, because it's merely using our algorithm for generating an Ethereum address from an MSA ID. Do we just leave this error mapping here and call it a day? If this fails it means somethings wrong with our address mapper algorithm itself (ie, it's not a problem with the extrinsic input). So maybe a new, louder error (that should hopefully never be surfaced...)?

mattheworris: There are a couple places where we extract `msa_account_id` from `EthereumAddressMapper`, maybe this one is better for this case? (And the code should use this everywhere this pattern is repeated-->might be a candidate for a small refactor, I think found 3 places with this code).

```rust
			let msa_account_id = T::AccountId::decode(&mut bytes).map_err(|_| {
				log::error!("Failed to decode MSA account ID from Ethereum address");
				Error::<T>::NoKeyExists
			})?;
```

wilwade: Yeah. The code here would throw the error, but wouldn't panic the wasm.

mattheworris, line 2044:
```suggestion
		let msa_account_id = T::AccountId::decode(&mut bytes).map_err(|_| {
			log::error!("Failed to decode MSA account ID from Ethereum address");
			InvalidTransaction::Custom(ValidityError::InvalidMsaKey as u8)
		})?;
```

pallets/msa/src/lib.rs
```diff
@@ -2120,8 +2123,10 @@ impl<T: Config + Send + Sync> CheckFreeExtrinsicUse<T> {
 
 		// - Convert to AccountId
 		let mut bytes = &EthereumAddressMapper::to_bytes32(&msa_address.0)[..];
-		let msa_account_id = T::AccountId::decode(&mut bytes)
-			.map_err(|_| InvalidTransaction::Custom(ValidityError::InvalidMsaKey as u8))?;
+		let msa_account_id = T::AccountId::decode(&mut bytes).map_err(|_| {
```
mattheworris, line 2126: Note: aligning all 3 uses of EthereumAddressMapper to the same error behavior.

PR #2496, opened 2025-07-09
Implement recover_account

mattheworris:
This pull request implements the `recover_account` extrinsic in the MSA pallet, enabling Recovery Providers to recover control of an MSA for a user who has the Recovery Secret and has maintained control over the Authentication Contact used when setting up the Recovery Commitment. Key updates include new storage types, extrinsics, events, tests, and benchmarks to support recovery operations.

### Recovery Provider System Enhancements:
* Added new events `AccountRecovered` and `RecoveryCommitmentInvalidated` for tracking recovery-related operations.
* Added extrinsic `recover_account`.

### Testing Improvements:
* Added a new e2e test `recoverAccount.test.ts` to validate account recovery functionality, including success and failure scenarios such as invalid intermediary hashes, unregistered providers, and mismatched MSA IDs.
* Enhanced the `transactions.test.ts` file with a new test case to verify account recovery using capacity for eligible transactions.

### Formatting Enhancements:
* Added `format-js` and `format-all` targets to the `Makefile` to streamline JavaScript code formatting across multiple directories.

### Documentation and Error Handling:
* Updated documentation to include recovery provider functionality and state queries.
* Added new error types for recovery operations, such as `NotAuthorizedRecoveryProvider` and `InvalidRecoveryCommitment`.

# Checklist
- [x] Updated Pallet Readme?
- [x] Design doc(s) updated?
- [x] Unit Tests added?
- [x] e2e Tests added?
- [x] Benchmarks added?


pallets/msa/src/tests/recovery_tests.rs
```diff
@@ -279,3 +264,459 @@ fn add_recovery_commitment_unsigned_origin_should_fail() {
...
+			Error::<Test>::NoRecoveryCommitment
+		);
+	});
+}
+
+#[test]
+fn recover_account_with_no_recovery_commitment_should_fail() {
+	new_test_ext().execute_with(|| {
+		// Create an MSA account without recovery commitment
+		let (msa_id, _msa_owner_key_pair) = create_account();
```
shannonwells, line 453: This seems like it exercises the same code path as the one above, but... I guess it should be here for completeness?

mattheworris: I had the same thought and almost removed it. 
The code fails the same for no MSA and an MSA with no RC.
I think I will remove the test for no MSA, as that test has no security implications if it fails.

pallets/frequency-tx-payment/src/tests/stable_weights_tests.rs
```diff
@@ -26,6 +26,7 @@ fn test_weights_are_stable() {
 				611897000,
 				5733,
 			),
+			("recover_account", SubstrateWeight::<Test>::recover_account(), 1041409000, 6531),
```
mattheworris, line 29: Computed from `recover_account` function in `weights.rs`, copied to `capacity_stable_weights.rs`.
Base weight comes from CI benchmarks.

```
Read = 47_280_000
Write = 74_376_000
Base weight = 169_633_000

                 169_633_000
9 * 47_280_000 = 425_520_000
6 * 74_376_000 = 446_256_000
----------------------------
               1_041_409_000
```

pallets/msa/src/benchmarking.rs
```diff
@@ -488,6 +488,105 @@ mod benchmarks {
 		Ok(())
 	}
 
+	#[benchmark]
+	fn recover_account() -> Result<(), BenchmarkError> {
+		frame_system::Pallet::<T>::set_block_number(1u32.into());
```
enddynayn, line 493: Everything looks great! I was just wondering the reason for setting the block number here?

pallets/msa/src/lib.rs
```diff
@@ -1134,6 +1161,104 @@ pub mod pallet {
...
+			intermediary_hash_a: RecoveryHash,
+			intermediary_hash_b: RecoveryHash,
+			new_control_key_proof: MultiSignature,
+			add_key_payload: AddKeyData<T>,
+		) -> DispatchResult {
+			let provider_key = ensure_signed(origin)?;
+
+			// Get the provider's MSA ID and ensure they're a registered provider
+			let provider_msa_id = Self::ensure_valid_msa_key(&provider_key)?;
+			ensure!(
```
enddynayn, line 1192: Not a big deal, but I think we could probably remove this check since becoming a recovery provider already requires being a registered provider. However, I am ok with keeping it.

pallets/msa/src/benchmarking.rs
mattheworris: In the mock setup, the signature registry is filled with signatures that expire on block 0. The test needs the block to be > 0, so that it can add a new signature to a full registry. This ensures the worst case scenario for the benchmarks.

pallets/msa/src/lib.rs
```diff
@@ -1134,6 +1161,104 @@ pub mod pallet {
...
+					&new_control_key_proof,
+					&add_key_payload.new_public_key,
+					&add_key_payload
+				),
+				Error::<T>::NewKeyOwnershipInvalidSignature
+			);
+			Self::register_signature(&new_control_key_proof, add_key_payload.expiration)?;
+
+			// Add the new control key to the MSA and the offchain index
+			Self::add_key(
```
enddynayn, line 1231: I think we should consider creating a ticket to refactor this function. It was originally written at a time when errors didn’t revert any state changes.

pallets/msa/src/lib.rs
```diff
@@ -1134,6 +1161,104 @@ pub mod pallet {
...
+			let provider_key = ensure_signed(origin)?;
+
+			// Get the provider's MSA ID and ensure they're a registered provider
+			let provider_msa_id = Self::ensure_valid_msa_key(&provider_key)?;
+			ensure!(
+				Self::is_registered_provider(provider_msa_id),
+				Error::<T>::ProviderNotRegistered
+			);
+
+			// Ensure the provider is approved for recovery operations
```
enddynayn, line 1197: I suggest breaking this function up so that we can unit test each component.
- ensure_provider_is_authorized
- ensure_commitment_matches
- verify_new_key_ownership

pallets/msa/src/lib.rs
```diff
@@ -1134,6 +1161,104 @@ pub mod pallet {
...
+			// Implement the signature verification for the new control key, which is usually done by
+			// add_public_key_to_msa, but in this case we do not have the MSA owner key.
+			ensure!(
+				Self::verify_signature(
+					&new_control_key_proof,
+					&add_key_payload.new_public_key,
+					&add_key_payload
+				),
+				Error::<T>::NewKeyOwnershipInvalidSignature
+			);
```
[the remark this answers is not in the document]
mattheworris: `add_key_payload.msa_id` is used to get the Recovery Commitment (RC).
If the RC does not match, then the code returns `InvalidRecoveryCommitment`.
This matches the `msa_id` to the RC and makes it impossible to use a different `msa_id` from the one that added the original RC.

There is an element of trust in the Recovery Provider (RP), as the RP has all the information needed to execute `recover_account`.

pallets/msa/src/lib.rs
mattheworris: Good idea, removed, fixed tests, and re-run benchmarks.

pallets/msa/src/lib.rs
```diff
@@ -1134,6 +1161,104 @@ pub mod pallet {
...
+
+			// Implement the signature verification for the new control key, which is usually done by
+			// add_public_key_to_msa, but in this case we do not have the MSA owner key.
+			ensure!(
+				Self::verify_signature(
+					&new_control_key_proof,
+					&add_key_payload.new_public_key,
+					&add_key_payload
+				),
+				Error::<T>::NewKeyOwnershipInvalidSignature
```
saraswatpuneet, line 1222: Not sure if some of these checks can be moved to extension side that might balance the weight better, i recall 6-9 rw

pallets/msa/src/lib.rs
```diff
@@ -1134,6 +1161,100 @@ pub mod pallet {
...
+		///
+		/// # Errors
+		/// * [`Error::NotAuthorizedRecoveryProvider`] - Caller is not an approved Recovery Provider.
+		/// * [`Error::ProviderNotRegistered`] - Recovery Provider is not registered.
+		/// * [`Error::NoRecoveryCommitment`] - No Recovery Commitment exists for the MSA.
+		/// * [`Error::InvalidRecoveryCommitment`] - Recovery Commitment does not match.
+		///
+		#[pallet::call_index(18)]
+		#[pallet::weight(T::WeightInfo::recover_account())]
+		pub fn recover_account(
```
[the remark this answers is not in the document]
mattheworris: Discussion notes:

- The extrinsic itself should not remove all the old keys, as it would limit other possible uses of the Recovery System, e.g. users of embedded wallets may want to add a more open key to use for login at other services.
- It is envisioned that the Recovery Provider User Interface could help guide a user through removing all existing keys, should that be desired by the user.
- If guided to remove all existing keys, that would also remove all existing delegations.

pallets/msa/src/lib.rs
```diff
@@ -1134,6 +1161,100 @@ pub mod pallet {
...
+			origin: OriginFor<T>,
+			intermediary_hash_a: RecoveryHash,
+			intermediary_hash_b: RecoveryHash,
+			new_control_key_proof: MultiSignature,
+			add_key_payload: AddKeyData<T>,
+		) -> DispatchResult {
+			let provider_key = ensure_signed(origin)?;
+
+			// Get the provider's MSA ID and ensure they're a registered provider
```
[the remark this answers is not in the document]
enddynayn: I suggest completely removing the comment.

e2e/msa/recoverAccount.test.ts
JoeCap08055, line 1: We should have an official policy for E2E tests going forward, whether we need to prefer testing with Ethereum keys, Sr25519 keys, or always both. I *think* we're moving toward, if only testing one key type, Ethereum-style keys.

pallets/msa/src/lib.rs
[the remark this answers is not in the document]
JoeCap08055: > * If guided to remove all existing keys, that would also remove all existing delegations.

Why? Delegations have nothing to do with public keys. We only require all delegations to be removed if you're retiring an MSA (which also requires removing all public keys from the MSA).

pallets/msa/src/lib.rs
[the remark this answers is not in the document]
mattheworris: Comment removed in refactor.

pallets/msa/src/lib.rs
mattheworris: `recover_account` refactored, and unit tests implmented for helper functions.

e2e/msa/recoverAccount.test.ts
mattheworris: - Refactored `generateSignedAddKeyProof` into `generateSignedAddKeyProofWithType`, which can sign with either sr25519 or EIP-712 based on the keypair type.
- Added e2e test for EIP-712 signed version.

pallets/msa/src/lib.rs
```diff
@@ -1134,6 +1161,73 @@ pub mod pallet {
...
+				add_key_payload.msa_id,
+			)?;
+
+			Self::ensure_valid_new_key_owner(&new_control_key_proof, &add_key_payload)?;
+
+			Self::add_key(
+				add_key_payload.msa_id,
+				&add_key_payload.new_public_key.clone(),
+				|msa_id| -> DispatchResult {
+					let event = Event::PublicKeyAdded {
```
enddynayn, line 1204: Minor: I think we can remove this event since adding a key is implied by AccountRecovered event.

enddynayn (approved):
Great work — clearly a lot of effort went into this. 👍

pallets/msa/src/lib.rs
mattheworris: Is this needed for the `offchain_index_event` and to keep the index in sync?

enddynayn: Ahh good question. @aramikm ?

aramikm: Yes, please keep it

pallets/msa/src/benchmarking.rs
```diff
@@ -488,6 +488,105 @@ mod benchmarks {
 		Ok(())
 	}
 
+	#[benchmark]
+	fn recover_account() -> Result<(), BenchmarkError> {
```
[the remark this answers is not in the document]
mattheworris: - Implemented a helper function `prep_recovery_benchmark_storage` which sets up the "lost" msa and a Recovery Provider and does not require executing the `add_recovery_commitment`.

aramikm (approved):

shannonwells (approved):
Good work.  Looks good to go.  I don't have anything else to add

PR #2506, opened 2025-07-18
Add a new failure check for when CheckingAccount is not funded

shannonwells:
If the CheckingAccount is never funded, a teleport from AssetHub to Frequency should fail.

Includes a little refactoring to eliminate some duplicate code, and fixes some warnings.


enddynayn (approved):
👍 Great!

PR #2516, opened 2025-07-23
feat: Add autogenerated weights for pallet_xcm

mattheworris:
# Goal
The goal of this PR is to add Benchmark Helpers for pallet-xcm.

# Discussion

- Asset Hub has an implementation like [this](https://github.com/paritytech/polkadot-sdk/blob/b4b019e4db0ef47b0952638388eba4958e1c4004/cumulus/parachains/runtimes/assets/asset-hub-westend/src/lib.rs#L2206).
- Moonbeam has their implementation [here](https://github.com/moonbeam-foundation/moonbeam/blob/2afb0c8f025b9a7e56f2652de9143af1fb6b6b2d/runtime/common/src/apis.rs#L862).


runtime/frequency/src/lib.rs
```diff
@@ -2095,6 +2099,44 @@ sp_api::impl_runtime_apis! {
...
+				/// A destination that `XcmRouter` can reach in your test setup.
+				/// The simplest is usually the relay‐chain:
+				fn reachable_dest() -> Option<staging_xcm::latest::Location> {
+					Some(staging_xcm::latest::prelude::Parent.into())
+				}
+
+				/// Return None to skip the teleport_assets benchmark (not properly configured).
+				fn teleportable_asset_and_dest() -> Option<(staging_xcm::latest::Asset, staging_xcm::latest::Location)> {
+					// Skip teleport benchmarks as the fee infrastructure isn't set up properly for benchmarking
+					None
```
mattheworris, line 2116: These functions returning `None` need to be implemented in order to get a correct weight.

runtime/frequency/src/lib.rs
```diff
@@ -2095,6 +2099,44 @@ sp_api::impl_runtime_apis! {
 			use cumulus_pallet_session_benchmarking::Pallet as SessionBench;
 			impl cumulus_pallet_session_benchmarking::Config for Runtime {}
 
+			// XCM benchmarking configuration
+			impl pallet_xcm::benchmarking::Config for Runtime {
+				/// Helper that makes sure `SendXcm` succeeds in the benches.
+				type DeliveryHelper = ();
```
mattheworris, line 2105: Moonbeam has implemented [TestDeliveryHelper](https://github.com/moonbeam-foundation/moonbeam/blob/2afb0c8f025b9a7e56f2652de9143af1fb6b6b2d/runtime/common/src/apis.rs#L835), which does some handling for ETH address and may be applicable for Frequency.

runtime/frequency/src/lib.rs
saraswatpuneet: Are we using weights available from polkadot for these then? Not sure if weights is natively generated

runtime/common/src/weights/pallet_xcm.rs
```diff
@@ -0,0 +1,490 @@
...
+		//  Measured:  `4`
+		//  Estimated: `1984`
+		// Minimum execution time: 17_032_000 picoseconds.
+		Weight::from_parts(17_793_000, 1984)
+			.saturating_add(T::DbWeight::get().reads(3_u64))
+			.saturating_add(T::DbWeight::get().writes(1_u64))
+	}
+	/// Storage: `Benchmark::Override` (r:0 w:0)
+	/// Proof: `Benchmark::Override` (`max_values`: None, `max_size`: None, mode: `Measured`)
+	fn teleport_assets() -> Weight {
```
enddynayn, line 53: We do not teleport any other asset using this extrinsic so does not matter.

enddynayn:
replaced by:
https://github.com/frequency-chain/frequency/pull/2364/commits/288f2607232c355383baf303aff572486dfd89fc

PR #2527, opened 2025-07-28
Fix E2E Tests Improve

wilwade:
# Goal
The goal of this PR is to make some of the e2e tests less flaky.

# Discussion

-  withdrawTokens doesn't change the nonce, so reset the autononce system for it
- Delegations don't need to be checked for order, just contents


enddynayn (approved):
great! 👍

saraswatpuneet (approved):

PR #2531, opened 2025-07-28
fix: E2E package deps for running locally

### LINKED ISSUE #2529: [chore] Checksum conflict when working with local JS packages for E2E tests
# Description
Currently, the script `generate_js_definitions.sh` used by the Makefile target `make js` builds local npm tarballs of the JS packages and installs them to the E2E test suite's `package.json` by doing:
```console
npm install <path-to-tarball>
```

While this works the first time it's done, this tends to cause problems when working with the same repo clone over time--especially if working on changes to the JS packages themselves. There are 2 issues:
1. Once the tarball package is installed, subsequent `npm install` invocations often encounter a checksum error with the `package-lock.json` if the tarball has since changed
2. Since the embedded npm package version of all of the generated tarballs is `0.0.0`, npm does not refresh the local package cache when doing another install. This is problematic, especially if testing changes to the JS packages themselves.

NOTE: This is not a problem for the CI pipeline, because it always uses a fresh repo checkout & npm cache.

# Suggested Solution
Amend the `generate_js_definitions.sh` script so that, instead of installing from the package tarball, we install from the built `dist` directory of the package. This causes a symlink to be generated in the `node_modules` folder, so that any time the JS packages are rebuilt, the changes will be reflected in the E2E runtime environment.

This should ONLY affect running E2E tests locally, as none of our CI pipelines use that script.

JoeCap08055:
# Goal
The goal of this PR is to eliminate workflow issues when running E2E tests locally.

Closes #2529 

# Discussion

- 

# Checklist
- [ ] Updated Pallet Readme?
- [ ] Design doc(s) updated?
- [ ] Unit Tests added?
- [ ] e2e Tests added?
- [ ] Benchmarks added?

### DIFF
--- a/e2e/package.json
+++ b/e2e/package.json
@@ -21,9 +21,9 @@
   "author": "",
   "license": "Apache-2.0",
   "dependencies": {
-    "@frequency-chain/api-augment": "file:../js/api-augment/dist/frequency-chain-api-augment-0.0.0.tgz",
-    "@frequency-chain/ethereum-utils": "file:../js/ethereum-utils/dist/frequency-chain-ethereum-utils-0.0.0.tgz",
-    "@frequency-chain/recovery-sdk": "file:../js/recovery-sdk/dist/frequency-chain-recovery-sdk-0.0.0.tgz",
+    "@frequency-chain/api-augment": "file:../js/api-augment/dist",
+    "@frequency-chain/ethereum-utils": "file:../js/ethereum-utils/dist",
+    "@frequency-chain/recovery-sdk": "file:../js/recovery-sdk/dist",
     "@helia/unixfs": "^5.0.3",
     "@noble/curves": "^1.9.2",
     "@polkadot-api/merkleize-metadata": "^1.1.18",


scripts/run_e2e_tests.sh
```diff
@@ -139,35 +139,27 @@ else
...
 
-
 cd e2e
-echo "Installing js/api-augment/dist/frequency-chain-api-augment-0.0.0.tgz"
-echo "Installing js/ethereum-utils/dist/frequency-chain-ethereum-utils-0.0.0.tgz"
-npm i ../js/api-augment/dist/frequency-chain-api-augment-0.0.0.tgz ../js/ethereum-utils/dist/frequency-chain-ethereum-utils-0.0.0.tgz
+echo "Installing local JS packages..."
+declare -a pre=( ${JS_PACKAGES[@]/#/..\/js\/} )
+packages=${pre[@]/%/\/dist}
+npm i ${packages}
```
wilwade, line 162: Do we need to update the package.json with the result?

```
"@frequency-chain/api-augment": "file:../js/api-augment/dist",
"@frequency-chain/ethereum-utils": "file:../js/ethereum-utils/dist",
"@frequency-chain/recovery-sdk": "file:../js/recovery-sdk/dist",
```

wilwade: I think for local running we want to have no changes that might be committed, and currently it points at the tarball, so I think pointing at the js/dist folder is the way to go. Mostly as someone will eventually commit that anyway.

shannonwells (approved):
Re: updating package.json, I'm good with whatever, also thanks for the shell script cleanup!

saraswatpuneet (approved):

PR #2537, opened 2025-08-01
[Provider Context] Update provider creation to intake new payload + changes around it

### LINKED ISSUE #2523: [Provider Context] Update `propose_to_be_provider` and modify `ProviderRegistryEntry`
## Details

`propose_to_be_provider` and `ProviderRegistryEntry` needs update inline with provider context [design doc](https://github.com/frequency-chain/frequency/blob/main/designdocs/provider_contexts.md#changes-and-additions-in-extrinsics-)

- [x] Update provider registry entry with new type fields such  as default name, logo hash, localized names map and localized logo hash map
- [x] Update `propose_to_be_provider` to accept new type.
- [x] should also insert all logo hashes inside payload to `ApprovedLogoHashes` upon successful approval
- [x] Update benchmarks, add unit tests, update e2e tests
- [x] Add a storage migration for `ProviderToRegistryEntry`

## Acceptance criteria

1. Updated provider proposal should work with new payload type
2. Upon approval the hashes should appear in `ApprovedLogoHashes` and check for failure cases
3. Benchmarks weight should be updated
4. State query for provider registry should return payload with new type fields

saraswatpuneet:
# Goal
This PR introduces changes required to support the new Provider Context model as outlined in the [Provider Contexts Design Doc](https://github.com/frequency-chain/frequency/blob/main/designdocs/provider_contexts.md). It updates the structure and handling of providers to allow them to represent a company with one or more applications, supporting internationalized metadata and trusted branding via content-addressed hashes.

Closes  #2523 

### Registry and Extrinsics
-  Updated ProviderRegistryEntry:
-  Replaces the old struct name with a new struct supporting:
            default_name: BoundedString
            default_logo: Option<Cid>
            localized_name: BoundedBTreeMap<LanguageCode, BoundedString>
            localized_logo: BoundedBTreeMap<LanguageCode, Cid>

- Updated propose_to_be_provider extrinsic:
- Updated create_provider_by_governance extrinsic
- Updated create_provider extrinsic
- Updated Benchmarks
- Updated Unit tests/e2e tests
- Added e2e tests
- Added unit tests
- Added `ProviderToRegistryEntry` to migration to storage version `2`
- Tested try-runtime migration with `Paseo` and `Mainnet` and it works

## In progress
- [x] add more unit tests
- [ ] run a full cycle of try-runtime to check migrations 

# Checklist
- [ ] Updated Pallet Readme?
- [ ] Design doc(s) updated?
- [x] Unit Tests added?
- [x] e2e Tests added?
- [x] Benchmarks added?

## Snaps
- Successful execution of `on_runtime_upgrade` via try-runtime cli on paseo
image


.github/workflows/verify-pr-commit.yml
```diff
@@ -9,6 +9,7 @@ on:
       - main
       - "**-development" # Feature Branches should suffix with -development
       - hotfix/**
+      - feat/provider-context
```
JoeCap08055, line 12: Could just rename the branch `feat/provider-context-development` or `provider-context-development` to avoid the need for this

pallets/capacity/src/tests/testing_utils.rs
```diff
@@ -71,7 +71,17 @@ pub fn system_run_blocks(n: u32) {
 
 pub fn register_provider(target_id: MessageSourceId, name: String) {
 	let name = Vec::from(name).try_into().expect("error");
-	assert_ok!(Msa::create_registered_provider(target_id.into(), name));
+	let cid = "bafkreidgvpkjawlxz6sffxzwgooowe5yt7i6wsyg236mfoks77nywkptdq"
```
JoeCap08055, line 74: Similar to the JS/E2E side, it would probably be useful to have a helper function to create a valid ProviderEntry.

pallets/msa/src/tests/creation_tests.rs
```diff
@@ -320,3 +359,137 @@ fn verify_signature_without_wrapped_bytes() {
...
+		// Fail to register provider with invalid CID
+		assert_noop!(
+			Msa::create_provider(RuntimeOrigin::signed(provider_account.into()), entry),
+			Error::<Test>::InvalidCid
+		);
+	});
+}
+
+#[test]
+pub fn create_provider_fails_with_invalid_locale() {
```
JoeCap08055, line 420:
```suggestion
pub fn create_provider_fails_with_invalid_logo_locale() {
```

runtime/common/src/constants.rs
```diff
@@ -85,13 +85,51 @@ pub type FrameSystemMaxConsumers = ConstU32<16>;
...
 /// The maximum number of public keys per MSA
 pub type MsaMaxPublicKeysPerMsa = ConstU8<25>;
-/// The maximum size of the provider name (in bytes)
-pub type MsaMaxProviderNameSize = ConstU32<16>;
 /// The number of blocks per virtual bucket
 pub type MSAMortalityWindowSize = ConstU32<{ 20 * MINUTES }>;
 /// The upper limit on total stored signatures.
 /// Set to an average of 50 signatures per block
 pub type MSAMaxSignaturesStored = ConstU32<50_000>;
+/// The maximum size of a provider name (in bytes)
```
JoeCap08055, line 93: Any reason these can't just be, say `ConstU32<256>`, for example?

pallets/msa/src/lib.rs
```diff
@@ -594,15 +633,23 @@ pub mod pallet {
 		///
 		/// # Errors
 		/// * [`Error::NoKeyExists`] - origin does not have an MSA
-		/// * [`Error::ExceedsMaxProviderNameSize`] - Too long of a provider name
 		/// * [`Error::DuplicateProviderRegistryEntry`] - a ProviderRegistryEntry associated with the given MSA id already exists.
 		///
 		#[pallet::call_index(2)]
 		#[pallet::weight(T::WeightInfo::create_provider())]
-		pub fn create_provider(origin: OriginFor<T>, provider_name: Vec<u8>) -> DispatchResult {
+		pub fn create_provider(
```
JoeCap08055, line 640: Did we make a decision to make these breaking changes, rather than do a `create_provider_v2`? This will, at minimum, break Provider Dashboard and many E2E tests across many repos...

saraswatpuneet: Good suggestion, This definitely a breaking change, thanks for putting in perspective. @wilwade  are you comfortable making this change all the way up.

My impression is to keep it in feature branch will all necessary changes arent not incorporated, since this only affects Provider Dashboard, and merge all together, if it was affecting gateway I would say yes go with v2 first

runtime/common/src/constants.rs
saraswatpuneet: ConstU32 does not implement some additional macros needed by pallet/extrinsic constraints, so need a custom type to satisfy those bounds

e2e/msa/provider.test.ts
```diff
@@ -0,0 +1,98 @@
...
+          ['es', 'ProveedorPrivado'],
+        ]),
+        defaultLogo250100PngCid: 'bafkreidgvpkjawlxz6sffxzwgooowe5yt7i6wsyg236mfoks77nywkptdq',
+        localizedLogo250100PngCids: new Map([
+          ['en', 'bafkreidgvpkjawlxz6sffxzwgooowe5yt7i6wsyg236mfoks77nywkptdq'],
+          ['es', 'bafkreidgvpkjawlxz6sffxzwgooowe5yt7i6wsyg236mfoks77nywkptdq'],
+        ]),
+      };
+      const createProviderOp = ExtrinsicHelper.createProvider(failureKeys, providerEntry);
+      await assert.rejects(createProviderOp.signAndSend(), undefined);
```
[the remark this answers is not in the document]
JoeCap08055: For extrinsics that fail on-chain, we typically use the following pattern:
```typescript
      await assert.rejects(addPublicKeyOp.fundAndSend(fundingSource), {
        name: 'KeyAlreadyRegistered',
      });
```

For extrinsics that fail due to an RpcError, we use this pattern:
```typscript
      await assert.rejects(retireOp.signAndSend('current'), {
        name: 'RpcError',
        message: /Custom error: 2/,
      });
```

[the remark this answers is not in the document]
JoeCap08055: Wait, are you saying that those tests I quoted work fine regardless of what the error argument is???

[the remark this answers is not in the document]
saraswatpuneet: I think so, they work regardless but let me know once you confirm it

runtime/common/src/constants.rs
JoeCap08055: Bummer.

pallets/msa/src/lib.rs
saraswatpuneet: On a second thought we should maintain both versions

1. Will still take name and put in registry 
2. Can accept a payload 
This could give good runway for changes to reach other dependencies

saraswatpuneet: so we can retrain old extrinsic and create a `_v2` for each and is just a re-routing? we can mark them `#[deprecated(note = "please use `verify` instead")]`

pallets/msa/src/lib.rs
```diff
@@ -953,31 +943,19 @@ pub mod pallet {
 		///
 		/// # Errors
 		/// - [`NoKeyExists`](Error::NoKeyExists) - If there is not MSA for `origin`.
+		/// - [`Error::ExceedsMaxProviderNameSize`] - Too long of a provider name
 		#[pallet::call_index(11)]
-		#[pallet::weight(T::WeightInfo::propose_to_be_provider())]
+		#[pallet::weight(T::WeightInfo::propose_to_be_provider_v2())]
```
JoeCap08055, line 948:
```suggestion
		#[pallet::weight(T::WeightInfo::propose_to_be_provider())]
```
I think this is the original, not the v2, correct?

JoeCap08055: Or are we just using the same weights for both?

pallets/msa/src/lib.rs
```diff
@@ -987,27 +965,21 @@ pub mod pallet {
 		///
 		/// # Errors
 		/// * [`Error::NoKeyExists`] - account does not have an MSA
+		/// * [`Error::ExceedsMaxProviderNameSize`] - Too long of a provider name
 		/// * [`Error::DuplicateProviderRegistryEntry`] - a ProviderRegistryEntry associated with the given MSA id already exists.
 		#[pallet::call_index(12)]
-		#[pallet::weight(T::WeightInfo::create_provider_via_governance())]
+		#[pallet::weight(T::WeightInfo::create_provider_via_governance_v2())]
```
JoeCap08055, line 971: same question as above

pallets/msa/src/lib.rs
JoeCap08055: That's the usual pattern when doing `_v2`, etc... unless there's a clear reason not to? Only Provider Dashboard is likely affected by `propose` and `..._via_governance`, but I'm not a fan of cutting corners on that kind of pattern. I'll defer to @wilwade .

e2e/msa/provider.test.ts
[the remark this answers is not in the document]
saraswatpuneet: Nice suggestions!

pallets/msa/src/lib.rs
saraswatpuneet: I assume we can since its calling underlying extrinsic (re-routing after size check), you suggest we have separate benchmarks for both?

Issue #2528, opened 2025-07-28
[Provider Context] Migration of `ProvideRegistryEntry`


saraswatpuneet:
## Details

Since `ProviderRegistryEntry` will have updates fields we will have to migrate state to update existing providers details updates 

- [ ] Write migration for `ProviderRegistryEntry` state to update existing data with new fields and default values.
- [ ] Test with try-runtime

## Acceptance Criteria 

1. Successful migration tests with try runtime 
2. Migrated entries should have default values initialized

Issue #2526, opened 2025-07-28
[Provider Context] Introduce RPC to fetch provider/application


saraswatpuneet:
## Details

Based on provider context design doc, add an RPC to fetch a full provider or application context with logos (not hashes) to make it easy to fetch full context in one call

- [ ] Add an rpc `get_provider_application_context` that takes a `ProviderId` and an option `ApplicationIndex`
- [ ] If application index is specified, read the state `ProviderToApplicationRegistryEntry` and fetch the application context with logo hashes and replace hashes with logo png bytes from `ApprovedLogos` , before returning full context else None
- [ ] If application index is not specified, read the state `ProviderRegistryEntry` and fetch the default provider context with logo hashes and replace hashes with logo png from `ApprovedLogos` before returning full provider default context else None

## Acceptance Criteria

1. A fully function functional rpc
2. e2e and unit tests

Issue #2525, opened 2025-07-28
[Provider Context] Add a new extrinsic `update_logo` for provider/application context


saraswatpuneet:
## Details

Refer [design doc](https://github.com/frequency-chain/frequency/blob/main/designdocs/provider_contexts.md#changes-and-additions-in-extrinsics-)

- [x] Add a new extrinsic `update_logo` that takes a logo bytes as input from provider MSA origin.
- [x] Validate provider, validate incoming logo bytes and compute hash from logo bytes.
- [x] Validate occurrence of logo hash in `ApprovedHashes` state.
- [x] Add a new state called `ApprovedLogos` that will contain a map of hash vs logo png bytes.
- [x] Upon successful validation remove hash from `ApprovedHashes` and add to the `ApprovedLogos` map.
- [x] Benchmark.
- [x] Unit test and e2e tests.

Note: this extrinsic is independent of application index and can be used to either update logo for hashes in ProviderEntry or ProviderToApplicationEntry states i.e. one extrinsic to handle both sides

## Acceptance Criteria

1. `update_logo` checks for valid provider, computes correct hash and ensure validity of approved hash.
2. `ApprovedLogos` is updated with correct hash vs logo png bytes information and respective hash removed from ApprovedHashes.
3. Hash should return correct logo.
4. Unit test, benchmarking and e2e tests for the same.

Issue #2524, opened 2025-07-28
[Provider Context] Introduce a new extrinsic `propose_to_add_application`


saraswatpuneet:
## Details

With respect to provider context design doc, complete the following:

- [ ] Add a new extrinsic `propose_to_add_application` that receives a `ProviderRegistryEntry` type payload from provider MSA
- [ ] Ensure the origin is registered provider
- [ ] Add a new state `ProviderToApplicationRegistry` douple map (ProviderId, ApplicationIndex) to store a new application once a goverance proposal is approved for application, where ApplicationIndex is monotically increasing number for each provider level application from 0..N
- [ ] Move all the approves hashes from payload to `ApprovedLogoHashes` state 
- [ ] Benchmark new extrinsic
- [ ] Add unit tests and e2e tests 

## Acceptance Criteria
1. Ability to propose and add new applications 
2. `ApprovedHashes` successfully contains approved logo hashes from 1
3. Tests emulating range of tests for adding and approving applications for providers
4. Benchmarking and e2e tests

PR #2549, opened 2025-08-12
Provider Context: Minimal Profile for Provider and Applications

### LINKED ISSUE #2577: [Provider Context] Add extrinsic to propose to edit existing provider registration
## Details

Based on provider context [design doc](https://github.com/frequency-chain/frequency/blob/main/designdocs/provider_contexts.md#runtime-rpc-changes-), add an extrinsic to do the following

- [ ] add extrinsic `propose_to_update_provider` which sets a governance proposal to update provider registry entry and mutate existing one
- [ ] Add unit tests
- [ ] Benchmarks
- [ ] E2E tests

### LINKED ISSUE #2578: [Provider Context] Add extrinsic to propose to edit existing provider application
Placeholder

saraswatpuneet:
# Details

[Design Doc](https://github.com/frequency-chain/frequency/blob/main/designdocs/provider_contexts.md)

Provider Context allows provider to now register a minimal profile `ProviderRegistryEntry` and `ApplicationContext` for provider's default profile and various applications providers want to register. 

Providers which want to register multiple applications under single provider delegation but distributed across multiple app. Following issues were resolved as part of this work with following salient points:

1. Migration existing `ProviderToRegistryEntry` to host new `ProviderRegistryEntry` payload
2. Introduce new state `ProviderToApplicationRegistry` and `ApprovedLogos` storage maps.
3. CID's are computed in deterministic way = `base(cid_v1(sha_256(logo_bytes)))` where logo_bytes are .png image bytes.
4. ApprovedLogos ensure it matches CID and logoBytes but recomputing and checking on chain.

Issues complete as part of  `Provider Context` work

- [x] #2523 
- [x] #2524 
- [x] #2525 
- [x] #2528 
- [x] #2526
- [x] #2577 
- [x] #2578 

Closes #2578 
Closes #2577 

# Discussion items
- [x] Do we feature flag than base filter out extrinsics @JoeCap08055  comment [here](https://github.com/frequency-chain/frequency/pull/2587#discussion_r2325375021)

# Checklist
- [x] Updated Pallet Readme?
- [x] Design doc(s) updated?
- [x] Unit Tests added?
- [x] e2e Tests added?
- [x] Benchmarks added?


pallets/msa/src/migration/v2.rs
```diff
@@ -0,0 +1,160 @@
...
+use frame_support::{pallet_prelude::*, traits::OnRuntimeUpgrade, weights::Weight};
+use frame_system::pallet_prelude::BlockNumberFor;
+use sp_runtime::Saturating;
+#[cfg(feature = "try-runtime")]
+use sp_runtime::TryRuntimeError;
+
+const LOG_TARGET: &str = "runtime::provider";
+
+/// Get known provider ids from paseo and mainnet
+pub fn get_known_provider_ids<T: Config>() -> vec::Vec<ProviderId> {
```
aramikm, line 16: nit: We only use it in `#[cfg(feature = "try-runtime")]`

pallets/msa/src/migration/v2.rs
```diff
@@ -0,0 +1,160 @@
...
+				return T::DbWeight::get().reads(1)
+			}
+			let mut reads = 1u64;
+			let mut writes = 0u64;
+			let mut bytes = 0u64;
+			let mut migrated_count = 0u64;
+
+			log::info!(target: LOG_TARGET, "Starting migration from v1 to v2...");
+			// Drain from OLD storage and migrate to NEW storage
+			for (provider_id, old_entry) in v1::ProviderToRegistryEntry::<T>::drain() {
```
aramikm, line 71: This is not going to work for paseo. We have around 3635 items which will not fit in a block. We need multi-block migration to make it work on paseo

pallets/msa/src/benchmarking.rs
```diff
@@ -700,6 +817,351 @@ mod benchmarks {
...
+		);
+
+		assert!(ApprovedLogos::<T>::get(input_bounded_cid.clone()).is_some());
+		let stored_logo_bytes = ApprovedLogos::<T>::get(&input_bounded_cid).unwrap();
+		assert_eq!(stored_logo_bytes, input_bounded_logo);
+		Ok(())
+	}
+
+	#[benchmark]
+	fn propose_to_update_provider() -> Result<(), BenchmarkError> {
```
aramikm, line 955: Benchmark for this function depends on two things

1. payload items have direct effect on the storage writes (logo adds)
2. stored state have direct effect on the storage writes (logo removals) (this should be maxed in benchmark and based on the actual value the rest should be refunded in extrinsic)

pallets/msa/src/benchmarking.rs
```diff
@@ -700,6 +817,351 @@ mod benchmarks {
...
+
+		#[extrinsic_call]
+		_(RawOrigin::Signed(provider_public_key), entry);
+
+		assert_eq!(frame_system::Pallet::<T>::events().len(), 1);
+		Ok(())
+	}
+
+	#[benchmark]
+	fn update_application_via_governance(
```
aramikm, line 1002: it's the same for this

pallets/msa/src/benchmarking.rs
```diff
@@ -700,6 +817,351 @@ mod benchmarks {
...
+
+		#[extrinsic_call]
+		_(RawOrigin::Signed(provider_public_key), 0u16, application_payload);
+
+		assert_eq!(frame_system::Pallet::<T>::events().len(), 1);
+		Ok(())
+	}
+
+	#[benchmark]
+	fn update_provider_via_governance(
```
aramikm, line 1118: same

common/primitives/src/msa.rs
```diff
@@ -202,14 +210,81 @@ impl From<ProviderId> for MessageSourceId {
...
-	T: Get<u32>,
+	NameSize: Get<u32> + Debug + PartialEq + Eq,
+	LangSize: Get<u32> + Debug + PartialEq + Eq,
+	CidSize: Get<u32> + Debug + PartialEq + Eq,
+	MaxLocaleCount: Get<u32> + Debug + PartialEq + Eq,
 {
-	/// The provider's name
-	pub provider_name: BoundedVec<u8, T>,
+	fn default() -> Self {
+		let name_bytes = b"default";
```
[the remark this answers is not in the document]
saraswatpuneet: Its not used in any code paths but tests thought i can do `new_with_name` to be explicit

pallets/msa/src/lib.rs
```diff
@@ -200,7 +221,12 @@ pub mod pallet {
 		_,
 		Twox64Concat,
 		ProviderId,
-		ProviderRegistryEntry<T::MaxProviderNameSize>,
+		ProviderRegistryEntry<
```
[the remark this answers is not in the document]
saraswatpuneet: We do but we have kep the older extrinsic till all dependent apps can change, they might have to update query part right away, for now its only Provider Dashboard that uses this so will work wtih claire

runtime/common/src/constants.rs
```diff
@@ -85,13 +85,51 @@ pub type FrameSystemMaxConsumers = ConstU32<16>;
...
 pub type MSAMortalityWindowSize = ConstU32<{ 20 * MINUTES }>;
 /// The upper limit on total stored signatures.
 /// Set to an average of 50 signatures per block
 pub type MSAMaxSignaturesStored = ConstU32<50_000>;
+/// The maximum size of a provider name (in bytes)
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub struct MsaMaxProviderNameSize;
+impl Get<u32> for MsaMaxProviderNameSize {
+	fn get() -> u32 {
+		256
```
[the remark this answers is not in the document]
saraswatpuneet: It was done to accomodate providers who would have longer names "Abc corpo ltd." etc.

runtime/frequency/src/lib.rs
```diff
@@ -818,12 +823,23 @@ impl pallet_msa::Config for Runtime {
 	#[cfg(not(any(feature = "frequency", feature = "runtime-benchmarks")))]
 	type RecoveryProviderApprovalOrigin = EnsureSigned<AccountId>;
 	// The origin that is allowed to create providers via governance
+	#[cfg(not(feature = "frequency-no-relay"))]
 	type CreateProviderViaGovernanceOrigin = EitherOfDiverse<
 		EnsureRoot<AccountId>,
 		pallet_collective::EnsureMembers<AccountId, CouncilCollective, 1>,
 	>;
+	#[cfg(feature = "frequency-no-relay")]
+	type CreateProviderViaGovernanceOrigin = EnsureSigned<AccountId>;
```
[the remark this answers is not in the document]
saraswatpuneet: Checking, i think you are right we can use sudo in e2e, will try removing this

pallets/msa/src/migration/v2.rs
saraswatpuneet: Do we have a defined way to run multi-block migration?

JoeCap08055: We might want to discuss this at midweek. My understanding was that Paseo was intended to be reset periodically. If that's an option, it seems a waste of time to do the work to code a multi-block migration only to support the state on testnet. Questions to be answered:
* Is it possible to "reset" Paseo? Level of effort?
    * What services could be affected by a reset? (Freesky, CI pipelines, FA, Provider Dashboard)
    * Would the effort involved in resetting Paseo be worth it vs. coding a more complex migration?
* Is it possible to simply prune the Provider storage on Paseo using a sudo script? Level of effort?
    * What would happen to staked tokens on testnet if an MSA is suddenly "de-authorized" as a Provider?
    * Same as above: worth the effort vs multi-block migration?

pallets/msa/src/lib.rs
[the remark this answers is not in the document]
JoeCap08055: There might be some other places where we query this, we should audit:
* Gateway
* frequency-scenario-template
* Reconnection service

Any usage would be only setup scripts and/or for the benefit of logging the Provider name, but worth a look.

pallets/msa/src/lib.rs
```diff
@@ -903,28 +1001,19 @@ pub mod pallet {
 		///
 		/// # Errors
 		/// - [`NoKeyExists`](Error::NoKeyExists) - If there is not MSA for `origin`.
+		/// - [`Error::ExceedsMaxProviderNameSize`] - Too long of a provider name
 		#[pallet::call_index(11)]
-		#[pallet::weight(T::WeightInfo::propose_to_be_provider())]
+		#[pallet::weight(T::WeightInfo::propose_to_be_provider_v2())]
```
JoeCap08055, line 1006: Same as above--do not rename extrinsics; use a _deprecate-and-add_ pattern

pallets/msa/src/lib.rs
```diff
@@ -597,17 +697,15 @@ pub mod pallet {
 		/// * [`Error::NoKeyExists`] - origin does not have an MSA
 		/// * [`Error::ExceedsMaxProviderNameSize`] - Too long of a provider name
 		/// * [`Error::DuplicateProviderRegistryEntry`] - a ProviderRegistryEntry associated with the given MSA id already exists.
-		///
 		#[pallet::call_index(2)]
-		#[pallet::weight(T::WeightInfo::create_provider())]
+		#[pallet::weight(T::WeightInfo::create_provider_v2(0u32, 0u32))]
```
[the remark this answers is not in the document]
JoeCap08055: nvm; I misread--I see now that this is just the weight reference. 👍🏻

e2e/msa/provider_applications.test.ts
```diff
@@ -0,0 +1,245 @@
...
+  let providerId: bigint;
+  let sudoKeys: KeyringPair;
+  let fundingSource: KeyringPair;
+
+  before(async function () {
+    sudoKeys = getSudo().keys;
+    fundingSource = await getFundingSource(import.meta.url);
+    keys = await createAndFundKeypair(fundingSource, 5n * DOLLARS);
+    nonProviderKeys = await createAndFundKeypair(fundingSource, 5n * DOLLARS);
+    const f = ExtrinsicHelper.createMsa(keys);
```
enddynayn, line 34: what does f stand for?

pallets/msa/src/benchmarking.rs
saraswatpuneet: I am filling up the `ApprovedLogos` so `remove_logo_storage` gets called when `updating_provider/application`

aramikm (commented):
Migration looks good to me but weights are still off

pallets/msa/src/lib.rs
```diff
@@ -1966,6 +2546,201 @@ impl<T: Config> Pallet<T> {
...
+		// Weight adjustment: refund weight for logos that were NOT removed
+		let max_logos_benchmark_assumed = T::MaxLocaleCount::get();
+		let removal_over_charged_by =
+			max_logos_benchmark_assumed.saturating_sub(total_logos_removed);
+		let weight_per_logo_removal = T::DbWeight::get().writes(1);
+		let weight_to_refund =
+			weight_per_logo_removal.saturating_mul(removal_over_charged_by as u64);
+		let actual_weight_used = base_weight.saturating_sub(weight_to_refund);
+		Ok(PostDispatchInfo { actual_weight: Some(actual_weight_used), pays_fee: Pays::Yes })
+	}
```
saraswatpuneet, line 2743: @aramikm  so in  both `update_provider/application` benchmarks I am not computing weight for worse case and refunding the weight that was not used (if provider upserts less than `MaxLocaleCount`), I have kept one paramters in benchmarks `n` which is total localized names as it is stored in registry and not separately

pallets/msa/src/lib.rs
```diff
@@ -196,11 +218,16 @@ pub mod pallet {
 	/// - Key: Provider MSA Id
 	/// - Value: [`ProviderRegistryEntry`](common_primitives::msa::ProviderRegistryEntry)
 	#[pallet::storage]
-	pub type ProviderToRegistryEntry<T: Config> = StorageMap<
+	pub type ProviderToRegistryEntryV2<T: Config> = StorageMap<
```
JoeCap08055, line 221: Now that I have more context surrounding migrations, is there a reason we're allocating a completely new storage here, versus migrating the existing storage in place?

Using `.drain()` + `.put()` costs 1 read + 2 writes per record migrated.
Using `.read()` old and `.insert()` new to the same storage map costs 1 read + 1 write.

Not a blocker, since mainnet only has a few providers and I guess we don't care so much about cost on Testnet, but just curious.

pallets/msa/src/lib.rs
```diff
@@ -1209,6 +1317,391 @@ pub mod pallet {
...
+			payload: ProviderRegistryEntry<
+				T::MaxProviderNameSize,
+				T::MaxLanguageCodeSize,
+				T::MaxLogoCidSize,
+				T::MaxLocaleCount,
+			>,
+		) -> DispatchResult {
+			let proposer = ensure_signed(origin)?;
+			Self::ensure_valid_msa_key(&proposer)?;
+			Self::ensure_correct_cids(&payload)?;
```
JoeCap08055, line 1338: I think maybe we can skip validating the CIDs here. It gets done in the wrapped extrinsic; why pay for the compute cost of the same calculations twice?

pallets/msa/src/lib.rs
saraswatpuneet: I agree, thinking from that perspective definitely not a big migration however I would be open to following any standard practice 💯

pallets/msa/src/lib.rs
```diff
@@ -1225,6 +1313,422 @@ pub mod pallet {
...
+
+		/// Adds an association between MSA id and ProviderRegistryEntry. As of now, the
+		/// only piece of metadata we are recording is provider name.
+		///
+		/// # Events
+		/// * [`Event::ProviderCreated`]
+		///
+		/// # Errors
+		/// * [`Error::NoKeyExists`] - origin does not have an MSA
+		/// * [`Error::DuplicateProviderRegistryEntry`] - a ProviderRegistryEntry associated with the given MSA id already exists.
```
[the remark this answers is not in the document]
enddynayn: I would prefer to improve our code based on what we have learned rather than stick to outdated patterns just for consistency. That said, I’m fine either way.

enddynayn (approved):
👍 amazing work!

pallets/msa/src/lib.rs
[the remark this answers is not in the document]
JoeCap08055: If we have consensus that the only real use for `create_provider` in a local dev chain is for provisioning a Provider for running other scenarios, and we're not concerned about name or logo localization, then I'm fine with skipping a "v2" version in this case.

JoeCap08055 (approved):

Issue #2550, opened 2025-08-12
EPIC: Schemas, Protocols, and Intent-based Delegation


p5150j:
# Epic: Schema, Protocols, and Intent-Based Delegation

## Summary
Implement a new architecture for Frequency's schema and delegation systems that introduces named intents with version tracking and intent-based delegation to address current limitations around coupling between schema versions and delegations.

## Background
The current implementation tightly binds delegations to specific SchemaIds, which has proven limiting:
- Strong coupling between schema versions and delegations
- Schemas represent data format but not purpose
- Lack of human-readable context for schemas

This epic introduces:
- **Named intents** with version tracking
- **Intent-based delegation** replacing schema-based delegation
- **Intent groups** for bundling related permissions
- **Protocol-based name registry** for off-chain resolution

## Goals
- **Schema Immutability**: Individual schema versions remain immutable once published
- **Minimal Delegation Churn**: Minor data format changes don't require new delegations
- **Minimal Storage Churn**: Format changes don't require mass data migration
- **Intent Separation**: Separate data purpose/action from format when permissioning
- **On-Chain Efficiency**: Maintain efficient on-chain operations

### Key Components
1. **Intents**: Represent consistent data meaning/purpose, can be associated with multiple schema versions
2. **Schemas**: Now only represent data format definitions, associated with a single Intent
3. **Intent Groups**: Bundles of IntentIds for easier permission management
4. **Name Registry**: Protocol-based naming system for off-chain discovery

### Major Changes
- Delegations now reference `IntentId` instead of `SchemaId`
- Storage location tied to `IntentId` rather than schemas
- Introduction of versioned storage for messages and stateful-storage pallets

## Dependencies
- Governance approval for all schema/intent/intent group operations
- Analysis of existing mainnet data for migration planning

## Risks & Mitigations
- **Risk**: Large-scale data migration complexity
  - **Mitigation**: Dual storage approach avoiding immediate migration needs
- **Risk**: Breaking existing delegations
  - **Mitigation**: 1:1 mapping of SchemaId to IntentId preserves all delegations
- **Risk**: Performance impact of additional lookups
  - **Mitigation**: Schema->Intent lookup mitigates performance impact

## Success Metrics
- Zero breaking changes to existing delegations
- No mass data migration required at launch
- Delegation updates reduced by >80% for schema version changes
- All existing functionality maintained with improved flexibility

## Testing Requirements
- Comprehensive unit tests for all new storage and extrinsics
[...]

PR #2552, opened 2025-08-13
feat: updated Schemas and Intents

JoeCap08055:
# Goal
The goal of this PR is to implement the design described [here](designdocs/schemas_protocols_intents.md)

Tracking epic: #2550 

# Checklist
- [ ] Updated Pallet Readme?
- [ ] Design doc(s) updated?
- [ ] Unit Tests added?
- [ ] e2e Tests added?
- [ ] Benchmarks added?


common/primitives/src/messages.rs
```diff
@@ -45,6 +45,58 @@ pub struct MessageResponse {
 	#[cfg_attr(feature = "std", serde(skip_serializing_if = "Option::is_none", default))]
 	pub payload_length: Option<u32>,
 }
+
+/// A type for responding with a single Message in an RPC-call dependent on schema model
+/// IPFS, Parquet: { index, block_number, provider_msa_id, cid, payload_length }
+/// Avro, OnChain: { index, block_number, provider_msa_id, msa_id, payload }
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+#[derive(Default, Clone, Encode, Decode, PartialEq, Debug, TypeInfo, Eq)]
+pub struct MessageResponseV2 {
```
JoeCap08055: Message storage is now indexed by `IntentId` instead of `SchemaId`. Since multiple Schemas can be stored under one Intent index, we added `schema_id` to the stored payload.

pallets/schemas/src/lib.rs
```diff
@@ -221,21 +283,68 @@ pub mod pallet {
...
 		Blake2_128Concat,
-		SchemaNamespace,
+		SchemaProtocolName,
 		Blake2_128Concat,
 		SchemaDescriptor,
 		SchemaVersionId,
 		ValueQuery,
 	>;
 
+	/// Storage for mapping names to IDs
```
JoeCap08055: WHy didn't we remove SchemaNameToIds?

pallets/schemas/src/lib.rs
```diff
@@ -625,19 +1016,57 @@ pub mod pallet {
 		/// Retrieve a schema info by id
 		pub fn get_schema_info_by_id(schema_id: SchemaId) -> Option<SchemaInfoResponse> {
 			if let Some(schema_info) = SchemaInfos::<T>::get(schema_id) {
-				let saved_settings = schema_info.settings;
-				let settings = saved_settings.0.iter().collect::<Vec<SchemaSetting>>();
 				let response = SchemaInfoResponse {
 					schema_id,
+					intent_id: schema_info.intent_id,
```
JoeCap08055: Make sure this goes at the end so SCALE decodes properly even for the old struct

pallets/messages/src/migration/v2/mod.rs
JoeCap08055: This file contains:
- "Old" storage types cut from `src/types.rs`, that can be isolated & fully removed once all deprecated endpoints supporting them are removed
- Storage aliases for:
    - "Old" storage that will be drained & removed at the completion of the migration
    - Temporary storage key used to track migration completion & killed at the end of all migrations in this set

pallets/messages/src/migration/v3.rs
```diff
@@ -0,0 +1,229 @@
...
+		if !messages_remain {
+			v2::DoneV3Migration::<T>::put(true);
+		}
+		Ok(messages_remain.then_some(last_cursor))
+	}
+}
+
+/// Finalize the migration of [`v2::MessagesV2`] map to [`crate::MessagesV3`]
+/// by updating the pallet storage version.
+pub struct FinalizeV3Migration<T: Config, W: weights::WeightInfo>(PhantomData<(T, W)>);
```
JoeCap08055: This second `SteppedMigration` simply bumps the `StorageVersion` in the pallet. It only runs after the first migration is complete. Might seem overkill; technically we could have just bumped this at the end of the main migration, but this provides more tracking via `pallet-migrations` and was a recommended pattern.

pallets/messages/src/rpc/src/lib.rs
JoeCap08055: Open question for this RPC API:

Do we:
1. Keep an RPC API for this pallet, deprecating the existing `messages_getBySchemaId` and adding `messages_getByIntentId`?
2. Since clients will eventually have to switch to a new, intent-based method anyway, do we mark the entire API  deprecated and remove in a future version, and force use of the Runtime API in the future?
    - if so, do we:
        1. change the meaning of the input to be `IntentId`
        2. continue to interpret as `SchemaId`, use it to look up the `IntentId` and return all messages for the `Intent`
        3. continue to interpret as `Schemaid`, use it to look up the `IntentId`, then filter the returned messages for only the requested `SchemaId`
4. Simply remove the RPC API and just update any clients  now (since only Gateway and maybe Wes's Freesky tool use it)

JoeCap08055: ☝🏻 note, regardless of what we intend to do with the RPC API in the future, we will preserve  the existing Runtime API surface so that old nodes won't break

pallets/stateful-storage/src/migration/v1/types.rs
JoeCap08055: This file contains the original types moved from `stateful-storage/src/types.rs`. They are used in the migration, as well as to support any legacy RPCs and extrinsics until such time as they can be removed.

pallets/stateful-storage/src/rpc/src/lib.rs
JoeCap08055: Question: do we want to deprecate this RPC API for future removal? It makes use of now-deprecated Runtime APIs. Or do we want to update it for the new page payloads?

pallets/stateful-storage/src/stateful_child_tree.rs
JoeCap08055: Added read/write "raw" methods that use pre-hashed keys; used in the migration.

pallets/stateful-storage/src/types.rs
```diff
@@ -75,13 +72,87 @@ pub enum ItemAction<DataSize: Get<u32> + Clone + core::fmt::Debug + PartialEq> {
...
+pub enum ItemVersion {
+	/// V1 variant can be removed after migration is complete
+	#[codec(index = 1)]
+	V1,
+
+	/// V2 (added schema_id)
+	#[codec(index = 2)]
+	#[default] // NOTE: move default when adding new variants
+	V2,
+}
```
JoeCap08055:
```suggestion
```

Remove; was an alternate approach I accidentally left in.

common/primitives/src/msa.rs
```diff
@@ -79,91 +79,89 @@ impl From<DelegatorId> for MessageSourceId {
...
 }
 
 /// RPC response for getting schema permission grants
 #[cfg_attr(feature = "std", derive(Deserialize, Serialize))]
 #[derive(TypeInfo, RuntimeDebug, Clone, Decode, Encode, MaxEncodedLen, Eq)]
-pub struct SchemaGrant<SchemaId, BlockNumber> {
+pub struct DelegationGrant<DelegationIdType, BlockNumber> {
 	/// SchemaId of schema for which permission is/was granted
-	pub schema_id: SchemaId,
+	#[cfg_attr(feature = "std", serde(rename = "schema_id"))]
```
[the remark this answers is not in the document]
JoeCap08055: Found a solution, will update.

pallets/messages/src/migration/v3.rs
```diff
@@ -0,0 +1,229 @@
...
+			version_to: 3,
+		}
+	}
+
+	/// The actual logic of the migration.
+	///
+	/// This function is called repeatedly until it returns `Ok(None)`, indicating that the
+	/// migration is complete. Ideally, the migration should be designed in such a way that each
+	/// step consumes as much weight as possible.
+	fn step(
```
JoeCap08055: Note, this migration does NOT block the pallet's extrinsics from executing while in progress. This is because stored messages are indexed first by block number, and the pallet will only write going forward to the new `MessagesV3` storage; so there is ZERO risk of a key conflict between old records being migrated and new records written by the pallet.

e2e/stateful-pallet-storage/handleSignatureRequiredPaginated.test.ts
```diff
@@ -64,13 +63,13 @@ describe('📗 Stateful Pallet Storage Signature Required Paginated', function (
...
       fundingSource,
-      [paginatedSchemaId],
+      [paginatedIntentId],
       delegatedProviderId,
       delegatedProviderKeys,
       'sr25519',
       delegatorKeys
     );
-    ExtrinsicHelper.transferFunds(fundingSource, delegatorKeys, 2n * DOLLARS);
+    // ExtrinsicHelper.transferFunds(fundingSource, delegatorKeys, 2n * DOLLARS);
```
aramikm: nit: leftover?

js/api-augment/definitions/schemas.ts
```diff
@@ -1,61 +1,71 @@
 export default {
   rpc: {
-    getBySchemaId: {
```
aramikm: is this one of those RPC's that Peter is using? If so might need to keep it

pallets/messages/src/weights.rs
```diff
@@ -20,7 +20,7 @@
 // --output=./scripts/../pallets/messages/src/weights.rs
 // --template=./scripts/../.maintain/frame-weight-template.hbs
 // --additional-trie-layers=3
-// --runtime=./scripts/../target/release/wbuild/frequency-runtime/frequency_runtime.wasm
+// --runtime=./scripts/../target/bench-dev/wbuild/frequency-runtime/frequency_runtime.wasm
```
aramikm: Reminder: we need to run this in release mode.

pallets/msa/src/types.rs
```diff
@@ -175,12 +175,52 @@ impl AddProvider {
...
+		let intent_ids = intent_ids.unwrap_or_default();
 
-		Self { authorized_msa_id, schema_ids, expiration }
+		Self { authorized_msa_id, intent_ids, expiration }
+	}
+}
+
+/// Structure that is signed for granting permissions to a Provider
+#[derive(TypeInfo, Clone, Debug, Decode, DecodeWithMemTracking, Encode, PartialEq, Eq)]
+pub struct AddProviderV2 {
```
aramikm: This is not used so far so the idea is to use it in future? We will need additional PRs to completely rollout this feature (including cleanup)

pallets/msa/src/types.rs
```diff
@@ -175,12 +175,52 @@ impl AddProvider {
...
+	/// The block number at which the proof for grant_delegation expires.
+	pub expiration: BlockNumber,
+}
+
+impl EIP712Encode for AddProviderV2 {
+	fn encode_eip_712(&self, chain_id: u32) -> Box<[u8]> {
+		lazy_static! {
+			// signed payload
+			static ref MAIN_TYPE_HASH: [u8; 32] = sp_io::hashing::keccak_256(
+				b"AddProvider(uint64 authorizedMsaId,uint16[] intentIds,uint32 expiration)"
```
aramikm:
```suggestion
				b"AddProviderV2(uint64 authorizedMsaId,uint16[] intentIds,uint32 expiration)"
```

I think in other places we tried to use the same struct name

pallets/schemas/src/lib.rs
aramikm: I think we have usecases that allows fetching schemas and intents by their name but I'm not sure why we want. remove that?

pallets/schemas/src/lib.rs
```diff
@@ -560,57 +846,148 @@ pub mod pallet {
...
 			model: BoundedVec<u8, T::SchemaModelMaxBytesBoundedVecLimit>,
 			model_type: ModelType,
 			payload_location: PayloadLocation,
-			settings: BoundedVec<SchemaSetting, T::MaxSchemaSettingsPerSchema>,
-			schema_name_option: Option<SchemaName>,
+			settings: IntentSettings,
+			status: SchemaStatus,
 		) -> Result<SchemaId, DispatchError> {
 			let schema_id = Self::get_next_schema_id()?;
-			let has_name = schema_name_option.is_some();
```
aramikm: question: So what happened with this field that we were storing in SchemaInfo? Are we enforcing all the schemas to be named ones and if so what are we going to do with the existing ones on chain? I'm assuming that might be part of migration

pallets/schemas/src/lib.rs
```diff
@@ -625,19 +1002,57 @@ pub mod pallet {
...
+		/// Retrieve an Intent by its ID
+		pub fn get_intent_by_id(
+			intent_id: IntentId,
+			include_schemas: bool,
+		) -> Option<IntentResponse> {
+			if let Some(intent_info) = IntentInfos::<T>::get(intent_id) {
+				let saved_settings = intent_info.settings;
+				let settings = saved_settings.0.iter().collect::<Vec<IntentSetting>>();
+				let schema_ids = include_schemas.then_some(
+					SchemaInfos::<T>::iter()
```
aramikm: This is dangerous if used incorrectly. I understand that include_schemas is false for all on-chain calls but we might need to add some kind of feature flag or at least a comment that should never be called with `include_schemas = true` from chain side.

pallets/stateful-storage/src/rpc/src/lib.rs
aramikm: I think we are using these everywhere. So I would be a bit careful about how we want to do that or should be do that under current circumstances.

pallets/stateful-storage/src/types.rs
```diff
@@ -194,7 +252,101 @@ impl<T: Config> EIP712Encode for ItemizedSignaturePayloadV2<T> {
...
+		<T as Config>::MaxItemizedActionsCount,
+	>,
+}
+
+impl<T: Config> EIP712Encode for ItemizedSignaturePayloadV3<T> {
+	fn encode_eip_712(&self, chain_id: u32) -> Box<[u8]> {
+		lazy_static! {
+			// signed payload
+			static ref MAIN_TYPE_HASH: [u8; 32] =
+				sp_io::hashing::keccak_256(b"ItemizedSignaturePayloadV3(uint16 intentId,uint32 targetHash,uint32 expiration,ItemAction[] actions)ItemActionV2(string actionType,uint16 schemaId,bytes data,uint16 index)");
```
aramikm:
```suggestion
				sp_io::hashing::keccak_256(b"ItemizedSignaturePayloadV3(uint16 intentId,uint32 targetHash,uint32 expiration,ItemActionV2[] actions)ItemActionV2(string actionType,uint16 schemaId,bytes data,uint16 index)");
```

pallets/messages/src/rpc/src/lib.rs
shannonwells: I vote for 3

pallets/stateful-storage/src/types.rs
```diff
@@ -450,26 +702,31 @@ impl<T: Config> ItemizedOperations<T> for ItemizedPage<T> {
...
-				offset + ItemHeader::max_encoded_len() <= self.data.len(),
-				PageError::ErrorParsing("wrong header size")
+				offset + ItemHeader::max_encoded_len() <= page_size,
+				PageError::ErrorParsing("incomplete item header")
 			);
 			let header = <ItemHeader>::decode(&mut &self.data[offset..])
-				.map_err(|_| PageError::ErrorParsing("decoding header"))?;
-			let item_total_length = ItemHeader::max_encoded_len() + header.payload_len as usize;
+				.map_err(|_| PageError::ErrorParsing("decoding item header"))?;
+			let item_total_length = ItemHeader::max_encoded_len() + header.payload_len() as usize;
```
aramikm: I don't think this is going to work. Previously ItemHeader had a deterministic max size but now there is difference in size for V1 and V2 and to be able to do correct max check you need to first discern the type and then get the max for that type.

node/cli/src/benchmarking.rs
```diff
@@ -131,6 +131,7 @@ pub fn create_benchmark_extrinsic(
 		pallet_frequency_tx_payment::ChargeFrqTransactionPayment::<runtime::Runtime>::from(0),
 		pallet_msa::CheckFreeExtrinsicUse::<runtime::Runtime>::new(),
 		pallet_handles::handles_signed_extension::HandlesSignedExtension::<runtime::Runtime>::new(),
+		pallet_stateful_storage::BlockDuringMigration::<runtime::Runtime>::new(),
```
shannonwells: why is this added as part of this PR?

pallets/stateful-storage/src/lib.rs
```diff
@@ -470,15 +506,176 @@ pub mod pallet {
...
+		#[pallet::call_index(11)]
+		#[pallet::weight(
+		T::WeightInfo::apply_item_actions_with_signature_v2_delete(payload.actions.len() as u32)
+		.max(T::WeightInfo::apply_item_actions_with_signature_v2_add(Pallet::<T>::sum_add_actions_bytes_v2(&payload.actions)))
+        )]
+		pub fn apply_item_actions_with_signature_v3(
+			origin: OriginFor<T>,
+			delegator_key: T::AccountId,
+			proof: MultiSignature,
+			payload: crate::types::ItemizedSignaturePayloadV3<T>,
```
aramikm: nit, might need to import

```suggestion
			payload: ItemizedSignaturePayloadV3<T>,
```

pallets/stateful-storage/src/lib.rs
```diff
@@ -470,15 +506,176 @@ pub mod pallet {
...
+		/// # Events
+		/// * [`Event::PaginatedPageDeleted`]
+		///
+		#[pallet::call_index(12)]
+		#[pallet::weight(T::WeightInfo::delete_page_with_signature_v3())]
+		pub fn delete_page_with_signature_v3(
+			origin: OriginFor<T>,
+			delegator_key: T::AccountId,
+			proof: MultiSignature,
+			payload: crate::types::PaginatedDeleteSignaturePayloadV3<T>,
```
aramikm:
```suggestion
			payload: PaginatedDeleteSignaturePayloadV3<T>,
```

pallets/stateful-storage/src/stateful_child_tree.rs
```diff
@@ -307,14 +307,26 @@ impl<H: MultipartKeyStorageHasher> StatefulChildTree<H> {
...
 		msa_id: &MessageSourceId,
 		pallet_name: &[u8],
 		storage_name: &[u8],
 		keys: &K,
 	) -> Result<Option<V>, ()> {
 		let child = Self::get_child_tree_for_storage(*msa_id, pallet_name, storage_name);
 		let value = child::get_raw(&child, &keys.hash());
+		match value {
+			None => Ok(None),
+			Some(v) => Ok(DecodeAll::decode_all(&mut &v[..]).map(Some).map_err(|_| ())?),
```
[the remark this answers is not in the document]
JoeCap08055: The old page struct did not have a version enum at the beginning, but theoretically an new page could exist that _might_ be able to be decoded into the parts of an old page. However, such a page would have a mismatch between the decoded payload length and the actual length of data in the page. By using `DecodeAll`, we ensure that the decoded page _exactly_ consumes 100% of the data in the page; no more, no less.

pallets/stateful-storage/src/lib.rs
```diff
@@ -819,31 +1131,174 @@ impl<T: Config> Pallet<T> {
...
+		Self(PhantomData)
+	}
+
+	/// Prevent an extrinsic from being executed while a migraiton is in progress.
+	/// Returns a `ValidTransaction` or wrapped [`ValidityError`]
+	///
+	/// # Errors
+	/// * [`ValidityError::InvalidStorageVersion`] - the on-chain storage version does not match the in-code storage version
+	///
+	pub fn block_extrinsics_during_migration() -> TransactionValidity {
```
aramikm: Q: Does this block all extrinsic calls for Stateful storage during migration? If that is the case how long is our estimate for migration time and we should also consider that we have other semi active providers

pallets/stateful-storage/Cargo.toml
```diff
@@ -28,12 +28,13 @@ sp-runtime = { workspace = true }
...
 common-primitives = { default-features = false, path = "../../common/primitives" }
 common-runtime = { path = "../../runtime/common", default-features = false }
+hex = { workspace = true, default-features = false, features = ["alloc"] }
 
 [dev-dependencies]
 env_logger = { workspace = true }
 pretty_assertions = { workspace = true }
 sp-keystore = { workspace = true }
-hex = { workspace = true, default-features = false, features = ["alloc"] }
+pallet-migrations = { workspace = true }
```
aramikm: Q: why is this only a `dev-dependency` if we are using some utilities from it to do multi-block migrations?

pallets/schemas/src/lib.rs
JoeCap08055: Schemas can no longer have names; only `Intents` and `IntentGroups` are named. Unlike with Schemas, where the name was optional, Intent/Group names are mandatory. A Schema implements one and only one _named_ Intent.

The migration ensures that for every existing Schema, we create an equivalent Intent with the same numeric ID, and the identical name (if the Schema had a name).

Q: Theoretically, if the original Schema was un-named, the resulting Intent will also not have a name. This is not necessarily problematic (and not an issue on mainnet where all schemas are named), but if we want to enforce (as the extrinsic does) that ALL Intents are named, we would have to either auto-create a name, or drop any unnamed schemas from Testnet instead of converting them.

resources/genesis-schemas.json
```diff
@@ -1,156 +1,308 @@
...
-    ],
-    "model": "{\"default\":{\"type\":\"record\",\"name\":\"DefaultTokenAddress\",\"namespace\":\"frequency\",\"fields\":[{\"name\":\"token_slip_0044\",\"type\":\"int\",\"doc\":\"Network for this token address using SLIP-0044 registered coin type integers\"},{\"name\":\"address\",\"type\":\"string\",\"doc\":\"The address as a string encoded in standard way for the given coin type\"}]}}",
-    "name": "frequency.default-token-address"
-  }
-]
+{
+  "schemas": [
+    {
+      "schema_id": 18,
+      "intent_id": 18,
```
aramikm: I'm not sure if these should be part of the payload in here. These schemas models are used in other JS tools to be imported or executed on chain along with on-chain genesis . So having these ids pre-determined is weird.
The association should ideally be by names but I'm not sure how much effort it is to make that change and if it will worth it.

js/api-augment/definitions/schemas.ts
JoeCap08055: Yes; restored in a later commit.

pallets/msa/src/types.rs
```diff
@@ -138,7 +138,7 @@ pub struct AddProvider {
 	pub authorized_msa_id: MessageSourceId,
 	/// Schemas for which publishing grants are authorized.
 	/// This is private intended for internal use only.
-	pub schema_ids: Vec<SchemaId>,
+	pub intent_ids: Vec<IntentId>,
```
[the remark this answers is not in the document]
JoeCap08055: This should be fine for any providers signing over SCALE payloads. The underlying SCALE representation hasn't changed; it's still `u16`.

If there are any providers using Eth keys to sign EIP-712, they would need to update to the new JS package to generate the correct type encoding to sign. However, I don't think we have any such providers. We can double-check.

runtime/frequency/src/lib.rs
```diff
@@ -252,28 +260,28 @@ pub struct BaseCallFilter;
...
+					StorageVersion::new(2),
+
+			#[cfg(feature = "frequency")]
+			// Filter out calls that are Governance actions on Mainnet
+			RuntimeCall::Msa(pallet_msa::Call::create_provider { .. }) |
+			RuntimeCall::Msa(pallet_msa::Call::create_application { .. }) |
+			RuntimeCall::Schemas(pallet_schemas::Call::create_schema_v4 { .. }) |
+			RuntimeCall::Schemas(pallet_schemas::Call::create_intent { .. }) |
+			RuntimeCall::Schemas(pallet_schemas::Call::create_intent_group { .. }) |
+			RuntimeCall::Schemas(pallet_schemas::Call::update_intent_group { .. }) => false,
```
aramikm: I would feel way safer if we can group these explicitly since I've seen weird behaviors when not doing so for multi-line feature flags

```suggestion
            {
				// Filter out calls that are Governance actions on Mainnet
				RuntimeCall::Msa(pallet_msa::Call::create_provider { .. }) |
				RuntimeCall::Msa(pallet_msa::Call::create_application { .. }) |
				RuntimeCall::Schemas(pallet_schemas::Call::create_schema_v4 { .. }) |
				RuntimeCall::Schemas(pallet_schemas::Call::create_intent { .. }) |
				RuntimeCall::Schemas(pallet_schemas::Call::create_intent_group { .. }) |
				RuntimeCall::Schemas(pallet_schemas::Call::update_intent_group { .. }) => false,
			}
```

runtime/frequency/src/lib.rs
```diff
@@ -252,28 +260,28 @@ pub struct BaseCallFilter;
...
+
+			#[cfg(all(feature = "frequency-bridging", feature = "frequency"))]
+			RuntimeCall::PolkadotXcm(pallet_xcm_call) => Self::is_xcm_call_allowed(pallet_xcm_call),
+			// Everything else is allowed
+			_ => true,
```
aramikm: Feature flags are weird and this help at least visually in the least
```suggestion

			// Everything else is allowed
			_ => true,
```

runtime/frequency/src/lib.rs
```diff
@@ -1976,25 +2048,100 @@ sp_api::impl_runtime_apis! {
...
+					schema_payload_location,
+					block_number,
+				).into_iter().map(|r| r.into()).collect(),
+				_ => vec![],
+			}
+		}
+
+		/// Retrieve the messages for a particular intent and block range (paginated)
+		fn get_messages_by_intent_id(intent_id: IntentId, pagination: BlockPaginationRequest) -> BlockPaginationResponse<MessageResponseV2> {
+			let mut response = BlockPaginationResponse::new();
```
aramikm: nit: This file is huge and I think it makes sense to keep these pallet specific details inside the pallet if possible

tools/eth-migration/metamask.html
```diff
@@ -426,7 +426,7 @@
 								"type": "uint64"
 							},
 							{
-								"name": "schemaIds",
+								"name": "intentIds",
```
aramikm: I think we need to add the new payloads in here such as `PaginatedDeleteSignaturePayloadV3` and etc and need to verify existing one for field name changes. These should also match with whatever we have onchain for eip712 compatbility.

JoeCap08055: Done

runtime/frequency/src/lib.rs
```diff
@@ -252,28 +260,28 @@ pub struct BaseCallFilter;
...
-				_ => true,
-			}
+		match call {
+			RuntimeCall::Utility(pallet_utility_call) =>
+				Self::is_utility_call_allowed(pallet_utility_call),
+			// Block stateful-storage extrinsics if V1->V2 migration is not complete
+			// May be removed once the migration has been completed on mainnet
+			RuntimeCall::StatefulStorage(..) =>
+				pallet_stateful_storage::Pallet::<Runtime>::on_chain_storage_version() >=
+					StorageVersion::new(2),
```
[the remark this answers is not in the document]
JoeCap08055: Fixed.

pallets/stateful-storage/src/lib.rs
JoeCap08055: We have no providers using any Stateful Storage extrinsics since MeWe, AFAIK, since the only stateful storage we currently use on-chain is for Graph, and only MeWe was doing graph stuff.
I'll confirm with @pfrank13 that we're not generating graph keys for current providers.

We'll have a better idea of the migration duration once we run the benchmarks in CI.

pallets/msa/src/types.rs
JoeCap08055: I think I left this in accidentally. Was trying to decide between renaming the field in the existing struct (only affects EIP-712 signing), or generating a complete new payload type.

I think I'll remove this--no clients are using EIP-712 (verified there are no Ethereum control keys registered on Mainnet; so nobody could be signing EIP-712 payloads), so renaming the existing `schema_ids` => `intent_ids` seems the less disruptive approach.

pallets/schemas/README.md
```diff
@@ -4,97 +4,170 @@ The Schemas Pallet provides universal schema registration for data flowing throu
...
+- [
+  `Itemized`](https://frequency-chain.github.io/frequency/common_primitives/schema/enum.PayloadLocation.html#variant.Itemized):
+  Data is stored in the Stateful Storage pallet as an array of individual items.
+- [
+  `Paginated`](https://frequency-chain.github.io/frequency/common_primitives/schema/enum.PayloadLocation.html#variant.Paginated):
+  Data is stored in the Stateful Storage pallet as a list of paged blobs.
+
+### Mainnet vs Testnet Entity Creation
+
+Mainnet Schemas, Intents, and Intent Groups must be approved by the Frequency Council.
```
shannonwells: I think it would be useful to summarize a little more specifically by saying on mainnet, those wishing to create schemas, intents and intent groups must call the `propose_to_create...` extrinsics, whereas on testnet they can simply call the `create_...` extrinsics directly.  Also that the `create_foo_via_governance` extrinsics aren't generally meant to be called directly.

pallets/messages/src/migration/v2/mod.rs
```diff
@@ -0,0 +1,54 @@
...
+	),
+	Message<<T as Config>::MessagesMaxPayloadSizeBytes>,
+	OptionQuery,
+>;
+
+/// Ephemeral storage key for tracking the completion status of the migration
+/// in order to perform the final step and for try-runtime. MUST be killed at
+/// the end of the migration!
+#[storage_alias]
+pub(crate) type DoneV3Migration<T: Config> = StorageValue<Pallet<T>, bool, ValueQuery>;
```
aramikm: nit: Shouldn't this be part of the Storage of the pallet? `storage_alias` is usually used to access existing storages with old type deserialization. In this case this is a brand new storage and might make sense to just move it into the pallet even though maybe `storage_alias` can be used to achieve the same.

pallets/schemas/src/lib.rs
```diff
@@ -328,228 +473,369 @@ pub mod pallet {
...
-				None => T::WeightInfo::create_schema_v3_without_name(model.len() as u32 + settings.len() as u32)
-			}
-		)]
-		pub fn create_schema_v3(
+		/// * [`Error::InvalidIntentGroupId`] - The specified [`IntentGroupId`] does not exist.
+		/// * [`Error::InvalidIntentId`] - At least one of the specified [`IntentId`]s does not exist.
+		/// * [`Error::TooManyIntentsInGroup`] - The update would result in too many Intents in the group.
+		#[pallet::call_index(16)]
+		#[pallet::weight(T::WeightInfo::update_intent_group(intent_ids.len() as u32))]
+		pub fn update_intent_group(
```
shannonwells: These extrinsics don't so much update as replace the intent group ids with the new collection, and that isn't obvious from the design doc / readme.

designdocs/schemas_protocols_intents.md
```diff
@@ -273,59 +274,59 @@ pub struct IntentInfo {
 }
 ```
 
-### 8. **Delegation Groups**<a id="delegation_groups"></a>
+### 8. **Intent Groups**<a id="intent_groups"></a>
 
 As mentioned <a href="#delegation_semantics">above</a>, other than changing the interpretation of a Delegation from
```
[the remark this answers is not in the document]
shannonwells: > The main motivation was that under the old model in which Schema represented both format and purpose, even a minor format change required a new delegation, which gets very cumbersome & expensive across an entire user base. Decoupling format from purpose (intent) allows data formats to evolve while keeping on-chain delegation churn to a minimum.

Right, I just meant, the design doc could use a concrete example, like, we have a collection of DSNP schemas that are pretty much used together all the time, but there was no way to group them on chain, and meaningful grouping was just done manually without any hint of why they belong together in the code, nor any external discoverability of that without asking a human.  But the ATProto one is completely separate and meaningless wrt DSNP, and again, there was no way to know that without asking someone.

pallets/messages/src/weights.rs
```diff
@@ -35,83 +35,133 @@ use core::marker::PhantomData;
...
 		//  Estimated: `4008`
-		// Minimum execution time: 31_646_000 picoseconds.
-		Weight::from_parts(32_704_000, 4008)
+		// Minimum execution time: 25_000_000 picoseconds.
+		Weight::from_parts(43_000_000, 4008)
 			.saturating_add(T::DbWeight::get().reads(2_u64))
 			.saturating_add(T::DbWeight::get().writes(1_u64))
 	}
+	/// Storage: UNKNOWN KEY `0x9ea3e2d10fdb9a071f2f534d51b0961f963740dfb0edb77dddcbb1e5542cf4d2` (r:2 w:1)
+	/// Proof: UNKNOWN KEY `0x9ea3e2d10fdb9a071f2f534d51b0961f963740dfb0edb77dddcbb1e5542cf4d2` (r:2 w:1)
```
aramikm: why is there 2 reads on this? Is it because of the `iter.next()` ? and what we talked about reading prefix keys. Since it is retrieving the key and value in the same call I was going to assume it's just 1 read for reading both and 1 write for drain functionality but this is still weird.

pallets/messages/src/lib.rs
```diff
@@ -337,27 +362,81 @@ impl<T: Config> Pallet<T> {
...
+
+		// Schema Fetch and Check
+		let intent = match <T>::SchemaProvider::get_intent_by_id(intent_id) {
+			Some(intent) => intent,
+			None => return response,
+		};
+
+		let mut from_index: u32 = pagination.from_index;
+
+		let _ = (pagination.from_block..pagination.to_block).try_for_each(|block_number| {
```
enddynayn: I think we can improve the readability by using a `for block_number in pagination.from_block..pagination.to_block` since we are mutating external variables

for block_number in pagination.from_block..pagination.to_block

pallets/messages/src/lib.rs
```diff
@@ -337,27 +362,81 @@ impl<T: Config> Pallet<T> {
...
+
+			// Max messages in a block are constrained to MessageIndex (u16) by the storage,
+			// so this is a safe type coercion. Just to be safe, we'll trap in in debug builds
+			let list_size: u32 = list.len() as u32;
+			debug_assert!(list_size <= u16::MAX.into(), "unexpected number of messages in block");
+
+			let iter = list.into_iter().skip((from_index as usize).saturating_sub(1));
+			// all subsequent blocks in this call should start at index 0
+			from_index = 0;
+			iter.enumerate().try_for_each(|(i, m)| {
```
enddynayn: same here.

pallets/messages/src/lib.rs
```diff
@@ -337,27 +362,81 @@ impl<T: Config> Pallet<T> {
...
+		let _ = (pagination.from_block..pagination.to_block).try_for_each(|block_number| {
+			let list: Vec<MessageResponseV2> = Self::get_messages_by_intent_and_block(
+				intent_id,
+				intent.payload_location,
+				block_number.into(),
+			);
+
+			// Max messages in a block are constrained to MessageIndex (u16) by the storage,
+			// so this is a safe type coercion. Just to be safe, we'll trap in in debug builds
+			let list_size: u32 = list.len() as u32;
```
enddynayn: Nit: I think we can avoids casting before asserting correctness. This also improves readability.
```
let list_size = list.len();
debug_assert!(list_size <= u16::MAX as usize, '....");
let list_size = list_size as u32;
```

pallets/messages/src/benchmarking.rs
```diff
@@ -118,22 +120,19 @@ mod benchmarks {
 			.as_bytes()
 			.to_vec();
 		let schema_id = IPFS_SCHEMA_ID;
+		let intent_id = IPFS_SCHEMA_ID as IntentId;
```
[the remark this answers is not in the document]
JoeCap08055: IPFS_SCHEMA_ID is defined as 20. In the mocks we've taken care of this; in benchmark mode with the regular runtime, the chain genesis contains the mainnet schemas; 20 is an IPFS schema.

The original benchmarks were written before we had the chain genesis seeded with schemas; it's no longer necessary to do so in the benchmark.

pallets/schemas/src/lib.rs
```diff
@@ -700,64 +1138,194 @@ pub mod pallet {
...
+		/// Update the list of [`IntentId`]s associated with a given [`IntentGroupId`]
+		///
+		/// # Errors
+		/// * [`Error::InvalidIntentGroupId`] - The specified [`IntentGroupId`] doesn't exist
+		/// * [`Error::InvalidIntentId`] - One of the [`IntentId`]s specified doesn't exist
+		pub fn update_intent_group_for(
+			intent_group_id: IntentGroupId,
+			intent_ids: BoundedVec<IntentId, T::MaxIntentsPerIntentGroup>,
+		) -> Result<(), DispatchError> {
+			ensure!(
```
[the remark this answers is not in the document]
JoeCap08055: It's not an error to have an empty intent group; we should discuss whether it would be better to simply allow the group to be deleted instead. The difference being that an empty group keeps the name reserved.

I think it's probably fine to file a separate issue to address this more fully later.

pallets/messages/src/migration/v2/mod.rs
JoeCap08055: The only real difference between a storage alias and "regular" storage is that the key prefix is registered in the pallet & exposed in metadata.

Interestingly, this could be a way to define "private" storage inside a pallet that's "hidden" from the metadata (though you could still retrieve it by hex key, if you compute the hash from the source).
Not sure what that buys anyone, aside from discouraging relying on the storage API vs runtime API for queries... but interesting.

pallets/schemas/src/weights.rs
```diff
@@ -151,18 +105,59 @@ impl<T: frame_system::Config> WeightInfo for SubstrateWeight<T> {
...
+	fn create_intent_group(m: u32, ) -> Weight {
+		// Proof Size summary in bytes:
+		//  Measured:  `210 + m * (39 ±0)`
+		//  Estimated: `4055 + m * (2488 ±0)`
+		// Minimum execution time: 14_000_000 picoseconds.
+		Weight::from_parts(17_499_362, 4055)
+			// Standard Error: 37_852
+			.saturating_add(Weight::from_parts(2_552_879, 0).saturating_mul(m.into()))
+			.saturating_add(T::DbWeight::get().reads(2_u64))
+			.saturating_add(T::DbWeight::get().reads((1_u64).saturating_mul(m.into())))
```
aramikm: This is weird. I think these weights are generated when we were reading all intents for verification and might need to re-generate this to be accurate.

common/primitives/src/msa.rs
```diff
@@ -318,45 +337,50 @@ pub trait MsaValidator {
...
 	>;
 }
 
 /// A behavior that allows for validating a schema grant
-pub trait SchemaGrantValidator<BlockNumber> {
-	/// Validates if the provider is allowed to use the particular schema id currently
-	fn ensure_valid_schema_grant(
+pub trait GrantValidator<DelegationIdType, BlockNumber> {
+	// /// The id type of the items being delegated
+	// type DelegationIdType;
```
enddynayn: Leftover?

pallets/stateful-storage/src/migration/v2.rs
```diff
@@ -0,0 +1,392 @@
...
+#[cfg(feature = "try-runtime")]
+use sp_runtime::TryRuntimeError;
+use sp_runtime::Weight;
+
+const LOG_TARGET: &str = "pallet::stateful-storage::migration::v2";
+
+/// The length of a PaginatedKey (twox_128, twox_128, u16, u16)
+pub type PaginatedKeyLength = ConstU32<72>;
+/// The length of an ItemizedKey (twox_128, u16)
+pub type ItemizedKeyLength = ConstU32<36>;
```
aramikm: Are these correct?

isn't twox_128 + u16 going to be 34 not 36?

pallets/stateful-storage/src/migration/v2.rs
```diff
@@ -0,0 +1,399 @@
...
+use sp_core::storage::ChildInfo;
+#[cfg(feature = "try-runtime")]
+use sp_runtime::TryRuntimeError;
+
+const LOG_TARGET: &str = "pallet::stateful-storage::migration::v2";
+
+/// The length of a PaginatedKey (twox_128, twox_128, u16, u16)
+pub type PaginatedKeyLength = ConstU32<72>;
+/// The length of an ItemizedKey (twox_128, u16)
+pub type ItemizedKeyLength = ConstU32<36>;
```
aramikm: nit: I think twox_128 + u16 = 34

runtime/frequency/src/lib.rs
```diff
@@ -1647,6 +1682,46 @@ impl pallet_utility::Config for Runtime {
...
+		>,
+		pallet_stateful_storage::migration::v2::MigrateItemizedV1ToV2<
+			Runtime,
+			pallet_stateful_storage::weights::SubstrateWeight<Runtime>,
+		>,
+		pallet_stateful_storage::migration::v2::FinalizeV2Migration<
+			Runtime,
+			pallet_stateful_storage::weights::SubstrateWeight<Runtime>,
+		>,
+	);
```
aramikm: Are the order of these migrations important? I think if they are maybe we want the storage to be first since we are locking the functionality.

pallets/stateful-storage/src/weights.rs
```diff
@@ -199,6 +204,70 @@ impl<T: frame_system::Config> WeightInfo for SubstrateWeight<T> {
...
+	/// Proof: `Schemas::SchemaInfos` (`max_values`: None, `max_size`: Some(17), added: 2492, mode: `MaxEncodedLen`)
+	/// Storage: UNKNOWN KEY `0x0763c98381dc89abe38627fe2f98cb7af1577fbf1d628fdddb4ebfc6e8d95fb1` (r:1 w:1)
+	/// Proof: UNKNOWN KEY `0x0763c98381dc89abe38627fe2f98cb7af1577fbf1d628fdddb4ebfc6e8d95fb1` (r:1 w:1)
+	fn delete_page_with_signature_v3() -> Weight {
+		// Proof Size summary in bytes:
+		//  Measured:  `2112`
+		//  Estimated: `60286`
+		// Minimum execution time: 168_000_000 picoseconds.
+		Weight::from_parts(175_000_000, 60286)
+			.saturating_add(T::DbWeight::get().reads(26_u64))
```
aramikm: This does not make sense. Reading 23 schemaInfos ?

aramikm: Some of these DB weights should be replaced with RocksDbWeightChild. We need to rerun the weights on this.

aramikm (changes requested):

designdocs/schemas_protocols_intents.md
```diff
@@ -273,59 +279,95 @@ pub struct IntentInfo {
...
+pub type IntentGroupId = u16;
 
-pub struct DelegationGroup {
+pub struct IntentGroup {
     /// List of Intents associated with this Delegation
     pub intent_ids: BoundedVec<IntentId, ConstU32<MAX_INTENTS_PER_GROUP>>,
 }
 ```
 
+#### Example
```
enddynayn: ❤️

pallets/stateful-storage/src/types.rs
```diff
@@ -449,30 +585,32 @@ impl<T: Config> ItemizedOperations<T> for ItemizedPage<T> {
...
-				PageError::ErrorParsing("wrong header size")
-			);
-			let header = <ItemHeader>::decode(&mut &self.data[offset..])
-				.map_err(|_| PageError::ErrorParsing("decoding header"))?;
-			let item_total_length = ItemHeader::max_encoded_len() + header.payload_len as usize;
+		let mut page_slice = &self.data[..];
+
+		while page_slice.len() > 0 {
+			let item_slice = page_slice;
+			let header = <ItemHeader>::decode(&mut page_slice)
```
aramikm: So the idea is that the decode modifies the page_slice and consumes the bytes that is decoded and it leaves the rest?

runtime/frequency/src/lib.rs
```diff
@@ -1647,6 +1678,46 @@ impl pallet_utility::Config for Runtime {
...
+			pallet_messages::weights::SubstrateWeight<Runtime>,
+		>,
+	);
+	// Benchmarks need mocked migrations to guarantee that they succeed.
+	#[cfg(feature = "runtime-benchmarks")]
+	type Migrations = pallet_migrations::mock_helpers::MockedMigrations;
+	type CursorMaxLen = ConstU32<65_536>;
+	type IdentifierMaxLen = ConstU32<256>;
+	type MigrationStatusHandler = ();
+	type FailedMigrationHandler = frame_support::migrations::FreezeChainOnFailedMigration;
```
aramikm: Q: Is there other strategies beside the freeze and what happens if in the worst case it get's frozen? Just a rough idea might be helpful to know

JoeCap08055: consensus: `Ignore`

pallets/msa/src/lib.rs
```diff
@@ -2883,7 +2904,7 @@ impl<T: Config> TargetValidator for Pallet<T> {
 	}
 }
 
-impl<T: Config> SchemaGrantValidator<BlockNumberFor<T>> for Pallet<T> {
+impl<T: Config> GrantValidator<IntentId, BlockNumberFor<T>> for Pallet<T> {
```
shannonwells: Should this use Self::BlockNumber?

```suggestion
impl<T: Config> GrantValidator<IntentId, Self::BlockNumber> for Pallet<T> {
```

common/primitives/src/schema.rs
```diff
@@ -120,54 +215,110 @@ pub struct SchemaInfoResponse {
...
+	/// The payload location associated with this Schema's associated Intent
 	pub payload_location: PayloadLocation,
-	/// grants for the schema
-	pub settings: Vec<SchemaSetting>,
+	/// The settings for this Schema's associated Intent
+	pub settings: Vec<IntentSetting>,
+	/// The IntentId of the Intent that this Schema implements
+	pub intent_id: IntentId,
+	/// The status of this Schema
+	pub status: SchemaStatus,
```
shannonwells: nit: truism - because "status" is a general term, this comment doesn't help a reader see the purpose of the the field.

shannonwells (approved):
Nothing blocking from me!

e2e/schemas/schemas.test.ts
```diff
@@ -0,0 +1,343 @@
...
+    const { target: createIntentEvent } = await f.fundAndSend(fundingSource);
+    if (createIntentEvent && ExtrinsicHelper.apiPromise.events.schemas.IntentCreated.is(createIntentEvent)) {
+      createdIntents.push({ intentId: createIntentEvent.data.intentId.toNumber(), intentName, schemaIds: [] });
+    }
+
+    assert.equal(createdIntents.length, 1);
+  });
+
+  describe('intents', function () {
+    it('create intent should fail if account does not have enough tokens', async function () {
```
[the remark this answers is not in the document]
JoeCap08055: ...sadly, Mocha doesn't support table tests out of the box

PR #2569, opened 2025-08-21
[DRAFT] refactor: remove frequency-bridging feature and related configurations

### LINKED ISSUE #2515: Remove feature = "frequency-bridging"
## Description

Bridging is a core feature of Frequency and once deployed will not need to be configured conditionally.
See also #2367 when re-factoring features.

## Acceptance Criteria

- [ ] `frequency-bridging` feature flag removed

mattheworris:
# Goal
The goal of this PR is to remove the `frequency-bridging` feature flag and refactor the code so that it is easier to read and maintain.

Closes #2515

# Discussion

- Must ensure the `frequency-no-relay` continues to work correctly. XCM features require some cumulus features.
- Code readability should improve.

# Checklist
- [ ] Updated Pallet Readme?
- [ ] Design doc(s) updated?
- [ ] Unit Tests added?
- [ ] e2e Tests added?
- [ ] Benchmarks added?

### DIFF
--- a/runtime/frequency/src/lib.rs
+++ b/runtime/frequency/src/lib.rs
@@ -459,9 +459,6 @@ pub type BlockId = generic::BlockId<Block>;
 /// Block type as expected by this runtime.
 pub type Block = generic::Block<Header, UncheckedExtrinsic>;
 
-#[cfg(any(not(feature = "frequency-no-relay"), feature = "frequency-lint-check"))]
-pub type AssetBalance = Balance; // kept for backwards compatibility; now centralized in relay::prelude
-
 /// Unchecked extrinsic type as expected by this runtime.
 pub type UncheckedExtrinsic =
 	generic::UncheckedExtrinsic<Address, RuntimeCall, Signature, TxExtension>;
--- a/runtime/frequency/src/relay.rs
+++ b/runtime/frequency/src/relay.rs
@@ -9,7 +9,6 @@ pub mod prelude {
 	use super::*;
 
 	// ---------------- Type Aliases ----------------
-	// AssetBalance previously re-exported; kept in lib for backwards compatibility if needed.
 	pub type RuntimeMigrations =
 		(MigratePalletsCurrentStorage<Runtime>, SetSafeXcmVersion<Runtime>);
 	pub type OnSetCodeHook = cumulus_pallet_parachain_system::ParachainSetCode<Runtime>;


Makefile
```diff
@@ -349,19 +347,22 @@ build-westend-release:
 build-mainnet-release:
 	cargo build --locked --features  frequency --release
 
-build-bridging-testnet:
-	cargo build --features frequency-testnet,frequency-bridging
+# build-bridging-testnet:
+# 	cargo build --features frequency-testnet,frequency-bridgin
```
shannonwells, line 351: non-blocking, but it is odd that the commented-out features are all misspelled

Makefile
JoeCap08055, line 1: Is there a reason we're keeping all of the commented-out stuff in the Makefile?

node/cli/src/command.rs
JoeCap08055, line 1: So, are we just getting rid of all Westend-related code, the idea being that we're focusing on bridging on Paseo (and then Mainnet)?

runtime/frequency/src/lib.rs
```diff
@@ -455,32 +459,31 @@ pub type BlockId = generic::BlockId<Block>;
...
-	Block,
-	frame_system::ChainContext<Runtime>,
-	Runtime,
-	AllPalletsWithSystem,
-	(MigratePalletsCurrentStorage<Runtime>, SetSafeXcmVersion<Runtime>),
->;
+#[cfg(feature = "frequency-no-relay")]
+use crate::no_relay::prelude as runtime_mode;
+#[cfg(any(not(feature = "frequency-no-relay"), feature = "frequency-lint-check"))]
+use crate::relay::prelude as runtime_mode;
```
JoeCap08055, line 469: Hmmm... initially I was confused by:
```rust
#[cfg(any(not(feature = "frequency-no-relay"), feature = "frequency-lint-check"))]
```
that's used in so many places; it seems like adding `feature = "frequency-lint-check"` to the `any(...)` ONLY enables the use case: `--features frequency-no-relay,frequency-lint-check`. I was going to let it slide, but here I think I see a problem. If we run `--features frequency-no-relay,frequency-lint-check`, here we'll end up with:
```rust
use crate::no_relay::prelude as runtime_mode; // enabled by "frequency-no-relay"
use crate::relay::prelude as runtime_mode; // enabled by "frequency-lint-check"
```

I think this would like be problematic...so, by that logic, it seems like we should NEVER allow both `frequency-no-relay` AND `frequency-lint-check` at the same time. And, therefore, we can reduce all of the instances of:
```rust
#[cfg(any(not(feature = "frequency-no-relay"), feature = "frequency-lint-check"))]
```
to simply
```rust
#[cfg(not(feature = "frequency-no-relay"))]
```

node/cli/src/command.rs
mattheworris: Westend-local was kind of a hack for testing with Snowbridge Zombienet, so removing it to clean that up. The westend release still remains.

PR #2573, opened 2025-08-25
[WIP] 2526 RPC for provider context fetching provider/application context give inputs

saraswatpuneet:
# Goal
The goal of this PR is 

Closes 

# Discussion

- 

# Checklist
- [ ] Updated Pallet Readme?
- [ ] Design doc(s) updated?
- [ ] Unit Tests added?
- [ ] e2e Tests added?
- [ ] Benchmarks added?


pallets/msa/src/lib.rs
```diff
@@ -2417,6 +2417,47 @@ impl<T: Config> Pallet<T> {
...
+			localized_name = provider_or_application_registry
+				.localized_names
+				.get(&locale)
+				.map(|bv| bv.to_vec());
+			let localized_logo_cid =
+				provider_or_application_registry.localized_logo_250_100_png_cids.get(&locale);
+			match localized_logo_cid {
+				Some(cid) => ApprovedLogos::<T>::get(cid).map(|bv| bv.to_vec()),
+				None => None,
+			}
```
aramikm, line 2449: nit: you might be able to use `and_then` in here. even for the following
`let localized_logo_250_100_png_bytes: Option<Vec<u8>> = if let Some(locale) = bounded_locale`
if looks more elegant in general

pallets/msa/src/rpc/src/lib.rs
```diff
@@ -72,6 +73,17 @@ pub trait MsaApi<BlockHash, AccountId> {
 		&self,
 		msa_id: MessageSourceId,
 	) -> RpcResult<Option<KeyInfoResponse<AccountId>>>;
+
+	/// Retrieve a provider's application context or default provider context
+	/// If application id is not specified, rpc will return the default provider context
+	/// The default provider context is registered at time of provider registration.
+	#[method(name = "msa_getProviderApplicationContext")]
```
aramikm, line 80: I think we had a consensus to not add new RPCs (only runtimes) unless there is a functionality that we can't do solely from inside runtime. This would simplify the upgrades and keep our node code simple. Please let me know if there are questions or concerns.

aramikm (approved):
Overall looks good but I think we should not add new RPC if not absolutely justified.

pallets/msa/src/rpc/src/lib.rs
saraswatpuneet: Good point, i will see how Provider Dashboard is using the provider state queries right now and if they are state queries we can get rid of RPC itself

saraswatpuneet: Ok makes me to not have RPC, i have removed the RPC and retained the runtime api for this and updated e2e tests

PR #2585, opened 2025-08-29
feat: update frequency runtime call for bridging support in metadata generation

### LINKED ISSUE #2584: [Bug]: Metadata does not include `frequency-bridging` feature
### What happened?

While working on #2515, it was discovered that `frequency-bridging` was not added to the metadata. `make js` failed because it could not find `FrequencyRuntimeRuntimeCall`.

## Acceptance Criteria

- [x] `FrequencyRuntimeRuntimeCall` is added to `js/api-augment/definitions` and `make js` compiles correctly.
- [x] `generate_js_definitions.sh` should include the `frequency-bridging` feature.

### Relevant log output

```shell

```

mattheworris:
# Goal
The goal of this PR is to fix metadata to include the bridging feature.

Closes #2584

# Checklist
- [ ] Updated Pallet Readme?
- [ ] Design doc(s) updated?
- [ ] Unit Tests added?
- [ ] e2e Tests added?
- [ ] Benchmarks added?


runtime/frequency/src/lib.rs
JoeCap08055, line 1: Do we really need to update the runtime for just a JS package update? Is there a way to just release the updated JS packages?

saraswatpuneet: I think it runs a full ci check for even .ts files, maybe we can limit the filters that in PR CI but yes we can just force push it

JoeCap08055: But, I guess then we'd have to manually trigger the JS release pipeline, and the package version wouldn't match the latest chain version... eh, probably easier to just do a Frequency release. We don't have to deploy the runtime, I suppose.

JoeCap08055: I'll approve & leave it to your best judgment.

mattheworris: I can try CI without bumping the `spec_version`. I'm not sure what requires a check. I proactively updated `spec_version` as I thought we decided on a liberal spec version update, in that basically any change  would update the spec.

However, this is basically an add-on to the bridging PR, so I don't think it would need any specific version update, but it needs to go in the release to Paseo.

mattheworris: A spec bump was not necessary, will merge without spec bump.

PR #2612, opened 2025-09-19
bugfix: fixed capacity estimation RPC

### LINKED ISSUE #2579: [Bug]: Capacity fee estimation RPC double-charges
### What happened?

# Description
The RPC `frequencyTxPayment.computeCapacityFeeDetails` is apparently double-charging weights for transactions. It's currently adding in the dispatch weight of the outer call, in addition to the Capacity stable weights of the inner calls. The payment extension ONLY relies on the stable weights of the inner calls when actually debiting Capacity, so the RPC should similarly only rely on the stable weights.

# Reproduction
* Run a local instant seal node
* Create an extrinsic wrapped with `frequencyTxPayment.payWithCapacity` _and sign it_.
* Submit it to `frequencyTxPayment.computeCapacityFeeDetails` and total the response to get the total Capacity fee
* Submit the actual extrinsic and retrieve the actual `capacity.CapacityWithdrawn` event from the block
* The two values should be equal or nearly equal

### Relevant log output

```shell

```

aramikm:
# Goal
The goal of this PR is to fix the bug in capacity estimation RPC.

Closes #2579

# Discussion
- The issue was using the weight from extrinsic which includes the inner calls instead of getting the bare capacity weights directly.

# Checklist

# Verification
As shown Calculated and deducted capacity values are really close.

Screenshot 2025-09-18 at 4 28 54 PM
Screenshot 2025-09-18 at 4 29 11 PM

### DIFF
--- a/pallets/frequency-tx-payment/src/lib.rs
+++ b/pallets/frequency-tx-payment/src/lib.rs
@@ -265,14 +265,14 @@ impl<T: Config> Pallet<T> {
 	/// Compute the capacity fee details for a transaction.
 	/// # Arguments
 	/// * `runtime_call` - The runtime call to be dispatched.
-	/// * `weight` - The weight of the transaction.
+	/// * `overhead_weight` - The overhead weight asociated with capacity transactions.
 	/// * `len` - The length of the transaction.
 	///
 	/// # Returns
 	/// `FeeDetails` - The fee details for the transaction.
 	pub fn compute_capacity_fee_details(
 		runtime_call: &<T as Config>::RuntimeCall,
-		dispatch_weight: &Weight,
+		overhead_weight: &Weight,
 		len: u32,
 	) -> FeeDetails<BalanceOf<T>> {
 		let calls = T::CapacityCalls::get_inner_calls(runtime_call)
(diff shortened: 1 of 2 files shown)


saraswatpuneet (approved):

JoeCap08055 (commented):
Good catch. Do we want to add/update any tests? I believe we have e2e tests that deal with stable weights...

e2e/capacity/capacity_rpc.test.ts
```diff
@@ -96,19 +100,31 @@ describe('Capacity RPC', function () {
       addProviderPayload
     );
     const tx = ExtrinsicHelper.api.tx.frequencyTxPayment.payWithCapacity(insideTx);
+    const signedTx = await firstValueFrom(tx.signAsync(capacityProviderKeys));
+    const signedHex = signedTx.toHex();
+    // it's important to submit a signed extrinsic to the rpc to get an accurate estimate due to the actual length of extrinsic and etc.
```
enddynayn, line 105: 👍

enddynayn (approved):
👍 superb!

PR #2616, opened 2025-09-30
2581 update messages and stateful storage pallets to support intents

JoeCap08055:
# Goal
The goal of this PR is to migrate the `messages` pallet to add `schema_id` to message storage, and use `intent_id` as the index instead.

Existing index values do not need to be changed, as the `schemas` pallet migration ensures that the chain is seeded with Intent IDs that match existing Schema IDs

Related to #2581 

# Checklist
- [ ] Updated Pallet Readme?
- [ ] Design doc(s) updated?
- [x] Unit Tests added?
- [ ] e2e Tests added?
- [x] Benchmarks added?


Makefile
JoeCap08055, line 1: Made the following enhancements for `try-runtime-*` targets:
- Only build the runtime, not the entire node; speeds up the build
- Consolidate targets using patterns to avoid duplicated targets/commands
- Fetch state for a minimal set of system pallets so that multi-block migrations can be properly tested
- Allow the use of the `PALLETS` env var to specify additional pallets for dumping/restoring state

pallets/messages/src/tests/mock.rs
```diff
@@ -68,25 +75,34 @@ impl system::Config for Test {
...
 pub type MaxSchemaGrantsPerDelegation = ConstU32<30>;
 
 // Needs parameter_types! for the impls below
 parameter_types! {
-	// Max payload size was picked specifically to be large enough to accommodate
-	// a CIDv1 using SHA2-256, but too small to accommodate CIDv1 w/SHA2-512.
-	// This is purely so that we can test the error condition. Real world configuration
-	// should have this set large enough to accommodate the largest possible CID.
-	// Take care when adding new tests for on-chain (not IPFS) messages that the payload
-	// is not too big.
```
JoeCap08055, line 87: This was an erroneous assumption. Since the configured value would never be this small on mainnet or testnet, it's better to allow a CIDv1/SHA2-512 by size, and have it fail for CID validation instead of for size reasons.

pallets/messages/src/tests/other_tests.rs
```diff
@@ -68,24 +60,9 @@ fn populate_messages(
 /// Helper function to generate message payloads
 ///
 /// # Arguments
-/// * `num_items` - Number of message to include in payload
```
JoeCap08055, line 71: This seemed overly complicated--a payload is just an array of bytes; why all the JSON stuff?

pallets/messages/src/types.rs
```diff
@@ -8,6 +11,12 @@ use parity_scale_codec::{Decode, Encode, MaxEncodedLen};
 use scale_info::TypeInfo;
 extern crate alloc;
 use alloc::vec::Vec;
+use common_primitives::messages::MessageResponseV2;
+
+/// Multihash hash function code for SHA2-256
+pub const SHA2_256: u64 = 0x12;
+/// Multihash hash function code for Blake3
+pub const BLAKE3: u64 = 0x1e;
```
saraswatpuneet, line 19: Move this to common? so msa pallet can use it too (i might add it already in my PR)

pallets/messages/src/migration/v3.rs
```diff
@@ -0,0 +1,173 @@
...
+			}
+
+			// If there's a next item in the iterator, perform the migration.
+			if let Some(((block_number, schema_id, index), value)) = iter.next() {
+				count += 1;
+				// Migrate the inner value to the new structure
+				let value = convert_from_old::<T>(schema_id, value);
+				// We can just insert here since the old and the new map share the same key-space.
+				MessagesV3::<T>::insert((block_number, schema_id, index), value);
+				last_cursor = Some((block_number, schema_id, index));
```
saraswatpuneet, line 84: is it better to use a cursor using the raw key pointing to last processed key in state? that way not sure you need to know block number

JoeCap08055: Honestly, we don't even use the cursor for anything, it just needs to be `Some(...)` or `None`. We don't inspect the cursor value at all, since on each iteration we're just taking the next value from a`Storage::drain()` iterator.

I just figured the key values are good to store in the cursor in case we ever need to inspect where we were in a failed migration.

common/primitives/src/cid.rs
JoeCap08055, line 1: This file is not really net-new; I just haven't merged from `main` in a while. I've added some additional consts, and moved the CID validation function and tests from the `messages` pallet.

PR #2618, opened 2025-10-01
2617: Retain `ProviderToRegistryEntry` till downstream migration

### LINKED ISSUE #2617: [Provider Context] Maintain backward compatibility for storage queries
### Current Issue

We renamed `ProviderToRegistryEntry` → `ProviderToRegistryEntryV2`, but this introduced some breaking changes across the stack (mainly around name resolution for providers created via V2).

### Proposed Fixes

* **Retain old registry (`ProviderToRegistryEntry`)** intact and mark it as **deprecated** (keeps backward compatibility for NC nodes and older tooling).
* **Skip write-through from V2 → V1**. New providers created via V2 won’t populate V1, but that’s acceptable since:

  * Existing V1 data remains untouched.
  * Migration already copied V1 → V2.
  * Internal tooling may need minor tweaks, but no re-migration is required.

### Next Steps

* Keep V1 around as a read-only, deprecated view.
* Move forward with V2 as the canonical source of truth.
* Deprecation notice should be added to docs + release notes.

saraswatpuneet:
# Goal
The goal of this PR is to retain `ProviderToRegistryEntry` for backward compatibility, to be deprecated post migration. Note: No new provider will be written to this storage.

Closes  #2617 
# Discussion

* Rationale is to support older state and notify downstream consumer applications to upgrade to new standard. This prevent breaking changes to some extent.
* `ProviderToRegistryEntry` remain frozen

# Checklist
- [ ] Updated Pallet Readme?
- [ ] Design doc(s) updated?
- [x] Unit Tests added?
- [ ] e2e Tests added?
- [ ] Benchmarks added?


pallets/msa/src/migration/v2.rs
```diff
@@ -16,25 +17,42 @@ pub struct MigrationStatus {
...
+	start_after_raw_key: Option<Vec<u8>>,
+) -> (Weight, u32, Option<Vec<u8>>) {
 	let mut reads = 0u64;
 	let mut writes = 0u64;
 	let mut bytes = 0u64;
 	let mut migrated_count = 0u32;
+	let mut last_raw_key = None;
+
+	let iter = match start_after_raw_key {
+		Some(raw_key) => ProviderToRegistryEntry::<T>::iter_from(raw_key),
```
saraswatpuneet, line 41: [iter_from](https://paritytech.github.io/polkadot-sdk/master/frame_support/storage/types/struct.StorageMap.html#method.iter_from)

pallets/msa/src/migration/v2.rs
```diff
@@ -44,14 +62,15 @@ pub fn migrate_provider_entries_batch<T: Config>(batch_size: usize) -> (Weight,
 		};
 		// Insert into new storage
 		ProviderToRegistryEntryV2::<T>::insert(provider_id, migrated_provider_entry);
-		reads += 1;
+		reads += 2;
 		writes += 1;
 		bytes += old_entry.encoded_size() as u64;
 		migrated_count += 1;
+		last_raw_key = Some(ProviderToRegistryEntry::<T>::hashed_key_for(provider_id));
```
saraswatpuneet, line 69: store the last processed raw key to start iterating from (in subsequent calls)

pallets/msa/src/migration/v2.rs
JoeCap08055: Had to look this up to verify: claims to start iterating at the key AFTER the specified key 👍🏻 
(Documentation isn't super clear on that point, might want to verify the behavior)

JoeCap08055 (approved):
Looks good.

Commenting here to have a record:
* Although we're leaving old storage in place for one more release cycle to give client apps time to migrate to the new API, we're _NOT_ doing a write-through from the new storage -> old storage on insert/update new ProviderRegistryEntry. So the `v1` storage is frozen in time as of this release. This is acceptable because we're not anticipating onboarding new Mainnet Providers in the interim. Apps on Testnet will need to update to the new API or self-manage updating the old storage if they want to work with new Providers.

pallets/msa/src/migration/v2.rs
saraswatpuneet: I will add unit test for this

pallets/msa/src/migration/v2.rs
```diff
@@ -16,25 +17,42 @@ pub struct MigrationStatus {
 	pub migrated_count: u32,
 	/// Whether migration is complete
 	pub completed: bool,
+	/// Last key processed
+	pub last_raw_key: Option<vec::Vec<u8>>,
 }
 
 /// Storage for tracking migration progress
 #[storage_alias]
 pub type MigrationProgressV2<T: Config> = StorageValue<Pallet<T>, MigrationStatus, ValueQuery>;
```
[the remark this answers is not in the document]
enddynayn: Curious, why don't we need to pass "pallet_name" to the macro such as: #[storage_alias(pallet_name)]


https://paritytech.github.io/polkadot-sdk/master/frame_support/attr.storage_alias.html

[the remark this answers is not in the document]
saraswatpuneet: I think our migrations aren't that complex to have any use for verbatim beyond a prefix in storage that will be removed once migration finished successfully 

But we can define some standards around such as part of our larger push for standardizing development

pallets/msa/src/migration/v2.rs
shannonwells: definitely needs a unit test

PR #2652, opened 2025-12-10
chore: create common action to setup Node.js and correct version of npm

JoeCap08055:
# Goal
The goal of this PR is to ensure that our CI pipeline is using at least version `>= 11.5.1` of `npm` in order to support OID auth from GitHub.

It refactors many individual references to `actions/setup-node` into a composite action that incorporates both `actions/setup-node` as well as the specific commands to ensure a recent enough version of `npm`.

Another benefit is that the default version of Node.js for all of our CI workflows is now specified in a single place (but can stlll be overridden per invocation if necessary)

### DIFF
--- a/.github/workflows/verify-pr-commit.yml
+++ b/.github/workflows/verify-pr-commit.yml
@@ -475,7 +475,7 @@ jobs:
         run: npm test
         working-directory: js/api-augment
       - name: Build & Publish Dry Run
-        run: npm publish --dry-run
+        run: npm publish --dry-run --tag ci-validation
         working-directory: js/api-augment/dist
       - name: Generate npm tarball
         run: npm pack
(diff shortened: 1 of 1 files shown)


.github/workflows/common/setup-node-and-npm/action.yml
```diff
@@ -0,0 +1,37 @@
+name: "Setup Node and npm"
+description: "Set up Node.js and ensure an npm supporting 'id-token' is installed"
+
+inputs:
+  node-version:
+    description: 'Version Spec of the version to use. Examples: 12.x, 10.15.1, >=10.15.0.'
+    required: false
+    default: '22.x'
```
shannonwells: non-blocking:  Is 22 the one we still want or should we go up to 24.x?

shannonwells (approved):
Just the one question but I think this is fine assuming Aramik's Q about NODE_AUTH_TOKEN is resolved.

PR #2658, opened 2025-12-26
Sort schemas for intent

### LINKED ISSUE #2657: [Bug]: schemasRuntimeApi.getIntentById does not return schemas for intent in order
### What happened?

# Description
`schemasRuntimeApi.getIntentById`, when called with a second argument of `true`, is supposed to return the list of schemas in order of oldest => newest (as indicated by increasing numeric schema ID). It does not. This is because the storage keys are returned in the lexical order of the _hashed_ key, not the underlying key value.

Additional *ON TESTNET ONLY*:
On Testnet, it's possible that the _most recent_ schema for an Intent may not be the largest `schemaID` numerically, due to:
* schemaIDs created organically on Testnet start at 16000, but schemas synced from Mainnet start at 1
However, since syncing schemas from Mainnet => Testnet is a manual process anyway, this can be mitigated operationally by:
* Any time we sync schemas from Mainnet => Testnet we delete schemas > 16000
_OR_
* Any time we sync schemas from Mainnet => Testnet we mark any schemas > 16000 as either *Deprecated* or *Unsupported*

This is not an issue on Mainnet because schema IDs are monotonically increasing, and it is not an issue on local Dev chains because they are short-lived--although there is a discontinuous gap between Genesis schemas (from 1) and locally-created schemas (from 16000), schemas would always monotonically increase from 16000 once the chain is started.

# AC
- [ ] `schemasRuntimeApi.getIntentById(intentId, true)` always returns the list of schemas in increasing sorted order by ID

JoeCap08055:
# Goal
The goal of this PR is to make sure that when retrieving the list of implementing schemas with an intent, that the returned list only includes currently-supported schemas in sorted order.

Closes #2657 

# Checklist
- [x] Updated Pallet Readme?
- [x] Design doc(s) updated?
- [x] Unit Tests added?


enddynayn (approved):
great! 👍

PR #2661, opened 2026-01-05
fix: distinguish explicit vs implicit delegation permission revocation for Intents

### LINKED ISSUE #2660: Delegation API is missing top-level revocation block
# Description
Recent changes to the permissions model have required some downstream projects like Gateway to make updates; as part of doing so, it is recommended that Runtime APIs (or custom RPCs where they exist) be preferred over Storage API queries, in order to help insulate clients from chain implementation changes in storage representation.

In light of this, we've come to realize that the MSA Runtime API currently provides no mechanism to retrieve the top-level revocation block for a Provider Delegation; currently, clients can only retrieve that information via a direct storage API query.

Currently, the runtime API adjusts the revocation block reported for each granular permission based on whether the entire top-level delegation has been revoked as of the current block. However, this response is therefore missing the information of whether the top-level delegation has been revoked, or just each individual Intent grant. It is possible to grant a delegation to a Provider without delegating any specific Intents.

JoeCap08055:
# Description
This PR does the following:
- Adds `revoked_at` block number to the overall `DelegationResponse`, indicating the revocation of the overall Provider relationship, rather than individual Intent(s)
- Adds `explicit_revoked_at` block number to `DelegationResponse` for each delegated Intent; the existing `revoked_at` maintains its original meaning of the earlier of the Intent or overall Delegation revocation block.
- Changes the newly-added runtime API method `get_granted_intents_by_msa_id` to `get_delegation_for_msa_and_provider`, to provide a stable method for obtaining an entire delegation rather than just the list of delegated intents. Previously getting the entire delegation would have required a direct storage query, or retrieving ALL delegations. As this was a recently-added API method, the only downstream effect is internal, as it had not been deployed yet.

# Goal
Closes #2660 

# Checklist
- [x] Updated Pallet Readme?
- [x] Unit Tests added?


common/primitives/src/msa.rs
```diff
@@ -92,8 +94,10 @@ pub struct DelegationResponse<DelegationIdType, BlockNumber> {
 pub struct DelegationGrant<DelegationIdType, BlockNumber> {
 	/// ID of the delegated entity for which permission is/was granted
 	pub granted_id: DelegationIdType,
-	/// Block number the permission was/will be revoked (0 = not revoked)
+	/// Block number when the permission was/will be EFFECTIVELY revoked, taking into consideration the overall provider delegation revocation.
 	pub revoked_at: BlockNumber,
+	/// Block number the permission was/will be explicitly revoked (i.e., not implicitly revoked by a top-level revocation) (0 = not revoked)
+	pub explicit_revoked_at: BlockNumber,
```
JoeCap08055: Opted to add new `explicit_revoked_at` instead of `effective_revoked_at`, because `revoked_at` in this context has historically ALWAYS meant the _effective_ revocation block.

shannonwells (approved):
Looks good, ship it!

common/primitives/src/msa.rs
```diff
@@ -111,6 +115,7 @@ impl<DelegationIdType: Serialize, BlockNumber: Serialize> Serialize
 		s.serialize_field("schema_id", &self.granted_id)?;
 		s.serialize_field("granted_id", &self.granted_id)?;
 		s.serialize_field("revoked_at", &self.revoked_at)?;
+		s.serialize_field("explicit_revoked_at", &self.explicit_revoked_at)?;
```
aramikm: for these do we need to implement deserialization also? Since the default would not work

js/api-augment/definitions/msa.ts
```diff
@@ -86,10 +86,17 @@ export default {
     SchemaGrantResponse: {
       schema_id: 'IntentId',
       revoked_at: 'BlockNumber',
+      explicit_revoked_at: 'BlockNumber',
```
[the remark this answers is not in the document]
JoeCap08055: So, originally I thought that we had changed the behavior of the endpoint to tweak the `revoked_at` to show the _effective_ revocation block, so @enddynayn suggested I instead keep `revoked_at` unmodified and add `effective_revoked_at`. However, on further investigation, we've ALWAYS tweaked the response of this endpoint to show the _effective_ revocation block. So, we could just leave it alone. But that would mean the only way to determine if a permission was explicitly revoked would be to use a storage query, and we're trying to avoid those by giving clients appropriate runtime APIs. So I figured if we add `explicit_revoked_at`, then we never need to use a storage query.

I'm happy to remove if there's strong consensus against, though, since there isn't really a strong use case for needing the `explicit_revoked_at` block.

pallets/msa/src/tests/chat.md
```diff
@@ -0,0 +1,503 @@
+I'm interested in developing a standard for transactional agentic AI. I've come across this: https://github.com/agentic-commerce-protocol/
```
aramikm: Why do we need this and why keeping this in this PR?

PR #2664, opened 2026-01-09
fix the npm token issues

aramikm:
# Description

Fix the npm token changes
https://github.com/frequency-chain/frequency/actions/runs/20860686128/job/59941049866

Closes


.github/workflows/merge-pr.yml
```diff
@@ -9,6 +9,7 @@ on:
       - main
 # Set default permissions as restrictive
 permissions:
+  id-token: write  # Required for OIDC
```
aramikm: @JoeCap08055 Is the the only change needed in here or there are other things I'm missing?

demisx (approved):
👍🏻

.github/workflows/merge-pr.yml
```diff
@@ -9,147 +9,11 @@ on:
...
   contents: read
   packages: read
 
 jobs:
-  publish-js-api-augment-rc:
```
aramikm: Open to arguments against this but I don't think we need to publish these on merge. We can always use an `rc` release if needed any of these packages and I think simplifying this makes more sense compared to creating a shared workflow which requires more effort without any tangible reward.

PR #2698, opened 2026-07-21
feat(xcm): enable bridging on mainnet and fix runtime upgrade migrations

enddynayn:
# Goal
Adds Session and AuraExt v0 to v1 migrations, folds frequency-bridging into
the mainnet feature the same way testnet already does, and cleans up the
Makefile so mainnet and paseo targets are not misleading. Also drops
generate-lockfile from the deny CI job for now because core2 0.4.0 is
yanked upstream (paritytech/polkadot-sdk#11769).


.github/workflows/verify-pr-commit.yml
```diff
@@ -240,9 +240,9 @@ jobs:
       - name: Check Out Repo
         uses: actions/checkout@v6
       - name: Set Up Cargo Deny
-        run: |
-          cargo install --force --locked cargo-deny@0.18.9
-          cargo generate-lockfile
+        # Use the committed Cargo.lock. `cargo generate-lockfile` fails because
```
enddynayn, line 243: The vulnerable crates job was dying on cargo generate-lockfile, not on cargo deny itself.

That line has been around since when we first added the audit check in [#520](https://github.com/frequency-chain/frequency/pull/520). Back then the job used cargo audit, and the usual setup was install the tool, generate a lockfile, then run the audit. That made more sense if you might not have a committed lockfile. We do commit Cargo.lock, and the job later moved to cargo deny, but it looks like the regenerate step just stayed. 

What happens now is we re resolve the whole dependency graph from scratch, and crates.io will not let you pick a yanked version in that situation. core2 0.4.0 got yanked, and we still pull it in through cid 0.9 from polkadot sdk sc-network. So fresh lockfile generation fails even though our committed Cargo.lock still builds fine under --locked.

Writeup from polkadot-sdk here: https://github.com/paritytech/polkadot-sdk/issues/11769

I took generate-lockfile out of this job for now so deny just checks the lockfile we already commit, which is what we actually build with. Once we bump polkadot sdk to a release that no longer needs yanked core2, we can put the regenerate step back if we still want it.

deny.toml
```diff
@@ -79,9 +79,28 @@ ignore = [
...
   { id = "RUSTSEC-2026-0020", reason = "Wasmtime resource exhaustion vulnerability, pending update in polkadot-sdk." },
-  { id = "RUSTSEC-2026-0021", reason = "Wasmtime panic vulnerability, pending update in polkadot-sdk." }
+  { id = "RUSTSEC-2026-0021", reason = "Wasmtime panic vulnerability, pending update in polkadot-sdk." },
+  { id = "RUSTSEC-2026-0085", reason = "Wasmtime advisory, pending update in polkadot-sdk." },
+  { id = "RUSTSEC-2026-0086", reason = "Wasmtime advisory, pending update in polkadot-sdk." },
+  { id = "RUSTSEC-2026-0087", reason = "Wasmtime advisory, pending update in polkadot-sdk." },
+  { id = "RUSTSEC-2026-0088", reason = "Wasmtime advisory, pending update in polkadot-sdk." },
+  { id = "RUSTSEC-2026-0089", reason = "Wasmtime advisory, pending update in polkadot-sdk." },
+  { id = "RUSTSEC-2026-0091", reason = "Wasmtime advisory, pending update in polkadot-sdk." },
+  { id = "RUSTSEC-2026-0092", reason = "Wasmtime advisory, pending update in polkadot-sdk." },
```
enddynayn, line 92: Polkadot-sdk upgrades are going to matter soon.

harry-evans (approved):
LGTM

sbendar:
Codex didn't seem have anything significant to say.

sbendar (approved):
Codex didn't have anything significant to say.

